PR #19, opened 2018-03-19
Allow Edge testing with webdriverio

alecf:
This is clearly a typo - not only is `i` not defined in the scope of this line, but it needs to use the edge regex directly, not look it up in list of all browser regexes.

Without this, you can't test MS Edge with webdriverio!


danielputerman:
Nice catch! Thanks

alecf:
Thanks @danielputerman - any chance of a release on NPM?

danielputerman:
Sure, probably today.

alecf:
Any chance for a release? (I still cannot test Edge without this..)

PR #36, opened 2018-08-01
Fix axios issue with empty buffer

amitzur:
Axios hangs when called with POST/PUT and an empty buffer as data.
Issue was submitted here:
https://github.com/axios/axios/issues/1701

This PR is a workaround until they fix the issue (or we switch http client).


delirius325:
This issue is related to https://github.com/olalonde/follow-redirects/pull/90 which is related to https://github.com/nodejs/node/issues/22066

Cheers,

PR #53, opened 2018-11-08
Jest performance test

corevo:
The test will fail unless completed within 8 seconds
Run through `npm run test:perf` either in the package or in the root
image

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,7 +3,8 @@
   "extends": ["airbnb-base"],
   "env": {
     "node": true,
-    "mocha": true
+    "mocha": true,
+    "jest": true
   },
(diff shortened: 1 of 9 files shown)


astappiev:
I don't like one more testing framework used in the SDK. Can I rewrite the test using mocha?

corevo:
No, this is a performance test targeted specifically at jest.  
For some reason jest's performance is hellish slow, unlike mocha, for that reason I wanna have a test that will make sure jest is as performant as mocha.

corevo:
In this example the test takes a single full page screenshot, it takes 40 seconds, while on mocha Daniel and I assume would take 8 seconds.

PR #54, opened 2018-11-20
Support deviceInfo, hostAppInfo, hostOSInfo

amitzur:
We lost this feature in Cypress and Storybook after the `visual-grid-client` updated to `eyes-sdk-core` v4.
It was added to v2.0.1 not through git :(


astappiev:
Published as 4.2.0

PR #55, opened 2018-11-27
Jsdoc update

yonatanLehman:
updates to jsdoc

### DIFF
--- a/eyes.sdk.core/lib/BatchInfo.js
+++ b/eyes.sdk.core/lib/BatchInfo.js
@@ -9,9 +9,9 @@ const { GeneralUtils } = require('./utils/GeneralUtils');
 class BatchInfo {
   /**
    * Creates a new BatchInfo instance.
-   * @param {string} [name] Name of batch or {@code null} if anonymous.
-   * @param {Date} [startedAt] Batch start time, defaults to the current time.
-   * @param {string} [id] ID of the batch, should be generated using GeneralUtils.guid().
+   * @param {string} [name] - Name of batch or {@code null} if anonymous.
+   * @param {Date} [startedAt] - Batch start time, defaults to the current time.
+   * @param {string} [id] - ID of the batch, should be generated using GeneralUtils.guid().
    */
   constructor(name, startedAt, id) {
     this._id = id || process.env.APPLITOOLS_BATCH_ID || GeneralUtils.guid();
@@ -20,6 +20,7 @@ class BatchInfo {
   }
 
   /**
+   * @protected
    * @param {object} object
    * @return {BatchInfo}
    */
@@ -28,7 +29,7 @@ class BatchInfo {
   }
 
   /**
-   * @return {string} The id of the current batch.
+   * @return {string} - The id of the current batch.
    */
   getId() {
     return this._id;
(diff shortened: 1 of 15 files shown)


astappiev:
Hi @yonatanLehman,

I have made a few changes in the direction in master branch:
1. Replaced all inline JSDoc blocks to multiline blocks
2. `fromObject`, `fromRegion` and other `from*` methods have been removed completely
3. Fixed few spelling mistakes

About other changes:
1. I don't like ' - ' in `@param` descriptions
2. I do not agree with most of the new `@protected` tags
3. If there is `@inheridDoc` tag, then `@override` tag is redundant http://usejsdoc.org/tags-inheritdoc.html
4. There is no `Integer` type in javascript https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures
5. There is no `@paramprivate` tag at all, especially if after this param comes another params

Partially, we have discussed the changes with @danielputerman. He told that you are using some parser and the formatting is preferred for it.

But, we need JSDoc not only for creating documentation or for reading by customers, it also uses by IDE for autocomplete and detecting mistakes. Like, when we mark a method as `@protected` it means that IDE will not allow to use it anywhere except this file and inherited ones. If we want to only use the method internally, we can use `@package` tag to make it package-private. But most of the methods used cross-packages like `eyes-sdk-core` methods used in `eyes-selenium`.

Can you try to use the latest version in your parser? Let me know what else I can do to help you.

Thanks,
Oleh

yonatanLehman:
Hi Oleh
Good to talk to you.


   1. I don't like ' - ' in @param descriptions

 I know you don't like it - I've spent a long time trying to get the parser
(Antlr4) to work without it, so far I have not been able to do it.
Do you know Antlr4 ? Maybe you can find a way to solve this problem.
I would appreciate for the moment if we can leave it the "-"  otherwise, I
can't pull out the types for the documentation.,
and need to do it manually which is error-prone and takes a long time.
On my part,  I will keep trying to remove this restriction maybe I'll ask
on Stack Exchange for a solution.
I'll even remove the "-" myself if I find a solution !!!

2. I do not agree with most of the new @protected tags
I don't care if it's @private or @protected but we need some way to
indicate which methods are available to the user and which are not.
What do you suggest? Maybe the correct solution is to have a Facade class
that only contains user accessible methods.

3. If there is @inheridDoc tag, then @override tag is redundant
http://usejsdoc.org/tags-inheritdoc.html
I don't user either of these tags.

4. There is no Integer type in javascript
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures

Did I add Integer somewhere ? I seem to recall that some methods will only
work, or only make sense if you pass them an integer as opposed to a
decimal (float). For examples
RectangleSize|{width: number, height: number}
width and height cannot be 10.3 and 11.5
I don't know what will happen if a non-integer value is passed, maybe they
will be truncated
but from a *semantic* point of view, only integer values are acceptable.

I could convey that in a comment I suppose, but I think it's cleaner
to make it explicit in the type, even if this has no formal meaning in
terms of
the Javascript implementation.
Let's discuss specific cases to see what makes sense.

5. There is no @paramprivate tag at all, especially if after this param
comes another params
I made this up to indicate that a parameter is intended for internal use
only.
I suppose the correct way is to have a separate API for user and internal
use, one without the parameter and one with.

Maybe we can discuss this by phone ? My number is +972-54-56634848

Regards
Yonatan

PR #56, opened 2018-11-28
return test results from abortIfNotClosed

amitzur:
right now this method always returns undefined

### DIFF
--- a/packages/eyes-sdk-core/lib/EyesBase.js
+++ b/packages/eyes-sdk-core/lib/EyesBase.js
@@ -991,7 +991,7 @@ class EyesBase {
       this._logger.verbose('Aborting server session...');
       try {
         // When aborting we do not save the test.
-        await this._serverConnector.stopSession(this._runningSession, true, false);
+        return await this._serverConnector.stopSession(this._runningSession, true, false);
         this._logger.log('--- Test aborted.');
       } catch (err) {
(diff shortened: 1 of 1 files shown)


astappiev:
`abortIfNotClosed` is void method without a return value. The method is inherited Java SDK's implementation.
https://github.com/applitools/eyes.sdk.java3/blob/master/eyes.sdk.core/src/main/java/com/applitools/eyes/EyesBase.java#L894

amitzur:
Thing is, before [this commit](https://github.com/applitools/eyes.sdk.javascript1/commit/57f4a7f47b85b3dfcb786dcae7711652c989056f#diff-dd9740fecc094ec0741c0ab0f5046bb5L936) it did return the test results from the server.
I think it's also the right behavior to return TestResults from this method because:
(a) there's data from the server to give to the user
(b) I need this info :-)

astappiev:
Okay, will release update today.

amitzur:
thank!
when can I expect it in npm?

astappiev:
I'm working on another feature requested by @corevo which is also needed today. I planed to release them together in few hours. If you need it now, I can release it earlier.

amitzur:
I don't need it now, next week is just fine. Thank you!

astappiev:
Published in v. 4.4.0

PR #58, opened 2018-12-12
interop deepmerge beetween commonjs and esmodules

corevo:
`deepmerge` returns an esmodule for browser packages, and a commonjs version for Node.  
Which means that for Node you get back a `function` while in the browser you get
```js
{
  default: function,
  __es_module: true
}
```

But since the file uses `require`, it only evaluates modules as commonjs, which means that in the browser when you call merge, you don't call the function, but an object.  
This will interop the two, if it finds the esmodule version of `deepmerge` it will set `merge` to the `default`, while otherwise to the function.

### DIFF
--- a/packages/eyes-sdk-core/lib/utils/GeneralUtils.js
+++ b/packages/eyes-sdk-core/lib/utils/GeneralUtils.js
@@ -1,10 +1,11 @@
 'use strict';
 
-const merge = require('deepmerge');
+const _merge = require('deepmerge');
 const dateFormat = require('dateformat');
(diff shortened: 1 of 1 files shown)


astappiev:
Looks like it was "fixed" (they removed es build altogether) a few days ago in version 3.0.

astappiev:
I updated package to version 3.0, can you try?

astappiev:
@corevo ping

corevo:
@astappiev thanks!

PR #60, opened 2019-01-21
Add configuration functions to eyes-rendring package

yuriieasternpeak:


packages/eyes-rendering/lib/Eyes.js
```diff
@@ -76,16 +77,16 @@ class Eyes {
...
       envName: renderingConfiguration.getEnvironmentName(),
-      // ignoreCaret,
-      // isDisabled,
-      // matchLevel,
-      // matchTimeout,
+      ignoreCaret: this._configuration.getIgnoreCaret(),
+      isDisabled: this._configuration.isDisabled(),
+      matchLevel: this._configuration.getMatchLevel(),
+      matchTimeout: this._configuration.getMatchTimeout(),
       parentBranchName: renderingConfiguration.getParentBranchName(),
```
corevo: The getters and setters swill won't affect this, this should be more like:
```js
this._configuration.getParentBranchName() ? this._configuration.getParentBranchName() : renderingConfiguration.getParentBranchName()
```

corevo: There's still an argument to be made which one takes precedence, the importance is that you would be able to define this with the getters and setters as well.

packages/eyes-rendering/lib/Eyes.js
```diff
@@ -257,6 +258,132 @@ class Eyes {
       sendDom: checkSettings.getSendDom(),
```
corevo: Make sure to send the match level if `this._matchLevel` is set, unless specified explicitly in `checkSettings`.

packages/eyes-rendering/lib/Eyes.js
```diff
@@ -257,6 +258,132 @@ class Eyes {
...
+    this._configuration.setIsDisabled(isDisabled);
+  }
+
+  /**
+   * @return {boolean} Whether eyes is disabled.
+   */
+  isDisabled() {
+    return this._configuration.getIsDisabled();
+  }
+
```
corevo: very important not to forget `setViewportSize` although it makes no difference for the baseline in the visual grid, it is important to resize the browser correctly.

PR #63, opened 2019-01-31
Feature: send dom using @applitools/dom-capture

yuriieasternpeak:


astappiev:
Try to use the latest version of `@applitools/dom-capture` (7.0.5).
And replace `@applitools/eyes-sdk-core` with `@applitools/eyes-common`.

Thanks

danielputerman:
@astappiev we'll first release a version based on capture 4, then update to version 7.

PR #64, opened 2019-02-10
Refactor eyes-selenium's Eyes to enable just taking a screenshot 

amirhalatzi:
We'd like to use this package inside the rendering grid,and need the functionality to take a screenshot without sending it to eyes.

Changes made:
- Move webdriver init from openEyes to a seperate, public, setDriver method (then called from openEyes)
- Allow to inject a different checkFunction (instead of super.checkWindowBase), used in getCheckSnapshot method


astappiev:
Hi Amir,

Why you can't just override the `checkWindowBase` method?

Thanks

amirhalatzi:
If I understand you correctly, you're suggesting is to override `checkWindowBase` in my external library. 
I prefer not to hack `Eyes` and have a straight-forward solution that will be maintained as part of this library

danielputerman:
Closed due to using an alternative method.

PR #65, opened 2019-02-10
Added general util function to get property of object by path

danielschwartz85:
for doing :
getPropertyByPath(obj, "user.name.email") // return undefined or value - no exceptions
* this is instead of doing: obj && obj.user && obj.user.email && obj.user.address ..


danielschwartz85:
Tnx.
btw - we can add an exception if path regex is invalid.. (works for my usage..)

PR #72, opened 2019-04-09
fixes IN jsdoc for EKB parse

yonatanLehman:
1) made a change so that the parser would run /*** to /*
2) Added signature that was missing

### DIFF
--- a/packages/eyes-common/lib/config/Configuration.js
+++ b/packages/eyes-common/lib/config/Configuration.js
@@ -397,10 +397,16 @@ class Configuration {
     }
     return this;
   }
-
+ 
   /**
    * Adds a property to be sent to the server.
-   *
+   * @signature `addProperty(property)`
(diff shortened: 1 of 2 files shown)


astappiev:
Added via cherry-pick  https://github.com/applitools/eyes.sdk.javascript1/commit/0ba1d8995e1f320dca6f4bd9b37642d09f94dbe0

yonatanLehman:
Hi
Thanks
When do you think you will have an update of VisualGridRunner for
VisualGridRunner getAllTestResults(shouldThrowException)

Will it be compatible with the Java method defined here in terms of symbol
names and behavior?

https://applitools.com/docs/api/eyes-sdk/classes-gen/class_visualgridrunner/method-visualgridrunner-getalltestresults-selenium-java.html?Highlight=getAllTestResults

PR #85, opened 2019-08-07
JSDOC 

yonatanLehman:
Adding missing @signatures to simplify API documentation

Note that if multiple @signature share the same parameter name you only need to define the @sigparam once (after the first @signature that uses it) and it is shared by all the following @signatures.
Note that you cant use the same name for a parameter for two different types
*The name of the parameter implies its type*


astappiev:
What did I say about tons of merges? I will create a new commit with these changes and put you as an author.

astappiev:
Why do we need this? https://github.com/applitools/eyes.sdk.javascript1/pull/85/commits/44a27e52d195fd00ab30caff2c31f12f5d00c881

astappiev:
Other changes committed to master

yonatanLehman:
Sorry - I tried to eliminate the intermediates and didn't get it right.

PR #88, opened 2019-09-02
vgc: takeScreenshot returns imageLocation in object along with renderId

giltayar:


giltayar:
This changes the interface of vgc's `take-screenshot`, nobody's really using it, so I didn't update the semver major version.

PR #99, opened 2019-11-19
Enable HTTP only corporate proxies

edmulraney:
Axios doesn't support HTTP only proxies. In order to use this in HTTP only corporate environment we can add an HTTPS over HTTP tunnel to the proxy.

This is not production ready but a POC hack to show how you can enable it.

See: https://janmolak.com/node-js-axios-behind-corporate-proxies-8b17a6f31f9d


danielputerman:
Thanks @edmulraney !
@amitzur Pls review.

amitzur:
Thanks for the contribution. This has been added to our SDK at about the same time following your suggestion.

PR #103, opened 2019-12-17
Add support for nested iframes

amitzur:
Couple of notes:
1) Not sure about the name `unserializeDomSnapshotResult`. Other options were `fromBlobsToResourceContents` or `blobsToResourceContents`.

2) Implementation turned a bit more verbose than I anticipated. My goals were (a) to be immutable, and (b) not to keep the `blobs` property.


danielschwartz85:
looks good to me.
(can be added to VGC render script as well..)

PR #107, opened 2020-01-13
Fix wait before screenshots

kyryloonufriiev:


packages/eyes-selenium/test/unit/Eyes.spec.js
```diff
@@ -13,9 +16,12 @@ const {
   ProxySettings,
   BatchInfo,
   PropertyData,
+  Target,
 } = require('../../index')
 
 describe('Eyes', function() {
+  this.timeout(60 * 1000)
```
amitzur: Since we run mocha with `--no-timeouts` this isn't needed. I think we can remove it when we see it in other places.

packages/eyes-selenium/test/unit/Eyes.spec.js
```diff
@@ -103,4 +109,41 @@ describe('Eyes', function() {
     assert.strictEqual(eyes.getBaselineEnvName(), 'baselineEnvName')
     assert.strictEqual(eyes.getSendDom(), false)
   })
+
+  it('should wait before screenshots', async function() {
+    try {
+      const driver = await new Builder().withCapabilities(Capabilities.chrome()).build()
+      const eyes = new Eyes()
+      const server = await fakeEyesServer({logger: {log: () => {}}})
+      const config = new Configuration()
```
amitzur: You can just do `eyes.setServerUrl`. No need for setting the API key since you are working with a fake server.

packages/eyes-selenium/test/unit/Eyes.spec.js
```diff
@@ -103,4 +109,41 @@ describe('Eyes', function() {
...
+      config.setServerUrl(`http://localhost:${server.port}`)
+      config.setApiKey('fakeApiKey')
+      eyes.setConfiguration(config)
+      eyes.setWaitBeforeScreenshots(1000)
+      let startTime
+      let duration
+      const thrownScreenshotDone = Symbol()
+      eyes.getScreenshot = function() {
+        duration = Date.now() - startTime
+        throw thrownScreenshotDone
```
amitzur: clever move!
This way you stop the process early and are able to inspect state.
It's a bit hacky though. I think with a more testable system we won't need these sort of smart tricks. But very nice, I learned something new!

packages/eyes-sdk-core/lib/EyesBase.js
```diff
@@ -615,6 +616,8 @@ class EyesBase extends EyesAbstract {
     // default result
     const validationResult = new ValidationResult()
 
+    await GeneralUtils.sleep(this._configuration.getWaitBeforeScreenshots())
```
amitzur: Regarding the default -
I think we should also test what happens when the user **doesn't** set this value explicitly. There's a default of 100ms (https://github.com/applitools/eyes.sdk.javascript1/blob/master/packages/eyes-common/lib/config/Configuration.js#L34).

In case the user explicitly sets this to null or undefined, I think maybe we should protect against that in the `setWaitBeforeScreenshots` setter.

It's not really that important to avoid a setTimeout at all. It's negligible with the time it takes for the whole `check` process.

kyryloonufriiev: For me, it doesn't have a lot of sense to wait by default 100ms before normal viewport screenshot if I even didn't specify it. Of course, the default value is important for full-page screenshots.
About protections of undefined and null, I think we can add type checking in the setter.
About `setTimeou` I think it doesn't cost anything to check how much we want to wait in `sleep` function and avoid not expensive but useless calls of `setTimeout` with 0 or nullish value.

packages/eyes-selenium/test/unit/Eyes.spec.js
kyryloonufriiev: I remove this in the latest commit

packages/eyes-selenium/test/unit/Eyes.spec.js
kyryloonufriiev: I think I still need to set the API key because otherwise, I get this
`Error: API key is missing! Please set it using setApiKey()`

packages/eyes-selenium/test/unit/Eyes.spec.js
kyryloonufriiev: In the new version, I come with a Proxy-based version of this trick.

kyryloonufriiev:
I update the current branch with with new test
1.  Default case without `setWaitBeforeScreenshots`
2. Null case with `setWaitBeforeScreenshots(null)`

Tests are failing not because of my changes, as far as I know.

packages/eyes-selenium/test/unit/Eyes.spec.js
amitzur: Oh, the SDK validates the input. I don't get this because I have an env variable set. This is actually an important point - I might be missing things because my environment has an API key set!
I suppose we need to think about this. Did you run other tests yet using `npm run test:unit` and `npm run test:it`?

packages/eyes-sdk-core/lib/EyesBase.js
amitzur: I agree that you can check the value and not call it if it's null or undefined. If it's `0` then you should still call it.
So we shouldn't add protection in the setter, so that the user would be able to avoid the timeout.

packages/eyes-selenium/test/unit/Eyes.spec.js
```diff
@@ -110,40 +113,86 @@ describe('Eyes', function() {
...
+      server = await fakeEyesServer({logger})
+      driver = await new Builder()
+        .withCapabilities(Capabilities.chrome())
+        .setChromeOptions(new ChromeOptions().addArguments('disable-infobars').headless())
+        .build()
+      eyes = new Proxy(new Eyes(), {
+        get(target, key, receiver) {
+          if (key === 'checkWindowBase') {
+            checkTimestamp = Date.now()
+          } else if (key === '_ensureRunningSession') {
```
amitzur: maybe 'startSession' instead?

packages/eyes-selenium/test/unit/Eyes.spec.js
```diff
@@ -110,40 +113,86 @@ describe('Eyes', function() {
...
+            networkTimestamp = Date.now()
+          } else if (key === 'getScreenshot') {
+            const screenshotTimestamp = Date.now()
+            duration =
+              screenshotTimestamp - checkTimestamp - (screenshotTimestamp - networkTimestamp)
+            throw thrownScreenshotDone
+          }
+          return Reflect.get(target, key, receiver)
```
amitzur: what's the benefit of this over `target[key]` ?

kyryloonufriiev: First of all, it is the recommended way to achieve default behavior inside a Proxy trap. We need to provide a receiver if we get a class method, the receiver will be `this` inside class method, otherwise, we couldn't handle all of nested calls.

packages/eyes-selenium/test/unit/Eyes.spec.js
kyryloonufriiev: I tried it first but after realizing that it isn't called when you doing check second time.

packages/eyes-selenium/test/unit/Eyes.spec.js
kyryloonufriiev: When I run it I explicitly set APPLITOOLS_API_KEY before a call.

packages/eyes-selenium/test/unit/Eyes.spec.js
```diff
@@ -103,4 +112,87 @@ describe('Eyes', function() {
...
+      eyes = new Proxy(new Eyes(), {
+        get(target, key, receiver) {
+          if (key === 'checkWindowBase') {
+            checkTimestamp = Date.now()
+          } else if (key === '_ensureRunningSession') {
+            networkTimestamp = Date.now()
+          } else if (key === 'getScreenshot') {
+            const screenshotTimestamp = Date.now()
+            duration =
+              screenshotTimestamp - checkTimestamp - (screenshotTimestamp - networkTimestamp)
```
amitzur: Why do you need to subtract `(screenshotTimestamp - networkTimestamp)`? isn't it negligible with the fake eyes server?

packages/eyes-selenium/test/unit/Eyes.spec.js
amitzur: Ah, got it. This behavior is because you call `eyes.close` only after all the tests, and not in `afterEach`. So all these tests are considered steps within the same session. So there's no need to call `startSession` again.
It's fine, we should just acknowledge it. Maybe insert a short comment about it.

amitzur:
The CI failure was due to a change in the baseline because the visual grid updated Firefox version. I've updated the baseline and restarted the job in Travis

packages/eyes-selenium/test/unit/Eyes.spec.js
kyryloonufriiev: No. It's take about 600 - 700 ms. If I didn't do that, I need to set really long delay to see that something was really happened.
I need to test default case when delay is 100ms, but networking takes much more.

PR #108, opened 2020-01-13
Test fluent api

Artem-applitools:
It's still Raw. I am still haven't finished updating test cases to apply assertions.
Also I am planning to put method for the assertion to the separate file, but haven't finish this yet


packages/eyes-selenium/test/coverage/custom/TestFluentApi.spec.js
```diff
@@ -0,0 +1,440 @@
+'use strict'
```
tourdedave: You don't need to cover all of these tests. Most of them are already handled in the coverage-tests DSL.

I linked you to the specific FluentApi tests that need to be covered. Here are the links again:

https://github.com/applitools/Eyes.Sdk.DotNet/blob/master/Tests/Test.Eyes.Selenium.DotNet/TestFluentApi.cs#L187-L193
https://github.com/applitools/Eyes.Sdk.DotNet/blob/master/Tests/Test.Eyes.Selenium.DotNet/TestFluentApi_Frames.cs#L165-L191
https://github.com/applitools/Eyes.Sdk.DotNet/blob/master/Tests/Test.Eyes.Selenium.DotNet/TestFluentApi_Frames.cs#L137-L163

If you're not sure if you should write a test or not, see if it's already specified in the coverage-tests DSL ([link](https://github.com/applitools/eyes.sdk.javascript1/blob/master/packages/sdk-test-kit/src/coverage-tests/tests.js))

packages/eyes-selenium/test/coverage/custom/util/FloatingRectangle.js
```diff
@@ -0,0 +1,14 @@
+'use strict'
```
tourdedave: I would try to avoid classes unless necessary. Rectangles are simple objects and we don't need to track side-effects with them. I would opt for a pure `make` function instead (e.g., `makeFloatingRectangle`). We could also combine `Rectangle.js` and `FloatingRectangle.js` into a single util file.

packages/eyes-selenium/test/coverage/custom/TestFluentApi.spec.js
```diff
@@ -0,0 +1,92 @@
...
+            currentY += 5000, c++
+          ) {
+            let region
+            if (rect.height > currentY + 5000) {
+              region = new Region(rect.x, currentY, rect.width, 5000)
+            } else {
+              region = new Region(rect.x, currentY, rect.width, rect.height - currentY)
+            }
+            await eyes.check('Check Long Out of bounds Iframe Modal', Target.region(region))
+          }
```
tourdedave: For easier understanding, I would pull this into its own function that receives a `rect` and call it something like `performChecksOnLongRegion`

packages/eyes-selenium/test/coverage/custom/util/TestSetup.js
```diff
@@ -50,7 +60,18 @@ function getEyes(runnerType, stitchMode, branchName) {
   }
 }
 
+function getSetups(...args) {
+  let setups = []
+  if (args.length !== 0) {
```
tourdedave: You can just say `if (args.length)`. It will evaluate the same and be more concise.

packages/eyes-selenium/test/coverage/custom/util/TestSetup.js
```diff
@@ -50,7 +60,18 @@ function getEyes(runnerType, stitchMode, branchName) {
...
+    args.forEach(arg => setups.push(SETUPS[arg]))
```
tourdedave: You could just use `map` and `return` that instead (e.g., `return args.map(arg => SETUPS[arg])`)

packages/eyes-selenium/test/coverage/custom/util/TestSetup.js
```diff
@@ -50,7 +60,18 @@ function getEyes(runnerType, stitchMode, branchName) {
...
+  } else {
+    setups.push(SETUPS.default, SETUPS.scroll, SETUPS.VG)
```
tourdedave: There's no need to use a `setup` object since you could just return an explicit array (e.g., `return [SETUPS.default, SETUPS.scroll, SETUPS.VG]`)

packages/eyes-selenium/test/coverage/custom/TestSendDom.spec.js
```diff
@@ -0,0 +1,114 @@
+'use strict'
+const {assertImage} = require('./util/ApiAssertions')
+const {getDriver, getEyes, getSetups} = require('./util/TestSetup')
+// const {expect} = require('chai')
+const {By} = require('selenium-webdriver')
+const {/*EyesSelenium,*/ BatchInfo, Target} = require('../../../index')
+const appName = 'Test Send Dom'
+
+/*class DomInterceptingEyes extends EyesSelenium {
```
tourdedave: Please delete commented out code before committing.

packages/eyes-selenium/test/coverage/custom/TestAbort.spec.js
```diff
@@ -0,0 +1,128 @@
...
+  ClassicRunner,
+  VisualGridRunner,
+  Eyes,
+} = require('../../../index')
+const appName = 'My application'
+const testedUrl = 'https://applitools.com/docs/topics/overview.html'
+describe(appName, () => {
+  let setups = getSetups()
+  let batch = new BatchInfo('JS My Batch')
+  setups.forEach(setup => {
```
tourdedave: This suite doesn't need all setups. Just VG and the classic runner (it doesn't matter which stitch mode). Also, generating the tests like this makes it harder to reason about and debug individual test cases for individual execution modes. It would be better to flatten the tests (even if there we end up with some code duplication initially).

tourdedave: The same thinking goes for the other test files too. It would be better to flatten them as well to make them more understandable and simpler to debug.

Artem-applitools: So tests should end up in the separate file per execution mode or it could be separate describe section per execution mode?

PR #109, opened 2020-01-14
Fix long running task polling

kyryloonufriiev:
Refactoring for the core logic of ServerConnector is finished. I separate helper functions to prepare and send a request. I inject the helper functions in Axios workflow through interceptors, to configure your request you can specify `_options` property in Axios config object.
Here is still space for some tests, I personally add test for the actual functionality I fixed and update some test wich use old `_createHttpOptions` method.

I want to know your opinion about this implementation.


packages/eyes-sdk-core/lib/server/requestHelpers.js
```diff
@@ -0,0 +1,262 @@
...
+  if (!('proxy' in config)) {
+    const proxy = configuration.getProxy()
+    if (TypeUtils.isNotNull(proxy)) {
+      configRequestProxy(config, {proxy, logger})
+    }
+  }
+
+  return config
+}
+const prepareRequest = async (config, {defaults, configuration, logger}) => {
```
amitzur: Generally i prefer to use one unary parameter. It's the closest we have to named parameters. So here you can bring config inside: `function({config, defaults, configuration, logger})`.

amitzur: The naming is confusing here: `config` vs. `configuration`. Since `configuration` is easily understood as our `Configuration` object, I'd change `config`. Here are a few suggestions: `options` or `httpOptions` or `axiosConfig`.

packages/eyes-sdk-core/lib/server/requestHelpers.js
```diff
@@ -0,0 +1,262 @@
...
+  configRequest(config, {defaults, configuration, logger})
```
amitzur: I prefer we use a function statement `function prepareRequest` instead of the arrow function expression. That way it's hoisted, it's consistent with our coding style (the newer one that's in visual-grid-client, not the one dominant in eyes-sdk-core), and makes it clear that it's a function.
I generally use arrow functions for smaller inline functions, esp. when I want to keep using `this` for the outside context.

packages/eyes-sdk-core/lib/server/requestHelpers.js
```diff
@@ -0,0 +1,262 @@
...
+      proxyAuth,
+    },
+  })
+  config.httpsAgent = agent
+  config.proxy = false // don't use the proxy, we use tunnel.
+
+  logger.log('proxy is set as http only, using tunnel', proxyObject.host, proxyObject.port)
+  return config
+}
+const configRequest = (config, {defaults, configuration, logger}) => {
```
amitzur: This is the replacement for `ServerConnector::_createHttpOptions`. Now that it's a utility function, I think it should be a pure function. It currently has a return value, but the caller in `prepareRequest` doesn't use it, and assume it mutates the `config` argument.

I also think that delegating behavior with defaults to this function is redundant. The ServerConnector (i.e. Business Logic - BL) is providing the defaults. So it might as well do the merge and then you'll have a cleaner API with the interceptor (for example, you won't need `isExternalRequest`).

packages/eyes-sdk-core/lib/server/requestHelpers.js
```diff
@@ -0,0 +1,262 @@
...
+const REQUEST_GUID = GeneralUtils.guid()
+const getRequestId = () => `${++counter}--${REQUEST_GUID}`
+
+const getPollingDelaySequence = () => {
+  let repeat = 0
+  return {
+    next() {
+      return {
+        value:
+          POLLING_DELAY_SEQUENCE[
```
amitzur: I think this is an over complication of the use case. It's not necessary to make an iterator out of this.
I imagine just an array of values (you can construct it programatically) and a then just use the `repeat` counter to access the element in that array. If you're out of bounds - use the last value. It's the same concept as what you did here, but the code as it is written now looks verbose and complicated.

kyryloonufriiev:
I fixed all of commented issues with new commit. Also, I remove unused functions such as `ServerConnector:: downloadResource` and `ServerConnector::_createHttpOptions`.
I decide to move `setProxyOptions` to requestHelpers file as `configRequestProxy` and fix issues with it.

One thing isn't done is about `configRequest` and `configRequestProxy` are look like pure functions but they are not (because of modifying `axiosConfig` argument). In my opinion, we don't really need them to be pure, because it will make us do a deep copy of the `axiosConfig` object, that will take extra logic to make it right because `axiosConfig` can contain nested objects, arrays, and functions.

packages/eyes-sdk-core/lib/server/ServerConnector.js
```diff
@@ -70,69 +69,34 @@ class ServerConnector {
...
-
-    this._axios = axios.create()
-    this._axios.interceptors.request.use(async config => {
-      return prepareRequest(config, {
-        defaults: this._defaultRequestConfig,
+    })
+    this._axios.interceptors.request.use(axiosConfig => {
+      axiosConfig._options = Object.assign(
+        {withApiKey: true, retry: 1, repeat: 0, delayBeforeRetry: false},
+        axiosConfig._options,
```
amitzur: what's the meaning of `_options`? is it all the options that are not axios-related? So we piggypack axiosConfig just to get them into the interceptor? If so, then I think we can make it cleaner.
I also want to isolate the logic that merges `configuration` into the axios config.
And if a function is not pure, then let's make that clear by not returning anything.

Let me describe what I mean in code:
```
function configRequestHeaders({axiosConfig, requestId, isLongRequest, isPollingRequest}) {
  if (!(CUSTOM_HEADER_NAMES.REQUEST_ID in axiosConfig.headers)) {
    requestId = requestId || getRequestId()
    axiosConfig.headers[CUSTOM_HEADER_NAMES.REQUEST_ID] = requestId
  }
  if (isLongRequest) {
    axiosConfig.headers[CUSTOM_HEADER_NAMES.EYES_EXPECT] = '202+location'
  }
  if (isLongRequest || isPollingRequest) {
    axiosConfig.headers[CUSTOM_HEADER_NAMES.EYES_DATE] = DateTimeUtils.toRfc1123DateTime()
  }

  // no return value
}

/*****************************/

function extractConfigurationForAxios({configuration, withApiKey, logger}) {
  const partialAxiosConfig = {
    params: {},
  }

  if (withApiKey) {
    partialAxiosConfig.params.apiKey = configuration.getApiKey()
  }

  const removeSession = configuration.getRemoveSession()
  if (TypeUtils.isNotNull(removeSession)) {
    partialAxiosConfig.params.removeSession = removeSession
  }

  const timeout = configuration.getConnectionTimeout()
  if (TypeUtils.isNotNull(timeout)) {
    partialAxiosConfig.timeout = timeout
  }

  const proxy = configuration.getProxy()
  if (TypeUtils.isNotNull(proxy)) {
    Object.assign(partialAxiosConfig, configRequestProxy({proxy, logger}))
  }

  return partialAxiosConfig
}

/*****************************/

async function prepareRequest({config, logger, name, requestId, delay, repeat}) {
  logger.verbose(
    `ServerConnector.${name} [${requestId}] will now call to ${
      config.url
    } with params ${JSON.stringify(config.params)}`,
  )

  if (delay) {
    const delay = TypeUtils.isArray(delay) ? delay[Math.min(repeat, delay.length - 1)] : delay
    logger.verbose(`ServerConnector.${name} request delayed for ${delay} ms.`)
    await GeneralUtils.sleep(delay)
  }

  return config
}

/*****************************/

this._axios.interceptors.request.use(async axiosConfig => {
      // 1. default options
      const options = Object.assign(
        {withApiKey: true, retry: 1, repeat: 0, delayBeforeRetry: false},
        axiosConfig._options,
      )

      // 2. merge configuration
      GeneralUtils.mergeDeep(
        extractConfigurationForAxios({
          configuration: this.configuration,
          logger,
          withApiKey: options.withApiKey,
          // or ...options
        }),
        axiosConfig,
      )

      // 3. configure headers
      configRequestHeaders({
        axiosConfig,
        requestId: options.requestId,
        isLongRequest: options.isLongRequest,
        isPollingRequest: options.isPollingRequest,

        // or ...options
      })

      // 4. prepare
      await prepareRequest({
        axiosConfig,
        logger,
        name: options.name,
        requestId: options.requestId,
        delay: options.delay,
        repeat: options.repeat,

        // or ...options
      })

      // 5. return config
      return axiosConfig
    })
```

packages/eyes-sdk-core/lib/server/requestHelpers.js
```diff
@@ -154,35 +129,42 @@ const prepareRequest = async (config, {defaults, configuration, logger}) => {
   )
 
   if (options.delay) {
-    const delay = TypeUtils.isIterator(options.delay) ? options.delay.next().value : options.delay
+    const delay = TypeUtils.isArray(options.delay)
+      ? options.delay[Math.min(options.repeat, options.delay.length - 1)]
```
amitzur: `options.repeat` is always a number, right?

kyryloonufriiev: yeah. It is always a number (0 by default). Of course, I'm not providing any kind of validation if the user passes on purpose invalid value because it is an internal thing user shouldn't know about.

packages/eyes-sdk-core/lib/server/requestHelpers.js
```diff
@@ -0,0 +1,231 @@
...
+async function prepareRequest({axiosConfig, logger}) {
+  const options = axiosConfig._options
+
+  logger.verbose(
+    `ServerConnector.${options.name} [${options.requestId}] will now call to ${
+      axiosConfig.url
+    } with params ${JSON.stringify(axiosConfig.params)}`,
+  )
+
+  if (options.delay) {
```
amitzur: I have a feeling that it's cleaner if `delay` is calculated outside, and this function only performs the delay. WDYT?

packages/eyes-sdk-core/lib/server/requestHelpers.js
```diff
@@ -0,0 +1,231 @@
...
+    default: {
+      throw new Error(`Unknown error during long request: ${JSON.stringify(response)}`)
+    }
+  }
+}
+async function handleRequestResponse({response, axios, logger}) {
+  const options = response.config._options
+
+  logger.verbose(
+    `ServerConnector.${options.name} [${options.requestId}] - result ${response.statusText}, status code ${response.status}, url ${response.config.url}`,
```
amitzur: logs with `ServerConnector` in this file should be changed to something like `axios response interceptor - ${name}...`

packages/eyes-sdk-core/lib/server/requestHelpers.js
```diff
@@ -0,0 +1,231 @@
...
+
+  if (options.isPollingRequest && response.status === HTTP_STATUS_CODES.OK) {
+    options.repeat += 1
+    return axios.request(response.config)
+  }
+
+  return response
+}
+
+async function handleRequestError({err, axios, logger}) {
```
amitzur: we need a test also for the error case

packages/eyes-sdk-core/lib/server/ServerConnector.js
```diff
@@ -237,50 +81,53 @@ class ServerConnector {
...
+          withApiKey: true,
+          retry: 1,
+          repeat: 0,
+          delayBeforeRetry: false,
+        },
+        axiosConfig._options,
+      )
 
-    options.maxContentLength = 20 * 1024 * 1024
+      axiosConfig._options = options
```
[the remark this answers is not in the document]
kyryloonufriiev: I have some thoughts about it. Maybe will be even better to provide our custom options plain as other axios options, this will let us possibility to define default state for them just the same as we doing for other options, and reduce step with `Object.assign`

packages/eyes-sdk-core/lib/server/requestHelpers.js
kyryloonufriiev: I will do it. I would like to rework how work `delayBeforeRetry` as well to pass not the boolean but duration. Is it make sense? I think it will be more clear and testable.

packages/eyes-sdk-core/lib/server/ServerConnector.js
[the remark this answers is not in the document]
amitzur: If it doesn't interfere with axios then yeah, great

packages/eyes-sdk-core/lib/server/requestHelpers.js
amitzur: sure, makes sense

packages/eyes-sdk-core/lib/server/requestHelpers.js
kyryloonufriiev: I add a test that ensures `handleRequestError` retry the requests.

packages/eyes-sdk-core/lib/server/requestHelpers.js
kyryloonufriiev: Change log titles in new commit

packages/eyes-sdk-core/lib/server/ServerConnector.js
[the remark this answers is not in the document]
kyryloonufriiev: You can see the changes in the latest commit. It works a little differently from how I thought, but I think to remove nested `_options` is the right decision.

packages/eyes-sdk-core/lib/server/requestHelpers.js
```diff
@@ -131,13 +120,14 @@ async function handleLongRequestResponse({response, axios}) {
...
     case HTTP_STATUS_CODES.ACCEPTED: {
-      const options = response.config._options
+      const prevConfig = response.config
       const config = {
-        _options: {
-          name: options.name,
-          isPollingRequest: true,
-          delay: options.pollingDelaySequence || POLLING_DELAY_SEQUENCE,
-        },
+        name: prevConfig.name,
```
amitzur: What about all the other properties from prevConfig? like `withApiKey, headers, timeout, responseType` etc. ?
Don't they need to be passed here as well?

packages/eyes-sdk-core/lib/server/ServerConnector.js
```diff
@@ -82,35 +88,40 @@ class ServerConnector {
     this._renderingInfo = undefined
 
     this._axios = axios.create({
+      withApiKey: true,
+      retry: 1,
+      repeat: 0,
+      delayBeforePolling: POLLING_DELAY_SEQUENCE,
+      get requestId() {
```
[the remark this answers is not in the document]
amitzur: Another reasoning is that in the (near?) future I want to add information to the log that the current request is actually a retry of a previous one, and include the previous request ID. So that I can understand what requests are part of the same API call. Right now I can't know from the logs, and I think it would be possible if we read `requestId` from the prevConfig as a separate thing from calling `createRequestId`

packages/eyes-sdk-core/lib/server/requestHelpers.js
```diff
@@ -169,53 +157,57 @@ async function handleLongRequestResponse({response, axios}) {
...
       REQUEST_FAILED_CODES.includes(err.code))
   ) {
-    logger.verbose(`ServerConnector retrying request with delay ${options.delayBeforeRetry}...`)
+    logger.verbose(
+      `axios error interceptor retrying request with delay ${config.delayBeforeRetry}...`,
+    )
 
-    if (options.delayBeforeRetry) {
-      options.delay = RETRY_REQUEST_INTERVAL
+    if (config.delayBeforeRetry) {
```
amitzur: as I wrote earlier, I eventually want to remove this. But not now :)

packages/eyes-sdk-core/lib/server/ServerConnector.js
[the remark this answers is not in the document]
kyryloonufriiev: It is actually calling only once when you merge config with default, but maybe it's really not so transparent as it could be. I will fix it.

packages/eyes-sdk-core/lib/server/requestHelpers.js
kyryloonufriiev: We don't need to specify them here because actually we are not duplicating our previous request we are sending the new one. So these values will be set by default.
Also, this logic copying the previous implementation, there these values were set by default.

packages/eyes-sdk-core/lib/server/ServerConnector.js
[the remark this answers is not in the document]
kyryloonufriiev: And about logging, requests are inheriting `requestId` in case of retry. I realize that it is wrong, and fix it. Also, I add logs for request retry and keep the original request id in field `originalRequestId`.

packages/eyes-sdk-core/lib/server/ServerConnector.js
```diff
@@ -82,35 +88,40 @@ class ServerConnector {
...
       headers: DEFAULT_HEADERS,
       timeout: DEFAULT_TIMEOUT_MS,
       responseType: 'json',
       maxContentLength: 20 * 1024 * 1024, // 20 MB
     })
 
-    this._axios.interceptors.request.use(axiosConfig => {
-      const options = Object.assign(
+    this._axios.interceptors.request.use(config => {
+      const axiosConfig = Object.assign(
```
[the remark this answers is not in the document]
kyryloonufriiev: A very good idea, already done.

packages/eyes-sdk-core/lib/server/ServerConnector.js
```diff
@@ -237,50 +87,58 @@ class ServerConnector {
...
+      })
+      configAxiosFromConfiguration({
+        axiosConfig,
+        withApiKey: axiosConfig.withApiKey,
+        configuration: this._configuration,
+        logger: this._logger,
+      })
 
-    return options
+      return prepareRequest({
```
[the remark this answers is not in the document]
kyryloonufriiev: Make sense for me

amitzur:
Verified manually that the all tests pass

PR #112, opened 2020-02-06
Getting branch info from server & scm

danielschwartz85:
Need more tests..


packages/eyes-sdk-core/lib/EyesBase.js
```diff
@@ -180,7 +180,53 @@ class EyesBase extends EyesAbstract {
...
+    const batchId = this.getUserSetBatchId()
+    let branchName = this._configuration.getBranchName()
+    let parentBranchName = this._configuration.getParentBranchName()
+
+    const isLocalBranchTest = branchName && parentBranchName && branchName !== parentBranchName
+    const isCiBranchTest = !isLocalBranchTest && batchId
+
+    let parentBranchBaselineSavedBefore, err
+    if (isCiBranchTest) {
+      ;[err, {branchName, parentBranchName, parentBranchBaselineSavedBefore}] = await presult(
```
danielschwartz85: From what I understand this: ```parentBranchBaselineSavedBefore``` is currently not returned from the server,
If it will (github will return it one day or something..) then we can remove this code all together, no ?
- so thinking to not try get ```parentBranchBaselineSavedBefore``` from server..wdyt ?

Also once ```parentBranchBaselineSavedBefore``` is out, why go to the server at all and not always check git ? don't I have this info in git ? (this info: ```branchName, parentBranchName, parentBranchBaselineSavedBefore```)
so maybe we can remove server stuff ??

packages/eyes-sdk-core/lib/EyesBase.js
```diff
@@ -945,6 +991,7 @@ class EyesBase extends EyesAbstract {
       if (!this._renderingInfoPromise) {
         this._renderingInfoPromise = this.getAndSaveRenderingInfo()
       }
+      this._batchInfoPromise = this.getAndSetBatchInfo()
```
danielschwartz85: Any thoughts about multiple ```open``` calls on the same eyes instance ? 
I.e. we get info ( per batchId x branchName x parentBranchName ) we wait on this promise on match as far as I see, but do we wait on match in close() ? I need to check this...

packages/eyes-sdk-core/lib/getScmInfo.js
```diff
@@ -0,0 +1,21 @@
+'use strict'
+
+const {
+  GeneralUtils: {pexec},
+} = require('@applitools/eyes-common')
+
+const getScmInfo = (function() {
```
amitzur: 1. why not call this `makeGetScmInfo` like `makeGetBatchInfo` ?
2. Also like I wrote for `getBatchInfo`, this can just be a utility function that does the caching.
3. The cache should hold reference to the promise, not to the value itself.
4. This function has weird API. Its input is `parentBranchName`. That's also how it should be cached. If you want to have a map from batchKey (batchId?) to the result, better do it outside. Batch information has nothing to do with this function.
5. what are _opts? the `_` is a convention for unused variables. Based on the usage here, I'd call it `execOptions`.

packages/eyes-sdk-core/lib/server/SessionStartInfo.js
```diff
@@ -180,6 +183,14 @@ class SessionStartInfo {
     return this._parentBranchName
   }
 
+  // noinspection JSUnusedGlobalSymbols
```
amitzur: no need for this line. It was to silence warnings that Oleh had in WebStorm. We can delete those wherever we want.

packages/eyes-selenium/lib/runner/ClassicRunner.js
```diff
@@ -30,11 +36,25 @@ class ClassicRunner extends EyesRunner {
...
+      return this._getBatchInfo(batchId)
+    } else {
+      throw new Error(
+        'Eyes runner could not get batch info since makeGetBatchInfo was not called before',
+      )
+    }
+  }
+
+  async getScmInfoWithCache(...args) {
+    return getScmInfo(...args)
```
amitzur: I think `getScmInfo` the way you wrote it returns a function (it's actually the `make...` function)

packages/eyes-sdk-core/lib/EyesBase.js
amitzur: Doesn't the cache live on the runner? Maybe I missed something

packages/eyes-selenium/lib/EyesSelenium.js
```diff
@@ -1274,6 +1275,17 @@ class EyesSelenium extends Eyes {
     const renderingInfo = await this._runner.getRenderingInfoWithCache()
     this._serverConnector.setRenderingInfo(renderingInfo)
   }
+
+  async _getAndSaveBatchInfoFromServer(batchId) {
```
amitzur: 1. I understand this is implemented in `EyesSelenium` and not `EyesBase` because of the runner, right? That was what happened to me with `getAndSaveRenderingInfo`. Then this needs to be implemented in WDIO 4+5 as well.

2. Actually, probably this should be implemented in `EyesRunner` and not `ClassicRunner`, since it's also for tests that are related to VG.

3. No need to call this `getAndSave...` because you're not saving anything (it was copied from `getAndSaveRenderingInfo`, and there it **is** saved to the server connector).

packages/eyes-selenium/lib/EyesSelenium.js
```diff
@@ -1274,6 +1275,17 @@ class EyesSelenium extends Eyes {
...
+    ArgumentGuard.notNullOrEmpty(batchId, 'batchId')
+    return this._runner.getBatchInfoWithCache(batchId)
+  }
+
+  async _getAndSaveScmMergeBaseTime(batchKey, parentBranchName) {
```
amitzur: 1. Also here, why call it `getAndSave...` ?
2. see how the use of batchKey isn't needed, and makes the API wrong also here? You pass it here just for the mapping, that can be done at a higher level.

packages/eyes-sdk-core/lib/EyesBase.js
danielschwartz85: Yes but eyes instance get's the data per batchId here : https://github.com/applitools/eyes.sdk.javascript1/pull/112/files#diff-5f08a678414bf530975114d4a703816eR195

so first eyes open can set the promise to batchId1 promise (from runner cache) and then another eyes open can set it batchId2 promise..(also from cache)

packages/eyes-sdk-core/lib/getScmInfo.js
danielschwartz85: 1. since it doesn't take any function (i.e. using the make here would be just for closure reasons)
2. makes sense
3. why cache the promise and not the value ?
4. assuming ```parentBranchName``` doesn't change in one node process..
5. for test usage only (not client api)

packages/eyes-selenium/lib/EyesSelenium.js
danielschwartz85: 1. yes. this should be implemented in WDIO 4/5  but note that if not implemented then it would not break..
2. ho.. I see..
3. I see.. I kept the name since we save it to cache etc (o.w EyesBase can basiccaly do this on it's own but it delegates this responsibility to implimentors since they know  in what scope to save it)

packages/eyes-selenium/lib/EyesSelenium.js
danielschwartz85: 2 - it's for caching - but now that I think about it.. maybe I can get rid of this per batch id caching..and just cache globally (i.e. if value exists) ; since I'm assuming branch name and parent branch name don't change per node process. rigth ?

packages/eyes-selenium/lib/runner/ClassicRunner.js
danielschwartz85: don't see this.. (iife returns/exports the getScmInfo func)

packages/eyes-selenium/lib/EyesSelenium.js
danielschwartz85: also needs impl in VGC & Testcafe..

PR #113, opened 2020-02-10
fix subregion region

kyryloonufriiev:
Fix a mistake.
Related to this  card https://trello.com/c/m7vRDQyP/207-js4-ignoreregions-not-working


amitzur:
@kyryloonufriiev can you add a test for this? Not sure about the right way to do that

PR #115, opened 2020-02-17
Clone configuration before set to the Eyes

kyryloonufriiev:
I add a call of `configuration.cloneConfig()` before setting it to the Eyes instance and test to ensure it works right.

Related to this card https://trello.com/c/d3rahmUd/220-setconfiguration-does-not-clone-the-configuration


amitzur:
This is a very desired change. However I'm afraid of regressions. So let's wait until @danielschwartz85 makes the build green and only then merge it

amitzur:
@kyryloonufriiev also here I want to consolidate implementations into one place. Right now `setConfiguration` and `getConfiguration` are implemented in 7 different places, part of them with bugs. Let's move it from `EyesAbstract` into `EyesBase` (so there's access to serverConnector) and delete all the other implementations (wdio4+5 EyesWDIO and EyesVisualGrid; eyes-testcafe; eyes-selenium)

kyryloonufriiev:
I didn't know that this function is overwritten in SDK Eyes implementations. I moved it to the EyesBase

packages/eyes-sdk-core/lib/EyesBase.js
```diff
@@ -135,6 +135,14 @@ class EyesBase extends EyesAbstract {
     this._userInputs = []
   }
 
+  /**
+   * @param {Configuration|object} configuration
+   */
+  setConfiguration(configuration) {
+    super.setConfiguration(configuration)
```
amitzur: No reason to leave the actual implementation in `EyesAbstract`. It's an artificial separation and eventually I want to delete `EyesAbstract`. The only class inheriting from it is `EyesBase`.
Another thing - I want to move away from `instanceOf` - it's creating a problem for us when users have multiple copies of node_modules. Let's use `constructor.name`. So the if here should be:
```
if (!(configuration && configuration.constructor.name === 'Configuration)) {...}

// or the de-morgan equivalent
if (!configuration || configuration.constructor.name !== 'Configuration') {...}
```

PR #116, opened 2020-02-20
Add fetch time limit

kyryloonufriiev:
Add time-limit for fetches to the `captureFrame`. I used `AbortController` to abort the request and combinate it with `Promise.race()` in case request will not be aborted (this bug exists in Safari browser)

### DIFF
--- a/packages/dom-capture/src/browser/captureFrame.js
+++ b/packages/dom-capture/src/browser/captureFrame.js
@@ -18,6 +18,7 @@ async function captureFrame(
   {styleProps, rectProps, ignoredTagNames} = defaultDomProps,
   doc = document,
   addStats = false,
+  fetchTimeLimit = 30000,
 ) {
   const performance = {total: {}, prefetchCss: {}, doCaptureFrame: {}, waitForImages: {}};
   function startTime(obj) {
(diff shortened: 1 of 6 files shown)


amitzur:
@kyryloonufriiev if the code works, then how come the fixture `longCss.dom.json` has the contents of `test.css` ?

kyryloonufriiev:
@amitzur I have a look at this test. It was really wrong. It happened because we have `cache: 'force-cache'` set on the fetch options. This makes it use a cached version of css file from another test. And another thing I realized, when we have long loaded css the `fetchCss` even not calling before the page is ready (all resources are loaded). So I decide to use `rel="prefetch"` for the test. This makes me add support for this kind of link to the `isLinkToStyleSheet` function.

I add the commit to this branch. https://github.com/applitools/eyes.sdk.javascript1/commit/21a84a42eddb6cbe4391f2087a472b781e9899c8

amitzur:
@kyryloonufriiev I cherry picked your commit. I'm glad we got support for prefetch as style now :)
But I didn't understand why `fetchCss` isn't called before the page is loaded. And I also didn't understand why cache busting with query param didn't work, and you needed to use a different port.

kyryloonufriiev:
@amitzur Oh, my bad, I didn't thought about just add query param to the css string to avoid cache.
About the another problem, I think our evaluated code doesn't call because default behaviour for links with `rel="stylesheet"` is blocking the call stack somehow.

amitzur:
ah, I get it. I did try changing it to query param but it didn't work. I thought you might have tried it and know.
I understand now that it's blocking the loading of the page. Maybe there's a way to make the request delayed in the test server only for ajax calls. Some http header or something.

kyryloonufriiev:
If I delay this only for xhr requests it will not work any way. Because `fetch` will get resource from cache.

PR #117, opened 2020-02-23
Runners refactoring

kyryloonufriiev:
In this PR I want to change the implementation of `VisualGridRunner`, to avoid some bugs related to the current implementation.
The first thing I do is attach `vgClient` creation to the `VisualGridRunner` with caching to be able to reuse `vgClient` if all of the inputs are the same. This makes me refactor the logic of `GeneralUtils.cachify` to avoid collisions.
The other thing I do is consolidate the logic of `.getAllTestResults()` to the `EyesRunner` and make it the same as `ClassicRunner` have.


kyryloonufriiev:
I have a problem with webdriverio 4 and 5 because it doesn't use `eyes-sdk-core` from the current branch but use own version from `node-modules`.

amitzur:
It's hard to understand in this PR what is the addition on top of #114 .
I'd like to land #114 as soon as possible, as it fixes 3 different issues.
Let's postpone these changes until after that.

kyryloonufriiev:
#114 is working well and fix all issues, but have the same problem with WDIO 4 and 5.
This PR adds two things:
1. Attach `vgClient` to the `VisualGridRunner`
2. Change the way how `VisualGridRunner` collect test results. This makes us change the `.close()` method of `EyesVisualGrid` in all related SDKs.

amitzur:
@danielschwartz85 please review

packages/eyes-selenium/lib/EyesVisualGrid.js
```diff
@@ -109,90 +110,42 @@ class EyesVisualGrid extends Eyes {
...
-    } finally {
-      this._isOpen = false
-      this._closePromise = undefined
-    }
-  }
-
-  /**
-   * @return {Promise}
-   */
-  async closeAsync() {
```
danielschwartz85: I think this is a public API and cannot be removed..
https://applitools.com/docs/api/eyes-sdk/classes-gen/class_eyes/method-eyes-abortasync-selenium4-javascript.html

kyryloonufriiev: This isn't removed. It uses implementation from the base class `EyesVisualGrid` doesn't need to override it anymore.

packages/eyes-webdriverio-4/src/EyesVisualGrid.js
```diff
@@ -172,79 +163,49 @@ class EyesVisualGrid extends EyesBase {
...
-        }
-      }
-
-      return testResultSummary
-    } finally {
-      this._isOpen = false
-      this._closePromise = undefined
-    }
+  async closeAsync() {
+    await this.close(false)
```
danielschwartz85: shouldn't we remove ```await``` ? (i.e. for the function being "async"; even though users don't have to wait on the returned promise but I think the function api is returning without waiting..)

packages/eyes-webdriverio-4/src/EyesVisualGrid.js
```diff
@@ -172,79 +163,49 @@ class EyesVisualGrid extends EyesBase {
...
   }
 
   /**
    * @return {Promise}
    */
-  async closeAsync() {
-    if (!this._closePromise) {
-      this._closePromise = this._closeCommand()
-    }
+  async abortAsync() {
```
danielschwartz85: same here, with "await"

packages/eyes-webdriverio-4/src/EyesVisualGrid.js
kyryloonufriiev: As far as I understand `closeAsync` method exists only because of compatibility with other SDKs (e.g Java). So in JS this method should be the same as `close` but without a return value. So we need to await the `close` method inside `closeAsync` to be sure that it will be resolved only when `close` is resolved.

packages/eyes-common/lib/utils/GeneralUtils.js
```diff
@@ -450,10 +450,14 @@ class GeneralUtils {
...
       const cache = {}
-      return function(arg1AndKey, ...args) {
-        const cacheKey = (!cacheRegardlessOfArgs && arg1AndKey) || 'default'
-        if (!cache[cacheKey]) {
-          cache[cacheKey] = getterFunction(arg1AndKey, ...args)
+      return function(...args) {
+        let cacheKey = 'default'
+        if (!cacheRegardlessOfArgs) {
+          const [key] = args
+          cacheKey = GeneralUtils.stringify(key)
```
danielschwartz85: 👍

packages/eyes-selenium/lib/EyesVisualGrid.js
danielschwartz85: I see 👍

packages/eyes-webdriverio-4/src/EyesVisualGrid.js
danielschwartz85: I see, cool.

PR #119, opened 2020-02-25
support future long running tasks

kyryloonufriiev:
This PR is changing the way how `ServerConnector` handle long requests related to changes in Eyes server version 10.9

https://trello.com/c/60Rm4xXG/240-support-future-long-running-tasks

### DIFF
--- a/packages/eyes-sdk-core/lib/server/requestHelpers.js
+++ b/packages/eyes-sdk-core/lib/server/requestHelpers.js
@@ -98,6 +98,7 @@ function configAxiosHeaders({axiosConfig}) {
       axiosConfig.timestamp,
     )
   }
+  // ---
 }
 
 async function delayRequest({axiosConfig, logger}) {
@@ -154,7 +155,7 @@ async function startPollingRequest({url, config, axios}) {
       const {config} = response
       const nextConfig = {
         name: config.name,
-        method: 'DELETE',
+        method: 'DELETE', // TODO should be changed to GET when Eyes server will be updated to 10.9
         url: response.headers.location,
         headers: {
           [CUSTOM_HEADER_NAMES.EYES_DATE]: DateTimeUtils.toRfc1123DateTime(),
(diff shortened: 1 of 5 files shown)


amitzur:
@tourdedave and @danielschwartz85 - please review this PR as well as the trello card. It's a bit long read, but it's a small code change and would be good for you to be acquainted with the protocol. Add your review to this PR.
@kyryloonufriiev - once this is approved, merge to master (make sure tests pass in at least eyes-sdk-core and eyes-selenium, and add a changelog entry for eyes-sdk-core).

Theoretically we should release all SDK's as a result, but there's no rush and I think we can batch it with other changes next week. WDYT (everyone)?

packages/eyes-sdk-core/lib/server/requestHelpers.js
```diff
@@ -108,70 +110,64 @@ async function delayRequest({axiosConfig, logger}) {
...
+    delay: TypeUtils.isArray(config.delayBeforePolling)
+      ? config.delayBeforePolling[0]
+      : config.delayBeforePolling,
+    method: 'GET',
+    url,
+  }
+  const response = await axios.request(pollingConfig)
+  switch (response.status) {
+    case HTTP_STATUS_CODES.OK:
+      return response
```
tourdedave: Based on what I read in the protocol specified in the Trello ticket -- if we're polling and the response is `200` we should make another call to continue polling. But this looks like we're just returning the response. Am I missing something?

kyryloonufriiev: I think you missed the place where this logic is handled, look at `handleRequestResponce` line 124. Line 150 will not be resolved until polling is finished (return non-200).

tourdedave: Ah, I see. This function is called from the `axios` receptor. That makes more sense.

packages/eyes-sdk-core/test/unit/server/ServerConnector.spec.js
```diff
@@ -103,6 +103,7 @@ describe('ServerConnector', () => {
       const response = {status: 200, config, data: {}, headers: {}, request: {}}
       if (config.isLongRequest) {
```
amitzur: I'd like to see a test that shows that the client initiates a request without the `isLongRequest` flag, the server responds with 202, and then we assert that long polling took place.

kyryloonufriiev: I add a specific test to check all nuances of polling protocol. Can I merge it?

PR #120, opened 2020-02-25
upload domsnapshot directly to Azure

kyryloonufriiev:
This PR changes the logic of `ServerConnector::postDomSnapshot`. The DOM will be uploaded to the `resultsUrl` webhook from rendering info.

Trello card says that it should be implemented in `eyes-sdk-core-legacy` as well, but I couldn't find any related logic there.

[trello] https://trello.com/c/ZCLJo8Fy/241-upload-dom-directly-to-azure


packages/eyes-sdk-core/test/unit/server/ServerConnector.spec.js
```diff
@@ -93,6 +93,16 @@ describe('ServerConnector', () => {
     assert.strictEqual(result, renderingInfo.getResultsUrl().replace('__random__', id))
   })
 
+  it('postDomSnapshot works', async () => {
```
amitzur:
```suggestion
  it('postDomSnapshot uploads to resultsUrl webhook', async () => {
```

PR #122, opened 2020-03-02
Fix issue with scaling on mobile browsers during FPS

kyryloonufriiev:
This PR fixes the issue founded by @Artem-applitools. The issue is happening during FPS on mobile browsers (created by chrome mobile emulator).
Since mobile browsers are using `viewport` to present a page this means that on the screenshot with width 382px can be 980px of information. To fix the issue I'm doing an additional check for User-Agent to detect mobile browsers.
I'm sure that regex can be improved and this feature can be implemented in other SDKs (it's selenium4 only for now). Do I need to invest extra time in it? Or it could be solved in a better way?

Test results to demonstrate wrong behavior: https://eyes.applitools.com/app/test-results/00000251819153792597/?accountId=UAujt6tHnEKUivQXIz7G6A~~

### DIFF
--- a/packages/eyes-selenium/lib/EyesSelenium.js
+++ b/packages/eyes-selenium/lib/EyesSelenium.js
@@ -783,13 +783,13 @@ class EyesSelenium extends Eyes {
   async _getScaleProviderFactory() {
     const element = await this._driver.findElement(By.css('html'))
     const entireSize = await EyesSeleniumUtils.getEntireElementSize(this._jsExecutor, element)
-
+    const isMobile = await this._driver.isMobile() || await this._driver.isMobileBrowser()
     return new ContextBasedScaleProviderFactory(
       this._logger,
       entireSize,
       this._viewportSizeHandler.get(),
       this._devicePixelRatio,
-      false,
+      isMobile,
       this._scaleProviderHandler,
     )
   }
--- a/packages/eyes-selenium/lib/EyesSeleniumUtils.js
+++ b/packages/eyes-selenium/lib/EyesSeleniumUtils.js
@@ -196,6 +196,22 @@ class EyesSeleniumUtils extends EyesJsBrowserUtils {
     const browserName = capabilities.get('browserName')
     return platformName && ['ANDROID', 'IOS'].includes(platformName.toUpperCase()) && !browserName
   }
+  /**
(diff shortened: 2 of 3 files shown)


amitzur:
@kyryloonufriiev @Artem-applitools what tests does this fix? are those tests in #157 ?
Kyrylo, do you want to do these changes as part of #167 ? Or do you want to merge this PR standalone?

Artem-applitools:
@amitzur Yeap, it's related to the tests https://github.com/applitools/eyes.sdk.javascript1/blob/master/packages/eyes-selenium/test/coverage/custom/TestMobileDevices.spec.js
They are currently skipped.

packages/eyes-selenium/lib/EyesSelenium.js
```diff
@@ -783,13 +783,13 @@ class EyesSelenium extends Eyes {
   async _getScaleProviderFactory() {
     const element = await this._driver.findElement(By.css('html'))
     const entireSize = await EyesSeleniumUtils.getEntireElementSize(this._jsExecutor, element)
-
+    const isMobile = await this._driver.isMobile() || await this._driver.isMobileBrowser()
```
amitzur, line 786: Can this be `EyesSeleniumUtils.isMobileDevice()` and then we don't need the utility with the user agent?

amitzur: I just checked this and seems to work fine. Driver caches the `capabilities` so it's the same optimization of not going twice to the browser.
Adding this fix should make tests in #157 pass, right

kyryloonufriiev: When I worked on this issue I tried to use `isMobileDevice` first. But it didn't work for me since the emulator is not a mobile device, but just desktop chrome.

amitzur:
@kyryloonufriiev is this implemented in the new core? can we close this PR?

amitzur:
@kyryloonufriiev how do you want to move forward with this PR?

kyryloonufriiev:
@amitzur I fixed this issue with https://github.com/applitools/eyes.sdk.javascript1/commit/5a15a038d8ae13b7145458a768bdd479f341c45b.
I didn't enable tests because they are still failing, because of the image metrics, but the images look much better.
So we can close this PR.

PR #128, opened 2020-03-10
eyes-storybook: add viewportSize property to set puppeteer page viewport

kyryloonufriiev:
This PR provides the logic to set the viewport size of a puppeteer page. The problem I faced is the inability to test function `initPage()`, which should apply the configuration to the page. I have a suggestion to move this function to the separate file and make it more testable. WDYT?


amitzur:
I merged this locally and pushed. Also added a commit with e2e test, default value of 1024x768 and a readme entry for the new parameter

PR #129, opened 2020-03-10
eyes-storybook: expand configurations

kyryloonufriiev:
This PR provides the possibility to add story level `properties`.


kyryloonufriiev:
Merge this PR with other storybook tasks related to the global and story-level configuration.
The complete list of covered issues is:

1. https://trello.com/c/IfkEZ4V3/45-storybook-add-set-get-ignoredisplacements-to-eyes-global-and-fluent?menu=filter&filter=*
2. https://trello.com/c/NNko9uQr/200-storybook-add-capability-to-add-layout-strict-content-floating-regions-via-the-config?menu=filter&filter=*
3. https://trello.com/c/yWbAZ2Fm/170-storybook-sdk-custom-properties-support

amitzur:
@kyryloonufriiev can you review my latest commit?
I don't want the local config to concat to global config, because that way it's not possible to override the global config, and also creates confusion.

I also refactored the `renderStory` tests so that each test is more concentrated on a single case.

kyryloonufriiev:
I totally agree about the regions. I didn't think about the overriding problem. But about `ignoreDisplacements` my idea was that we already pass global config to the `makeRenderingGridClient` so if we didn't specify this on the story level config, so it will use the value we pass in `makeRenderingGridClient`. Same for properties.

amitzur:
@kyryloonufriiev absolutely right. I'll modify the code around `ignoreDisplacements` and `properties`

amitzur:
@kyryloonufriiev done, including some testing goodies - I added ability to inspect the input for Eyes through the fake eyes server. Plus, ability to control the selector regions, and options sent for each step.

I'm just waiting for product's answer on this:
https://trello.com/c/NNko9uQr/200-storybook-add-capability-to-add-layout-strict-content-floating-regions-via-the-config#comment-5e80f8081d97b92b67c39d65

Let me know what you think about the changes.

kyryloonufriiev:
@amitzur I think your chances are great, now it much better covers all cases with global and local regions config. And I really like your suggestion about naming improvement. But I think if we replace `ignore` with `ignoreRegions` then the new naming will be counter-intuitive in case of providing a single region. But I think it is not a big deal, users still can provide single region as an array.

amitzur:
Isn't it great that we're able to inspect the input to the server for open and check ?!

kyryloonufriiev:
Yeah. It's what I mean in my comment before

Issue #127, opened 2020-03-09
Installing @applitools/eyes-selenium leads to duplicated @applitools/eyes-common


ZauberNerd:
Installing `@applitools/eyes-selenium@4.33.24` leads to `@applitools/eyes-common` being installed in multiple version.

```
 /tmp/foo $ yarn list --pattern "applitools"
yarn list v1.22.0
├─ @applitools/dom-capture@7.1.3
├─ @applitools/dom-snapshot@3.3.3
├─ @applitools/dom-utils@4.7.7
│  └─ @applitools/eyes-common@3.19.0
├─ @applitools/eyes-common@3.20.0
├─ @applitools/eyes-sdk-core@8.1.1
├─ @applitools/eyes-selenium@4.33.24
│  └─ @applitools/eyes-common@3.20.1
├─ @applitools/functional-commons@1.5.5
├─ @applitools/http-commons@2.3.12
├─ @applitools/isomorphic-fetch@3.0.0
├─ @applitools/jsdom@1.0.2
├─ @applitools/monitoring-commons@1.0.19
└─ @applitools/visual-grid-client@13.6.0
   └─ @applitools/functional-commons@1.5.4
Done in 0.16s.
```

Which leads to the following error when executing tests:

```
[2020-03-08T23:33:51.370Z]      Error: IllegalType: x is not a number
[2020-03-08T23:33:51.370Z]       at Function.isNumber (/app/node_modules/@applitools/eyes-selenium/node_modules/@applitools/eyes-common/lib/utils/ArgumentGuard.js:158:13)
[2020-03-08T23:33:51.370Z]       at new Location (/app/node_modules/@applitools/eyes-selenium/node_modules/@applitools/eyes-common/lib/geometry/Location.js:39:19)
[2020-03-08T23:33:51.370Z]       at EyesWebDriverScreenshot.convertLocation (/app/node_modules/@applitools/eyes-selenium/lib/capture/EyesWebDriverScreenshot.js:431:18)
[2020-03-08T23:33:51.370Z]       at EyesWebDriverScreenshot.convertRegionLocation (/app/node_modules/@applitools/eyes-sdk-core/lib/capture/EyesScreenshot.js:95:34)
[2020-03-08T23:33:51.370Z]       at EyesWebDriverScreenshot.getIntersectedRegion (/app/node_modules/@applitools/eyes-selenium/lib/capture/EyesWebDriverScreenshot.js:605:30)
[...]

PR #131, opened 2020-03-12
WIP - Monolith

tourdedave:
tl;dr -- Our mono wants to be a monolith. So let's help give it what it wants.

## Motivation

There are 3 types of "packages" we have - internal, customer facing, and hybrid.

Right now there is an implicit need for customer facing SDK to having a unique version of internal packages. That and given the accidental complexity of how all packages work together, there is too much burden put either onto the developer or the build tooling (which is still a WIP) in order to publish packages.

As a result, we run the risk of shipping packages with inconsistent versions that will cause a break for the customer (just like [this](https://github.com/applitools/eyes.sdk.javascript1/issues/127)). Also, given that internal packages are only consumed by customer facing packages within the monorepo - there's no need for them to be packages published to a registry.

## Direction

This PR will be used to test out a new structure for the monorepo that will be part monolith, part modular packages.

We will move the internal packages into a `/src` directory in the root of the mono and update the customer facing SDK dependencies to use these with the `file:` option in their package.json.

We will also add a build step to each customer facing SDK that will copy the necessary code out of `/src` and into a build folder, creating a final package.json, and pack/unpack/verify that the SDK is loadable.

## Drawbacks

- Traceability for the internal packages

Since there are effectively no versions for internal packages with this approach, keeping track of which customer facing SDKs have received code changes from internal packages will need to be thought through.

- Blocking code changes

If a customer facing package isn't in a working state due to a code change in an internal package, then it will be difficult to ship any code for that customer facing package (e.g., bug fix, new feature, etc.) until the issue with the internal package is resolved. Worth a discussion to make sure we're all aware of this potential outcome and some ideas of how to address it.

[x] Need install a package again after making a change to an internal package

~~If we stick with `file:` to reference internal packages -- in order for customer facing packages to pick up a change made in an internal package, you need to run `yarn install --check-files` (or simply `yarn --check-files`) in the customer facing package.~~

~~Alternatively, we can look into functionality offered in yarn v2 to address this ([link](https://yarnpkg.com/features/protocols))~~

yarn v2 with the portal protocol 💯 

[x] Debugging with yarn v2 cannot be done with `ndb`

~~Normally you would rely on a globally installed version of `ndb` and then prefix it to a command you're about to run (e.g., `ndb yarn test:unit`). This errors in yarn v2 due to the implicit dependency. Adding the dep to the project doesn't resolve the issue either. It just leads to a more obscure error.~~

~~To debug instead, you need to run run yarn with a `--inspect` flag, and have it attach to an existing debug session running (e.g., `chrome://inspect` and click `Open dedicated DevTools for Node` in Chrome).~~

This was fixed by disabling PNP in Yarn.

## Remaining Items

- script to prepare build folder (e.g., create if not present, clean from last build, etc.)
- script to copy package-to-be-built items into build folder (e.g., package.json, mirror?)
- script to update package.json (e.g., update dep relative paths to point to build folder, create concatenated deps)
- script to pack and verify


tourdedave:
Seeing odd behavior with `file:` and `yarn install`.

It looks like the contents aren't symlinked, but installed once (seems reasonable in hindsight). The problem is running `yarn install` again to try and pick up changes made in `src` -- they don't show up in the package's `node_modules`.

This is without workspaces enabled and relying solely on `file:`. Curious to see what happens with workspaces turned back on.

amitzur:
I don't want `pnp`. I think it's a rabbit hole (if we weren't in one already). Is there a way to turn it off? maybe that's why ndb doesn't work

tourdedave:
@amitzur Yes, the `ndb` errors were coming from PNP. There is a way to turn it off in the `.yarnrc.yml` and turn `node_modules` back on. I've just done that.

I also dropped the zero install/cache. Curious your thoughts on that.

tourdedave:
Haven't fully wrapped my head around the setup we want with `portal:` and workspaces yet -- or what's fully possible.

Right now I've specified the customer facing packages in workspaces (seems to be a hard requirement for yarn -- unless I'm missing something). Since these packages reference internal packages in `src` through `portal:` yarn appears to try and install the dependencies for them. Based on my initial checks with this approach, customer facing SDKs seem to work.

However, when you `cd` into an internal package and try to run a script for it (e.g., `yarn test` in `release-kit`) yarn gives an error about not being listed in the workspace and having unmet dependencies. Running `yarn` in the internal package resolves this and adds a `yarn.lock` file and `.yarn` folder.

tourdedave:
NOTE: had to remove `eyes-leanft` from the workspaces due to an error on install.

```
Error: leanft.sdk.core@14.03.x isn't supported by any available resolver
```

tourdedave:
Closing for now since the need for this approach is unclear. That and there is still a fair bit to sort out as part of this PR if we wanted to land it. We can always re-open it if we want to.

PR #133, opened 2020-03-14
add refresh element to the EyesWebElement

kyryloonufriiev:
This PR tries to handle error `StaleElementReference` for WDIO4. During the investigation, I face issues with EyesWebElement:

1. This extends from WebElement but keep the WebElement object on the state and work with it as a wrapper.
2. getLocation method works wrong with elements inside frames because return location related to the page, not frame. 

This PR isn't ready. We still need to handle the case when we refresh an element is inside a frame but the driver is switched to the default content.


packages/eyes-webdriverio-4/src/EyesWDIO.js
```diff
@@ -1432,8 +1442,8 @@ class EyesWDIO extends EyesBase {
     const eyesRemoteWebElement = new EyesWebElement(this._logger, this._driver, element)
     return eyesRemoteWebElement
       .getBounds()
-      .then(bounds => {
-        const currentFrameOffset = originalFC.getCurrentFrameOffset()
+      .then(async bounds => {
+        const currentFrameOffset = await originalFC.getCurrentFrameOffset()
```
amitzur: @kyryloonufriiev why? this is not an async operation

packages/eyes-webdriverio-4/src/wrappers/EyesWebElement.js
```diff
@@ -305,47 +338,51 @@ class EyesWebElement extends WebElement {
...
   /**
    * @Override
    * @inheritDoc
    */
-  sendKeys(keysToSend) {
+  async sendKeys(keysToSend) {
     return super.sendKeys(keysToSend)
   }
 
+  async getRect() {
```
amitzur: why go in this direction instead of actually removing the methods that just call super?

packages/eyes-webdriverio-4/src/EyesWDIO.js
kyryloonufriiev: totally right. I have a problem, so add it. But it was because of my code. So I fix my code and didn't remove it here.

packages/eyes-webdriverio-4/src/wrappers/EyesWebElement.js
kyryloonufriiev: You are right I thought I need to do only the fix with refresh call.

amitzur:
@tourdedave @kyryloonufriiev I found the issue in `getLocation`. We are trying to address 2 use cases here:
1) Dave initially provided support for the user switching to frame without our wrapped driver. While it's questionable we should support this, in wdio it's hard to argue against since this forces the user to use selenium-webdriver API's. Since the case in question has only same-origin iframes, it's possible to solve by walking up the frames inside the browser and calculate absolute position.
I've added the test `TestCustom: check element when switch to frame directly using wdio browser` to verify we don't break that scenario.

2) Kyrylo added support for handling StaleElementReference, but the code for (1) breaks the tests he added: `TestRefreshableDom: refresh element inside iframe after StaleElementReference`. It has nothing to with stale elements, it's the use case of **using** the wrapped driver, and checking an element inside a frame.

My conclusion: we need to calculate location relative to the frame chain. So I'm thinking about calling `getAbsoluteElementLocation` when `frameChain.size() === 0`, and `getBoundingClientRect` otherwise.

WDYT?

amitzur:
@kyryloonufriiev @tourdedave well, that indeed solves both cases.
2 things to note:
1) Assuming that getLocation will refresh the element for both getLocation and getSize, so I'm not changing getSize to use something like executeScript (even though it doesn't throw on StaleElementReference)
2) I'm not consolidating getLocation and getSize into getRect because of the usage of `getAbsoluteElementLocation`. We need to revisit this, because current fix feels very fragile.

PR #134, opened 2020-03-19
wdio5: clear target element

kyryloonufriiev:


kyryloonufriiev:
@amitzur add custom coverage test to check if a region is cleared after check

amitzur:
This needs to be fixed in wdio4 as well. Plus the same test. We can do it once we figure the way to share custom coverage tests code between SDK's

PR #135, opened 2020-03-19
Capture viewport on native apps with Appium

tourdedave:
re: [Trello 229](https://trello.com/c/Ph3FndE1/229-wdio-5-appium-status-bar-showing-in-test-results)

Based on [the Appium docs](http://appium.io/docs/en/commands/mobile-command/), it's possible to capture a screenshot of a native app's viewport in a platform agnostic way through the JS executor.

This PR is an attempt to confirm this, and look into an approach that will work for WDIO5, and potentially other relevant JS SDKs which can be run on Appium (e.g., Selenium 3, Selenium 4, and WDIO4).

## TODO

- [x] Confirm on Android (only works on UIAutomator2)
- [x] Confirm on iOS
- ~~[ ] Determine way to make this configurable to the end-user (currently using an env variable)~~
- ~~[ ] Determine reasonable structure for all JS SDKs to leverage it w/o needless duplication~~ (already done thanks to the new core)


tourdedave:
Some things to consider:

- On by default w/ the option to disable (e.g., by env var)? Or vice-versa?

It's possible that on by default could break existing baselines for users. Need to check.

- Can we determine the Appium version?

Log a warning if not using a version that supports this functionality. Appium falls back to taking the device viewport with the status bars, but it would be helpful to have a log entry.

tourdedave:
@kyryloonufriiev Thoughts on the tests for this PR being turned into generic coverage tests instead? It looks like generic tests support alternate setups. It's just not clear how best to do it.

PR #136, opened 2020-03-20
Support refreshable DOM without the driver wrapper

tourdedave:


tourdedave:
This PR is a subset of behavior which is already addressed #151. Closing.

PR #137, opened 2020-03-23
Selenium support for typed selectors

kyryloonufriiev:
This PR is adding functionality which allows `VisualGridEyes` class to send typed selectors to the VG server. This makes the unification step as well. To make `VisualGridEyes` code match with more advanced implementation from `eyes-webdriverio-5`.


packages/eyes-sdk-core/lib/EyesJsBrowserUtils.js
```diff
@@ -161,7 +161,7 @@ class EyesJsBrowserUtils {
         if (frameElement) {
           xpath = genXpath(frameElement) + ',' + xpath;
         }
-        return xpath.replace(/\/$/, '');
+        return xpath.replace(/\\/$/, '');
```
danielschwartz85: :) had the same fix..

packages/eyes-selenium/lib/EyesVisualGrid.js
```diff
@@ -189,6 +186,11 @@ class EyesVisualGrid extends Eyes {
 
       const source = await this._driver.getCurrentUrl()
 
+      const [ignore, floating] = await Promise.all([
+        this._persistRegions(checkSettings.getIgnoreRegions()),
+        this._persistRegions(checkSettings.getFloatingRegions()),
```
danielschwartz85: In the current impl - I put _persistedRegions in toCheckSetings(eyesWebDriver):
https://github.com/applitools/eyes.sdk.javascript1/pull/140/files#diff-7e529d0671fecfeb92b241a8d9e15468R257
* u can use the new version (if we decide it's good to put it in core, not sure..)
* just make sure to implement "toPersistedRegions" in all the regions of eyes-selenium

packages/eyes-selenium/lib/fluent/AccessibilityRegionByElement.js
```diff
@@ -40,6 +44,17 @@ class AccessibilityRegionByElement extends GetAccessibilityRegion {
     })
     return [accessibilityRegion]
   }
+
+  async toPersistedRegions(driver) {
+    const xpath = await EyesJsBrowserUtils.getElementXpath(driver, this._element)
```
danielschwartz85: If this._element is EyesWebElement then I think this._element.element is needed (not sure, but if it was copied from WDIO5 then it was a bug there, and fixed in the latest PR)

kyryloonufriiev: This is a bug for wdio because `EyesWebElement` is wrapping wdio `WebElement JSON` and you have access to it using `element` getter. But in Selenium this is works because `EyesWebElement` is extending Selenium `WebElement`.

danielschwartz85:
Other than considering using the new toCheckConfiguraion() (that will be merged soon) looks good 👍

kyryloonufriiev:
I moved to use new `ChackSettings::toCheckConfiguraion`. As I understand now it should support all of this `ignore, floating, strict, layout, content, accessibility` properties. Am I right?

danielschwartz85:
yes 👍

danielschwartz85:
@kyryloonufriiev one more thing, you can remove ```getSizeMode``` from seleniumCheckSettings.

PR #138, opened 2020-03-23
Add verify-commits script

amitzur:
Added script:
```
bongo --verify-commits
```
This runs:
```
git log --oneline ${tagName}..HEAD -- ${pkgPath}

// for example
git log --oneline @applitools/eyes-selenium@4.33.29..HEAD -- /home/bettyboop/dev/eyes.sdk.javascript1/packages/eyes-selenium
```

If there are unreleased commits, then the process fails and the released is blocked.
If there is no tag, a warning is output in yellow and the release continues.

### DIFF
--- a/packages/dom-capture/package.json
+++ b/packages/dom-capture/package.json
@@ -7,6 +7,7 @@
     "vv": "bongo --verify-versions",
     "vc": "bongo --verify-changelog",
     "uc": "bongo --update-changelog",
+    "verify-commits": "bongo --verify-commits",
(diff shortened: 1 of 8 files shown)


amitzur:
@tourdedave if I get your approval I will add `verify-commits` to the rest of the packages

amitzur:
So true! I didn't even think about it, I just put it in the same place as `vv` and `vc`. Will fix

amitzur:
I call this - the traffic light
Screenshot from 2020-03-23 22-52-51

PR #139, opened 2020-03-23
Test kit: ability to filter by name with glob pattern

amitzur:
Currently, running the following will result in 12 executions:
```
npx coverage-tests run --fn TestCheckRegion --fm isScrollStitching
```
(It's essentially `*TestCheckRegion*`)

With this PR, it will result in 1 execution, which is the desired state (otherwise there's no way to just run TestCheckRegion except by index which is not user friendly).


tourdedave:
💯

PR #140, opened 2020-03-23
WDIO5 target regions

danielschwartz85:
* Added WDIO 5 target regions support for VG eyes
* Fix xpath script in core
* Fix EyesWebElements regsions
* Moved persisted regions calculation to checkSettings (will write comments about this)  
* refactor region tests to use real driver

### DIFF
--- a/packages/eyes-sdk-core/lib/fluent/CheckSettings.js
+++ b/packages/eyes-sdk-core/lib/fluent/CheckSettings.js
@@ -346,6 +346,10 @@ class CheckSettings {
     throw new TypeError('Method called with argument of unknown type!')
   }
 
+  _getTargetType() {
+    return !this._targetRegion ? 'window' : 'region'
+  }
+
   /**
    * Adds one or more ignore regions.
    *
@@ -559,6 +563,37 @@ class CheckSettings {
    */
   toString() {
     return `${this.constructor.name} ${GeneralUtils.toString(this)}`
+  }
+
+  async toCheckWindowConfiguration(eyesWebDriver) {
+    const regions = await this._getPersistedRegions(eyesWebDriver)
+    return {
+      ...regions,
+      target: this._getTargetType(),
+      fully: this.getStitchContent(),
+      tag: this.getName(),
+      scriptHooks: this.getScriptHooks(),
+      sendDom: this.getSendDom(),
+      matchLevel: this.getMatchLevel(),
+    }
+  }
+
+  async _getPersistedRegions(eyesWebDriver) {
+    const [ignore, floating, strict, layout, content, accessibility] = await Promise.all([
+      persistRegions(this.getIgnoreRegions()),
+      persistRegions(this.getFloatingRegions()),
+      persistRegions(this.getStrictRegions()),
+      persistRegions(this.getLayoutRegions()),
+      persistRegions(this.getContentRegions()),
+      persistRegions(this.getAccessibilityRegions()),
+    ])
+
+    return {ignore, floating, strict, layout, content, accessibility}
+
+    async function persistRegions(regions) {
+      const persisted = await Promise.all(regions.map(r => r.toPersistedRegions(eyesWebDriver)))
+      return [].concat(...persisted)
+    }
   }
 }
 
--- a/packages/eyes-webdriverio-5/src/EyesVisualGrid.js
+++ b/packages/eyes-webdriverio-5/src/EyesVisualGrid.js
@@ -9,7 +9,6 @@ const {
   CorsIframeHandle,
   CorsIframeHandler,
   TypeUtils,
-  IgnoreRegionByRectangle,
   RectangleSize,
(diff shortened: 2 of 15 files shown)


packages/eyes-sdk-core/lib/fluent/CheckSettings.js
```diff
@@ -346,6 +346,10 @@ class CheckSettings {
     throw new TypeError('Method called with argument of unknown type!')
   }
 
+  _getTargetType() {
```
danielschwartz85: wdio5 checksettings overrides this

packages/eyes-sdk-core/lib/fluent/CheckSettings.js
```diff
@@ -560,6 +564,37 @@ class CheckSettings {
   toString() {
     return `${this.constructor.name} ${GeneralUtils.toString(this)}`
   }
+
```
danielschwartz85: I think other SDKs would also use _getPersistedregions(webDriver) method so moved it now to checkSettings

packages/eyes-webdriverio-5/src/EyesVisualGrid.js
```diff
@@ -259,46 +254,46 @@ class EyesVisualGrid extends EyesBase {
...
-    const [ignore, floating, strict, layout, content, accessibility] = await Promise.all([
-      this._persistRegions(checkSettings.getIgnoreRegions()),
-      this._persistRegions(checkSettings.getFloatingRegions()),
-      this._persistRegions(checkSettings.getStrictRegions()),
-      this._persistRegions(checkSettings.getLayoutRegions()),
-      this._persistRegions(checkSettings.getContentRegions()),
-      this._persistRegions(checkSettings.getAccessibilityRegions()),
+    const [config, {region, selector}] = await Promise.all([
+      checkSettings.toCheckWindowConfiguration(this._driver),
+      this._getTargetConfiguration(checkSettings),
```
danielschwartz85: notice getTargetConfiguration (the persisted target regions are still in SDK responsibility and not core, since SDK checksettings has properties regarding this which core/checkSettings doesn't know about (some of this can maybe also be moved to core in the future..)

packages/eyes-webdriverio-5/src/fluent/AccessibilityRegionByElement.js
```diff
@@ -46,7 +46,7 @@ class AccessibilityRegionByElement extends GetAccessibilityRegion {
   }
 
   async toPersistedRegions(driver) {
-    const xpath = await EyesJsBrowserUtils.getElementXpath(driver, this._element)
+    const xpath = await EyesJsBrowserUtils.getElementXpath(driver, this._element.element)
```
danielschwartz85: this is unrelated to refactor - just a fix to make regions of EyesWebElement work

packages/eyes-webdriverio-5/src/EyesVisualGrid.js
```diff
@@ -259,46 +254,46 @@ class EyesVisualGrid extends EyesBase {
...
-        ? checkSettings.getMatchLevel()
-        : this.getMatchLevel(),
       source,
     })
   }
 
-  async _persistRegions(regions) {
-    const persisted = await Promise.all(regions.map(r => r.toPersistedRegions(this._driver)))
-    return [].concat(...persisted)
+  async _getTargetConfiguration(checkSettings) {
```
kyryloonufriiev: Why `getTargetProvider` and `getTargetRegion` are separate functions? As I understand there couldn't be both at the same time.

danielschwartz85: It seems that getTargetRegion is sdk-core and getTargetProvider is SDKs.
They are used in different ways in Classic - when user want's element as target is a different code path from use who want's region (src/EyesWDIO.js:362).
Also the VG needs to differentiate, 

So we could say let's make a function that returns either ```{selector}``` or ```{region}``` and then we could tell the difference but perhaps even saving them in the same property on checkSettings is the right thing.. anyway here also I preferred not handle this since it's a bit of straying away..

danielschwartz85:
@kyryloonufriiev 
* removing unused ```_prepareRegions```
* regrading ```getSelector```:
 Many regions have SelectorByLocator and SelectorByElement and so they use it's ```getSelector``` method, but many regions don't have it (I thought about just using these classes as Util functions and adding them to regions that don't use them - so we can call getSelector).
Also note that ```getSelector``` is not really returning a real selector (since they generate a selector set it in the DOM and return the generated selector - which then the VG doesn't see).
We could maybe override this functionality and calculate xpath etc..
So not sure how to go about this, but generally I think this issue straying from this PR.

Note also that regarding ```getSelector``` we can probably do more to have a cleaner way to persist all the regions - this is very related to upcoming code changes we want to do which may have a bigger scope :)

PR #126, opened 2020-03-08
Coverage tests v2

tourdedave:
Work to move from an interpreted runner to a test code generator which will rely on existing test frameworks (e.g., Jest, testcafe, etc.) that will also be usable in other programming languages.

High-level requirements:
- Specify tests in a language agnostic way (e.g., JSON)
- Generate tests to a given language/framework from JSON
- Aggregate test results from runs using a standard format (e.g., JUnit XML)
- Enable tests to be specified in a reusable way within JavaScript (TBD) -- e.g., tests that are difficult to specify with JSON

PR #143, opened 2020-03-24
add way to provide shared custom coverage tests

kyryloonufriiev:
This PR is related to what @tourdedave do in PR https://github.com/applitools/eyes.sdk.javascript1/pull/126. Since we would like to create a language-independent way to describe a coverage test I think that it still could be a good idea to keep the current infostructure to share tests between SDKs that couldn't be described as a JSON.

I would like to know your vision of CLI changes, if any, when https://github.com/applitools/eyes.sdk.javascript1/pull/126 is done.


tourdedave:
Nice! Worth a discussion since it's a different direction than I was planning on taking things.

Here's a quick breakdown of my thinking.

The general idea of "custom" tests was to specify one-off tests that don't fit cleanly into being specified by the DSL -- which means they would be written with a traditional approach (e.g., per SDK with a test framework like Mocha, etc.) and need to be run separately (for now) with the hopes of eventually being incorporated into the runner (which is part of the work for the rewrite) so that both the DSL and custom tests could be run and provide aggregated results.

The main point here is that if a test is able to be specified with the DSL then we want to do so. But if we want to offer a way to both add tests to the DSL *and* distinguish between a set of tests (e.g., ones that are meant to verify the documented customer contract, ones to verify edge cases, and/or ones to verify that regressions don't get reintroduced) then we can go about it a different way (e.g., with metadata) instead of introducing a new primitive type of tests with a duplicate DSL.

The idea behind the coverage tests rewrite is to specify the tests in a language agnostic way (e.g., JSON) and to generate test code into a target test framework which would handle running the tests with and write the results to disk in a standard format that all test frameworks in all languages support (e.g., likely JUnit XML). We'd then parse those files to create an aggregate report that can be sent off to the test results dashboard. This would enable us to handle both tests specified in JSON (that get generated) and the "custom" tests written by hand -- and also enable SDKs in other languages to do the same.

amitzur:
The discussion is taking off in different directions in #176 
@kyryloonufriiev can I close this PR?

kyryloonufriiev:
@amitzur sure

PR #144, opened 2020-03-25
Fix generated batch ID

amitzur:
The issue surfaced in storybook, which still sends unnecessary requests to Eyes from `_getAndSaveBatchInfoFromServer` that get 404:

There was a bug that `openEyes` received `batchId` in `makeOpenEyes`, and then in `configureWrappers` it created a new `BatchInfo` object that was marked as `isGeneratedId: false`.

As far as I understand, this is because since commit 31877b37fe6ce8f9c08062ce3a00ae891e49d549 if `openEyes` doesn't get a `batch` property as input, it will always be `isGeneratedId: false`. This is because `batchId` was always passed in `makeOpenEyes`.

### The fix
- `makeOpenEyes` and `configureWrappers` don't receive the individual properties anymore. Only an object of type `BatchInfo`.
- any individual property of the batch that is passed to `openEyes` is merged with the `BatchInfo` instance that was either passed from `makeOpenEyes` or passed directly to `openEyes`.


packages/visual-grid-client/test/it/openEyes.it.test.js
```diff
@@ -138,6 +138,41 @@ describe('openEyes', () => {
...
+    expect(batch.getIsGeneratedId()).to.be.true
+  })
+
+  it('sets batch properties passed to makeRenderingGridClient', async () => {
+    const now = Date.now()
+    const batchId = `batchId_${now}`
+    const batchName = `batchName_${now}`
+    const batchSequence = `batchSequence_${now}`
+
+    openEyes = makeRenderingGridClient({
```
danielschwartz85: how is this not ```{openEyes} = ..```, am I missing something..

packages/visual-grid-client/src/sdk/openEyes.js
```diff
@@ -345,4 +338,22 @@ function makeOpenEyes({
   }
 }
 
+function setIndividualBatchProperties({batch, batchId, batchName, batchSequence, batchNotify}) {
+  if (batchId !== undefined) {
+    batch.setId(batchId)
```
danielschwartz85: Note that if some one does 
```
const {openEyes} = makeRenderingGridClient({})
openEyes({batchId: 'someNotGeneratedId'})
``` 
then I think this ```batch``` here; would have ```_isAutoGenerated === true``` (since it's auto generated in renderingGridClient)

could add here:
```batch.setIsGeneratedId(false)```

File: packages/visual-grid-client/src/sdk/renderingGridClient.js [hunk 3]
```diff
@@ -152,11 +152,12 @@ function makeRenderingGridClient({
...
-    batchId: defaultBatchId,
-    batchName: defaultBatchName,
-    batchSequence: defaultBatchSequence,
-  } = getBatch({batchSequence, batchName, batchId})
+  const batch = new BatchInfo({
+    name: batchName,
+    id: batchId,
+    sequenceName: batchSequence,
+    notifyOnCompletion: batchNotify,
+  })
```
danielschwartz85: c co coo cool

packages/visual-grid-client/src/sdk/openEyes.js
amitzur: arghh... good point.
I really wanted to avoid duplicating the code in `BatchInfo`. But it seems impossible to avoid.

Now that I think about it there's a severe race condition here - I'm reusing the `batch` instance passed in `makeOpenEyes` across different `openEyes` calls. So they potentially override each other.
:(

amitzur: I'm going to write an additional test case for what you mentioned and for the race condition and fix them

packages/visual-grid-client/test/it/openEyes.it.test.js
amitzur: because in line 165 there is `.openEyes` :)

PR #147, opened 2020-03-26
Pub workflow improvements: pack -> install -> verify installed versions

tourdedave:
https://trello.com/c/gVFxTv0w/272-publishing-workflow-improvements


packages/sdk-release-kit/src/cli/index.js
```diff
@@ -3,39 +3,65 @@
...
+    execute(updateChangelog.bind(undefined, cwd))
+  } else if (args['send-release-notification']) {
+    execute(sendReleaseNotification.bind(undefined, args.recipient))
+  } else if (args['verify-versions']) {
+    execute(verifyVersions.bind(undefined, {isFix: args.fix, pkgPath: cwd}))
+  } else if (args['verify-commits']) {
+    const isForce = args.force || process.env.BONGO_VERIFY_COMMITS_FORCE
+    execute(verifyCommits.bind(undefined, {pkgPath: cwd, isForce}))
+  } else if (args['verify-installed-versions']) {
+    execute(createDotFolder.bind(undefined, cwd))
```
amitzur: you're not awaiting `createDotFolder`, so there will be a race condition with `packInstall`. This also happens below in `release-pre-check`.
I think the `execute` is nice, but just overly complicates things a bit. Why not wrap all this function in try/catch and call the actual script functions directly?
```
if (args['verify-changelog']) {
  await verifyChangelog(cwd)
} else if (args['update-changelog']) {
  await updateChangelog(cwd)
}
```

packages/sdk-release-kit/src/dry-run/scripts/pack-install.js
```diff
@@ -0,0 +1,14 @@
+const fs = require('fs')
+const path = require('path')
+const {exec} = require('child_process')
+const {promisify} = require('util')
+const pexec = promisify(exec)
+
+module.exports = async packageDir => {
+  const workingDir = path.join(packageDir, '.bongo', 'dry-run')
+  await pexec(`rm -rf ${workingDir}`)
```
amitzur: just fyi there's `fs.rmdirSync(path, {recursive:true})` but this is fine and clean

packages/sdk-release-kit/src/dry-run/scripts/pack-install.js
```diff
@@ -0,0 +1,14 @@
...
+  await pexec(`rm -rf dry-run.tgz`)
+  fs.mkdirSync(workingDir)
+  await pexec(`yarn pack --filename dry-run.tgz`)
```
amitzur: maybe output onto `.bongo`?

packages/sdk-release-kit/src/dry-run/scripts/pack-install.js
```diff
@@ -0,0 +1,14 @@
...
+  await pexec(`npm install dry-run.tgz --prefix ${workingDir}`)
```
amitzur: per this answer: https://stackoverflow.com/a/14469516/2023596
shouldn't you create `.bongo/dry-run/node_modules` so that npm doesn't walk up?

packages/sdk-release-kit/src/cli/index.js
```diff
@@ -3,39 +3,65 @@
...
-// - verify-changelog
-// - verify versions
-// - verify-commits
-// - verify-installed-versions
+  if (args['verify-changelog']) {
+    execute(verifyChangelog.bind(undefined, cwd))
+  } else if (args['update-changelog']) {
+    execute(updateChangelog.bind(undefined, cwd))
+  } else if (args['send-release-notification']) {
+    execute(sendReleaseNotification.bind(undefined, args.recipient))
```
amitzur: I think you missed sending `cwd` here

packages/sdk-release-kit/src/versions/scripts/verify-installed-versions.js
```diff
@@ -12,8 +13,19 @@ async function npmLs() {
   }
 }
 
-async function main(packageNames) {
-  checkPackagesForUniqueVersions(await npmLs(), packageNames)
+async function main({pkgPath, installedDirectory}) {
+  debugger
```
amitzur: bugger

packages/sdk-release-kit/src/versions/scripts/verify-installed-versions.js
```diff
@@ -12,8 +13,19 @@ async function npmLs() {
...
+  const {dependencies} = require(path.join(pkgPath, 'package.json'))
+  const packageNames = Object.keys(dependencies)
```
amitzur: including non-applitools dependencies?

packages/sdk-release-kit/src/versions/scripts/verify-installed-versions.js
```diff
@@ -12,8 +13,19 @@ async function npmLs() {
...
+  if (installedDirectory) {
+    process.chdir(installedDirectory)
+  }
+  try {
+    checkPackagesForUniqueVersions(await npmLs(), packageNames)
+  } catch (error) {
+    process.chdir(pkgPath)
```
amitzur: do you want to chdir back also if there wasn't an error? maybe it should be in `finally`?

packages/sdk-release-kit/src/versions/index.js
```diff
@@ -54,27 +55,34 @@ function checkPackagesForUniqueVersions(input, packageNames) {
...
   return new Set(versionNumbers)
 }
 
-function _grep(input, target) {
+function findEntryByPackageName(input, target) {
   if (typeof input === 'string') input = input.split('\n')
-  return input.filter(entry => entry.includes(target))
+  return input.filter(entry => {
+    const result = entry.match(/(@?\w+\/?\w+\-?\w+)@/)
+    const _entry = result ? result[1] : ''
```
amitzur: normally we use the `_` prefix for unused variables. So here i'd call it `entryName` or something

packages/sdk-release-kit/src/dry-run/scripts/pack-install.js
tourdedave: Only works in Node 12, also is experimental. I got mixed results with it.

packages/sdk-release-kit/src/versions/scripts/verify-installed-versions.js
tourdedave: Yeah, good point. We'll want to filter this against the workspaces list.

packages/sdk-release-kit/src/versions/index.js
tourdedave: To me it also denotes private variables/functions. Isn't that idiomatic?

tourdedave:
Thanks for the feedback! I addressed most of the comments. I'll address the ones not directly related to this PR (e.g., package structure, commands vs args, etc) later.

re: `ls-dry-run` was added to get the `npm ls` output from a dry-run which has errored because there are multiple versions of an internal package. Right now the error output just tells you which internal packages have multiple versions installed, but not which package has installed them. `ls-dry-run` is a stop gap until better output is added, which is on my punch list for later.

packages/sdk-release-kit/src/cli/index.js
tourdedave: `createDotFolder` is sync. Agreed, though. Dropping `execute` would make this easier to reason about.

packages/sdk-release-kit/src/versions/scripts/verify-installed-versions.js
tourdedave: Turns out the `chdir` isn't necessary since it happens in the context of the script, not the current terminal session. :clap:

PR #148, opened 2020-03-26
Add 'edgelegacy' and 'edgechromium' to supported browsers

amirhalatzi:
VG supports these new names. We'd like to add it to enable testing

### DIFF
--- a/packages/visual-grid-client/src/sdk/openEyes.js
+++ b/packages/visual-grid-client/src/sdk/openEyes.js
@@ -25,6 +25,8 @@ const SUPPORTED_BROWSERS = {
   ie10: 'ie10',
   ie11: 'ie11',
   edge: 'edge',
+  edgelegacy: 'edgelegacy',
+  edgechromium: 'edgechromium',
   ie: 'ie',
   safari: 'safari',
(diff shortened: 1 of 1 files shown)


packages/visual-grid-client/src/sdk/openEyes.js
```diff
@@ -25,6 +25,8 @@ const SUPPORTED_BROWSERS = {
   ie10: 'ie10',
   ie11: 'ie11',
   edge: 'edge',
+  edgelegacy: 'edgelegacy',
```
amitzur: I understand the intent, but I think that we don't plan to get `edgelegacy` from the user.
See this PR: https://github.com/applitools/eyes.sdk.javascript1/pull/150

Still in draft since I didn't hear from product yet about the exposed strings

amitzur: Ohh now I see that you wanted this to start testing. I'll merge this in the meantime.

amitzur: Actually users will see a message about this. So we need to hide it for the meantime. Let's discuss today.

PR #153, opened 2020-04-06
coverage-tests++

tourdedave:
The aim of this pull request is to update how the coverage tests create and run tests.

Up until now they have worked as a real-time interpreter with a custom test runner.

The focus of this PR is to remove this existing approach and replace it with a mechanism to construct test files for a given test framework that (when run) will output to a standard report format that can be consumed/transformed/sent where needed.

This would enable us to run both generic and custom coverage tests for each of the SDKs as part of the same run, and report the results for both (whereas right now we can only report the results from the generic tests).

### Punch list

- [x] Generate test files from existing test DSL for Jest
- [x] Wire up Jest execution w/ JUnit XML output from the test-kit
- [x] Wire up initial POC with eyes-selenium
- [x] Finish wiring up testcafe
- [x] Finish wiring up eyes-selenium
- [x] Generate test files from existing test DSL for alternate test framework
- [x] Wire up support for alternative test framework execution from the test-kit 
- [x] Wire up initial POC with test-cafe (w/ JUnit XML output)
- [x] Process XML result output
- [x] Send processed results to QA dashboard
- [x] Wire up WDIO-4
- [x] Wire up WDIO-5
- [ ] ~~Include `test/coverage/custom` tests in runs~~ (will do this after merging)
- [x] Ability to re-run just failed/passed tests (or alt. way to specify enabled tests - TBD)


tourdedave:
Noting some additional things that will need to be addressed:

#### Improved output

- real-time stdout when running tests through a provided executor (e.g., `testcafe`)
- duplicate failure output when running tests with `jest`

#### Improved execution speed

- `testcafe` does not run concurrently even though the `--concurrency` option is set

PR #154, opened 2020-04-07
Support iosDeviceInfo

amitzur:
The challenge here is to provide a clean API, also taking into consideration the following:
1) render jobs are mapped 1:1 with elements in the array `browser` that is passed to `openEyes`.
2) the entries in the `browser` array are already pretty complicated with device emulation.
3) `screenOrientation` is an example that is right now flat on the browser object.
4) the decision today if something is a device emulation is based on implicit parameters.
5) there's a "mobile" property that can be specified that indicates device emulation explicitly

I don't mind doing a breaking change in vgc. We just need to decide on the correct structure of the input.

This PR is just WIP that I didn't finish because of the issues above.

trello: https://trello.com/c/ToLqWI30/301-add-ios-related-render-info-to-vgc-under-a-flag


packages/visual-grid-client/test/it/openEyes.it.test.js
```diff
@@ -1960,6 +1960,24 @@ Received: 'firefox-1'.`,
...
   })
 
+  it.only('renders iosDeviceInfo', async () => {
+    const deviceName = 'iPhone 4'
+    const {checkWindow, close} = await openEyes({
+      wrappers: [wrapper],
+      browser: {
+        platform: 'ios',
+        name: 'safari',
+        iosDeviceInfo: {screenOrientation: 'portrait', version: 'latest'},
```
corevo, line 1970: name: `iPhone 11 Pro`

PR #155, opened 2020-04-13
added agent id to eyes requests

danielschwartz85:
* Now base agent id is saved on Configuration (since Configuration is passed to ServerConnector)
* This means that baseAgent id is needed on creation time
* EyesBase implementors can override getBaseAgentId method for setting the agent id
* EyesBase implementors can use EyesBase static setter for setting getBaseAgentId (so we can create a string dynamic override for agent id; used by VGC)

### DIFF
--- a/packages/eyes-sdk-core-legacy/src/ServerConnector.js
+++ b/packages/eyes-sdk-core-legacy/src/ServerConnector.js
@@ -1,20 +1,20 @@
 ;(function() {
   'use strict'
 
-  let request = require('request')
-  let GeneralUtils = require('eyes.utils').GeneralUtils
-  let ArgumentGuard = require('eyes.utils').ArgumentGuard
-  let RenderingInfo = require('./RenderingInfo').RenderingInfo
+  const request = require('request')
+  const GeneralUtils = require('eyes.utils').GeneralUtils
+  const ArgumentGuard = require('eyes.utils').ArgumentGuard
+  const RenderingInfo = require('./RenderingInfo').RenderingInfo
 
   // constants
-  let TIMEOUT = 5 * 60 * 1000,
+  const TIMEOUT = 5 * 60 * 1000,
     API_PATH = '/api/sessions/running',
     DEFAULT_HEADERS = {
       Accept: 'application/json',
       'Content-Type': 'application/json',
     }
 
-  let LONG_REQUEST_DELAY = 2000, // ms
+  const LONG_REQUEST_DELAY = 2000, // ms
     MAX_LONG_REQUEST_DELAY = 10000, // ms
     LONG_REQUEST_DELAY_MULTIPLICATIVE_INCREASE_FACTOR = 1.5
 
--- a/packages/eyes-sdk-core-legacy/src/ServerConnector.js
+++ b/packages/eyes-sdk-core-legacy/src/ServerConnector.js
@@ -49,17 +49,16 @@
    * @param {Logger} logger
    * @constructor
    **/
-  function ServerConnector({promiseFactory, serverUrl, logger, agentId}) {
-    ArgumentGuard.notNullOrEmpty(agentId, 'agentId')
-
+  function ServerConnector({promiseFactory, serverUrl, logger, getAgentId}) {
     this._promiseFactory = promiseFactory
+    this._getAgentId = getAgentId
     this._logger = logger
     this._runKey = undefined
     this._renderingInfo = undefined
     this._httpOptions = {
       proxy: null,
       strictSSL: false,
-      headers: {...DEFAULT_HEADERS, [AGENT_ID_HEADER]: agentId},
+      headers: DEFAULT_HEADERS,
       timeout: TIMEOUT,
       qs: {},
     }
@@ -517,6 +516,7 @@
     options = options || {}
 
     const req = GeneralUtils.clone(that._httpOptions)
+    req.headers = {...req.headers, [AGENT_ID_HEADER]: that._getAgentId()}
     req.uri = uri
     req.method = method
     if (options.query) req.qs = GeneralUtils.objectAssign(req.qs, options.query)
--- a/packages/eyes-sdk-core-legacy/src/ServerConnector.js
+++ b/packages/eyes-sdk-core-legacy/src/ServerConnector.js
@@ -1,19 +1,20 @@
 ;(function() {
   'use strict'
 
-  var request = require('request')
-  var GeneralUtils = require('eyes.utils').GeneralUtils
-  var RenderingInfo = require('./RenderingInfo').RenderingInfo
+  let request = require('request')
+  let GeneralUtils = require('eyes.utils').GeneralUtils
+  let ArgumentGuard = require('eyes.utils').ArgumentGuard
+  let RenderingInfo = require('./RenderingInfo').RenderingInfo
 
   // constants
-  var TIMEOUT = 5 * 60 * 1000,
+  let TIMEOUT = 5 * 60 * 1000,
     API_PATH = '/api/sessions/running',
     DEFAULT_HEADERS = {
       Accept: 'application/json',
       'Content-Type': 'application/json',
     }
 
-  var LONG_REQUEST_DELAY = 2000, // ms
+  let LONG_REQUEST_DELAY = 2000, // ms
     MAX_LONG_REQUEST_DELAY = 10000, // ms
     LONG_REQUEST_DELAY_MULTIPLICATIVE_INCREASE_FACTOR = 1.5
 
@@ -38,6 +39,7 @@
   ]
 
   const REQUEST_FAILED_CODES = ['ECONNRESET', 'ECONNABORTED', 'ETIMEDOUT', 'ENOTFOUND', 'EAI_AGAIN']
+  const AGENT_ID_HEADER = 'x-applitools-eyes-client'
 
   /**
    * Provides an API for communication with the Applitools server.
@@ -47,7 +49,9 @@
    * @param {Logger} logger
    * @constructor
    **/
-  function ServerConnector(promiseFactory, serverUrl, logger) {
+  function ServerConnector({promiseFactory, serverUrl, logger, agentId}) {
+    ArgumentGuard.notNullOrEmpty(agentId, 'agentId')
+
     this._promiseFactory = promiseFactory
     this._logger = logger
     this._runKey = undefined
@@ -55,7 +59,7 @@
     this._httpOptions = {
       proxy: null,
       strictSSL: false,
-      headers: DEFAULT_HEADERS,
+      headers: {...DEFAULT_HEADERS, [AGENT_ID_HEADER]: agentId},
       timeout: TIMEOUT,
       qs: {},
     }
@@ -278,7 +282,9 @@
       },
     }
 
-    return _sendLongRequest(this, 'uploadScreenshot', uri, 'put', options, 3).then(function(results) {
+    return _sendLongRequest(this, 'uploadScreenshot', uri, 'put', options, 3).then(function(
+      results,
+    ) {
       if (results.status !== HTTP_STATUS_CODES.CREATED) {
         throw new Error(
           `ServerConnector.uploadScreenshot - unexpected status ${_makeResponseOutputString(
--- a/packages/eyes-sdk-core/lib/EyesBase.js
+++ b/packages/eyes-sdk-core/lib/EyesBase.js
@@ -82,6 +82,7 @@ class EyesBase {
 
     this._configuration.setServerUrl(serverUrl)
     this._configuration.setIsDisabled(isDisabled)
+    this._configuration.setBaseAgentId(this.getBaseAgentId())
 
     /** @type {ServerConnector} */
     this._serverConnector = new ServerConnector(this._logger, this._configuration)
@@ -186,6 +187,7 @@ class EyesBase {
     }
 
     this._configuration = configuration.cloneConfig()
+    this._configuration.setBaseAgentId(this.getBaseAgentId())
     this._serverConnector._configuration = this._configuration
   }
 
@@ -980,15 +982,11 @@ class EyesBase {
   }
 
   /**
+   * @protected
    * @return {string} - The full agent id composed of both the base agent id and the user given agent id.
    */
   getFullAgentId() {
-    const agentId = this._configuration.getAgentId()
-    if (!agentId) {
-      return this.getBaseAgentId()
-    }
-
-    return `${agentId} [${this.getBaseAgentId()}]`
+    return this._configuration.getFullAgentId()
   }
 
   /**
@@ -2240,7 +2238,14 @@ class EyesBase {
    * @return {string} - The base agent id of the SDK.
    */
   getBaseAgentId() {
-    throw new TypeError('The method is not implemented!')
+    if (EyesBase._baseAgentId) {
+      return EyesBase._baseAgentId
+    }
+    throw new TypeError('Base agent id is not implemented!')
+  }
+
+  static set baseAgentId(baseAgentId) {
+    EyesBase._baseAgentId = baseAgentId
   }
 
   /**
--- a/packages/eyes-sdk-core/lib/EyesBase.js
+++ b/packages/eyes-sdk-core/lib/EyesBase.js
@@ -82,10 +82,13 @@ class EyesBase {
 
     this._configuration.setServerUrl(serverUrl)
     this._configuration.setIsDisabled(isDisabled)
-    this._configuration.setBaseAgentId(this.getBaseAgentId())
 
     /** @type {ServerConnector} */
-    this._serverConnector = new ServerConnector(this._logger, this._configuration)
+    this._serverConnector = new ServerConnector({
+      logger: this._logger,
+      configuration: this._configuration,
+      getAgentId: this.getFullAgentId.bind(this),
+    })
 
     if (this._configuration.getIsDisabled()) {
       this._userInputs = []
@@ -187,7 +190,6 @@ class EyesBase {
     }
 
     this._configuration = configuration.cloneConfig()
-    this._configuration.setBaseAgentId(this.getBaseAgentId())
     this._serverConnector._configuration = this._configuration
   }
 
@@ -197,14 +199,14 @@ class EyesBase {
    * @param {string} agentId - The agent ID to set.
    */
   setAgentId(agentId) {
-    this._configuration.setAgentId(agentId)
+    this._agentId = agentId
   }
 
   /**
    * @return {string} - The user given agent id of the SDK.
    */
   getAgentId() {
-    return this._configuration.getAgentId()
+    return this._agentId
   }
 
   /**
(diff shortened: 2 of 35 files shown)


packages/eyes-sdk-core-legacy/src/ServerConnector.js
```diff
@@ -1,19 +1,20 @@
...
-  var GeneralUtils = require('eyes.utils').GeneralUtils
-  var RenderingInfo = require('./RenderingInfo').RenderingInfo
+  let request = require('request')
+  let GeneralUtils = require('eyes.utils').GeneralUtils
+  let ArgumentGuard = require('eyes.utils').ArgumentGuard
+  let RenderingInfo = require('./RenderingInfo').RenderingInfo
 
   // constants
-  var TIMEOUT = 5 * 60 * 1000,
+  let TIMEOUT = 5 * 60 * 1000,
```
amitzur, line 10: why not `const`?

packages/eyes-sdk-core/lib/server/ServerConnector.js
```diff
@@ -48,6 +48,11 @@ const HTTP_STATUS_CODES = {
   GATEWAY_TIMEOUT: 504,
 }
 
+const SERVER_CONFIG = {
```
amitzur, line 51: I think we should send the agent ID to the VG as well. No reason not to. Just like we send the request ID.

packages/eyes-selenium-3/src/Eyes.js
```diff
@@ -77,14 +77,10 @@
     EyesBase.call(this, this._promiseFactory, serverUrl || EyesBase.DEFAULT_EYES_SERVER, isDisabled)
   }
 
+  EyesBase.prototype._getBaseAgentId = () => 'eyes.selenium.javascript/' + VERSION
```
amitzur, line 80: I understand why it was needed to move the line. But why was it necessary to change `EyesBase.prototype` and not `Eyes.prototype` ? Not that it makes any difference, i'm just curious.

packages/eyes-selenium/scripts/render.js
```diff
@@ -11,6 +11,7 @@ const {
   Configuration,
   StitchMode,
   EyesSeleniumUtils,
+  // ProxySettings,
```
amitzur, line 14: see this commit :) 627aa2af50f2bb1f202418071aa315f06d8efea1

packages/visual-grid-client/test/it/openEyes.it.test.js
```diff
@@ -1230,7 +1230,6 @@ Received: 'firefox-1'.`,
       renderWrapper: wrapper,
       serverUrl: 'serverUrl',
       proxy: 'proxy',
-      agentId: 'agentId',
```
amitzur, line 1233: why did you take these out?

packages/visual-grid-client/src/sdk/EyesWrapper.js
```diff
@@ -12,7 +12,8 @@ const {presult} = require('@applitools/functional-commons')
 const VERSION = require('../../package.json').version
 
 class EyesWrapper extends EyesBase {
-  constructor({apiKey, logHandler, getBatchInfoWithCache} = {}) {
+  constructor({agentId, apiKey, logHandler, getBatchInfoWithCache} = {}) {
+    EyesBase.baseAgentId = agentId || `visual-grid-client/${VERSION}`
```
amitzur, line 16: I want to offer a different way to implement this. In order to not do things differently.
In wrapperUtils-->configureWrappers, Instead of doing `wrapper.set...()` for all properties, we change it to:
```
const configuration = wrapper.getConfiguration()

...
displayName !== undefined && configuration.setDisplayName(displayName)
...
agentId !== undefined && configuration.setBaseAgentId(agentId)

wrapper.setConfiguration(configuration)
```
And then it works like all other SDK's

packages/eyes-sdk-core/lib/EyesBase.js
```diff
@@ -2240,7 +2238,14 @@ class EyesBase {
    * @return {string} - The base agent id of the SDK.
    */
   getBaseAgentId() {
-    throw new TypeError('The method is not implemented!')
+    if (EyesBase._baseAgentId) {
```
amitzur, line 2241: I would like to avoid this, as it's a hack. See my suggestion below in `EyesWrapper`

packages/visual-grid-client/src/sdk/openEyes.js
```diff
@@ -171,6 +171,7 @@ function makeOpenEyes({
         apiKey,
         logHandler: logger.getLogHandler(),
         getBatchInfoWithCache,
+        agentId,
```
amitzur, line 174: with my suggestion in this is not needed

packages/visual-grid-client/src/sdk/wrapperUtils.js
```diff
@@ -2,10 +2,10 @@
 const EyesWrapper = require('./EyesWrapper')
 const {RectangleSize, TypeUtils} = require('@applitools/eyes-sdk-core')
 
-function initWrappers({count, apiKey, logHandler, getBatchInfoWithCache}) {
+function initWrappers({count, agentId, apiKey, logHandler, getBatchInfoWithCache}) {
```
amitzur, line 5: with my suggestion this is not needed

packages/visual-grid-client/test/util/FakeEyesWrapper.js
```diff
@@ -45,6 +45,7 @@ class FakeEyesWrapper extends EventEmitter {
     goodResources = [],
     closeErr = false,
     failRender,
+    agentId,
```
amitzur, line 48: with my suggestion this is not needed

packages/eyes-sdk-core/lib/server/requestHelpers.js
```diff
@@ -56,13 +57,23 @@ function configAxiosProxy({axiosConfig, proxy, logger}) {
 
   logger.log('proxy is set as http only, using tunnel', proxyObject.host, proxyObject.port)
 }
+
 function configAxiosFromConfiguration({axiosConfig, configuration, logger}) {
-  if (axiosConfig.params === undefined) {
-    axiosConfig.params = {}
-  }
-  if (axiosConfig.withApiKey && !('apiKey' in axiosConfig.params)) {
+  axiosConfig.params = axiosConfig.params || {}
```
amitzur, line 62: what's the reasoning behind changing this function?

packages/eyes-webdriverio-4/src/EyesVisualGrid.js
```diff
@@ -39,7 +38,7 @@ class EyesVisualGrid extends EyesBase {
    * @param {EyesRunner} [runner] - Set {@code true} to disable Applitools Eyes and use the WebDriver directly.
    */
   constructor(serverUrl, isDisabled, runner = new VisualGridRunner()) {
-    super(serverUrl, isDisabled, new Configuration())
+    super(serverUrl, isDisabled)
```
amitzur, line 41: Is this necessary or just cleanup?
question is relevant to all other places with this change

packages/eyes-sdk-core-legacy/src/ServerConnector.js
danielschwartz85: no reason, just did a search and replace on vars for that file (so didn't go into details), but will change these.

packages/eyes-sdk-core/lib/server/ServerConnector.js
danielschwartz85: ok will make changes (in EyesBase as well)

packages/eyes-selenium-3/src/Eyes.js
danielschwartz85: hmm no special reason actually, didn't think about this too much (since we do new EyesBase()). but I see how Eyes might be more fitting with the rest of the lines here..(will change this)

packages/eyes-sdk-core/lib/server/requestHelpers.js
danielschwartz85: since I set headers here I make sure the headers is initialized (and so since I have 2 initializers at start of the function I changed it to one liners).

* I do see something that shouldn't be here (deleting the apiKey and agentIdHeader; no need for this - I saw it in requests sent to VG that shouldn't have been there and thought initially that their not deleted - but then  saw other bug for this.. so the delete part can be removed)

packages/eyes-selenium/scripts/render.js
danielschwartz85: :+1:

packages/eyes-webdriverio-4/src/EyesVisualGrid.js
danielschwartz85: just for cleanup (left from change I did to CTR of EyesBase that eventually didn't keep)

packages/visual-grid-client/src/sdk/EyesWrapper.js
danielschwartz85: Thing is by that time the EyesBase ctr was already called and agent id was already needed (since it is set on teh configuration object); and so an exception is thrown if there is no baseAgentId at creation time.

but I think I can create a new Configuration obj (in EyesWrapper ctr) and pass it with baseAgentId to EyesBase (then in EyesBase not search for baseAgentId since it's already set)  --  this instead of setting with hacky static method.. wdty ?

packages/visual-grid-client/test/it/openEyes.it.test.js
danielschwartz85: since the in this test the wrapper is supplied to VGC and since now we don't set the wrapper with agent id (we set it in ctr of wrpper which is not the flow here) - and so I don't want to create false impression that this is tested.

packages/visual-grid-client/src/sdk/EyesWrapper.js
danielschwartz85: Actually not sure about the configuration option in ctr since maybe a hack somewhat better in that it's explicit / documented in code about what should be done for implementing an EyesBase if u have a dynamic baseagentID. 
(also the configuration obj is not meant for dev purposes, i.e. it's for user usage..)

danielschwartz85: pushed the prototype method override way..

packages/eyes-sdk-core/lib/server/requestHelpers.js
danielschwartz85: removed the delete parts..

packages/eyes-sdk-core/lib/EyesBase.js
```diff
@@ -980,15 +982,11 @@ class EyesBase {
   }
 
   /**
+   * @protected
```
amitzur, line 985: who cares?

packages/eyes-sdk-core/test/it/EyesBase.spec.js
```diff
@@ -13,7 +13,7 @@ describe('EyesBase', () => {
   let eyes
   describe('getAndSetBatchInfo()', () => {
     beforeEach(() => {
-      eyes = new EyesBase()
+      eyes = new FakeEyes()
```
amitzur, line 16: how FakeEyes is written, there's no need to do `new FakeEyes()`, just `FakeEyes()`

packages/eyes-sdk-core/test/testUtils.js
```diff
@@ -12,9 +12,9 @@ const getResource = fileName => fs.readFileSync(getResourcePath(fileName))
 const getResourceAsText = fileName => fs.readFileSync(getResourcePath(fileName), 'utf8').trim()
 
 function FakeEyes(...args) {
+  EyesBase.prototype.getBaseAgentId = () => 'fakeAgentId'
```
amitzur, line 15: why not inherit from EyesBase and be done with?

packages/eyes-sdk-core/test/unit/server/ServerConnector.spec.js
```diff
@@ -11,7 +11,7 @@ describe('ServerConnector', () => {
     const {port, close} = await fakeEyesServer({logger})
     try {
       const serverUrl = `http://localhost:${port}`
-      const configuration = new Configuration()
+      const configuration = new Configuration({baseAgentId: 'test'})
```
amitzur, line 14: why do you need to do this? Is it now mandatory to pass in Configuration c'tor

packages/eyes-selenium/scripts/render.js
```diff
@@ -126,6 +126,8 @@ if (!url) {
     configuration.setApiKey(args.apiKey)
   }
   eyes.setConfiguration(configuration)
+  // eyes.setProxy(new ProxySettings('localhost:8888'))
```
amitzur, line 129: can you remove this?

amitzur: probably merge from master first

packages/visual-grid-client/test/util/FakeEyesWrapper.js
```diff
@@ -388,12 +388,8 @@ class FakeEyesWrapper extends EventEmitter {
     this.deviceInfo = value
   }
 
-  setBaseAgentId(value) {
-    this.agentId = value
-  }
-
   getBaseAgentId() {
-    return this.agentId || 'fake wrapper'
+    return 'fake warpper'
```
amitzur, line 392: wrapper instead of warpper

packages/eyes-sdk-core/lib/EyesBase.js
danielschwartz85: I don't that's for sure.
this is just the format of the file - I'm guessing to say that it's not public API for users but still usable by impl.

danielschwartz85: (I initially removed the function then returned it and so i thought it was there before)

packages/eyes-sdk-core/test/unit/server/ServerConnector.spec.js
danielschwartz85: it is mandatory if you give the configuration to server connector and make requests

danielschwartz85:
I added serverConfiguration because I notices that there is a connection between where the request is going to what configuration it's getting.
This allows configuring all requests and also overriding them if needed.

Now we send agentId also to VG so this might be less valuable... I still kept it since it makes sens to me.

packages/visual-grid-client/test/util/FakeEyesWrapper.js
danielschwartz85: :+1:

packages/visual-grid-client/test/it/openEyes.it.test.js
amitzur: fair enough. is this still relevant? or can we bring it back?

danielschwartz85: yes, still relevant (not creating the wrappers..)

packages/eyes-sdk-core/test/unit/server/ServerConnector.spec.js
amitzur: It makes me realize we didn't do the right thing :/
`configuration.setBaseAgentId` shouldn't be exposed to the user. We should think of a different way to do this.

danielschwartz85: true, were riding on a user object for convince.. (note though that same can be said about the API between ServerConnector and Eyes - using a user object)

amitzur:
> I added serverConfiguration because I notices that there is a connection between where the request is going to what configuration it's getting.
> This allows configuring all requests and also overriding them if needed.

Yeah, I guess it makes sense. What was weird for me was keeping the distinction between `configureAxiosHeaders` and `configureAxios` and assuming that `axiosConfig.headers` might not be defined.

amitzur:
The implementation in C#:
https://github.com/applitools/Eyes.Sdk.DotNet/commit/f0f123f854acd823dcd6f7f547ab71f227b2630f

danielschwartz85:
@amitzur yes, saw the commit; also talked with itay about what they do..

danielschwartz85:
> > I added serverConfiguration because I notices that there is a connection between where the request is going to what configuration it's getting.
> > This allows configuring all requests and also overriding them if needed.
> 
> Yeah, I guess it makes sense. What was weird for me was keeping the distinction between `configureAxiosHeaders` and `configureAxios` and assuming that `axiosConfig.headers` might not be defined.

hmm actually not sure (by looking at the code), but the assumption was there before (also I got an exception for this at some point if i'm not mistaken).
regarding the split between functions I think it's good to merge them..

packages/eyes-sdk-core/lib/server/ServerConnector.js
```diff
@@ -48,6 +48,11 @@ const HTTP_STATUS_CODES = {
...
+  EYES: {withApiKey: true, withAgentId: true},
+  VG: {withApiKey: false, withAgentId: true},
+}
```
kyryloonufriiev, line 54: I think since we want to send agent id header to the VG, we don't need this object and `serverConfig` field anymore.

packages/eyes-sdk-core/test/unit/server/ServerConnector.spec.js
danielschwartz85: > It makes me realize we didn't do the right thing :/
> `configuration.setBaseAgentId` shouldn't be exposed to the user. We should think of a different way to do this.

we can add @protected
;)

amitzur:
> regarding the split between functions I think it's good to merge them.. (it was kept because one takes args from configuration and I didn't change this)

Ok so lets merge them

packages/eyes-sdk-core/test/unit/server/ServerConnector.spec.js
amitzur: So I propose a new solution:
`EyesBase` instantiates `ServerConnector` and provides it with a function to get the agent id:
```
this._serverConnector = new ServerConnector({logger: this._logger, configuration: this._configuration, getAgentId: this.getFullAgentId.bind(this)})
```
wdyt?

amitzur: I don't like the C# solution so much. It creates confusion IMO

danielschwartz85: > So I propose a new solution:
> `EyesBase` instantiates `ServerConnector` and provides it with a function to get the agent id:
> 
> ```
> this._serverConnector = new ServerConnector({logger: this._logger, configuration: this._configuration, getAgentId: this.getFullAgentId.bind(this)})
> ```
> 
> wdyt?
sounds ok to me (we discussed this sort of thing before implementing this, we talked about sending a param to the server connector ctr.  but it had the drawback of requiring baseAgentId at EyesBase creation time [if you remember] - which eventually is now the case anyhow, so makes sense)

PR #159, opened 2020-04-17
Set viewport size logic

kyryloonufriiev:
As I described in [Trello card](https://trello.com/c/khwqmB3D/318-ms-edge-missing-screenshots-fatal-errors?menu=filter&filter=member:kyrylo36) we have a problem with viewport size setting logic.
I want to discuss how do we want to fix it.
If we need a fast fix, I suggest saving to the cache an actual value, not one which user pass. As a more holistic solution, I think we need to develop a more transparent logic around a viewport size setting. Fortunately, I already rewrite an algorithm itself as a part of moving it to the core. I hope it could happen after we move eyes implementation to the core.

What do you think?


amitzur:
There's no test here that validates the exception for the customer doesn't happen. This is the exception:
```
[MicrosoftEdge  windows #0-0] OutOfBoundsError: Region [(514, 965) 626x243] is out of screenshot bounds [(0, 0) 1280x960]
```
If I comment out lines 1108-1109 in EyesWDIO and the assertion in the test, it passes. I would like it to fail.

kyryloonufriiev:
This PR is closed because the bug is already fixed.

PR #164, opened 2020-04-27
Changes to make compatible with JSDOC and the EKB parserNote - I reca…

yonatanLehman:
All changes impact only jJSDOV except for the removal of the _ from parameter names in eyes.check to match JSDOC and other instances of the check call as agreed with Amit.

Note that some of these changes were already fixed in the past in the initial EKB integration with WDIO5 - it seems that there was some regression at some point. 

Changes were:
1) Don't document the possibility of passing a Configuration object to eyes.open
2) Add '-' before free text in @param and @return if the line is extended then the - can't be at the end of the line - it should be followed by at least one word of the free text.
3) Added @private on non user-accessible methods
4) Removed incorrect '.' in Promise.<*>to Promise<*>
5) moved incorrect ... from after type to before type
6) Added @return and @param value where it is missing and the type is obvious
7) Added {TBDTYPE} where jsdoc is missing type information and it is not obvious what the type should be. This needs to be updated by the developer

### DIFF
--- a/packages/eyes-selenium/lib/Eyes.js
+++ b/packages/eyes-selenium/lib/Eyes.js
@@ -142,7 +142,8 @@ class Eyes extends EyesBase {
    * @param {SeleniumCheckSettings} checkSettings - Target instance which describes whether we want a window/region/frame
    * @return {Promise<MatchResult>} - A promise which is resolved when the validation is finished.
    */
-  async check(_name, _checkSettings) {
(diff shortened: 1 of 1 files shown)


amitzur:
@yonatanLehman can you recreate the PR? I don't know why git is failing for me and it closed the PR automatically..

amitzur:
@yonatanLehman forget about it. I'm pushing it to master and we'll continue from there

PR #165, opened 2020-04-27
Custom tests updates

Artem-applitools:
Adding new tests for the WDIO5 and updated tests to run only success tests


packages/eyes-webdriverio-5/package.json
```diff
@@ -36,7 +36,9 @@
     "@applitools/eyes-sdk-core": "9.2.1",
     "@applitools/visual-grid-client": "13.7.2",
     "selenium-webdriver": "^4.0.0-alpha.5",
-    "webdriverio": "^5.18.6"
+    "webdriverio": "^5.18.6",
+    "chai": "^4.2.0",
```
amitzur, line 40: chai and chai-as-promised should be a dev dependency

packages/eyes-selenium/test/coverage/custom/TestAPI.spec.js
```diff
@@ -0,0 +1,27 @@
...
+  beforeEach(async () => {
+    webDriver = await getDriver('CHROME')
+    ;({eyes} = await getEyes('classic', 'CSS'))
+  })
+  afterEach(async () => {
+    await eyes.abortIfNotClosed()
+    await webDriver.quit()
+  })
+
+  it('TestIsDisabled', async () => {
```
amitzur, line 16: What is this testing? That we don't get an exception? We might as well ignore the isDisabled flag and the test will still pass.

Artem-applitools: Yeap, it check that there will be no exceptions. There are the difference on the page used in the tests, different number on refresh. So if eyes aren't disabled then test will fail. Also there used locators which are absent on the page. On Selenium 4 SDK if the flag wasn't set test will fail.

PR #170, opened 2020-05-04
coverage-tests: `--all` flag

kyryloonufriiev:
Here I added an `--all` for ignoring the `disabled` flag on supported tests.
I have a question in addition. Why the filtering out of disabled tests is not a part of `filterTests` function? Does it have some reason?

### DIFF
--- a/packages/eyes-webdriverio-4/package.json
+++ b/packages/eyes-webdriverio-4/package.json
@@ -11,7 +11,7 @@
     "test:e2e": "mocha --no-timeouts \"test/e2e/*.js\"",
     "test:it": "mocha --no-timeouts \"test/it/**/*.js\"",
     "test:unit": "mocha \"test/unit/**/*.js\"",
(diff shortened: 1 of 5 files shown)


packages/eyes-webdriverio-4/package.json
```diff
@@ -11,7 +11,7 @@
     "test:e2e": "mocha --no-timeouts \"test/e2e/*.js\"",
     "test:it": "mocha --no-timeouts \"test/it/**/*.js\"",
     "test:unit": "mocha \"test/unit/**/*.js\"",
-    "coverage:sandbox": "coverage-tests create-tests --remote=$CVG_TEST_REMOTE && yarn test:coverage && coverage-tests process-report",
+    "coverage:sandbox": "coverage-tests create-tests --remote=$CVG_TEST_REMOTE --all && yarn test:coverage && coverage-tests process-report",
```
tourdedave, line 14: I'm not sure we'd want to add `--all` to this script. It's meant to mirror what would be run for a production run, but send the results to the `sandbox` QA report.

kyryloonufriiev: Absolutely agree. I just forget to remove it from the script.

PR #171, opened 2020-05-05
fix coverage tests

amitzur:
1. Baselines were updated by Daniel P. for the following tests:
  - TestCheckFrameInFrame_Fully_Fluent
  - TestCheckFrameInFrame_Fully_Fluent2
So I renamed them to have prefix of `_v1` to show that it was first (it's not `_v2`)

2. The following tests were written with error after #153:
  - TestCheckElementFully_Fluent
  - TestCheckRegionByCoordinates_Fluent

3. I changed `TestCheckFrameHideScrollbars_Fluent` to `TestCheckFrame_Fluent_v2` and deleted `TestCheckFrame_Fluent` as it will not pass in JS SDK's.


tourdedave:
LGTM

amitzur:
I addressed @tourdedave 's comment 1. I think we can merge without comment 2, and solve that later. I'm also in favor of option (a). Not sure about the parens format. Where do we parse the test name today?

PR #172, opened 2020-05-06
Test updates

Artem-applitools:
This PR contains optimizations of the tests for the mocha-parallel-tests and adding usage of the 'APPLITOOLS_API_KEY_SDK' env variable for the S3 and S4 packages.


Artem-applitools:
I have added env variable CVG_TESTS_REMOTE to the custom tests so it was easy use them as it done with generic tests.

PR #173, opened 2020-05-06
XUnit report formatter

tourdedave:


packages/eyes-storybook/src/handleXmlFile.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+const fs = require('fs');
+const {resolve} = require('path');
+
+function handleXmlFile(xmlFilePath, formatter, {suiteName = 'Eyes Storybook', totalTime} = {}) {
```
amitzur, line 6: I don't see much point in exposing `suiteName` to this function. You can hard code it inside the function. It would be easy enough to later extract that into external configuration (and do the same as what you did here), but right now since it's not actually used I found it more confusing and feels premature.

packages/eyes-sdk-core/test/lib/TestResultsFormatter.spec.js
```diff
@@ -0,0 +1,279 @@
...
+          duration: 10,
+        }),
+        new TestResultsError({
+          name: 'My Component | Button1',
+          message: 'some error message',
+        }),
+      ]
+      const expected = `<?xml version="1.0" encoding="UTF-8" ?>
+<testsuite name="blah" tests="2" time="20">
+<testcase name="My Component | Button2" time=10>
```
amitzur, line 266: should be `time="10"` ?

packages/eyes-sdk-core/lib/TestResultsFormatter.js
```diff
@@ -240,6 +240,28 @@ class TestResultsFormatter {
...
     return tapString
   }
+
+  toXmlOutput({suiteName, totalTime} = {}) {
+    let output = `<?xml version="1.0" encoding="UTF-8" ?>`
+    const testResults = this._resultsList
+    output += `\n<testsuite name="${suiteName}" tests="${testResults.length}" time="${totalTime}">`
+    testResults.forEach(result => {
+      const duration = result.getDuration()
+      output += `\n<testcase name="${result.getName()}"${duration ? ' time=' + duration : ''}>`
```
amitzur, line 250: is this missing a `"`? `${duration ? ' time="' + duration + '"' : ''}`

packages/eyes-sdk-core/lib/TestResultsFormatter.js
```diff
@@ -240,6 +240,28 @@ class TestResultsFormatter {
...
+      if (result.getIsDifferent()) {
```
amitzur, line 251: what about `if (result.isAborted())` ?

packages/eyes-sdk-core/lib/TestResults.js
```diff
@@ -974,5 +974,14 @@ class TestResults {
   }
 }
 
+class TestResultsError extends TestResults {
+  constructor({name, message} = {}) {
```
amitzur, line 978: I think the error itself should be passed here. If a consumer is only interested in the message, they can get it from the error.
I understand the need to keep the `name`. So I'd do:
```
constructor({name, error})
```

packages/eyes-sdk-core/lib/TestResultsFormatter.js
danielschwartz85: or if failed:
see packages/eyes-sdk-core/lib/TestResultsFormatter.js:146

tourdedave: I was going off of what was done in processResults in eyes-storybook. There's no explicit distinction of it there. But I suppose it's probably handled implicitly somehow. I'll take a look.

packages/eyes-sdk-core/lib/TestResultsFormatter.js
tourdedave: It would appear so.

packages/eyes-sdk-core/lib/TestResults.js
tourdedave: 👍

packages/eyes-storybook/src/handleXmlFile.js
tourdedave: Makes sense.

packages/eyes-sdk-core/lib/TestResultsFormatter.js
amitzur: I actually don't know how `processResults` behaves with aborted tests. Curious to find out...

packages/eyes-storybook/src/processResults.js
```diff
@@ -19,10 +19,10 @@ function processResults({results = [], totalTime, concurrency}) {
 
   outputStr += '[EYES: TEST RESULTS]:\n\n';
   if (passedOrNew.length > 0) {
-    outputStr += testResultsOutpot(passedOrNew);
+    outputStr += testResultsOutput(passedOrNew);
   }
   if (unresolved.length > 0) {
-    outputStr += testResultsOutpot(unresolved);
```
danielschwartz85, line 25: re the function name It was meant to suggest that pot is bad (therefor it should be out). but I can see how this can be confusion..

packages/eyes-storybook/src/processResults.js
```diff
@@ -75,6 +75,14 @@ function processResults({results = [], totalTime, concurrency}) {
 
   passedOrNew.forEach(formatter.addTestResults.bind(formatter));
   unresolved.forEach(formatter.addTestResults.bind(formatter));
+  errors.forEach(error => {
+    formatter.addTestResults(
```
danielschwartz85, line 79: basically we can now use the TestResultsError to save the story names instead of creating objects like this `{name, errorOrResult}`, and then here in processResults we would just do `errors.forEach(formatter.addTestResults.bind(formatter));`
(and change the way we determine if we have an error for the CLI output)

See, how we save story name and error here (can create a TestResultsError instead):  
/eyes-storybook/src/renderStories.js@onDoneStory

but we can leave this to once we implement TestResultsError in the VGC (then we would need change the processResults function to account for TestResultsError).

danielschwartz85:
> Nice and clean.
> Only one change that I think should be done in passing the error to `TestResultsError`.
> 
> @danielschwartz85 FYI - I think this now lets us return `TestResultsError` from vgc instead of just the error object. The VGC would connect the error with the test name provided to `openEyes` and return homogenous array of `TestResults` instances (TestResultsError is an instance of TestResults).
> This might be a big change - it's not urgent, but it would make things a lot cleaner. For example storybook's `processResults` won't have to create `new TestResultsError` from the error because it will already have it.

right, and once done in VGC we can then create TestResultsError in storybook and then processResults  would just get an array of TestResults (i.e. this function would not create the error objects like it does today and SB would not need to create objects like `{errorOrResuslt, storyName}`)

tourdedave:
Seeing some test failures when I run `yarn test` which seem unrelated to my changes. Need to spend some time looking into that.

I also want to step through a sandbox example using eyes-storybook with the XML output config.

PR #174, opened 2020-05-07
startInfo-displayName-support

rannn505:
Instead of doing something like this:

```javascript
matchSingleWindowData._startInfo._displayName = displayName;
```

### DIFF
--- a/packages/eyes-sdk-core/lib/metadata/StartInfo.js
+++ b/packages/eyes-sdk-core/lib/metadata/StartInfo.js
@@ -14,6 +14,7 @@ class StartInfo {
    * @param {string} appIdOrName
    * @param {boolean} compareWithParentBranch
    * @param {string} scenarioIdOrName
+   * @param {string} displayName
    * @param {BatchInfo|object} batchInfo
    * @param {AppEnvironment|object} environment
    * @param {MatchLevel|string} matchLevel
@@ -29,6 +30,7 @@ class StartInfo {
     appIdOrName,
     compareWithParentBranch,
     scenarioIdOrName,
+    displayName,
     batchInfo,
     environment,
     matchLevel,
@@ -55,6 +57,7 @@ class StartInfo {
     this._appIdOrName = appIdOrName
     this._compareWithParentBranch = compareWithParentBranch
     this._scenarioIdOrName = scenarioIdOrName
+    this._displayName = displayName
     this._batchInfo = batchInfo
     this._environment = environment
     this._matchLevel = matchLevel
@@ -146,6 +149,20 @@ class StartInfo {
    */
   setScenarioIdOrName(value) {
     this._scenarioIdOrName = value
+  }
+
+    /**
+   * @return {string}
+   */
+  getDisplayName() {
+    return this._displayName
+  }
+
+  /**
+   * @param {string} value
+   */
+  setDisplayName(value) {
+    this._displayName = value
   }
 
   /**


amitzur:
@rannn505 this class represents the value returned from an API call to retrieve the test's metadata. Is that your intention in this PR? Is it something you use?

While this PR is definitely valid, from your comment above I think this file is more relevant: https://github.com/applitools/eyes.sdk.javascript1/blob/master/packages/eyes-sdk-core/lib/server/SessionStartInfo.js#L15
That file consists of the information sent to Eyes server in the *open* operation.

I'm going to merge this either case.

rannn505:
hey @amitzur 
First of all thank you for merging my PR (Can you tell when the NPM version is due?).

Secondly, I use eyes SDK as follows:
```js
    const {
      ServerConnector,
      Configuration,
      Logger,
      MatchSingleWindowData,
      metadata: {
          StartInfo,
      },
      AppOutput,
    } = require('@applitools/eyes-sdk-core');

    const eyes = new ServerConnector(
        new Logger(),
        new Configuration({
            serverUrl: url,
            apiKey: runKey,
        }),
    );

    const startInfo = new StartInfo({
        agentId: 'testim.io',
        verId: '3.0.0',
        appIdOrName: appId,
        batchInfo: {
            id: batchInfoId,
            name: `${testimTestName} - ${batchInfoId}`,
            startedAt: (new Date()).toUTCString(),
        },
        scenarioIdOrName: applitoolsTestId, // testim step === applitools test
        **displayName, // step name - this was missing ** 
        environment: {
            displaySize: { height, width },
            hostingApp: browser,
            inferred: false,
            os,
        },
        defaultMatchSettings: {
            matchLevel,
        },
        isTransient: forceBaseline,
        branchName: branch,
        parentBranchName: null,
    });

    const appOutput = new AppOutput({
        title: displayName,
    });
    if(isBase64(image)) {
        appOutput.setScreenshot64(new Buffer(image, 'base64'));
    } else {
        appOutput.setScreenshotUrl(image);
    }

    const matchSingleWindowData = new MatchSingleWindowData({
        startInfo,
        userInputs: [],
        appOutput,
        ignoreMismatch,
    });
    matchSingleWindowData.setUpdateBaselineIfNew(forceBaseline);
    matchSingleWindowData.setUpdateBaselineIfDifferent(forceBaseline);
    **matchSingleWindowData._startInfo._displayName = displayName; // workaround which is the reason for this PR**

    try {
        logger.info("matchExpectedOutputAsSession start", { metadata, image, ignoreMismatch, forceBaseline });
        **let response = await eyes.matchSingleWindow(matchSingleWindowData);**
        response = response.toJSON();
        logger.info("matchExpectedOutputAsSession done", { response });
        return response;
    } catch (error) {
        logger.info("matchExpectedOutputAsSession fail", { error });
        handleError(error);
    }
```

Is there a better way to do this? Basically this code is pretty much built on looking at your code (unfortunately there are no examples).

Thx,
Ran

amitzur:
One of the reasons there are no examples is that this is an internal package and we never made effort to document it.
I recommend replacing `StartInfo` with `SessionStartInfo` like so:
```
    const {
      ServerConnector,
      Configuration,
      Logger,
      MatchSingleWindowData,
      SessionStartInfo,
      AppOutput,
    } = require('@applitools/eyes-sdk-core');
```
The swap would be seamless and `displayName` is supported even without a new NPM version.

PR #176, opened 2020-05-12
Shared custom coverage tests

amitzur:
This is a quick and dirty demo for how we can do shared custom coverage tests.

1. The SDK needs to externalize its EyesWrappedDriver in `wrappers/interface.js` and provide a `getDriver` function in `test/util` (everything is debatable of course).

2. The controller gets another function, `visit`.

3. The actual custom tests (and its future utilities) are written in the `sdk-coverage-tests` package.

### DIFF
--- a/packages/eyes-webdriverio-4/package.json
+++ b/packages/eyes-webdriverio-4/package.json
@@ -14,7 +14,7 @@
     "coverage:sandbox": "coverage-tests create-tests --remote=$CVG_TEST_REMOTE && yarn test:coverage && coverage-tests process-report",
     "coverage:prod": "coverage-tests create-tests --remote=$CVG_TEST_REMOTE && yarn test:coverage && coverage-tests process-report --send-report prod",
     "test:coverage": "yarn start-chromedriver && env APPLITOOLS_BATCH_NAME='JS Coverage Tests: eyes-webdriverio-4' APPLITOOLS_BATCH_ID=$(uuidgen) XUNIT_FILE=coverage-test-report.xml mocha-parallel-tests --no-timeouts --reporter spec-xunit-file --max-parallel 15 test/coverage/generic/*.spec.js",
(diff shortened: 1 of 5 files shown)


kyryloonufriiev:
Realy great idea!
As far as I understand, we will able to manipulate the page only through our abstractions. 
This means that we need to implement every function that we would like to use in the test just next to the functions which serve the business logic. I'm not a big fan of the file which serves for test and bl purposes.
Another point I want to raise, if we will use only our internal abstractions this means that we are not really testing a user API, but our internal.
I would like to suggest something that will consolidate custom and generic tests under the same interface. I think generic tests created a very good basis, which shows us how we can wrap frameworks API and use it in the same way for all SDKs. I would like to create another interface for test purposes only, let's say it will call `TestController`, it will describe a contract for full (as much as we need for the tests) control of the driver. Each SDK will have two implementations of this interface. The first one is `GenericTestController`, pretty much the same as we have in the `test/coverage/index.js`. The second one is `CustomTestController` for the implementations using plain js.
In my opinion, this will make both types of coverage tests are similar in implementation, and the most important it will distantiate custom coverage tests from the internal API.

WDYT @amitzur @danielschwartz85 @tourdedave?

amitzur:
@kyryloonufriiev very valid points, I am very much in agreement with your comment.
I'm less in agreement with the code demo :)

Here's another perspective - I pushed your commit to another branch, `shared-custom-coverage-tests-2` so it's not lost. Have a look at `TestDriver` which replaces the controller.

I might go so far as actually using this test driver in the generic tests, and then code emission is the same for all mocha-run SDK's

kyryloonufriiev:
I like very much the idea to use `TesDriver` in generic tests as well. This frees us from having to do the same for both types of tests (e.g. `switchToFrame`, `setup`...). How do you want to achieve this?

PR #177, opened 2020-05-12
Azure retry

danielschwartz85:
Today we retry once on every failure, for 404, 500, 502 and 504.
Under current impl - we would retry also on 503 for all request.
Under current impl - if we change retry fo upload snapshot to 10 times we would get this retry also for 404 (for upload snapshot). maybe we should customize a solution for per status retry ?
Under current impl - we would get retry also for non Azure upload snapshot, is that ok ?


packages/eyes-sdk-core/lib/server/ServerConnector.js
```diff
@@ -300,6 +304,7 @@ class ServerConnector {
         'x-ms-blob-type': 'BlockBlob',
         'content-type': 'application/octet-stream',
       },
+      ...AZURE_RETRY_CONFIG,
```
danielschwartz85, line 307: Can we retry on this request regardless of weather the url is of Azure ?
(i.e. for simplicity instead of checking the url domain for weather that's Azure or not)

packages/eyes-sdk-core/lib/server/requestHelpers.js
```diff
@@ -10,13 +10,15 @@ const HTTP_STATUS_CODES = {
   NOT_FOUND: 404,
   INTERNAL_SERVER_ERROR: 500,
   BAD_GATEWAY: 502,
+  SERVICE_UNAVAILABLE: 503,
```
danielschwartz85, line 13: So now every request that uses the retry (on fail) mechanism will also do a retry on 503 as well, wdty ?

see:
https://github.com/applitools/eyes.sdk.javascript1/pull/177/files#diff-763eec627d678a376f8ab2e4e27cfe78R171

packages/eyes-sdk-core/lib/server/ServerConnector.js
```diff
@@ -47,6 +47,11 @@ const HTTP_STATUS_CODES = {
   GATEWAY_TIMEOUT: 504,
 }
 
+const AZURE_RETRY_CONFIG = {
```
danielschwartz85, line 50: Also all failures of the ```uploadScreenshot``` will now do a retry - so also a 404 (since this is how the err retry mechanism works, see notes below). So if would change these parameters to reties (which makes sense for 503) this would do a retry of 10 on 404s as well.

We can create a more "complicated" mechanism something like the request would send ```retryForStatus: [{status: 503, delayBeforeRetry: 500}]``` and then be more accurate on failures.
(I think this is a good idea). wdyt ?

amitzur:
> Under current impl - if we change retry fo upload snapshot to 10 times we would get this retry also for 404

I didn't understand this sentence. Can you explain?

I think the changes in `ServerConnector` are unnecessary. Only the ones in `requestHelpers`. It's okay to have the same retry behavior for all reqeusts, IMO.

danielschwartz85:
> > Under current impl - if we change retry fo upload snapshot to 10 times we would get this retry also for 404
> 
> I didn't understand this sentence. Can you explain?

I meant that any change I make to the retry count in uploadScreenshot() method would affect all retries of that method (i.e. a retry also for 404 and not only 503) - but based on you other comment I think it's ok.

Will remove serverConnector stuff; so we will do retries for 503 for all request (and we will have 3 retries for Azure uploads). :+1:

packages/eyes-sdk-core/lib/server/ServerConnector.js
amitzur: yes, sure

packages/eyes-sdk-core/lib/server/ServerConnector.js
amitzur: ahhh now I understand. I think 3-5 retries with 500ms between them is a good catch-all solution. WDYT?

amitzur:
Ok, sorry to make a mess out of this. I didn't fully understand what you meant at first. I thought the changes in `ServerConnector` create a precedent of different behavior between Azure and other endpoints. But I see now that it's already defining a different number of retries, and you just made that 5 instead of 3, and with a delay before retry (which is a much more sensible default IMO).

What caught my eye is that the variable is called `AZURE_RETRY_CONFIG`. I think this is the sensible retry configuration for all requests that should have retries. And I see also that some requests that shouldn't have retries (matchWindow, close, matchSingleWindow, render(?)) - do have them. But that's a different story.

PR #178, opened 2020-05-18
dom-capture IE fix

tourdedave:
Fixes all JS SDKs that use dom-utils to leverage dom-capture when running on IE.

Also fixes WDIO4 dom-capture, which was broken due to an unimplemented API call.


packages/eyes-webdriverio-4/src/wrappers/WDIODriver.js
```diff
@@ -68,6 +68,14 @@ class WDIODriver extends EyesWrappedDriver {
...
     return this._driver.url(url)
   }
+
+  async getCapabilities() {
+    return this._driver.desiredCapabilities
+  }
+
+  async getCurrentUrl() {
+    return this._driver.getUrl()
+  }
```
kyryloonufriiev, line 78: Please move these methods to the `LegacyApiDriver`.

tourdedave: 👍

PR #179, opened 2020-05-19
selenium4: migrate to new core

kyryloonufriiev:
I already сreated all the necessary files to specialize abstractions from the `core`. But it still should wait until `core` will be updated to support async `specs.extractId(el)`.


packages/eyes-selenium/src/selenium4/SpecWrappedElement.js
```diff
@@ -30,7 +30,8 @@ module.exports = {
     return (
       selector instanceof By ||
       TypeUtils.has(selector, ['using', 'value']) ||
-      Object.keys(selector).some(key => key in By)
+      Object.keys(selector).some(key => key in By) ||
```
amitzur, line 33: @kyryloonufriiev function returns true if there's at least one key in `selector` that is also in `By`? How is this logic correct?

packages/eyes-selenium/test/it/VisualGridResources.spec.js
```diff
@@ -54,7 +54,7 @@ describe('VisualGrid Resources', () => {
 
     await eyes.check('region', Target.region(new Region(200, 200, 50, 100)))
 
-    await eyes.check('selector', Target.region('#scroll1'))
+    await eyes.check('selector', Target.region({css: '#scroll1'}))
```
amitzur, line 57: @kyryloonufriiev wait, the API used to support getting a string and now it doesn't?

packages/eyes-selenium/src/selenium4/SpecWrappedElement.js
kyryloonufriiev: You can see it in the Selenium docs. They support objects like this `{css: '.some'}`

packages/eyes-selenium/test/it/VisualGridResources.spec.js
kyryloonufriiev: Only Selenium4. I preserve strings support on the Selenium4, but I fix the test to work also with Selenium3.

PR #180, opened 2020-05-19
Edge screenshot retry

tourdedave:
Taking screenshots on Microsoft Edge sometimes yields blank images. This adds a one-time retry when that happens.

Note that it's hard to reliably reproduce this issue. So to start, there is a placeholder test that walks through the new code path end-to-end.

### DIFF
--- a/packages/eyes-webdriverio-5/package.json
+++ b/packages/eyes-webdriverio-5/package.json
@@ -35,6 +35,7 @@
     "@applitools/dom-utils": "4.7.17",
     "@applitools/eyes-sdk-core": "10.2.0",
     "@applitools/visual-grid-client": "14.1.0",
+    "@applitools/bitmap-commons": "1.0.0",
     "selenium-webdriver": "^4.0.0-alpha.5",
     "webdriverio": "^5.18.6"
   },
--- a/packages/eyes-webdriverio-5/src/capture/TakesScreenshotImageProvider.js
+++ b/packages/eyes-webdriverio-5/src/capture/TakesScreenshotImageProvider.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const {ImageProvider, MutableImage} = require('@applitools/eyes-sdk-core')
+const {isBlankImage} = require('@applitools/bitmap-commons')
 
 /**
  * An image provider based on WebDriver's interface.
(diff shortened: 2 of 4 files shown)


amitzur:
I agree with the implementation.

1. I'd rather not use `@applitools/bitmap-commons`, since its code sits in a private repo, and really not needed fully but only for a specific need. So package versioning etc. just for a utility function.
We already have image related utilities in `ImageUtils.js` that use `png-async`. We can borrow code that's missing into that file.

2. In order to test stably we would need to mock.

3. In #167 this is already in core, so maybe create a PR onto that branch?

kyryloonufriiev:
I would like to see a separate `EdgeImageProvider` for this. I think this make sense since we already do this for Firefox and Safari.
Does this happens on EdgeLegacy or EdgeChromium as well? We could check it in `ImageProviderFactory`

tourdedave:
re: 1:
- How is this different than us using the feature-flags library from mono?
- True, it is a utility function. But it's well tested and is used for the exact same use case in other parts of the org. Seems like we would all benefit from sharing this code rather than duplicating it.

re: 2 & 3. -- Sounds good.

tourdedave:
New PR -> https://github.com/applitools/eyes.sdk.javascript1/pull/181

amitzur:
> - How is this different than us using the feature-flags library from mono?
> - True, it is a utility function. But it's well tested and is used for the exact same use case in other parts of the org. Seems like we would all benefit from sharing this code rather than duplicating it.

Fair points, I'll explain:
1. Using feature-flags was intended as a service to the vg team, to solve the problem of early release in vg-cli. As they were the target audience, it made sense to use their own mechanism for their convenience. I was still facing the same issue of git being private. But in my head I rationalized it as a tradeoff.

2. bitmap-commons also depends on `png-async`. Which means if versions are not aligned we will have 2 copies of that package. Not terrible, but avoidable. And if we want to mitigate it we'll have to occasionally go to mono, update dep, release from there, hope we don't break vg, etc.

3. isBlankImage is indifferent to what previous work was done on the image. It parses the buffer into a png object regardless of whether it was already parsed. This may incur a performance overhead. In `MutableImage` there is a check `if _isParsed` to avoid parsing again.

This is analogous to the decision to use `left-pad` or any other granular npm module. Sure, it's fully tested, but there is sometimes a tax to pay, and the decision to do it should be done with proper reasoning. Sometimes duplicating code is a fine solution.

Don't get me wrong - I think everyone in the VG team produces pure gold code (no cynicism) - but it's not about the quality of the code.

Makes sense?

PR #181, opened 2020-05-20
Edge screenshot retry v2

tourdedave:
Reimplemented screenshot retries for EdgeClassic on top of the eyes-core branch.

Still need to:
- [ ] confirm if this happens on EdgeChromium as well as EdgeClassic (?)
- [ ] add test coverage for this (with mocking)
- [ ] close the loop on using bitmap-commons or pulling the code into core


tourdedave:
This is no longer needed.

PR #182, opened 2020-05-20
VGC - 409 Conflict fix

danielschwartz85:
Clearing resource cache content only after render request is done.


danielschwartz85:
merged manually

PR #183, opened 2020-05-25
Chromedriver multi-version support

tourdedave:
With yarn we can install multiple versions of the same package -- each with their own alias that can be required on its own ([link](https://classic.yarnpkg.com/en/docs/cli/add/#toc-yarn-add-alias)).

This is an example of doing it for `chromedriver`, wired up for use in the WDIO5 SDK.

### DIFF
--- a/packages/eyes-webdriverio-5/example/EyesClassicRunnerExample.js
+++ b/packages/eyes-webdriverio-5/example/EyesClassicRunnerExample.js
@@ -1,6 +1,7 @@
 'use strict'
 
-const chromedriver = require('chromedriver')
+const {makeChromeDriver} = require('@applitools/sdk-shared')
+const chromedriver = makeChromeDriver()
 const {remote} = require('webdriverio')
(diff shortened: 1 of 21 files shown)


kyryloonufriiev:
I thought we want to do that with docker containers.

tourdedave:
Ideally, yes. This is an incremental improvement that will help fix the immediate issue of differences between each of our machines (e.g., Linux tends to lag behind in the Chrome version that's available). There are also still a fair number of tests that rely on spinning up chromedriver locally across all SDKs.

danielschwartz85:
Looks good

amitzur:
@tourdedave can we close this in favor of working on docker containers?

tourdedave:
Depends on how long until we implement containers.

amitzur:
Next week max

tourdedave:
Works for me. We can close this.

amitzur:
Cool. I think the API of `CHROME_MAJOR_VERSION` still has merit. I just imagine it spinning the right docker container.
I'm aware that it's a performance overhead though and this PR might be a better solution for local chrome.
Let's not close this yet, and revisit when work on docker is done.

tourdedave:
Closing this since I don't think we'll need it.

PR #186, opened 2020-05-27
Show edgechromium versions in help message

amirhalatzi:
Edge Chromium was filtered out, and now that it is officially supported should not be filtered


danielschwartz85:
@amirhalatzi `edgechromium` and `edgelegacy` are shown in available browser list message.
by removing the filter line you enable `edge` to be in the available browser list message, but `edge`is deprecated.

amirhalatzi:
Sorry. Apparently I was using an older version

PR #189, opened 2020-06-07
wdio5: added fallback support for window size get/set functions

tourdedave:
re: [Trello 382](https://trello.com/c/UBr0w3UF/382-wdio6-with-wdio5-service-viewport-size-not-being-applied)

When using WDIO against certain drivers the functions used to get and set the window size are not implemented which causes setting the viewport size to fail silently for users.

This appears to be an issue specific to WDIO5 & later.


packages/eyes-webdriverio-5/test/it/SpecWrappedDriver.spec.js
```diff
@@ -179,6 +179,41 @@ describe('SpecWrappedDriver', async () => {
     })
   })
 
+  describe('browser with unimplemented methods', async () => {
```
amitzur, line 182: this test will be relevant until (if at all) Sauce updates the driver version, right?

packages/eyes-webdriverio-5/src/SpecWrappedDriver.js
```diff
@@ -52,11 +63,11 @@ module.exports = {
     return driver.setWindowRect(location.x, location.y, null, null)
   },
   async getWindowSize(driver) {
-    const rect = await driver.getWindowRect()
```
amitzur, line 55: I personally think it's fine putting the ternary inside this method and avoid creating the oneline function above.

packages/eyes-webdriverio-5/src/SpecWrappedDriver.js
```diff
@@ -52,11 +63,11 @@ module.exports = {
...
+    const rect = await getWindowRect(driver)
     return {width: rect.width, height: rect.height}
   },
   async setWindowSize(driver, size) {
-    return driver.setWindowRect(null, null, size.width, size.height)
```
amitzur, line 59: same comment as in `getWindowSize`

packages/eyes-webdriverio-5/test/it/SpecWrappedDriver.spec.js
tourdedave: Yes. But, I'm assuming that this setup has the problem based on customer feedback. I didn't confirm. I added the test after implementing the fix. Worth checking since it adds some overhead to the run-time.

packages/eyes-webdriverio-5/src/SpecWrappedDriver.js
tourdedave: Agreed.

I separated it when I thought it would be used more than once (e.g., also for `getWindowLocation` and `setWindowLocation`) but I realized that wasn't possible since the fallback doesn't get you `x` and `y`. I just never put it back.

tourdedave: Confirmed.

packages/eyes-webdriverio-5/test/it/SpecWrappedDriver.spec.js
tourdedave: Confirmed.

PR #190, opened 2020-06-07
Selenium 3 with new core

amitzur:
this PR fixes some tests so that running the following with `SELENIUM_MAJOR_VERSION=3` passes:
1. `yarn coverage:sandbox`
2. `yarn test`

The only point left hanging is how do we run the suite of tests for both Selenium 3 and 4 before publishing this package. Any suggestions?


amitzur:
Worth noting that the agent ID is the same. Need to communicate this with product

tourdedave:
> Worth noting that the agent ID is the same. Need to communicate this with product

Seems like we'd want to track if the user is running with a different major driver version, no?

amitzur:
Oh, there's one very **big** item left - how we identify which driver version to use? Right now it's based on an environment variable, but we need to change it to read from `selenium-webdriver` package.json.

I'll do my suggestion in the following commit, please let me know what you think @kyryloonufriiev @tourdedave

amitzur:
Sorry, I now realized that this is implemented in `index.js`:
```
if (!process.env.SELENIUM_MAJOR_VERSION) {
  const {version} = require('selenium-webdriver/package.json')
  const [major] = version.split('.', 1)
  process.env.SELENIUM_MAJOR_VERSION = major
}
```
I think it's good enough. I don't mind risking setting the process' env var. Do you think we should rename it to `APPLITOOLS_SELENIUM_MAJOR_VERSION` ?

amitzur:
Regarding the agent ID:
The decision with product is to keep both the SDK version and the driver's major version inside the agent ID, in a way that's inspectable from the marketing tools we use. So:
- Selenium 4: `eyes.selenium.javascript/4.x.x--4`
- Selenium 3: `eyes.selenium.javascript/4.x.x--3`

And the same for the visual grid equivalents. We can go ahead and implement this and merge this PR so that we formally have this as the SDK for Selenium 3!

tourdedave:
Seems like a pretty good practice to use a vendor prefix for env vars that we introduce.

amitzur:
@kyryloonufriiev please review - PR is done

packages/eyes-selenium/src/selenium3/SpecWrappedDriver.js
```diff
@@ -62,11 +70,18 @@ module.exports = {
...
+        .manage()
+        .window()
+        .getSize()
+      return {width, height}
+    } catch (err) {
+      // workaround for Appium
+      return driver.execute(
+        new cmd.Command(cmd.Name.GET_WINDOW_SIZE).setParameter('windowHandle', 'current'),
+      )
+    }
```
kyryloonufriiev, line 84: Why do we need this workaround? I added it for selenium 4 because it doesn't have `.getSize()` method.

packages/eyes-selenium/src/SeleniumSpecializedEyes.js
```diff
@@ -7,13 +7,13 @@ const CheckSettings = require('./SeleniumCheckSettings')
 const {version} = require('../package.json')
 
 const SeleniumEyesClassic = EyesClassic.specialize({
-  agentId: `eyes.selenium.javascript/${version}`,
+  agentId: `eyes.selenium.javascript/${version}--${process.env.APPLITOOLS_SELENIUM_MAJOR_VERSION}`,
```
kyryloonufriiev, line 10: Maybe it should be `eyes-selenium`, same as the package name?

amitzur: that doesn't really matter. Product team tracks this value to see usage stats. it makes more sense in the overall picture since other languages have `eyes.selenium.ruby`, `eyes.selenium.java` etc.

packages/eyes-selenium/src/selenium3/SpecWrappedDriver.js
amitzur: no need, since there were other things I copy pasted from Selenium 4 spec driver, I also took this one. I will remove it

PR #191, opened 2020-06-10
Stitching improvements

kyryloonufriiev:
This PR will improve the way how we take screenshots of the element with the fixed position also it will add support of the scroll root elements with a position different from (0, 0).


packages/sdk-coverage-tests/coverage-tests/custom/TestFluentApi.spec.js
```diff
@@ -14,41 +14,12 @@ describe(appName, () => {
...
 
-  describe.skip(`Test`, () => {
+  describe(`Test`, () => {
     beforeEach(async () => {
       webDriver = await spec.build({capabilities: Browsers.chrome()})
       await spec.visit(webDriver, 'https://applitools.github.io/demo/TestPages/FramesTestPage/')
       eyes = await getEyes({isCssStitching: true})
-    })
-
-    it('TestCheckRegionInFrame2_Fluent', async () => {
```
amitzur, line 24: This test was removed because it is redundant? I'm just verifying, since I see that `TestCheckRegionInFrame_Fluent` is identical. And in addition the c# implementation is different (has multiple check of the same region only with different ignore regions, which in itself looks like a pretty unnecessary test).

packages/sdk-coverage-tests/coverage-tests/custom/TestFluentApi.spec.js
```diff
@@ -14,41 +14,12 @@ describe(appName, () => {
...
-        'Fluent - Inner frame div 1',
-        Target.frame('frame1')
-          .region('#inner-frame-div')
-          .fully()
-          .ignoreRegions(new Region(50, 50, 100, 100)),
-      )
-      await eyes.close()
-    })
-
-    it('TestCheckRegionInFrameInFrame_Fluent', async () => {
```
amitzur, line 39: why remove this test? I don't see an equivalent test in our suite. What replaces it?

packages/sdk-coverage-tests/coverage-tests/custom/TestFluentApi.spec.js
kyryloonufriiev: Yes. I remove this test because it duplicates the functionality from the generic tests.

packages/sdk-coverage-tests/coverage-tests/custom/TestFluentApi.spec.js
kyryloonufriiev: Oh. Only now I realized that this test is different from what I thought. We can reimplement it in a generic way.

PR #194, opened 2020-06-16
Fix Stale SRE error

amitzur:


packages/eyes-sdk-core/test/utils/MockDriver.js
```diff
@@ -118,6 +111,7 @@ class MockDriver {
       script => /^\/\* @applitools\/dom-snapshot@[\d.]+ \*\//.test(script),
       () => FakeDomSnapshot.generateDomSnapshot(this),
     )
+    this.mockElement('html', {scrollPosition: {x: 0, y: 0}})
```
kyryloonufriiev, line 114: What about all other nested frames?

packages/eyes-selenium/test/it/selenium4/SpecWrappedElement.spec.js
```diff
@@ -18,7 +18,10 @@ describe('SpecWrappedElement Selenium4', () => {
   })
 
   after(async () => {
-    await driver.quit()
+    // mocha version 8 still runs this code even though the suite is skipped in Selenium 3
+    if (driver) {
+      await driver.quit()
+    }
```
kyryloonufriiev, line 24: I think the reason why it happened here and not in `SpecWrappedDriver.spec.js` is that we try to skip it and start the driver in the same `describe `. I would like to move `skip` out of `describe` and then it will not even call `after`.

packages/eyes-sdk-core/lib/EyesClassic.js
```diff
@@ -320,7 +320,7 @@ class EyesClassic extends EyesCore {
     if (this._context.topContext.scrollRootElement) {
       await this._context.topContext.scrollRootElement.init(this._driver)
     } else {
-      const element = await EyesUtils.getScrollRootElement(this._logger, this._executor)
+      const element = await this._finder.findElement({type: 'css', selector: 'html'})
       this._context.topContext.scrollRootElement = new this.constructor.WrappedElement(
```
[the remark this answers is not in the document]
amitzur: Not sure I understand. This is the fix itself, no?

packages/eyes-sdk-core/lib/EyesClassic.js
```diff
@@ -739,7 +739,7 @@ class EyesClassic extends EyesCore {
 
       let scrollRootElement = positionProvider.scrollRootElement
       if (!scrollRootElement) {
-        const element = await EyesUtils.getScrollRootElement(this._logger, this._executor)
+        const element = await this._finder.findElement({type: 'css', selector: 'html'})
         scrollRootElement = new this.constructor.WrappedElement(this._logger, this._driver, element)
```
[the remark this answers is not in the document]
amitzur: why? I must be missing something

packages/eyes-sdk-core/test/utils/MockDriver.js
amitzur: You mean that every mocked frame needs to have this mocked element? sounds about right, how do I do that?

packages/eyes-selenium/test/it/selenium4/SpecWrappedElement.spec.js
amitzur: ahh, right

amitzur: I will do that

packages/eyes-sdk-core/lib/EyesClassic.js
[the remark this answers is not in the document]
kyryloonufriiev: Line 324 wraps an element, but we don't need to do that because an element is already wrapped.
I thought about something like this.
```js
this._context.topContext.scrollRootElement = await this._finder.findElement({type: 'css', selector: 'html'})
```

packages/eyes-selenium/test/it/selenium4/SpecWrappedElement.spec.js
amitzur: done

packages/eyes-sdk-core/lib/EyesClassic.js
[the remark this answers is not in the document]
kyryloonufriiev: Oh. I meant "We __need__ to remove this..."

packages/eyes-sdk-core/test/utils/MockDriver.js
kyryloonufriiev: I did it

packages/eyes-sdk-core/lib/EyesClassic.js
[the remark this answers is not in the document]
amitzur: ah, great. now i understand. I also got confused because I thought this comment was about line 323

packages/eyes-sdk-core/lib/EyesClassic.js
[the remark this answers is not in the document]
amitzur: done

packages/eyes-sdk-core/test/utils/MockDriver.js
amitzur: thanks

PR #195, opened 2020-06-22
Documentation generation

kyryloonufriiev:
This PR should solve the problems of generation type definitions (`.d.ts` files) from JSDoc.
The steps to make it done are:

- [x]  Rewrite core JSDocs using generic types for almost all abstractions.
- [x]  Solve the problem of generation `.d.ts` in one file. This is already could be generated but still couldn't be used because of the bug with typescript ([Typescript issue](https://github.com/microsoft/TypeScript/issues/37832)). Probably we will need to live with multiple `.d.ts` until the typescript team will fix this issue.
- [x] Investigate how to generate documentation for the WDIO service.


packages/eyes-sdk-core/lib/EyesClassic.js
```diff
@@ -34,26 +34,58 @@ const EyesUtils = require('./EyesUtils')
...
  */
 
 const UNKNOWN_DEVICE_PIXEL_RATIO = 0
 const DEFAULT_DEVICE_PIXEL_RATIO = 1
 
+/**
+ * @template Driver
+ * @template Element
+ * @template Selector
+ * @extends {never}
```
amitzur, line 78: what does `@extends {never}` do here?

PR #196, opened 2020-06-24
dom-snapshot LWC adjustments

amitzur:
[Trello link](https://trello.com/c/wfQzNryP)
LWC framework has a lot of quirks that require adjustments in dom-snapshot.

1. Custom implementation of shadow DOM:
https://github.com/salesforce/lwc/blob/f44b7cb6c6891c31b0eeb9d834af89d6c913d133/packages/%40lwc/synthetic-shadow/src/faux-shadow/shadow-root.ts
This requires us to check if the `shadowRoot` is a native implementation, and if not we run a heuristic of adding the `shadowRoot`'s child nodes as the shadow host's child nodes.

2. Because of #1, we now have elements in the cdt which are not discoverable by `document.querySelectorAll`. So we need to go over the cdt array to find resources.

3. CSS variables behave differently when asked for `getPropertyValue`. This is both for asking the value of `--some-var` and also when getting a value of format `var(--some-var,<some-fallback>)`. It treats the value as a string, and therefore escapes it. So we need to unescape it before we absolutize and try to fetch the content.


packages/dom-snapshot/src/browser/extractResourcesFromStyleSheet.js
```diff
@@ -19,7 +19,12 @@ function makeExtractResourcesFromStyleSheet({styleSheetCache, CSSRule = window.C
           [CSSRule.STYLE_RULE]: () => {
             let rv = [];
             for (let i = 0, ii = rule.style.length; i < ii; i++) {
-              const urls = getUrlFromCssText(rule.style.getPropertyValue(rule.style[i]));
+              const property = rule.style[i];
```
amitzur, line 22: // TODO - try using `unescapeCss` from vg code here

packages/dom-snapshot/src/browser/domNodesToCdt.js
```diff
@@ -70,8 +70,14 @@ function domNodesToCdt(docNode, baseUrl, log = noop) {
           (elementNode.childNodes.length ? childrenFactory(cdt, elementNode.childNodes) : []);
 
         if (elementNode.shadowRoot) {
-          node.shadowRootIndex = elementNodeFactory(cdt, elementNode.shadowRoot);
-          docRoots.push(elementNode.shadowRoot);
+          if (/native code/.test(elementNode.shadowRoot.toString())) {
```
amitzur, line 73: // TODO check `typeof elementNode.shadowRoot.toString === 'function'` - we don't want to crash on exceptions here

test.html
```diff
@@ -0,0 +1,45 @@
+<!DOCTYPE html>
```
amitzur, line 1: // TODO this file doesn't reproduce the issue. Need to keep trying to get the same result (or just use LWC)

packages/dom-snapshot/src/browser/extractResourcesFromStyleSheet.js
amitzur: worked great

amitzur:
Some details about the solution:

1. Now, `domNodesToCdt` also returns the `linkUrls`. So instead of those being 2 separate operations, we extract resource urls from the DOM while traversing. It's also better perf, but also better correlation between the data structures of DOM and resources. The API for extracting then (`extractLinks` --> `extractLinksFromElement`), is no longer accepting a document, but rather an element, and it's now possible to test an element for the output it provides.
There's a technical task to modularize the tests for this module. But it was important for me to also verify that existing behavior was preserved.

2. We also get resources URLs from style attributes during the traversal, instead of iterating over the entire CDT after it's constructed.

3. Better testing for `extractResourcesFromStyleSheet`

4. unescape CSS - in cases of css variables, the CSSOM doesn't return a well-parsed value. It has no knowledge that a URL is a URL, so strings are returned escaped. We need to unescape them in this case. I copied code from applitools/mono to do the unescaping.

PR #199, opened 2020-06-30
update xml parsing for sdk-coverage-tests package

Artem-applitools:
it junit report have a <testsutes> tag it was parsed to the array of the arrays instead of array of the testcases.
For the report with a structure 
```
<testsuites>
 <testsuite>
  <testcase>
  </testcase>
 </testsuite>
 <testsuite>
  <testcase>
  </testcase>
 </testsuite>
</testsuites>
```
process-report command is failed with an error "Cannot read property 'name' of undefined"
Parsing such junit reports is required for the some SDKs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,8 @@
       "packages/eyes-storybook",
       "packages/eyes-leanft",
       "packages/side-eyes",
-      "packages/sdk-shared"
+      "packages/sdk-shared",
+      "packages/ruby"
     ],
     "nohoist": [
       "**/@wdio/*",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -29,7 +29,8 @@
       "packages/eyes-storybook",
       "packages/eyes-leanft",
       "packages/side-eyes",
-      "packages/sdk-shared"
+      "packages/sdk-shared",
+      "packages/ruby"
```
kyryloonufriiev, line 33: Why do we need this?

Artem-applitools: @kyryloonufriiev We don't, It's left from the first day working on the ruby sdks. I will remove it.

PR #201, opened 2020-07-02
Fix for WebDriver with 'jasmine'

ojkelly-vha:
When running with `wdio` and `jasmine`, `appName` and `testName` are not set. From digging into the code it appears they are in a different property.

So this PR enables a fallback to use the jasmine props if the initial ones are unset.

### DIFF
--- a/packages/eyes-webdriverio-5/src/EyesService.js
+++ b/packages/eyes-webdriverio-5/src/EyesService.js
@@ -39,7 +39,7 @@ class EyesService {
    */
   beforeSession(config, _capabilities, _specs) {
     this._eyes.setConfiguration(this._eyesConfig)
-    this._appName = this._eyes.getConfiguration().getAppName()
+    this._appName = this._eyes.getConfiguration().getAppName() || config.appName
     if (config.enableEyesLogs) {
       this._eyes.setLogHandler(new ConsoleLogHandler(true))
     }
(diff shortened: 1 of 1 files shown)


amitzur:
Thanks for the contribution @ojkelly-vha !
We'll check this out and verify compatibility with both versions 5 and 6 of wedriver.io and merge this.

packages/eyes-webdriverio-5/src/EyesService.js
```diff
@@ -39,7 +39,7 @@ class EyesService {
    */
   beforeSession(config, _capabilities, _specs) {
     this._eyes.setConfiguration(this._eyesConfig)
-    this._appName = this._eyes.getConfiguration().getAppName()
+    this._appName = this._eyes.getConfiguration().getAppName() || config.appName
```
amitzur, line 42: This change is not needed. `appName` is already populated from the service's configuration.
In webdriver.io version 5 it's like this:
```
exports.config = {
  // ...
  eyes: {
    appName: 'your app name'
  }
}
```
And in version 6 it's this:
```
exports.config = {
  // ...
  services: [['@applitools/eyes-webdriverio5-service', {
    appName: 'your app name'
  }]]
}
```

ojkelly-vha: 🎉 awesome, this is missing from the WebDriver docs, I'll raise a pr there too

https://github.com/webdriverio/webdriverio/pull/5583

PR #202, opened 2020-07-02
Support static method: Eyes.setViewportSize

amitzur:
I tried to see how to test this only with core, but:
1) I thought it was hard to implement using MockDriver and FakeEyesFactory
2) This should be a coverage test, implemented also in non-JS SDK's


kyryloonufriiev:
Looks good.
I think we can move the test to the generics. Just need to come up with a syntax for static methods of eyes, perhaps something like `eyes.constructor.setViewportSize(...)`

amitzur:
@kyryloonufriiev definitely! I thought about it (although I didn't think about c'tor). But I hate to say that due to time constraints I didn't have motivation to go all the way.

amitzur:
:clap:

PR #203, opened 2020-07-03
Xml parser skipped

Artem-applitools:
Updated xml parser to filter tests which have property `skipped` so they weren't sent to the dashboard 
Added the tests for the new testcases
Examples of the xml with `skipped` property added to the fixtures

### DIFF
--- a/packages/sdk-coverage-tests/src/report/xml.js
+++ b/packages/sdk-coverage-tests/src/report/xml.js
@@ -3,7 +3,7 @@ const {logDebug} = require('../log')
 
 function convertJunitXmlToResultSchema({xmlResult, browser}) {
   let result = []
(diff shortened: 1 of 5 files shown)


amitzur:
@Artem-applitools how do tests get this `skipped` property in the XML?

Artem-applitools:
JUnit reporters for the ruby and python add skipped tests to the report. So
if we implemented test but disabled it cause it failing. Processing report
with skipped tags, send skipped tests as passed. Not sure we can generate
such junit xml report with mocha.

вс, 5 июл. 2020 г., 11:01 Amit Zur <notifications@github.com>:

amitzur:
Great, thanks! I'll let @kyryloonufriiev review and approve.

PR #206, opened 2020-07-06
Support visual locators

amitzur:
- Add `eyes.locate` API.
- Types are provided instead of fluent API or data classes.
- Generic coverage test for classic and VG.

The tricky part was to take a viewport screenshot with `EyesVisualGrid`. For that I moved all the necessary properties and methods from `EyesClassic` to `EyesCore`.
I didn't want to take it all the way to merge the Eyes classes. That's a desired refactor, but not critical ATM. So only the minimal changes were made.

The part I was forced to do and is ugly is to call `this.getAndSaveRenderingInfo` from inside `eyes.locate`, since `EyesVisualGrid` doesn't call `serverConnector.renderInfo()` - because it does everything inside VGC. Because it's cached it's not a redundant network call.

### DIFF
--- a/packages/eyes-sdk-core/CHANGELOG.md
+++ b/packages/eyes-sdk-core/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- support visual locators
 
 ## 11.2.2 - 2020/7/5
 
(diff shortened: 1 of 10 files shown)


amitzur:
Implementation inspired by https://github.com/applitools/eyes.sdk.java3/pull/270/files?file-filters%5B%5D=.java&file-filters%5B%5D=.md

PR #207, opened 2020-07-07
Fix empty test

amitzur:


amitzur:
seems like this is causing coverage tests to fail. I reverted the commit on master until we investigate

PR #211, opened 2020-07-09
Avoid error.getTestResults error

amitzur:
https://trello.com/c/XAmA255U
https://trello.com/c/1Bl2EaDE

Suppose this case:
```
eyes.close(false)
```
And there was an internal error. It might be due to our bug and it might be something else, but it's not a `TestFailedError`.
What should be the return value? Should the method still throw in spite of the false argument?
Ultimately if the user calls `runner.getAllTestResults` then they would get the exception inside the `TestResultContainer`, but the question is regarding the `eyes.close` itself.
I'm inclined to say that it should throw the exception. The alternative is to return a `TestResultsError` instance, which is something we already use in storybook.


amitzur:
@kyryloonufriiev this is still draft - I need to add tests

PR #214, opened 2020-07-15
Coverage tool is generic

Artem-applitools:
Added support for the "isGeneric" field by adding new metaData json file

### DIFF
--- a/packages/sdk-coverage-tests/src/cli/commands/process-report.js
+++ b/packages/sdk-coverage-tests/src/cli/commands/process-report.js
@@ -6,8 +6,7 @@ const {logDebug} = require('../../log')
 
 async function processReport(args) {
   const {name: sdkName, meta: metaPath} = require(path.join(process.cwd(), args.path))
-  logDebug(args.reportName)
-  const results = readFileSync(path.resolve(process.cwd(), args.reportName), {
+  const results = readFileSync(path.resolve(process.cwd(), 'coverage-test-report.xml'), {
     encoding: 'utf-8',
   })
   const metaDataFile = readFileSync(
--- a/packages/sdk-coverage-tests/src/cli/commands/process-report.js
+++ b/packages/sdk-coverage-tests/src/cli/commands/process-report.js
@@ -5,13 +5,16 @@ const {sendReport} = require('../../send-report')
 const {logDebug} = require('../../log')
 
 async function processReport(args) {
-  const {name: sdkName} = require(path.join(process.cwd(), args.path))
+  const {name: sdkName, meta: metaPath} = require(path.join(process.cwd(), args.path))
   logDebug(args.reportName)
   const results = readFileSync(path.resolve(process.cwd(), args.reportName), {
     encoding: 'utf-8',
   })
-  const metaDataFile = readFileSync(path.resolve(process.cwd(), 'coverage-tests-metadata.json'))
+  const metaDataFile = readFileSync(
+    path.resolve(process.cwd(), metaPath || 'coverage-tests-metadata.json'),
+  )
   const metaData = JSON.parse(metaDataFile)
+  logDebug(metaData)
   const isSandbox = args.sendReport === 'sandbox'
   process.stdout.write(`\nSending report to QA dashboard ${isSandbox ? '(sandbox)' : ''}... `)
   const report = createReport({
@@ -19,7 +22,7 @@ async function processReport(args) {
     xmlResult: results,
     sandbox: isSandbox,
     id: args.reportId,
-    metaData: metaData
+    metaData: metaData,
   })
   logDebug(report)
   const result = await sendReport(report)
--- a/packages/sdk-coverage-tests/src/cli/commands/process-report.js
+++ b/packages/sdk-coverage-tests/src/cli/commands/process-report.js
@@ -9,9 +9,15 @@ async function processReport(args) {
   const results = readFileSync(path.resolve(process.cwd(), 'coverage-test-report.xml'), {
     encoding: 'utf-8',
   })
-  const isSandbox = args.sendReport === 'sandbox' ? true : false
+  const isSandbox = args.sendReport === 'sandbox'
   process.stdout.write(`\nSending report to QA dashboard ${isSandbox ? '(sandbox)' : ''}... `)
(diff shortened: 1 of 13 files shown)


packages/sdk-coverage-tests/src/cli/commands/process-report.js
```diff
@@ -6,12 +6,21 @@ const {logDebug} = require('../../log')
 
 async function processReport(args) {
   const {name: sdkName} = require(path.join(process.cwd(), args.path))
-  const results = readFileSync(path.resolve(process.cwd(), 'coverage-test-report.xml'), {
+  logDebug(args.reportName)
+  const results = readFileSync(path.resolve(process.cwd(), args.reportName), {
     encoding: 'utf-8',
   })
-  const isSandbox = args.sendReport === 'sandbox' ? true : false
+  const metaDataFile = readFileSync(path.resolve(process.cwd(), 'coverage-tests-metadata.json'))
```
kyryloonufriiev, line 13: I think we would want to customize this as well as "reportName"

packages/sdk-coverage-tests/src/report/xml.js
```diff
@@ -10,6 +10,7 @@ function convertJunitXmlToResultSchema({xmlResult, browser}) {
     const testNameWithoutSuffix = removeSuffix(testName)
     result.push({
       test_name: testNameWithoutSuffix,
+      isGeneric: metaData[testName] ? metaData[testName].isGeneric : false,
```
kyryloonufriiev, line 13: I would like to see all metadata for a specific test and not only "specific properties. Maybe something like `...metaData[testName]`?

packages/sdk-coverage-tests/src/code-export/save.js
```diff
@@ -12,7 +13,14 @@ async function createTestFiles(emittedTests, sdkImplementation) {
     fs.writeFileSync(filePath, payload)
   })
 }
+async function createTestMetaData(emittedTests) {
+  let metaData = {}
+  emittedTests.forEach(test => metaData[test.name] = {isGeneric: true} )
+  const filePath = path.resolve(process.cwd(), 'coverage-tests-metadata.json')
```
kyryloonufriiev, line 19: Also, I want to be able to customize this in generation time.

packages/sdk-coverage-tests/src/report/xml.js
Artem-applitools: Updated

packages/sdk-coverage-tests/src/code-export/save.js
Artem-applitools: Added to the configuration index.js file option `meta` to identify path to the meta file. In such case it will be always the same during test generation and reporting

amitzur: why should this be customizable?

packages/sdk-coverage-tests/src/code-export/save.js
```diff
@@ -12,7 +13,17 @@ async function createTestFiles(emittedTests, sdkImplementation) {
     fs.writeFileSync(filePath, payload)
   })
 }
+async function createTestMetaData(emittedTests, sdkImplementation) {
+  let metaData = {}
```
amitzur, line 17: prefer `const` over `let` when possible

packages/sdk-coverage-tests/src/cli/commands/process-report.js
```diff
@@ -5,13 +5,25 @@ const {sendReport} = require('../../send-report')
...
+  logDebug(metaData)
+  const isSandbox = args.sendReport === 'sandbox'
   process.stdout.write(`\nSending report to QA dashboard ${isSandbox ? '(sandbox)' : ''}... `)
-  const report = createReport({sdkName, xmlResult: results, sandbox: isSandbox, id: args.reportId})
+  const report = createReport({
+    sdkName,
+    xmlResult: results,
+    sandbox: isSandbox,
+    id: args.reportId,
+    metaData: metaData,
```
amitzur, line 25: no need for `metaData: metaData`. It can be written just as `metaData`. It's called "shorthand property names": https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#Property_definitions

amitzur:
I think this is good. I made some fixes and code style improvements.
Only question that remains is why should the metadata file name be customizable?
And it should be added to gitignore in every repo it is run (I added it to this repo's gitignore)

packages/sdk-coverage-tests/src/cli/commands/process-report.js
Artem-applitools: It also could be added to the eslint config rules, to fix such issues automatically.

packages/sdk-coverage-tests/src/code-export/save.js
kyryloonufriiev: We want to specify a path for metadata file. I think other teams will want to specify a certain location for all generic tests stuff.

amitzur: ah, so the *path* should be configurable. I wouldn't make the file name configurable.

Artem-applitools: Updated to configurable path at last commit.

kyryloonufriiev:
I think we could use the same path for generated tests output and their metadata.

amitzur:
@kyryloonufriiev I agree. Let's make the `outputPath` configurable, and keep the names of files that we use unchanged.

kyryloonufriiev:
I have it added some time ago. Just need to use it for metadata as well

amitzur:
So no need for `reportName`

Artem-applitools:
xml report is generated by the external tools, having the `reportName` options gives some flexibility to work with them.

amitzur:
but surely they all have a way to specify an output folder

amitzur:
So do we have `outputPath`? I see that we're saving inside `cwd`

Artem-applitools:
I don't see need in `outputPath` at the moment, lets make it a part of another tool release

amitzur:
I changed `meta` to `metaPath` since I felt it's more explicit

PR #217, opened 2020-07-21
Take out cvg tests

amitzur:
new location: https://github.com/applitools/sdk.coverage.tests/blob/master/tests.js


kyryloonufriiev:
I have a suggestion about the implementation. Instead of `write-require-unlink` chain we could use node native `vm` module to execute the script from a string.

PR #219, opened 2020-07-24
Fix duplicate copies of SDK

amitzur:
It seems like the Cucumber testing framework is creating a copy of the SDK's code, and also invalidates require cache. It's not 100% certain, it could be either that or something else - what we do know is that checks for `instanceof` are failing, and when we added console logs for customers that experienced this, the path to the file was the same. Meaning, it's not a duplicate node_module somewhere. It's just loaded again into memory.

This PR serves to fix the particular issue of the customer. After this, we should change all occurences of `instanceof` to `TypeUtils.instanceOf`

In order to do that, we need to add `static get __ConstructorName() { return true; }` to all classes.
We could do this in a utility function, so something like:

```
class Region() {
}

module.exports = markType(Region)
```

where `markType` is something like:

```
function markType(ctor) {
  Object.defineProperty(ctor, `__${ctor.name}`, {value: true})
}
```

### DIFF
--- a/packages/eyes-sdk-core/lib/geometry/Location.js
+++ b/packages/eyes-sdk-core/lib/geometry/Location.js
@@ -1,4 +1,5 @@
 const ArgumentGuard = require('../utils/ArgumentGuard')
+const TypeUtils = require('../utils/TypeUtils')
 
 /**
  * @typedef {{x: number, y: number}} LocationObject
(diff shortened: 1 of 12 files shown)


amitzur:
The test itself is also a bit dirty. It manually creates a copy of the SDK by copying the core into the SDK, copying `src` and `index` in the SDK itself, and fixing some references.
I would love to revisit it with docker later, and it would be even easier if we fade out yarn workspaces (so node_modules folder could be copied from the SDK without overlap with other packages)

Issue #218, opened 2020-07-23
Add the ability to set a custom result formatter


zalohcin:
The output from the eyes API provides a lot more data than is currently exposed by the TAP and XML files.

Would it be possible to make it so I can pass in a function to manipulate that data how I see fit. Or otherwise is it possible to access the RAW data?

zalohcin:
I opened a PR to resolve this issue. https://github.com/applitools/eyes.sdk.javascript1/pull/224

zalohcin:
merged

PR #224, opened 2020-07-27
Adding jsonFilePath config option

zalohcin:
This allows the user to set a jsonFilePath in the same manner as tapFilePath and xmlFilePath


zalohcin:
This resolves the issue I was having. https://github.com/applitools/eyes.sdk.javascript1/issues/218

tourdedave:
Looks great, thank you!

PR #227, opened 2020-07-29
JSDOC fixes  + added 'use strict'

yonatanLehman:


packages/eyes-selenium/src/SeleniumCheckSettings.js
```diff
@@ -30,7 +30,10 @@ const SpecCheckSettings = {
     return SeleniumFrame.fromReference(reference)
   },
 }
-
+/**
+  *  @class SeleniumCheckSettings
+  *  @instantiate SeleniumCheckSettings from DriverCheckSettings
```
amitzur, line 35: this is the only thing that I don't agree on in this PR

yonatanLehman: You can leave it out - for the moment till we work out how we do the instantiation.
I have more minor JSDOC  changes (standard...) based on the Master from 29/7/20 16:20 (089ac59fff444b67095d1506f698eb75cbf7109a)

Can you accept these as they are or do I first need to merge with new updates?

yonatanLehman: Not sure how to move them - just delete them after the merge - they have changed in any case

amitzur:
Adding these changes manually because of conflicts with master.
Another thing I omitted - `@return` doesn't need a dash after the type

amitzur:
reference to commit:
https://github.com/applitools/eyes.sdk.javascript1/commit/65e8fddd083a9e56fc0ef198b1387ad4f4d9b37c

PR #231, opened 2020-07-31
dom-snapshot: expose function to fetch and process resourceUrls

tourdedave:
In some SDKs we want to improve the performance of `dom-snapshot` (like `eyes-testcafe`).

One way to do that is to instruct dom-snapshot to not fetch the resources.

This delegates the responsibility down the SDK to fetch and process the resources. This PR is an attempt to make the implementation for the SDK relatively straight-forward.

Still a WIP. Feedback welcome!


amitzur:
I may be missing something, but I believe this is unnecessary.
DS supports the `dontFetchResources` argument, which will make it return only resourceUrls and no blobs. Nothing else is needed besides calling with `dontFetchResources: true`, as the SDK will take it from there.

tourdedave:
What happens in the SDK if there are resourceUrls but no blobs?

tourdedave:
Had a call to discuss this (thanks @amitzur)!

The phrase `the SDK will take it from there` is meant as -- the SDK uses the visual-grid-client to download the resourceUrls out of band from the browser.

So, this is unnecessary. Closing.

PR #233, opened 2020-08-04
Fix user agent parsing for Edge Chromium

amitzur:
[Trello](https://trello.com/c/bOUk7oj8/182-edge-browserstack-user-agent-is-considered-as-chrome)


tourdedave:
+1

PR #234, opened 2020-08-05
Support adoptedStyleSheets

amitzur:
This is a new non-standard feature in Chrome.
Spec:
https://developers.google.com/web/updates/2019/02/constructable-stylesheets

Reference to decision not to currently support the `sheet.replace` async API:
https://github.com/WICG/construct-stylesheets/issues/119#issuecomment-588352418

Spec for SDK-VG API:
https://trello.com/c/cX4q0DzH/390-adoptedstylesheets-feature-not-supported#comment-5f1b44f7530156669ebbe228


packages/dom-snapshot/src/browser/domNodesToCdt.js
```diff
@@ -253,4 +263,12 @@ function domNodesToCdt(docNode, baseUrl, log = noop) {
   }
 }
 
+function getAdoptedStyleSheets(node) {
+  return Array.from(node.adoptedStyleSheets).map(sheet => {
+    const cssomAst = createAstFromCssom(sheet.cssRules);
+    const mergedRules = mergeRules(cssomAst, cssomAst);
+    return csstree.generate(csstree.fromPlainObject({type: 'StyleSheet', children: mergedRules}));
```
amitzur, line 270: @kyryloonufriiev as you can see I hacked this. I wasn't sure what should be done to get all the goodies of cssom to string that we have in `processInlineCss`. You can see in the test that I check one scenario of shorthand_replaceable (I specify `justify-content:center;align-content:center` and expect it not to become `place-content:center`. That's a sign for me that the entire logic is turned on.

Can you recommend a better way to do this?

kyryloonufriiev: I believe that's not right.
First of all that's make no sense to merge `cssomAst` with itself, since they are the same.
You see `justify-content:center;align-content:center` in the cdt because it how cssom works, it splits everything to the lowes level as possible (longhand).
So to see an actual problem you need to say `place-content:center` in the adopted stylesheet, cdt will have two longhands in this case. And this is what we want to avoid.
It seems like in case of chrome we can not avoid it, because the actual string provided to the `replaceSync` is not accessible. We have `cssText` property which is not exactly what we want, since in chrome it will "compress" text to make it as short as possible using all kinds of shorthands (even if you didn't use it initially).

amitzur: Right.
I think I should just iterate on the lowest level longhand properties. I don't mind that it's not as the text. What I mind is browser compatibility. This is a compromise.
I just wasn't sure if there's any business logic going on in either `mergeRules` or `createAstFromCssom` that makes me need to use them instead.

kyryloonufriiev: If you don't mind about it to be exactly the same text, then you doing everything right. `mergeRules` doesn't modify an ast it just decide what to choose in a conflict between two asts. So you just need to remove `mergeRules`, everything else handled in `createAstFromCssom`, including the fact that shorthands are not always preferred since some of them are experimental and have very poor support.

amitzur: Thanks @kyryloonufriiev !
I incorporated this and merging. If you have anything to add after the merge feel free.

PR #239, opened 2020-08-06
Update package.json

amirhalatzi:
Update dependencies


amitzur:
@amirhalatzi we'll update deps at another time. It needs a bit more tooling. We currently have a different workflow than vg. For example, chalk 4 only supports Node.js >= 10 but we support >=8.9 so we need to test it in different Node.js versions.

PR #240, opened 2020-08-07
Improved execute script logging

tourdedave:
Log what executeScript was trying to do when it threw an exception

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,4 @@ coverage-tests-metadata.json
 # other
 tags
 support
+eyes-testcafe-bak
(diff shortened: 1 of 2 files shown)


tourdedave:
I initially wanted to just `JSON.stringify(args)` but got a circular error when `args` included an element.

kyryloonufriiev:
@tourdedave Maybe `JSON.stringify` is a good approach. We just need to have a reasonable `toJSON` and `toString` methods in `EyesElement` then we would benefit not only here but always when we want to log an element.

amitzur:
@kyryloonufriiev @tourdedave what do you think we should have in our `EyesElement.toJSON` and `toString` in order to have this functionality?

kyryloonufriiev:
I think `toJSON` should contain selector, parent context and framework element itself if it could be serialized. `toJSON` for context should contain parent context, frame element/reference, scroll root element and also there might be calculated metrics such as scroll position. Then `toSstring` could be just `${this.constructor.name} ${JSON.stringify(this, null, 2)}`

tourdedave:
Works for me.

PR #242, opened 2020-08-10
upload cvg test result data to azure

amitzur:
Need to define env var `AZURE_STORAGE_CONNECTION_STRING` when running `coverage-tests process report`.
Blob container will be organized like so:
`<sdk name>/<sandbox|prod>/<timestamp>-<version>.json`

For example:
`eyes.webdriverio.javascript4/sandbox/2020-08-10-17-10-34-234-sandbox.json`
`eyes.webdriverio.javascript5/prod/2020-08-10-17-10-34-234-5.21.0.json`

### DIFF
--- a/packages/eyes-webdriverio-4/package.json
+++ b/packages/eyes-webdriverio-4/package.json
@@ -33,7 +33,7 @@
     "webdriverio": "~4.14.4"
   },
   "devDependencies": {
-    "@applitools/sdk-coverage-tests": "1.0.12",
+    "@applitools/sdk-coverage-tests": "^1.0.13",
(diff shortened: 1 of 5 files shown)


amitzur:
@kyryloonufriiev @tourdedave I'm wondering if timestamp is good for the prod results. It would make it hard to directly reach results for a given version. But it provides uniqueness. If we know for sure that no version is reported twice we can remove timestamps from prod and keep it only for sandbox.

amitzur:
What I learned now is that the `reportId` is also important, as some SDK's report the results in chunks (either because they run custom/generic tests separately or because they run different packages like C#)

PR #243, opened 2020-08-12
Improve handling of fractional metrics

kyryloonufriiev:
Please pay attention on next tests:
TestCheckScrollableModal
TestCheckScrollableModal_Scroll
CheckRegionWithFractionalMetrics

The image on the `TestCheckScrollableModal` test is changed for the better, so I think we could safely update a baseline since we already have our own branch for these tests.
The new test is `CheckRegionWithFractionalMetrics` so please approve a baseline.


packages/eyes-sdk-core/lib/geometry/RectangleSize.js
```diff
@@ -27,8 +27,8 @@ class RectangleSize {
     }
 
     const {width, height} = varArg1
-    ArgumentGuard.greaterThanOrEqualToZero(width, 'width', true)
-    ArgumentGuard.greaterThanOrEqualToZero(height, 'height', true)
+    ArgumentGuard.isNumber(width, 'width')
```
amitzur, line 30: lack of rounding can cause us to have a negative number here?

packages/eyes-webdriverio-5/.mocharc.cvg.js
```diff
@@ -16,7 +16,7 @@ const protocol = process.env.APPLITOOLS_WDIO_PROTOCOL in tags ? process.env.APPL
 module.exports = {
   spec: [
     './test/coverage/generic/*.spec.js',
-    // '../sdk-shared/coverage-tests/custom/**.spec.js',
+    '../sdk-shared/coverage-tests/custom/**.spec.js',
```
amitzur, line 19: already done on master

packages/eyes-sdk-core/lib/geometry/RectangleSize.js
kyryloonufriiev: Return this check back just without a third argument which triggers a check for integer.

amitzur:
@kyryloonufriiev I love the visual impact of these changes in the mentioned tests. They are indeed better. And I like very much the CheckRegionWithFractionalMetrics test. I accepted that baseline.

A few points:

1. TestCheckScrollableModal and TestCheckScrollableModal_Scroll are both currently disabled. And they don't have a different branch baseline, so I'm not sure why you mentioned that. How did you get to run them? Did you try to find out what tests can be enabled?

2. Change in TestCheckScrollableModal and in CheckRegionWithFractionalMetrics (I ran in also in master branch to see the bug) is not detected by Strict algorithm, so I reported that. I will update the baseline in a branch once you answer (1) and report in this card: https://trello.com/c/llGLV3i9/2054-potentially-wrong-baselines

PR #244, opened 2020-08-14
Tutorial docker

Artem-applitools:
Add the Dockerfiles, Docker-compose.yml to run the tutorials for the Selenium4, Selenium3, WebdriverIO-4, WebdriverIO-5, cypress, testcafe. 
To run the tutorials we need follow next steps: 
1) (Required for Selenium, cypress and testcafe) Run ./build.sh inside sdk-shared/tutorial
2) Open eyes-[name]/tutorial , and them execute ./run.sh

Script executes all steps to build and run docker image with a tutorial. Current state of the package is packed from the repo and installed to the tutorial.


Artem-applitools:
I have updated the tutorials.
Now they built from the ubuntu official image, all required dockerfiles a located in the `sdk-shared/tutorial/` folder. 
For all SDK images build from one source. 
Added to the SDK `tutorial.sh` script to build, run and report results. 
Added scripts to package.json to invoke the script. `tutorial:test:sandbox` and `tutorial:test:prod` which send results to sandbox and prod accordingly. 
Also there is a readme with some details in `sdk-shared/tutorial/` folder 

There also some tutorials which failing due to diffs, 
Ultrafastgrid: https://eyes.applitools.com/app/test-results/00000251790480196872/00000251790480196747?accountId=xIpd7EWjhU6cjJzDGrMcUw~~
I suppose those diffs could be accepted, but I don't have permissions to do so 
And for the selenium4 SDK basic tutorial fails due to missing check 
https://eyes.applitools.com/app/test-results/00000251790480226830/?accountId=xIpd7EWjhU6cjJzDGrMcUw~~

amitzur:
@Artem-applitools a few points:

1. Can you verify this doesn't take a long time to run in github actions? On my machine it took a long time to download the ubuntu base. I'm not sure if that will happen on every github action job.
2. It's going to fail on github actions because it's only supported in Mac OS, because of `host.docker.internal`. This property isn't supported in Linux.
3. Can you explain how the tutorial is tested against the source code? Because if you just `git clone` the tutorial and then `npm install`, it installs what was already published. What do you do in order to run what is actually **going to be published**?

Artem-applitools:
1. Haven't checked the github actions, yet. I was assuming you will run them locally. Need to check this one. If it will download the images from the docker every time it's anyway an issue as it also will always download the selenium images as well. They take not less than a ubuntu image. It you solved this already i could be applied to the base image for the tutorials. As it the same for all. 
2. It only the issue for the few SDKs, and there it can be updated to run differently on linux. (Simple change the network for the container to the host and it will work on linux)
3. There builds image with all the dependencies which actually will be common to all the SDKs, and then builded the image which will have the current version of the SDKs in the `//SDK` folder. And cloned version of the tutorial is installed all the dependencies with the current version of the SDK pointed to the file on the system. After the install the package.json looks like this 
{
  "name": "tutorial-selenium-javascript-basic",
  "version": "1.0.0",
  "description": "",
  "homepage": "https://applitools.com",
  "author": {
    "name": "Applitools Team",
    "email": "team@applitools.com"
  },
  "repository": {
    "type": "git",
    "url": "git://github.com/applitools/tutorial-selenium-javascript-basic.git"
  },
  "bugs": {
    "url": "https://github.com/applitools/tutorial-selenium-javascript-basic/issues"
  },
  "dependencies": {
    "@applitools/eyes-selenium": "file:../../../SDK",
    "chromedriver": "latest",
    "mocha": "^8.0.1",
    "selenium-webdriver": "latest"
  },
  "scripts": {
    "test": "mocha --no-timeouts \"test/*.test.js\""
  },
  "license": "ISC",
  "engines": {
    "node": ">=8.9.0"
  }
}

amitzur:
@Artem-applitools @Yarden-Ingber 
1. A requirement of this is to run in GH actions. We will not run it locally unless there's some failure. So it needs to run in a reasonable time as part of our release pipeline. You had a good idea there - to use the same Selenium docker image we already use for tests. You need to verify this.
2. Until this is handled we cannot merge this PR.
3. I'm not sure this is the right thing to do. What this requires is that the package.json of the SDK will be updated with the latest dependencies (e.g. eyes-sdk-core, visual-grid-client). But this is not the case. When the GH action is run, a part of the release pipeline is to update dependencies. Unless you do that, you are using outdated dependencies. In order to overcome this, a lot of patching need to happen, which make it questionable whether we went in the right direction here to begin with. I'd be happy to setup a meeting to discuss this.

PR #245, opened 2020-08-17
TestCafe++

tourdedave:
An eyes-testcafe rewrite to use the new core, with a current focus on replacing [the existing vg-based eyes-testcafe SDK](https://www.npmjs.com/package/@applitools/eyes-testcafe) (classic execution to be finished later)


packages/eyes-testcafe/src/spec-driver.js
```diff
@@ -0,0 +1,377 @@
...
+// no-op for coverage-tests
+exports.build = () => {
+  return [undefined, () => {}]
+}
+exports.scrollIntoView = () => {} // TestCafe does this implicitly
+exports.isStaleElementError = () => {} // TestCafe doesn't have a stale element error
+// no-op for core
+exports.toEyesSelector = () => {
+  return {type: false}
+}
```
kyryloonufriiev, line 375: We don't use this function anymore

packages/eyes-testcafe/src/spec-driver.js
```diff
@@ -0,0 +1,377 @@
...
+    await executeScript(
+      driver,
+      `if (!window.outerWidth && !window.outerHeight) {
+    window.outerWidth = ${width}
+    window.outerHeight = ${height}
+  }`,
+    )
+  }
+  if (Number.isInteger(x) && Number.isInteger(y))
+    await executeScript(driver, `window.scroll(${x}, ${y})`)
```
kyryloonufriiev, line 335: You don't need to do this at all, window location is the location of the browser window on the screen and not the page's scroll location.

packages/eyes-testcafe/src/spec-driver.js
```diff
@@ -0,0 +1,377 @@
...
+// https://devexpress.github.io/testcafe/documentation/guides/concepts/built-in-wait-mechanisms.html#wait-mechanism-for-selectors
+async function waitUntilDisplayed(_driver, element, timeout) {
+  await element.with({visibilityCheck: true, timeout})
+}
+// NOTE:
+// This is an interim solution until it's properly implemented in core
+async function getWindowRect(driver) {
+  const rect = await executeScript(
+    driver,
+    'return {x: window.scrollX, y: scrollY, width: window.outerWidth, height: window.outerHeight}',
```
kyryloonufriiev, line 313: The window location is the location of the browser window on the screen and not the page's scroll location.

packages/eyes-sdk-core/lib/sdk/EyesDriver.js
```diff
@@ -217,18 +217,62 @@ class EyesDriver {
...
+      ) {
+        return path
+      }
+    }
+    const framesInfo = await EyesUtils.getChildFramesInfo(this._logger, this)
+    for (const frameInfo of framesInfo) {
+      const possiblePath = [...path, frameInfo.element]
+      const result = await this.findPathToChildContext(contextInfo, possiblePath)
+      if (result && result.length) return result
+    }
```
kyryloonufriiev, line 274: This part should be optimized using a smarter snippet that will traverse through the frame chain in the browser. So in the best case (when the active context doesn't have cors frames in the path), everything could be done with one script execution.

packages/eyes-sdk-core/lib/sdk/EyesDriver.js
```diff
@@ -217,18 +217,62 @@ class EyesDriver {
...
+      await this.spec.childContext(this._driver, element)
+    }
+    if (contextInfo.selector) {
+      const element = await this.spec.findElement(this._driver, {
+        type: 'xpath',
+        selector: contextInfo.selector,
+      })
+      if (element) {
+        return [...path, element]
+      }
```
kyryloonufriiev, line 257: I don't think we can assume that this is our context we looking for only if an element was found. Since it could be another frame. We still need to compare `documentElement`'s

packages/eyes-testcafe/src/spec-driver.js
```diff
@@ -0,0 +1,377 @@
...
+// To compensate we create normalized copies of the elements and do a
+// stringified comparison.
+function isEqualElements(_driver, element1, element2) {
+  if (!element1 || !element2) return false
+  const elements = [{...element1}, {...element2}].map(el => {
+    delete el.propType
+    return el
+  })
+  return JSON.stringify(elements[0]) === JSON.stringify(elements[1])
+}
```
kyryloonufriiev, line 167: Why not compare in browser?

kyryloonufriiev (changes requested):
Looks really good and solid. I'm not sure I fully understood all of the testcafe specific parts. But I think we are safe since we have good coverage in this package. 
Please pay attention to the part my comment in for `findPathToChildContext`, it might be a bug, everything excepts this is minor. 
💯

packages/eyes-testcafe/src/spec-driver.js
tourdedave: I don't remember if there was a reason for not doing it (it was long time ago since this was function was written). It's worth checking though.

packages/eyes-sdk-core/lib/sdk/EyesDriver.js
tourdedave: 👍 sounds like a technical task for later

packages/eyes-sdk-core/lib/sdk/EyesDriver.js
tourdedave: Do you mean something like this? 

```js
if (
  element && 
  (await this.spec.isEqualElements(this._driver, contentDocument, contextInfo.documentElement))
) {
```

tourdedave:
Going to address the unresolved items as part of this ticket: https://trello.com/c/rTSTc8Mw

PR #251, opened 2020-09-02
Add disableBrowserFetching configuration

amitzur:
This PR adds `disableBrowserFetching` as a configuration parameter to both `Configuration` and `CheckSettings`.
It can serve as a guide to how to add configuration properties.
I was thinking how to test this, and thought it would be best to do an e2e test. Mainly because I wanted to verify that dom-snapshot receives `dontFetchResources`, but it's hard to validate without changing something fundamental: dom-snapshot is run today with no args, so it's hard to test its arguments if they are encoded in the script string. And I didn't want to change it and find myself in a rabbit hole.

So the test I did uses a local test server with a middleware that would serve an image only if it's fetched from Node.js. So I know that DS cannot fetch it and therefore the visual test is an indicator for passing the right configuration.

### DIFF
--- a/packages/eyes-sdk-core/lib/config/Configuration.js
+++ b/packages/eyes-sdk-core/lib/config/Configuration.js
@@ -245,6 +245,9 @@ class Configuration {
     /** @type {number[]|boolean} */
     this._layoutBreakpoints = undefined
 
+    /** @type {boolean} */
+    this._disableBrowserFetching = undefined
+
     if (configuration) {
       this.mergeConfig(configuration)
     }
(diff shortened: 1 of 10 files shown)


kyryloonufriiev:
Looks good to me.
Did you think to test it in the core? With MockDriver we can check which arguments were passed to the dom-snapshot.

amitzur:
@kyryloonufriiev I thought about it and tried it, but it's what I wrote in the description above - it was hard to check which arguments were passed to DS

amitzur:
So we are left with a technical task of sending dom snapshot as a function with arguments, so it can be better checked as a core test. That test would `mockScript` in the MockDriver and check that the correct parameters are passed to dom snapshot.

PR #252, opened 2020-09-02
Waits for session to start in order to perform renderings

kyryloonufriiev:


kyryloonufriiev:
Storybook measurements:
- 1500 stroies x 1 browser
  - before PR: 
    - Time: ~8.3m (~500s)
    - Memory: ~(95-110) MB total heap
  - after PR:
    - Time: ~32.5m (~1950s)
    - Memory: ~110 MB total heap
- 100 stroies x 15 browser
  - before PR: 
    - Time: ~15m (~900s)
    - Memory: ~86 MB total heap
  - after PR:
    - Time: ~27.5m (~1650s)
    - Memory: ~86 MB 

**1500 x 1**
The time difference is understandable since the number of parallel renders (50) is now blocked by the number of parallel sessions (10). So the increase of the execution time by 4 times is expected. The memory usage isn't growing because we do only one browser for each session and the sizes of the dom snapshots are tiny since this is not a big page but only one component.

**100 x 15**
This run is more interesting. The time "after PR" is almost the same as **1500**, but before PR time is increased almost 2 times. The **100x15** and **1500** need to perform the same number of renders, so the time for both of these runs should be almost the same, instead, there is a difference of 5 minutes. The difference is reasonable because in **1500** `one story = one session = one render` and only after the finish of the session we could start another render. In the case of **100x15**, each story initiates 15 sessions and the first of them initiate 15 renders. Since only 10 sessions could be performed concurrently 5 renders will be ready even before the next session is started. This is an explanation for 5 minute difference between **1500** and **100x15**

packages/visual-grid-client/src/sdk/makeTestWindow.js
```diff
@@ -3,7 +3,7 @@
 const makeOpenEyes = require('./openEyes')
 
 function makeTestWindow(openConfig) {
-  const openEyes = makeOpenEyes({...openConfig, isSingleWindow: true})
+  const openEyes = makeOpenEyes({...openConfig, isSingleWindow: false})
```
amitzur, line 6: Since this is the only place that uses `isSingleWindow`, I think we can remove it. And also `testWindow` and `closeTestWindow` from EyesWrapper

packages/visual-grid-client/test/it/openEyes.it.test.js
```diff
@@ -750,6 +751,92 @@ Received: 'firefox-1'.`,
...
+
+      wrapper.open = async () => {
+        await psetTimeout(50)
+        counters.open++
+      }
+
+      wrapper.renderBatch = async () => {
+        await psetTimeout(50)
+        counters.render++
+        return [new FakeRunningRender(`renderId${counters.render}`, RenderStatus.RENDERED)]
```
amitzur, line 773: To be purist, render status here can't be rendered. It can at most be rendering

packages/visual-grid-client/test/it/openEyes.it.test.js
```diff
@@ -750,6 +751,92 @@ Received: 'firefox-1'.`,
...
 
+    it('waits for session to start in order to perform renderings', async () => {
+      const wrapper = createFakeWrapper(baseUrl)
+
+      const counters = {
+        open: 0,
+        session: 0,
+        checkWindow: 0,
+        render: 0,
+        close: 0,
```
amitzur, line 762: this is not used

packages/visual-grid-client/test/it/openEyes.it.test.js
```diff
@@ -750,6 +751,92 @@ Received: 'firefox-1'.`,
...
+      expect(counters.open).to.equal(1)
+      expect(counters.session).to.equal(1)
+      expect(counters.render).to.equal(1)
+      expect(counters.checkWindow).to.equal(1)
+
+      await close()
+    })
+
     // TODO statuser makes this test impossible to fix. need to check concurrency
     it.skip("doesn't wait for eyes test to open in order to perform renderings", async () => {
```
amitzur, line 841: haha, I guess this test can be deleted as it's exactly what we're changing.

amitzur (changes requested):
The results of storybook make me realize that there is wrong behavior when multiple browsers are specified.
Renderings cannot start before the Eyes session. This means we need a more complicated solution, since we can't ask for 15 renderings if there is only 10 concurrency.
We need to think of a design that accommodates this requirement.

kyryloonufriiev:
@amitzur please correct me if I am wrong. Would we really want to not start the session on open right now? I believe we could remove `skipStartingSession` from open, and avoid a need to implement our own `ensureRunningSession`.

Also regarding a renders now you on the same page with me. I tried to explain to you exactly this case, but maybe I did a bad job on it. Do you think we need to send render requests separately for each session? or maybe it makes sense to collect a few and only then send?

amitzur:
Regarding starting the session in open - this is a very valid suggestion. We discussed this f2f, but I want to document the arguments.
For:
1. Symmetry with classic runner. There we already start the session in eyes.open. (I don't like this so much, but it is intentional and based on feedback from users.)
2. Easier implementation - we would already know that a session is started when `eyes.check` is called, so we don't need to wait on anything.

Against:
The new concurrency model is already holding back the tests. Between `eyes.open` and `eyes.check` there may be a lot of time spent in which our servers don't do anything. A lot of users put `eyes.open` in `beforeEach` so the calls are not close to each other. This would make tests even slower, and may cause complaints, as there's no real need to enforce concurrency before the first check.

The argument against is stronger, therefore we will start the session lazily.

amitzur:
@kyryloonufriiev I viewed the changes briefly. I want to go over this more thoroghly, but in the meantime you can proceed with measurements:
1. The same as what you did before
2. Raise concurrency to 20, 50, and if it still doesn't match previous performance then also 100.
3. Extreme flow of 1 concurrency, 2 browsers, 100 steps (measure perf and memory)

kyryloonufriiev:
I took some measurements after the changes I made. I think the most interesting of those is the first one (**1500 stories x 1 browser with 10 concurrencies**). We could see that time grew a little bit compared to what we had before the last changes. I believe this happened because of a small delay before start actual rendering which we use to collect and send independent render requests as a batch. The other results as expected just increase performance by a certain number of times. To achieve a performance comparable with what we had before the PR we need to increase concarancy by 5 times.
- 1500 stories x 1 browser / 10 concurrency
  Time: ~36.6m (~2220s)
  Memory: ~109 MB total heap
- 1500 stories x 1 browser / 20 concurrency
  Time: ~17.5m (~1050s)
  Memory: ~109 MB total heap
- 1500 stories x 1 browser / 50 concurrency
  Time: ~6.8m (~410s)
  Memory: ~107 MB total heap

kyryloonufriiev:
Also as expected this is doesn't have a real difference compared to **1500 x 1**
- 100 stories x 15 browser / 10 concurrency
  Time: ~35.1m (~2110s)
  Memory: ~86 MB total heap

kyryloonufriiev:
@amitzur I add `startSessionBeforeRender` flag which is `true` by default. Does it right? Or by default, we should preserve the current behavior?

amitzur:
I made a change in commit 3070bf7 that sets the inferred environment on the wrappers at the very beginning.
I made it so it's part of the same existing flow for getting the rendering info. This is done lazy at the first open and cached in the same place.
I tried to simplify how it's written as well.

amitzur:
2 bugs detected:
1. for devices, there is no waiting for startSession before rendering, because in `EyesWrapper.ensureRunningSession` there is no viewport size. This requires to get this information from the grid. They are adding another endpoint instead of `/user-agents`: https://github.com/applitools/mono/issues/112.
But I would like to change behavior in `ensureRunningSession` so that it doesn't pass through if there's no running session. The new endpoint will make sure there is a viewport size before this is called.

2. debounce in `render` is not correct logic, it can potentially lead to starvation. There should be throttling there.

packages/visual-grid-client/src/sdk/renderingGridClient.js
```diff
@@ -254,7 +258,7 @@ function makeRenderingGridClient({
 
     setRenderingInfo(renderInfo)
     await doLogEvents([
-      BatchStartEvent({concurrency, testConcurrency, finalConcurrency}),
+      RunnerStartedEvent({concurrency, testConcurrency, defaultConcurrency}),
```
kyryloonufriiev, line 261: @amitzur I thought we would like to see `finalConcurrency`.  We will still able to know what we use as default and also see if we correctly multiply legacy concurrency and other things. I'm saving that know the final value we are using is something that would be helpful.

PR #253, opened 2020-09-03
chore: adding docker-compose for chrome and firefox containers

tool3:
adding `docker-compose` shared file, that will spin up `chrome` and `firefox` selenium containers, for the following sdks:
- `eyes-playwright`
- `eyes-protractor`
- `eyes-selenium`
- `eyes-webdriverio-4`
- `eyes-webdriverio-5`


amitzur:
I can report that this works on my machine, and the only failing tests are TestDisableBrowserFetching and TestVisualGridRefererHeader which require additional docker work

amitzur:
ah, sorry, reported too soon.
If tests fail, then docker-compose down doesn't happen. There should probably be a `;` there instead of `&&`

amitzur:
@kyryloonufriiev my bad, I said to add it to playwright as well.

tourdedave (approved):
💯

PR #254, opened 2020-09-04
Fix for iOS Safari

kyryloonufriiev:
[Trello Asos](https://trello.com/c/UumHPGlt/484-asos-wdio5-regression-errors)
[Trello NAB](https://trello.com/c/ZI0UqZqN/500-nabprotractorelement-does-not-exist-in-cache-ios-and-android-devices)
These two cards wait for this fix. 

The original issue happens when you try to call `executeScript` with an object where one of the fields has a dom element as a value. I fixed the original problem by revoking all snippets to work with arrays instead of objects.

During working on this PR I also found other problems for ios safari which our users will face.
1. If you return an object from js snippet and one of the fields is dom element the script will return a wrong element id which is not referred on en actual element. (fixed)
2. Element ids are not stable. If you will find the same element twice you will get two different ids. (fixed)
3. Dom capture script will stack the whole process even if we will use IE script. (need more investigations)


amitzur:
@kyryloonufriiev I added a test for dom-capture on ios: https://github.com/applitools/dom-scripts/commit/0a72fedc0e4a8e2d4739eddada899919506857a1
Seems to work fine. Maybe something in this page. I'm using the poll variant

kyryloonufriiev:
I already find out why it doesn't work and I know how to fix it. We can consult.

kyryloonufriiev:
I moved `DomCapture` tests to the `dom-capture` package since these tests don't have any relation to the `core`. Also, these tests make us use `selenium-webdriver`, but this is unwanted in the core.

Please merge this PR before moving on https://github.com/applitools/dom-scripts/pull/6

Also, this PR is required https://github.com/applitools/dom-scripts/pull/5 to be merged.

packages/eyes-sdk-core/lib/sdk/EyesDriver.js
```diff
@@ -227,7 +227,7 @@ class EyesDriver {
       })
     }
     const framesInfo = await EyesUtils.getChildFramesInfo(this._logger, this)
-    const contextDocument = contextInfo.contentDocument
+    const contextDocument = contextInfo.documentElement
```
amitzur, line 230: was this a bug already in master?

packages/eyes-sdk-core/lib/takeDomCapture.js
```diff
@@ -0,0 +1,164 @@
+'use strict'
+
+const Axios = require('axios')
+const {URL} = require('url')
+const {
+  getCaptureDomAndPollScript,
+  getPollScript,
+  getCaptureDomAndPollForIE,
+  getPollForIE,
+} = require('/Users/ep/Documents/applitools/dom-scripts/packages/dom-capture/index.js')
```
amitzur, line 10: whoops

packages/eyes-sdk-core/lib/takeDomCapture.js
```diff
@@ -0,0 +1,164 @@
...
+    if (!capture) return {}
+    const raws = capture.split('\n')
+    const tokens = JSON.parse(raws[0])
+    const cssEndIndex = raws.indexOf(tokens.separator)
+    const frameEndIndex = raws.indexOf(tokens.separator, cssEndIndex + 1)
+    let dom = raws[frameEndIndex + 1]
+
+    const cssResources = await Promise.all(
+      raws.slice(1, cssEndIndex).reduce((cssResources, href) => {
+        return href ? cssResources.concat(fetchCss(url, href)) : cssResources
```
amitzur, line 62: just curious - filter and map doesn't work here?

packages/eyes-sdk-core/test/unit/takeDomCapture.spec.js
```diff
@@ -0,0 +1,152 @@
...
+        frame: true,
+        children: [{selector: 'frame2-2', frame: true, isCORS: true}],
+      },
+    ])
+    driver = new Driver(logger, mock)
+    await driver.init()
+  })
+
+  it('works', async () => {
+    mock.mockScript('dom-snapshot', () => {
```
amitzur, line 64: shouldn't this be `dom-capture` ? How does it work then?

amitzur (commented):
Looks :100: !
See my other comments, but one general question:
Where is the scroll functionality before/after capturing?

packages/eyes-sdk-core/lib/takeDomCapture.js
kyryloonufriiev: Of course, it does. I just don't want it to be `O(2N)`  when it can be `O(N)`. I understand that in practice the difference isn't big at all.

packages/eyes-sdk-core/lib/takeDomCapture.js
kyryloonufriiev: Very weird. Probably it happened when I unlink `dom-capture` from the core...

packages/eyes-sdk-core/lib/sdk/EyesDriver.js
kyryloonufriiev: No. It wasn't a bug until I changed snippets in this PR.

packages/eyes-sdk-core/test/unit/takeDomCapture.spec.js
kyryloonufriiev: Hmmm. Sure. The sequence of stupid mistakes.

kyryloonufriiev:
Answering this 
>Where is the scroll functionality before/after capturing?

We didn't use it so I didn't implement it.

PR #255, opened 2020-09-06
Add mocha hook to wait for docker to be ready

amitzur:


tourdedave (approved):
I ran into the same problem. Tried solving it with a wait-for-it shell script (commonly referenced in Docker docs) but it wasn't enough. This looks great!

PR #258, opened 2020-09-09
Alt. docker-compose config

tourdedave:
We need to run coverage tests inside of a container from both Linux and Mac, but we also need to have the container access some ports from the host machine.

Sadly our current configuration only supports Mac, and there's no clear path on how to specify a single, platform agnostic configuration.

Specifically, the problems are:

- `network_mode: 'host'` only works on Linux ([link](https://docs.docker.com/network/host/))
- specifying ports as `network: 'host'` works on Mac but not Linux (unable to find supporting documentation)

This PR addresses these limitations by generating a docker-compose file based on the platform.


tourdedave:
Here's an example of it working w/ TestDisableBrowserFetching on Linux - [link](https://eyes.applitools.com/app/test-results/00000251802643567256/00000251802643566959/steps/1?accountId=UAujt6tHnEKUivQXIz7G6A~~&mode=step-editor)

packages/sdk-shared/src/generate-docker-compose-config.js
```diff
@@ -0,0 +1,53 @@
+const fs = require('fs')
+const path = require('path')
+
+function generateDockerComposeConfig({saveToDisk, platform = process.platform} = {}) {
+  let config = {}
```
amitzur, line 5: suggestion:
```
const config = {
  version: '3.4',
  services: {
  // ...
  }
}
```

Use let only in rare cases where you really need to change the value of a variable

packages/sdk-shared/src/generate-docker-compose-config.js
```diff
@@ -0,0 +1,53 @@
...
+      ...generateNetworkConfigForPlatform(platform),
+    },
+    firefox: {
+      image: 'selenium/standalone-firefox',
+      volumes: ['/dev/shm:/dev/shm'],
+      ports: ['4445:4444'],
+    },
+  }
+  const result = JSON.stringify(config)
+  if (!saveToDisk) return result
```
amitzur, line 20: suggestion: 
```
if (saveToDisk) {
fs.writeFileSync(path.resolve(process.cwd(), 'docker-compose.yaml'), JSON.stringify(result))
}
return result
```

That way you always return the configuration and the parameter `saveToDisk` affects exactly what it's expected to affect

packages/sdk-shared/src/generate-docker-compose-config.js
```diff
@@ -0,0 +1,53 @@
...
+}
+
+function generateNetworkConfigForPlatform(platform) {
+  return platform === 'linux'
+    ? {network_mode: 'host'}
+    : {
+        ports: [
+          '4444:4444',
+          {
+            target: '5555',
```
amitzur, line 31: what if we start using port 0? Are we locked into knowing the port in advance on Mac?

amitzur (commented):
I thought about trying this out myself.
It's an ugly solution, really, and I wish there was another way. But nicely done.
Did you call the file `.yaml` so it's possible to run `docker-compose up/down` without specifying `-f`?

If this solves the problem we have right now, IMO we can merge it. But let's try to keep thinking if we can do something else so we don't get locked into it. What concerns me most is the need to know the ports in advance on Mac.

tourdedave:
It's `.yaml` since `docker-compose down` implicitly looks for this and only supports `.yaml`/`.yml`. Didn't try it for up - we might not need to specify `-f` at all, which would be neat.

packages/sdk-shared/src/generate-docker-compose-config.js
tourdedave: Since network_mode host isn't supported on Mac, we're locked into knowing the ports with how the testServer is currently used today.

packages/sdk-shared/src/generate-docker-compose-config.js
```diff
@@ -0,0 +1,53 @@
+const fs = require('fs')
+const path = require('path')
+
+function generateDockerComposeConfig({saveToDisk, platform = process.platform} = {}) {
+  let config = {}
+  config.version = '3.4'
+  config.services = {
+    chrome: {
+      image: 'selenium/standalone-chrome:3.141.59-20200515',
```
tool3: going forward (and given it's actually needed) we can can make this an argument and pass it in the npm script

tool3 (approved):
awesome stuff buddy! 🤘🏼

amitzur:
I believe `docker-compose -f docker-compose.json` should work. Just noting

tourdedave:
It does, and it's what I originally had. But it won't work for `docker-compose down`

amitzur:
Oh sorry, I meant `down`

tourdedave:
> I believe `docker-compose -f docker-compose.json` should work. Just noting

Confirmed. This works for both up and down w/ .json. The error I hit with the .json extension was just when you don't specify the file for down.

PR #259, opened 2020-09-10
feat: cross origin iframes

tool3:
# support for cross origin iframes - sdk
this is the second part of the support for cross origin iframes
see the work done in `dom-scripts`:
https://github.com/applitools/dom-scripts/pull/4


packages/eyes-sdk-core/lib/utils/takeDomSnapshot.js
```diff
@@ -26,13 +26,30 @@ async function getScriptForIE({disableBrowserFetching}) {
...
+      snapshot.frames.push(crossFrameSnapshot)
+      await driver.switchToParentContext()
+    }
+  }
+}
+
+async function getFrame(driver, xpath) {
+  const element = await driver.element({type: 'xpath', selector: xpath})
+  await driver.switchToChildContext(element)
+  return await _takeDomSnapshot({driver})
```
amitzur, line 43: original arguments are not preserved: browser, disableBrowserFetching and startTime.

Regarding `startTime`: current behavior times out after 5 minutes. I don't think we ever hit this limit, but I think we should preserve this behavior, and the 5 minutes limit (or whatever we set it to in the future) should be on the overall operation, including all the cross origin frames work.

amitzur (commented):
Very nice work.
I think that for cleanness sake, `crossFramesXpaths` property should be deleted after it's consumed.
The most outstanding issue right now is that original arguments are not preserved.
I think the way to solve it is to have the cross-origin frames functions **inside** `_takeDomSnapshot` so that the parameters are accessible in the closure.

packages/eyes-sdk-core/lib/utils/takeDomSnapshot.js
```diff
@@ -26,13 +26,30 @@ async function getScriptForIE({disableBrowserFetching}) {
...
+  if (crossFramesXPaths.length > 0) {
+    for (const xpath of crossFramesXPaths) {
+      const crossFrameSnapshot = await getFrame(driver, xpath)
+      snapshot.frames.push(crossFrameSnapshot)
+      await driver.switchToParentContext()
+    }
+  }
+}
+
+async function getFrame(driver, xpath) {
```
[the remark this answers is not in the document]
tool3: I agree !

packages/eyes-sdk-core/lib/utils/takeDomSnapshot.js
```diff
@@ -27,33 +27,58 @@ async function getScriptForIE({disableBrowserFetching}) {
 }
 
 async function takeDomSnapshot({driver, startTime = Date.now(), browser, disableBrowserFetching}) {
```
kyryloonufriiev, line 29: I would love to not pass a `browser` property here. But extract the information we need from a `driver`.

packages/eyes-sdk-core/lib/utils/takeDomSnapshot.js
```diff
@@ -27,33 +27,58 @@ async function getScriptForIE({disableBrowserFetching}) {
...
+    return snapshot
   }
 
-  await GeneralUtils.sleep(PULL_TIMEOUT)
-  return takeDomSnapshot({driver, startTime})
+  async function _takeDomSnapshot() {
+    const processPageAndPollScript =
+      browser === 'IE'
+        ? await getScriptForIE({disableBrowserFetching})
+        : await getScript({disableBrowserFetching})
```
kyryloonufriiev, line 53: I think you need to do it once at the start of the whole `takeDomSnapshot` function. Otherwise, you will read from file N times.

packages/eyes-sdk-core/lib/utils/takeDomSnapshot.js
```diff
@@ -27,33 +27,58 @@ async function getScriptForIE({disableBrowserFetching}) {
...
-    )
+  async function getCrossOriginFrames(driver, snapshot) {
+    const {crossFramesXPaths} = snapshot
+    if (crossFramesXPaths.length > 0) {
+      for (const xpath of crossFramesXPaths) {
+        const crossFrameSnapshot = await takeFrameSnapshot(driver, xpath)
+        snapshot.frames.push(crossFrameSnapshot)
+      }
+    }
+    delete snapshot.crossFramesXPaths
```
kyryloonufriiev, line 38: What if you have a frame (not cors) with cors frame inside?

kyryloonufriiev (changes requested):
Well done. I leave some comments for you

packages/eyes-sdk-core/lib/utils/takeDomSnapshot.js
tool3: all frames are handled by the `processPage` script so when we get here we already have the structure of all frames in the page (including any nested and\or cors frames)

kyryloonufriiev: Let's say you have a page like this
```html
<!-- localhost:3000 -->
<iframe src="localhost:3000/frame">
<iframe src="localhost:3001">
```
This is a frame with cors
```html
<!-- localhost:3000/frame -->
<iframe src="localhost:3002">
```

As far as I understand `processPage` scrip will return you a structure like this
```js
{
  // ...
  frames: [{
    // ...
    frames: [],
    crossFramesXPaths: ['/IFRAME[1]']
    // ...
  }],
  crossFramesXPaths: ['/IFRAME[2]']
  // ...
}
```

kyryloonufriiev: Am I right? If so, will you go to the nested frame to get information about nested cors frame?

tool3: I understand what you mean now... no I am not doing that ! great catch !

amitzur (changes requested):
Reviewed everything except `takeDomSnapshot.js` :)
I will submit a PR to explain my suggestions for `takeDomSnapshot.spec.js`

packages/sdk-shared/coverage-tests/custom/TestCheckDomSnapshotCrossOriginFrames_VG.spec.js
```diff
@@ -0,0 +1,49 @@
...
+  beforeEach(async () => {
+    url = adjustUrlToDocker('http://localhost:7373/handles/cors_frames/cors.hbs')
+    const staticPath = path.join(__dirname, '../fixtures')
+    serverA = await testServer({
+      port: 7373,
+      staticPath,
+      allowCors: false,
+      middlewareFile: path.resolve(__dirname, '../util/cors-snapshot-middleware.js'),
+      hbData: {
+        title: 'Hello, I have a designer cross origin frame',
```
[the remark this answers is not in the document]
tool3: just a joke.. the inner frame says: 'Made by Michael CORS'

packages/sdk-shared/coverage-tests/util/cors-snapshot-middleware.js
```diff
@@ -0,0 +1,18 @@
+const Handlebars = require('handlebars')
+const {readFileSync} = require('fs')
+
+const getFilePath = (url, staticPath) => {
+  return `${staticPath}${url.replace('/handles', '')}`
```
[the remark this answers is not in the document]
tool3: we actually don't need `path.resolve` as I am getting the `staticPath` as a parameter.. so I literally just append `url` to `staticPath`

packages/sdk-shared/src/test-server.js
```diff
@@ -6,24 +6,31 @@ const morgan = require('morgan')
...
+
   const app = express()
   app.use(cookieParser())
   if (allowCors) {
     app.use(cors())
   }
+
   if (middlewareFile) {
-    app.use(require(middlewareFile))
+    const middleWare = hbData ? require(middlewareFile)(argv) : require(middlewareFile)
```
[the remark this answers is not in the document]
tool3: yeah I have been trying to resolve that issue - this works and is straightforward

amitzur (changes requested):
I added to the other PR also suggested changes for `takeDomSnapshot.js`

packages/eyes-sdk-core/lib/utils/takeDomSnapshot.js
```diff
@@ -12,48 +12,96 @@ const CAPTURE_DOM_TIMEOUT_MS = 5 * 60 * 1000 // 5 min
...
+async function takeDomSnapshot({driver, startTime = Date.now(), disableBrowserFetching}) {
+  const {browserName, browserVersion} = driver
+  const isIE = browserName === 'internet explorer'
+  const isEdgeLegacy = browserName.toLowerCase().includes('edge') && browserVersion <= 44
+  const processPageAndPollScript = isIE || isEdgeLegacy ? await getScriptForIE() : await getScript()
+
+  async function getCrossOriginFrames(context, selectorMap) {
+    for (const {path, replace} of selectorMap) {
+      const references = path.reduce((parent, reference) => {
+        return {reference, parent}
```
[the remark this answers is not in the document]
kyryloonufriiev: I would like to explain what does it do. Actualy it converts an array of selectors to a backward linked list. This is a structure supported by `EyesContext#context` to get a nested context.
So if you have `['a', 'b', 'c']`
You will get
```
{
  reference: 'c',
  parent: {
    reference: 'b',
    parent: {
      reference: 'c',
      parent: null
    }
  }
}
```

PR #261, opened 2020-09-21
chore: handlebars

tool3:
support dynamic URLs using handlebars for test server


amitzur:
First of all, well done. I'm going to wave hands a little to explain how I was thinking in a slightly different approach. I didn't think this thoroughly.

`express-handlebars` brings with it a ton of stuff we don't really care about. It's a whole view engine for express, and while you abstract away most of it inside test-server, i think it can be simpler.

The direction I was thinking is to use the already existing `middlewareFile` API and have this sort of middleware (using your `/handles` convention):
```
// your-specific-test's-middleware-file.js
const Handlebars = require('handlebars')
module.exports = ({staticPath, hbData}) => (req, res, next) => {
  if (!req.url.startsWith('/handles')) {
    next()
  } else {
    const html = fs.readFileSync(findPathFromUrl(req.url, staticPath))
    res.render(Handlebars.compile(html)(hbData))
  }
}
```

This requires a change in test-server to accept a middleware file that exports a function that returns the middleware:
```
// test-server.js line 21
if (middlewareFile) {
  app.use(require(middlewareFile)({staticPath, hbData}))
}
```

Now that I'm writing it all I see that it's practically the same amount of changes as in this PR. So maybe not such a big hit. But less abstraction. And no need for layouts and stuff.
But at this point - if it does the work then great.
@kyryloonufriiev @tourdedave wdyt?

tool3:
I like your solution as it gives people the ability to define their own paths and custom behaviour...

Initially I struggled with using a view engine for express (the handlebars package does not provide an engine) but if we use `Handlebars.compile` than it works great with `res.send` (as oppose to `res.render` that works with a view engine)

another thing I took out is the docker URL processing - I prefer doing this in "every" test until we resolve it and the port business altogether

packages/sdk-shared/coverage-tests/custom/TestCheckDomSnapshotCrossOriginFrames_VG.spec.js
```diff
@@ -10,9 +10,18 @@ describe('Coverage Tests', () => {
...
-    url = adjustUrlToDocker('http://localhost:7373/cors_frames/cors.html')
+    url = adjustUrlToDocker('http://localhost:7373/handles/cors_frames/cors.hbs')
     const staticPath = path.join(__dirname, '../fixtures')
-    serverA = await testServer({port: 7373, staticPath, allowCors: false})
+    serverA = await testServer({
+      port: 7373,
+      staticPath,
+      allowCors: false,
+      middlewareFile: path.resolve(__dirname, '../util/cors-snapshot-middleware.js'),
+      hbData: JSON.stringify({
```
amitzur, line 20: why do you need to JSON.stringify it?

packages/sdk-shared/coverage-tests/util/cors-snapshot-middleware.js
```diff
@@ -0,0 +1,19 @@
+const Handlebars = require('handlebars')
+const {readFileSync} = require('fs')
+
+const getFilePath = (url, staticPath) => {
+  url = !url.includes('.hbs') ? `${url}.hbs` : url
+  return `${staticPath}${url.replace('/handles', '')}`
+}
+
+module.exports = ({hbData, staticPath}) => (req, res, next) => {
+  if (req.url.startsWith('/handles')) {
```
amitzur, line 10: maybe `if (/\.hbs$/.test(req.url))` ?

packages/sdk-shared/package.json
```diff
@@ -34,6 +34,8 @@
     "cookie-parser": "^1.4.5",
     "cors": "^2.8.5",
     "express": "^4.17.1",
+    "express-handlebars": "^5.1.0",
```
amitzur, line 37: you don't need this anymore, right?

packages/sdk-shared/src/test-server.js
```diff
@@ -6,24 +6,33 @@ const morgan = require('morgan')
...
   const app = express()
   app.use(cookieParser())
   if (allowCors) {
     app.use(cors())
   }
+
   if (middlewareFile) {
-    app.use(require(middlewareFile))
+    const middleWare = hbData
+      ? require(middlewareFile)({hbData, staticPath})
```
amitzur, line 27: we can pass all the `argv` here i think

packages/sdk-shared/src/test-server.js
```diff
@@ -6,24 +6,33 @@ const morgan = require('morgan')
...
+
   const app = express()
   app.use(cookieParser())
   if (allowCors) {
     app.use(cors())
   }
+
   if (middlewareFile) {
-    app.use(require(middlewareFile))
+    const middleWare = hbData
```
amitzur, line 26: i wouldn't want to differentiate. I think you can check if the middleware is a function or not

amitzur (commented):
Great!
I think there are some nuances that can be done differently which would make this perfect. See comments below

packages/sdk-shared/coverage-tests/custom/TestCheckDomSnapshotCrossOriginFrames_VG.spec.js
tool3: since it's running wrapped in node through yargs - simply providing an object results in `[Object object]`

packages/sdk-shared/src/test-server.js
tool3: I think we have to differentiate... a middleware is just another function.. no ?

the reason I differentiated is due to the nature of the require which is different in both cases.. but please do correct me if i'm wrong

packages/sdk-shared/coverage-tests/util/cors-snapshot-middleware.js
tool3: since this is the "user" middleware - I don't think it's critical.
I thought about supporting both with `.hbs` and without (so that `/handles/cors_frames/cors` would work just the same as `/handles/cors_frames/cors.hbs` - and your suggestion will kind of enforce the `.hbs` (or simply won't run) but i'm down if you think it's more legible

packages/sdk-shared/coverage-tests/custom/TestCheckDomSnapshotCrossOriginFrames_VG.spec.js
amitzur: I think forcing the consumer to stringify makes the abstraction more leaky.
You could send an object here, then in `run-test-server.js` do:
```
const spawnArgs = [
  filepath,
  ...Object.entries(args).reduce((acc, [key, value]) => acc.concat([`--${key}`, typeof value === 'object' ? JSON.stringify(value) : value]), []),
]
```

And in `test-server.js`:
```
const {argv} = require('yargs').option('hbData', {corece: JSON.stringify})
```

packages/sdk-shared/coverage-tests/util/cors-snapshot-middleware.js
amitzur: I just think this can be a general middleware that we have, unrelated to the cors use case.
It actually already is, but you require the file to be placed in a `handles` folder. I think the `.hbs` suffix provides more flexibility to the placement of the hbs file. Plus, you could remove line number 5 above.

amitzur (approved):
Still a couple of comments, but it's really fine nuances. Take it in the direction you want - generally this looks great.

packages/sdk-shared/coverage-tests/util/cors-snapshot-middleware.js
tool3: a valid point indeed 👍🏼

PR #262, opened 2020-09-23
takeDomSnapshot.spec.js suggestion

amitzur:
There are several points here:

1. Be more explicit in tests. Don't try to DRY. Specifying what dom-snapshot returns should be specific to each test, so it should be mocked in each test, rather than in `beforeEach`.

2. Expecting a function to throw: `presult` is a helpful utility when you have an async function that might throw and you want to handle the error.

3. No need for try/catch if catching only throws the error

4. `generateSnapshotObject` is more general and you can specify any overriden property. This leaves room to test also blobs and resourceContents in the future.

5. `should throw an error if timeout is reached` didn't really do anything because it wasn't failing on timeout.

6. `generateSnapshotResponse` has less functionality than the previous `generateSnapshotObject`. If I want to generate an error or WIP response, I don't need that function. Clearer API.

### DIFF
--- a/packages/eyes-sdk-core/lib/utils/createFramesPaths.js
+++ b/packages/eyes-sdk-core/lib/utils/createFramesPaths.js
@@ -1,0 +1,21 @@
+function createFramesPaths(snapshot, path = []) {
+  const paths = snapshot.crossFramesXPaths
+    ? snapshot.crossFramesXPaths.map(selector => ({
+        path: path.concat(selector),
+        parentSnapshot: snapshot,
+      }))
+    : []
+
+  for (const frame of snapshot.frames) {
+    if (frame.selector) {
+      paths.push(...createFramesPaths(frame, path.concat(frame.selector)))
+    }
+  }
+
+  delete snapshot.selector
+  delete snapshot.crossFramesXPaths
+
+  return paths
+}
+
+module.exports = createFramesPaths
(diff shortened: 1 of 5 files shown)


amitzur:
And one additional thing about coding style -
Rarely do we define functions like this:
```
const f = () => {...}
```

I prefer `function f() {}` which makes it hoisted. This correlates with another coding style which is - if you have closures, put them at the end. This way when you read a function you can read it from the beginning, and after the last statement in the function you'll function all the inner functions.

amitzur:
Regarding `takeDomSnapshot.js`:
Coding style:
1. Prefer using `map` instead of `foreach`+`push`
2. Prefer `for .. of` over `foreach`
3. Prefer `if` over `condition && ...` which is less readable.

More meaningful:
1. In order to keep a reference to the parent snapshot, you save a `replace` function. That could easily turn into spaghetti code. The desire is to see code executed serially without jumping too much. What you actually need is just a reference to the `parentSnapshot`.
2. Instead of saving an object of shape `{type: 'xpath', selector}` in the selector map, it makes it clearer to save just the selector. That's the data you want to save. Later, you want to use that selector for a frame reference, so I created an object with that shape at the moment it is needed, and not before.

Still open:
1. I think (but not sure) there's a bug in line 57 - the `.reduce` there confuses me.
2. Unit tests for `createSelectorMap`
3. The name `createSelectorMap` is also confusing to me, since it doesn't really create a map. And the variable `map` inside is actually an array. What is it doing really?

PR #263, opened 2020-09-23
Add "See details at ..." section at the beginning

### LINKED ISSUE #256: eyes-storybook(feature request): show only non-passing tests in the output report
As we've reached the point of 1k+ tests, some machines cannot handle the output log properly, cutting the report's output.
This means that we can't see the bottom line including the link to the applitools dashboard.

I would like to suggest showing an output that includes only the **non-passing tests**, which means it will extremely reduce the log and also focus only on what requires my attention.

This could also be configurable in case you don't think this should be the default behavior.

thanks

EgoziE:
We have over 1000 tests and output gets cut so we don't see the link to applitools.
I added the link to the beginning of the output as well so at least we can click it even the end is cut.

resolves #256 

This is an example for the output:
image

### DIFF
--- a/packages/eyes-storybook/src/processResults.js
+++ b/packages/eyes-storybook/src/processResults.js
@@ -21,6 +21,8 @@ function processResults({results = [], totalTime, concurrency}) {
   );
   errors = flatten(errors).filter(({err}) => err.constructor.name === 'Error');
 
+  outputStr += `See details at ${(passedOrNew[0] || unresolved[0]).getAppUrls().getBatch()}\n\n`;
+
   outputStr += '[EYES: TEST RESULTS]:\n\n';
   if (passedOrNew.length > 0) {
     outputStr += testResultsOutput(passedOrNew);


amitzur:
@EgoziE thanks for your creativity for pushing this forward!
I'll pass this through our product shortly and we'll update.
Do you think after this the #256 issue still needs to be resolved?

EgoziE:
This PR will fix only the most urgent issue: "link is not reachable when log is cut (too long)"
But I still think we need a global solution for the bigger problem, which is the log is too long.

The solution I think is best is a "Log level" option that **for example** can be set to:
- silent - no logs except errors
- diffs only - logs only errors and tests with diffs 
- verbose (default) - logs every test

amitzur:
@EgoziE Thanks again for your help with pushing this.
The "See details..." message is merged in #269 and released in version `3.8.11`.

I'd like to ask about the "log level" solution. In your suggestion there is a missing level for a "trace" level which is the current behavior when setting `showLogs:true`.
To avoid this, we may want to skip the "diffs only" level. I'd like to ask if seeing what stories have diffs in the SDK's output is valuable to you at all, or do you skip it and just head to the dashboard to understand what tests are unresolved?

 If so, the alternative is:
- silent - minimal logs (this could also be called `quiet`)
```
Using @applitools/eyes-storybook version 3.8.11

✔ Storybook was started
✔ Reading stories
✔ Done 1000 stories out of 1000

A total of 8 differences were found.
See details at <some_url>
Total time: <some_time> seconds
```
- info: also shows the list of all tests, including passed (this is what you specifically wanted to avoid)
- verbose: the current behavior when setting `showLogs: true`

amitzur:
Closing this PR as it was released in version `3.8.11`, but would still be interested to hear the answer to the question above.

EgoziE:
First of all i'd like to thank you for accepting my solution and merge it ❤️ this issue was really frustrating.

About your question:
First let's disregard names, I don't mind if you call it silent or quiet or none.
About the `diffs only` option, in short, it's not really mandatory.

**But**, there's a case I see that `diffs only` can be helpful for us.
Visual tests are not 100% stable, therefore knowing in advanced which test failed is really helpful.

Why visual tests are not stable?
- Some tests include textual elements, fonts takes time to load and sometimes doesn't load at all.
- Some tests are "interactive" which means before we take a snapshot, we need to take some action, like hover or click.

We do try to reduce our flaky tests as best as we can, but it's a long process and we have over 1000 tests already. (don't worry we use your API to take snapshot at a specific time, it's not always the case)
Bottom line, `diffs only` is nice to have to save some clicks when CI fails, but not mandatory.

PR #267, opened 2020-09-28
Handling large snapshots

kyryloonufriiev:
This PR is simply an implementation of the chunking feature implemented in https://github.com/applitools/dom-scripts/pull/7.

I also create a common function which implements dom scripts polling protocol. I have it use for both `take Snapshot` and `takeDomCapture`.

I didn't implement the compression part of https://github.com/applitools/dom-scripts/pull/7 on the SDK side, since I would prefer to do it after https://github.com/applitools/eyes.sdk.javascript1/pull/252 is merged. Otherwise, we will need to reimplement it after the merge. The main idea is to "inflate" buffers exactly before sending it to the grid and not early to get a benefit of compressed resources even on the SDK side. Since the "compression" is just a "nice to have" feature we don't urgent with releasing it with this PR.


packages/eyes-sdk-core/test/utils/MockDriver.js
```diff
@@ -5,8 +5,8 @@ const {TypeUtils} = require('../../index')
 const FakeDomSnapshot = require('./FakeDomSnapshot')
 
 const WELL_KNOWN_SCRIPTS = {
-  'dom-snapshot': script => /^\/\* @applitools\/dom-snapshot@[\d.]+ \*\//.test(script),
-  'dom-capture': script => /^\/\* @applitools\/dom-capture@[\d.]+ \*\//.test(script),
+  'dom-snapshot': script => /\/\* @applitools\/dom-snapshot@[\d.]+ \*\//.test(script),
```
amitzur, line 8: why was this needed?

amitzur (approved):
:clap: :clap: :clap: 
I like the API for `EyesUtils.executePollScript` with `scripts.main` and `scripts.poll`
There's no test for that function, but I think it's not a showstopper as this function is in the happy path for every VG based coverage test

amitzur:
@kyryloonufriiev There are some changes in the PR which are actually changes in master. I think those are the places you resolved conflicts.
I can't really review every line from those, but if tests pass then I trust it.

packages/eyes-sdk-core/test/utils/MockDriver.js
kyryloonufriiev: Because of this
```js
`return (${isLegacyBrowser ? await getProcessPagePollForIE() : await getProcessPagePoll()}).apply(null, arguments);`
```
And previously it was
```js
`${await getProcessPagePoll()} return __processPagePoll.apply(null, arguments)`
```

I didn't want to depend on the function name. This is why the `dom-snapshot` comment-marker not on the start of the string.

kyryloonufriiev:
@amit I ran tests in the core. Of course, they are passing.

PR #270, opened 2020-10-01
fix: don't crash when switch to frame fails

tool3:
# take dom snapshot
sometimes switching to a frame failed for whatever reason and `takeDomSnapshot` did not handle it very well.


packages/eyes-sdk-core/lib/utils/takeDomSnapshot.js
```diff
@@ -36,17 +36,24 @@ async function takeDomSnapshot({logger, driver, startTime = Date.now(), disableB
...
-
-      const frameContext = await context.context(references)
-      const contextSnapshot = await _takeDomSnapshot(frameContext)
-      parentSnapshot.frames.push(contextSnapshot)
+      try {
+        const frameContext = await context.context(references)
+        const contextSnapshot = await _takeDomSnapshot(frameContext)
+        parentSnapshot.frames.push(contextSnapshot)
+      } catch (error) {
+        logger.verbose(`could not switch to frame: ${error}`)
```
amitzur, line 46: message should be more descriptive to the operation:
```
could not switch to frame during takeDomSnapshot. Path to frame: ...
```
(I'm not sure what's the best way to describe the path to the frame in the message string. Maybe something like the logging of `paths` in `createFramesPaths.js`)

packages/eyes-sdk-core/lib/utils/takeDomSnapshot.js
```diff
@@ -36,17 +36,24 @@ async function takeDomSnapshot({logger, driver, startTime = Date.now(), disableB
...
+      }
     }
   }
 
   async function _takeDomSnapshot(context) {
-    logger.verbose(`taking dom snapshot. ${context._reference ? `context referece: ${JSON.stringify(context._reference)}` : ''}`)
+    logger.verbose(
+      `taking dom snapshot. ${
+        context._reference ? `context referece: ${JSON.stringify(context._reference)}` : ''
```
amitzur, line 54: this is still a dirty hack. @kyryloonufriiev any better suggestion? Should we do `EyesContext.toString` at this point?

packages/eyes-sdk-core/test/utils/MockDriver.js
```diff
@@ -250,7 +250,7 @@ class MockDriver {
     if (TypeUtils.isString(reference)) {
       reference = await this.findElement(reference)
     }
-    const frame = this._contexts.get(reference.contextId)
+    const frame = reference && this._contexts.get(reference.contextId)
```
amitzur, line 253: I think this is not needed. I checked and `reference` is actually an object in all cases tested by `takeDomSnapshot.spec.js`.
What happens when the frame is not mocked is that `reference.contextId` is undefined, and then the frame is not found and the correct error is thrown.

amitzur (changes requested):
almost there. I added a commit which removes a bit of noise in the test's mocked elements.

PR #273, opened 2020-10-09
Override cors frame's src after dom snapshot is taken

kyryloonufriiev:
[Trello](https://trello.com/c/iBY0LLki)
Related to https://github.com/applitools/dom-scripts/pull/10

- Fix the try-catch block around the context switching, since it catches and ignores all other errors as well. 
- Insert `data-applitools-src` attribute to each cors frame with a value of `frameSnapshot.url`

### DIFF
--- a/packages/eyes-sdk-core/test/it/takeDomSnapshot.spec.js
+++ b/packages/eyes-sdk-core/test/it/takeDomSnapshot.spec.js
@@ -55,20 +55,25 @@ describe('takeDomSnapshot', () => {
   })
 
   it('should take a dom snapshot with cross origin frames', async () => {
-    driver.mockScript('dom-snapshot', function() {
-      return this.name === '[data-applitools-selector="123"]'
-        ? generateSnapshotResponse({cdt: 'frame-cdt'})
-        : generateSnapshotResponse({crossFramesSelectors: ['[data-applitools-selector="123"]']})
-    })
     driver.mockElements([
       {
         selector: '[data-applitools-selector="123"]',
         frame: true,
       },
     ])
+    driver.mockScript('dom-snapshot', function() {
+      return this.name === '[data-applitools-selector="123"]'
+        ? generateSnapshotResponse({cdt: 'frame-cdt', url: 'http://cors.com'})
+        : generateSnapshotResponse({
+            cdt: [{nodeName: 'IFRAME', attributes: []}],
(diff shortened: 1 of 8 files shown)


packages/eyes-sdk-core/test/it/takeDomSnapshot.spec.js
```diff
@@ -232,6 +258,7 @@ describe('takeDomSnapshot', () => {
       {
         selector: '[data-applitools-selector="123"]',
         frame: true,
+        isCORS: true,
```
amitzur, line 261: This parameter is used in the process of taking dom snapshot?

packages/eyes-sdk-core/test/it/takeDomSnapshot.spec.js
```diff
@@ -266,6 +293,36 @@ describe('takeDomSnapshot', () => {
...
+        default:
+          return generateSnapshotResponse({
+            cdt: [{nodeName: 'IFRAME', attributes: []}],
+            crossFrames: [{selector: '[data-applitools-selector="1"]', index: 0}],
+          })
+      }
+    })
+
+    const {cdt} = await takeDomSnapshot(logger, eyesDriver)
+    expect(cdt).to.deep.equal([
```
amitzur, line 322: just noting that `.eql` is an alias for `.deep.equal`

amitzur (approved):
Looks great.
Do you think it's worth adding this to the coverage test `TestCheckDomSnapshotCrossOriginFrames_VG`? (add it to cors.hbs)

packages/eyes-sdk-core/test/it/takeDomSnapshot.spec.js
kyryloonufriiev: No. I just add it to be a purist. Since dom snapshot itself is mocked and it will say which frames need to be treated as a cors.

amitzur: Got it. I removed those from Tal's previous PR because I felt that it might be confusing to think that this property influences the process. I think from the test's perspective it shouldn't be here, but then again it *is* what's created in the real world scenario so maybe it should.

kyryloonufriiev:
@amitzur done!
Now we have all the possible variants. Inline frame with a cross-origin redirect and same-origin frame with cross-origin redirect.

PR #276, opened 2020-10-15
eyes-nightwatch

tourdedave:


packages/eyes-nightwatch/src/spec-driver.js
```diff
@@ -0,0 +1,221 @@
+const {TypeUtils} = require('@applitools/eyes-sdk-core')
+
+//// #region HELPERS
+function extractElementId(element) {
+  const _element = element.value ? element.value : element
```
amitzur, line 5: `element.value || element` - no need for ternary expresssion

packages/eyes-nightwatch/src/spec-driver.js
```diff
@@ -0,0 +1,221 @@
...
+  return Object.values(_element)[0]
```
amitzur, line 6: see `spec-driver` in wdio - it should probably be the same logic. It's an element if it has `element-6066-11e4-a52e-4f735466cecf` or `ELEMENT` property

packages/eyes-nightwatch/nightwatch.conf.js
```diff
@@ -0,0 +1,295 @@
+// Refer to the online docs for more details: https://nightwatchjs.org/gettingstarted/configuration/
+const Services = {}
```
amitzur, line 2: 1. I think this file should be under the `test` folder, as it's used for tests.
2. We should ultimately work against the docker-selenium browsers to be consistent with other SDK's.

amitzur (commented):
WIP review - let's go over this online today.

packages/eyes-nightwatch/nightwatch.conf.js
tourdedave: 1. Fair point. I'll take a look.
2. Of course. Planning to do this after I've sorted getting all of the coverage tests to run.

PR #277, opened 2020-10-17
Kill save debug data

amitzur:
The `saveDebugData` flag hasn't been useful for a very long time. It was initially used to troubleshoot what is received from the SDK as the DOM snapshot, with the CDT and all resources saved on the file system.

It was a very nice feature for troubleshooting, BUT:
1. `vg-cli` is now much more mature than it was, and can provide similar functionality.
2. We didn't productize the way to get these assets from customers, so we didn't really save time on troubleshooting sessions.
3. It wasn't maintained properly over time.
4. Customers kept the flag turned on unknowingly and so received errors and maybe also inflated file system data.

I propose to remove this with a deprecation warning.

In this PR I also introduce `GeneralUtils.deprecationWarning` which was taken from storybook. It has a new variant of `isDead` for features that are no longer supported.


tourdedave:
👏

PR #279, opened 2020-10-20
feat: batch close

tool3:
# Batch Close
provide the ability to manually close batches using an array of `batchIds`

### API
```
const {BatchClose} = require('@applitools/eyes-sdk-core')
async function someFunction() {
...
await BatchClose().setBatchIds(['123', '456']).close()
...
}
```

you can also provide a custom URL:
```
const {BatchClose} = require('@applitools/eyes-sdk-core')
async function someFunction() {
...
await BatchClose().setUrl('example.com').setBatchIds(['123', '456']).close()
...
}
```

### DIFF
--- a/packages/eyes-sdk-core/lib/close/BatchClose.js
+++ b/packages/eyes-sdk-core/lib/close/BatchClose.js
@@ -1,39 +1,38 @@
 function makeBatchClose(closeBatch) {
   return function BatchClose() {
-    this._batchIds = null
-    this._serverUrl = null
-    this._apiKey = null
-    this._proxy = null
-
-    this.setBatchIds = function(batchIds) {
-      this._batchIds = batchIds
-      return this
+    const that = {
+      batchIds: null,
+      serverUrl: null,
+      apiKey: null,
+      proxy: null,
     }
 
-    this.setUrl = function(serverUrl) {
-      this._serverUrl = serverUrl
-      return this
+    that.setBatchIds = function(batchIds) {
+      that.batchIds = batchIds
+      return that
     }
 
-    this.setApiKey = function(apiKey) {
-      this._apiKey = apiKey
-      return this
+    that.setUrl = function(serverUrl) {
+      that.serverUrl = serverUrl
+      return that
     }
 
-    this.setProxy = function(proxy) {
-      this._proxy = proxy
-      return this
+    that.setApiKey = function(apiKey) {
+      that.apiKey = apiKey
+      return that
     }
 
-    this.close = async function() {
-      const serverUrl = this._serverUrl
-      const batchIds = this._batchIds
-      const apiKey = this._apiKey
-      const proxy = this._proxy
+    that.setProxy = function(proxy) {
+      that.proxy = proxy
+      return that
+    }
+
+    that.close = async function() {
+      const {batchIds, serverUrl, apiKey, proxy} = that
       await closeBatch({batchIds, serverUrl, apiKey, proxy})
     }
 
-    return this
+    return that
   }
 }
 
--- a/packages/eyes-sdk-core/lib/close/BatchClose.js
+++ b/packages/eyes-sdk-core/lib/close/BatchClose.js
@@ -1,29 +1,22 @@
-'use strict'
+function BatchClose(closeBatch) {
+  this._batchIds = null
+  this._serverUrl = null
 
-const closeBatch = require('./closeBatch')
-
-class BatchClose {
-  constructor() {
-    this._batchIds = null
-    this._serverUrl = null
-    this._apiKey = null
-  }
-
-  setBatchIds(batchIds) {
+  this.setBatchIds = function(batchIds) {
     if (batchIds && batchIds.length > 0) {
       this._batchIds = batchIds
-      return this
     }
+    return this
   }
 
-  setUrl(serverUrl) {
+  this.setUrl = function(serverUrl) {
     if (serverUrl) {
       this._serverUrl = serverUrl
-      return this
     }
+    return this
   }
 
-  async close() {
+  this.close = async function() {
     try {
       const serverUrl = this._serverUrl
       const batchIds = this._batchIds
@@ -32,6 +25,8 @@ class BatchClose {
       throw new Error(error.message)
     }
   }
+
+  return this
 }
 
 module.exports = BatchClose
--- a/packages/eyes-sdk-core/lib/close/BatchClose.js
+++ b/packages/eyes-sdk-core/lib/close/BatchClose.js
@@ -2,6 +2,8 @@ function makeBatchClose(closeBatch) {
   return function BatchClose() {
     this._batchIds = null
     this._serverUrl = null
+    this._apiKey = null
+    this._proxy = null
 
     this.setBatchIds = function(batchIds) {
       if (batchIds && batchIds.length > 0) {
@@ -17,11 +19,27 @@ function makeBatchClose(closeBatch) {
       return this
     }
 
+    this.setApiKey = function(apiKey) {
+      if (apiKey) {
+        this._apiKey = apiKey
+      }
+      return this
+    }
+
+    this.setProxy = function(proxy) {
+      if (proxy) {
+        this._proxy = proxy
+      }
+      return this
+    }
+
     this.close = async function() {
       try {
         const serverUrl = this._serverUrl
         const batchIds = this._batchIds
-        await closeBatch({batchIds, serverUrl})
+        const apiKey = this._apiKey
+        const proxy = this._apiKey
+        await closeBatch({batchIds, serverUrl, apiKey, proxy})
       } catch (error) {
         throw new Error(error.message)
       }
--- a/packages/eyes-sdk-core/lib/close/BatchClose.js
+++ b/packages/eyes-sdk-core/lib/close/BatchClose.js
@@ -1,23 +1,18 @@
 'use strict'
 
-const ServerConnector = require('../server/ServerConnector')
-const RunningSession = require('../server/RunningSession')
-const Configuration = require('../config/Configuration')
-const Logger = require('../logging/Logger')
+const closeBatch = require('./closeBatch')
 
 class BatchClose {
-  constructor({batchIds = [], serverUrl, apiKey}) {
-    this._batchIds = batchIds
-    this._serverUrl = serverUrl
-    this._apiKey = apiKey
+  constructor() {
+    this._batchIds = null
+    this._serverUrl = null
+    this._apiKey = null
   }
 
   setBatchIds(batchIds) {
     if (batchIds && batchIds.length > 0) {
       this._batchIds = batchIds
       return this
-    } else {
-      // err handling ?
     }
   }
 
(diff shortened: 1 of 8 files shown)


packages/eyes-sdk-core/test/unit/close/closeBatch.spec.js
```diff
@@ -0,0 +1,50 @@
...
+      .replyWithError({message: 'something went wrong', code: 500})
+    const [err] = await presult(closeBatch({batchIds: ['678'], serverUrl, apiKey}))
+    expect(err.message).to.equal('Error: something went wrong')
+  })
+
+  it('should setup configuration', async () => {
+    const [err] = await presult(closeBatch({batchIds, serverUrl, apiKey}))
+    expect(err).to.be.undefined
+    batchIds.forEach((batchId, index) => {
+      expect(scopes[index].basePath).to.equal(serverUrl)
```
amitzur, line 38: I understand you use `scopes` because `serverConnector.deleteBatchSessions` doesn't return anything, right?
Otherwise you could use the reply to communicate the original request parameters back and not need this scope.
Just a thought, but indeed nothing to do here.

packages/eyes-sdk-core/test/unit/close/closeBatch.spec.js
```diff
@@ -0,0 +1,50 @@
...
+      expect(scopes[index].interceptors[0].path).to.include(batchId)
+      expect(scopes[index].interceptors[0].queries).to.eql({apiKey})
+    })
+  })
+
+  it('should call deleteBatchSession per batchId', async () => {
+    await closeBatch({batchIds, serverUrl, apiKey})
+    batchIds.forEach((batchId, index) => {
+      expect(scopes[index].interceptors[0].path).to.include(batchId[index])
```
amitzur, line 47: shouldn't it be `batchIds[index]`? How does this pass?

packages/eyes-sdk-core/test/unit/close/closeBatch.spec.js
```diff
@@ -0,0 +1,50 @@
...
+    const [err] = await presult(closeBatch({batchIds: ['678'], serverUrl, apiKey}))
+    expect(err.message).to.equal('Error: something went wrong')
+  })
+
+  it('should setup configuration', async () => {
+    const [err] = await presult(closeBatch({batchIds, serverUrl, apiKey}))
+    expect(err).to.be.undefined
+    batchIds.forEach((batchId, index) => {
+      expect(scopes[index].basePath).to.equal(serverUrl)
+      expect(scopes[index].interceptors[0].path).to.include(batchId)
```
amitzur, line 39: maybe you can be more explicit and do ``.to.equal(`/api/sessions/batches/${batchId}/close/bypointerid`)``

packages/eyes-sdk-core/test/unit/close/closeBatch.spec.js
```diff
@@ -0,0 +1,50 @@
...
+      expect(scopes[index].interceptors[0].queries).to.eql({apiKey})
+    })
+  })
+
+  it('should call deleteBatchSession per batchId', async () => {
```
amitzur, line 44: what does this test do that `should setup configuration` doesn't? I think it can be removed

packages/eyes-sdk-core/test/unit/close/closeBatch.spec.js
```diff
@@ -0,0 +1,50 @@
...
+  it('should throw', async () => {
+    nock(serverUrl)
+      .delete(`/api/sessions/batches/678/close/bypointerid`)
+      .query({apiKey})
+      .replyWithError({message: 'something went wrong', code: 500})
+    const [err] = await presult(closeBatch({batchIds: ['678'], serverUrl, apiKey}))
+    expect(err.message).to.equal('Error: something went wrong')
+  })
+
+  it('should setup configuration', async () => {
```
amitzur, line 34: A better name would be "should send the correct close batch requests to the server"

amitzur (changes requested):
There are changes to be made.
Also new requirements:
1. add `setApiKey` method to fluent API
2. add `setProxy` method to fluent API and `proxy` param to `closeBatch`. The type of the proxy argument shouldn't matter here. It just passes through into the `Configuration` object.

packages/eyes-sdk-core/test/unit/close/closeBatch.spec.js
tool3: yeah since it doesn't return anything I needed a way to verify it was done successfully.

packages/eyes-sdk-core/test/unit/close/closeBatch.spec.js
tool3: because I am using `.includes` (and so index 0 for string `123` also works)
indeed a warning sign - updating this to explicit string

packages/eyes-sdk-core/test/unit/close/closeBatch.spec.js
tool3: I did exactly that before reading this comment 😆

packages/eyes-sdk-core/lib/close/BatchClose.js
```diff
@@ -0,0 +1,40 @@
+function makeBatchClose(closeBatch) {
+  return function BatchClose() {
+    this._batchIds = null
```
amitzur, line 3: Since this is not called with `new`, then this code currently pollutes `global`.
What you want is to initialize `const ret = {}` and return that.

packages/eyes-sdk-core/lib/close/closeBatch.js
```diff
@@ -0,0 +1,33 @@
...
+  apiKey = process.env.APPLITOOLS_API_KEY,
+  proxy,
+}) {
+  const getAgentId = () => ''
+  const {logger, configuration} = setupServerConfig({serverUrl, apiKey, proxy})
+  const serverConnector = new ServerConnector({logger, configuration, getAgentId})
+
+  const promises = batchIds.map(batchId => serverConnector.deleteBatchSessions(batchId))
+  const results = await Promise.all(promises.map(presult))
+  const err = results.find(([err]) => !!err)
```
amitzur, line 18: it's confusing to call the argument in the inner function `err` and also assign the return value to a variable `err`.
I suggest renaming one of them, since they are not referring to the same thing.

packages/eyes-sdk-core/lib/close/closeBatch.js
```diff
@@ -0,0 +1,33 @@
...
+  if (err) throw err[0]
```
amitzur, line 19: why `err[0]` and not `err`? How did the tests pass? Isn't this a bug?

packages/eyes-sdk-core/lib/close/closeBatch.js
```diff
@@ -0,0 +1,33 @@
...
+  const logger = new Logger(!!process.env.APPLITOOLS_SHOW_LOGS)
+  const configuration = new Configuration()
+  if (serverUrl) configuration.setServerUrl(serverUrl)
+  if (apiKey) configuration.setApiKey(apiKey)
+  if (proxy) configuration.setProxy(proxy)
+  return {configuration, logger}
+}
+
+async function closeBatch({
+  batchIds = [],
```
amitzur: Not good. If user doesn't set batch ID's in the fluent API then you will silently not do anything.
You need to remove the default value and also have a test for this case (using the fluent API).
Notice the nuance in the test: the test for `BatchClose` which doesn't set batch ID's should throw an error that no batch ID's are set. So the `closeBatch` function that you inject needs to throw if there are not `batchIds`.

packages/eyes-sdk-core/test/unit/close/closeBatch.spec.js
```diff
@@ -0,0 +1,57 @@
...
+    nock(serverUrl)
+      .delete(`/api/sessions/batches/999/close/bypointerid`)
+      .query({apiKey})
+      .replyWithError({message: 'something went wrong', code: 500})
+    const [err] = await presult(closeBatch({batchIds: ['888', '999'], serverUrl, apiKey}))
+    expect(err.message).to.equal('something went wrong')
+  })
+
+  it('should send the correct close batch requests to the server', async () => {
+    scopes = []
```
amitzur, line 38: `scopes` is initialized here, so no need to use `let scopes` 30 lines above. Just do `const scopes = []`.
While we're at it, how about:
```
const scopes = batchIds.map(...)
```

packages/eyes-sdk-core/test/unit/close/closeBatch.spec.js
```diff
@@ -0,0 +1,57 @@
+const nock = require('nock')
+const closeBatch = require('../../../lib/close/closeBatch')
+const {expect} = require('chai')
+const {presult} = require('../../../lib/troubleshoot/utils')
+
+describe('closeBatch', () => {
+  let serverUrl, apiKey, batchIds, scopes
```
amitzur, line 7: I think all these variables need to be defined in each test individually. Or, at least define them here with `const`, so it's clear they don't change. But right now the code looks like they are meant to be changed. And there's no reason to initialize their values in `beforeEach`.

amitzur (changes requested):

packages/eyes-sdk-core/lib/close/closeBatch.js
amitzur: Sorry, it's my mistake about the test. It should only verify that `closeBatch` was called with `undefined`.

packages/eyes-sdk-core/lib/close/closeBatch.js
tool3: since the return type of `presult` is `[err, value]`

packages/eyes-sdk-core/lib/close/closeBatch.js
```diff
@@ -0,0 +1,23 @@
...
+const Configuration = require('../config/Configuration')
+const Logger = require('../logging/Logger')
+const {presult} = require('../troubleshoot/utils')
+
+async function closeBatch({batchIds, serverUrl, apiKey, proxy}) {
+  if (!batchIds) throw new Error('no batchIds were set')
+  const serverConnector = new ServerConnector({
+    logger: new Logger(!!process.env.APPLITOOLS_SHOW_LOGS),
+    configuration: new Configuration({serverUrl, apiKey}),
+    proxy,
```
amitzur, line 13: proxy is an argument for `Configuration`, like `serverUrl`. I wonder if this can be tested...

packages/eyes-sdk-core/test/unit/close/BatchClose.spec.js
```diff
@@ -0,0 +1,99 @@
...
+  it('should set proxy', async () => {
+    let proxy = {
+      protocol: 'https',
+      host: 'localhost',
+      port: 1234,
+      isHttpOnly: false,
+      url: 'http://localhost:1234',
+    }
+    let expectedProxy = {}
+    const BatchClose = makeBatchClose(({proxy}) => Object.assign(expectedProxy, proxy))
```
amitzur, line 54: Any special reason for doing `Object.assign` and not `expectedProxy = proxy`?
In any case, if it's like this, prefer `const` over `let`.

packages/eyes-sdk-core/test/unit/close/BatchClose.spec.js
```diff
@@ -0,0 +1,99 @@
...
+      .setUrl(serverUrl)
+      .setBatchIds(batchIds)
+      .setApiKey(apiKey)
+      .setProxy(proxy)
+      .close()
+
+    expect(response).to.eql({serverUrl, batchIds, apiKey, proxy})
+  })
+
+  it('should throw if no batchIds were provided', async () => {
```
amitzur, line 91: `BatchClose` doesn't throw on its own. And there's already a test for `should set batchIds` so maybe you want to add it to that or rename this test to `should set empty batchIds`?

amitzur:
:clap: :clap: :clap: :clap:

PR #281, opened 2020-10-22
Automatically link dev dependencies

amitzur:
This PR changes references to all dev dependencies inside the monorepo to be symlinks.
After merging this, need to run `yarn install --force` to get the symlink.
So from this point on, `yarn link` will only be needed for dependencies, and not dev dependencies.

We will still need to release versions of sdk-release-kit and sdk-shared because of dom-scripts, unless we want to agree on a relative path between these repos, and then we can specify a `link:` in the other repo as well.


tourdedave (approved):
I haven't used `link:` before, but LGTM

amitzur:
@tourdedave because `file:` actually copies the files

tourdedave:
Yeah, that's what I gathered. Much better.

amitzur:
@tourdedave Why should TestCafe pose an issue? Because of Babel? If so, the `link:` is only for dev deps, why would it matter?

PR #283, opened 2020-10-25
fix: remove cyclic dependencies from index

tool3:
# cyclic dependencies
throughout the codebase we had requires like this one: 
`const {SeverConnector, GeneralUtils , ...} = require('../..')`

this is convenient (because otherwise you would have to use 2 different requires from 2 different locations in the example above) but results in many cyclic dependencies:
image

after this PR - **0** cyclic dependencies are created from `index.js` 😃 

going forward - I think it's better to have an `index.js` per folder which will allow for greater grouping when requiring from the same folder, and also give us the opportunity to wrap exported modules (so we wouldn't have to do that in the main `index.js`)

### DIFF
--- a/packages/eyes-sdk-core/lib/AppEnvironment.js
+++ b/packages/eyes-sdk-core/lib/AppEnvironment.js
@@ -1,6 +1,7 @@
 'use strict'
 
-const {GeneralUtils, RectangleSize} = require('..')
+const RectangleSize = require('../lib/geometry/RectangleSize')
+const GeneralUtils = require('../lib/utils/GeneralUtils')
 
 /**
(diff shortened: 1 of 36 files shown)


amitzur:
I think this is a good change. It'a very nice graph :smile: 

I don't like the idea of having `index.js` in every folder. I think it creates an inflation of `index.js` files which then makes it hard to identify which `index.js` file you want to open. I don't see a big problem with how it's done right now in the PR.

And in general I am not for folders at all. Why do they matter? In Java they dictate the class's package path, but in JS it's just for organizing files. But IMO if there's something that's worth a folder - then it's worth a package.
Happy to hear other opinions.

PR #289, opened 2020-11-03
concurrency backoff

amitzur:
For tests blocked by concurrency, the requirement is to have a different backoff than normal failures or polling requests.
The backoff that was decided is 2s x 5, 5s x 4, 10, ....

### DIFF
--- a/packages/eyes-sdk-core/lib/server/requestHelpers.js
+++ b/packages/eyes-sdk-core/lib/server/requestHelpers.js
@@ -158,15 +158,6 @@ async function startPollingRequest({url, config, axios}) {
   }
 }
 
-async function startConcurrencyPolling({config, axios}) {
-  return axios.request({
-    isConcurrencyPolling: true,
-    delay: config.concurrencyBackoff[0],
-    repeat: 0,
-    ...config,
-  })
-}
-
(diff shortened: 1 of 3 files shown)


packages/eyes-sdk-core/lib/server/requestHelpers.js
```diff
@@ -156,7 +157,18 @@ async function startPollingRequest({url, config, axios}) {
...
 }
+
+async function startConcurrencyPolling({config, axios}) {
+  return axios.request({
+    isConcurrencyPolling: true,
+    delay: config.concurrencyBackoff[0],
+    repeat: 0,
+    ...config,
+  })
+}
```
kyryloonufriiev, line 168: I'm not sure why we would need this function. We need it for long requests only because we want to do something after it ended. But in this case, you call it only in the `else` block and it is doing almost the same thing as code in `if` block.

kyryloonufriiev:
I left some minor comment. But regardless it is good to move forward.

kyryloonufriiev (approved):

amitzur:
@kyryloonufriiev thanks! I agree and will fix it

amitzur:
@kyryloonufriiev i tried several options and this one looked to be formatted the best

PR #290, opened 2020-11-03
fix: correct layout region calculation within a region

tool3:
when using eyes classic with a region and specifying a layout region - the calculation didn't consider the offset of the region - and therefore our layout regions would turn out slightly off their respective positions.


amitzur:
@tool3  @kyryloonufriiev  feels like a coverage test is needed here. or a core test. WDYT?

tool3:
> @tool3 @kyryloonufriiev feels like a coverage test is needed here. or a core test. WDYT?

I created a generic coverage test which adds padding to `body` and verifies the layout region coordinates.

https://github.com/applitools/sdk.coverage.tests/pull/11

PR #291, opened 2020-11-05
fix: remove firefox position compensation

tool3:
removing firefox region position compensation which was only needed in old firefox versions where the region would have to be compensated due to inconsistent browser behaviour.

in rare cases it will cause a region out of bounds error (for example when running on a Retina display)


tourdedave:
👏

PR #293, opened 2020-11-06
eyes-puppeteer

tourdedave:
All tests pass and everything runs in docker!

To run a single test, there's a helper script: `yarn moca path/to/spec.js`

There's a different script for running the coverage tests - `yarn test:coverage:local`.

There is a `yarn test:coverage` script, which is used by GitHub Actions.


packages/eyes-puppeteer/CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+# Change Log
+
+## Unreleased
+- updated to @applitools/eyes-sdk-core@12.5.7 (from 12.5.5)
```
amitzur, line 4: can remove this

packages/eyes-puppeteer/.mocharc.cvg.js
```diff
@@ -0,0 +1,13 @@
+const tags = ['chrome', 'chromium']
+module.exports = {
+  spec: [
+    './test/generic/*.spec.js',
+    'node_modules/@applitools/sdk-shared/coverage-tests/custom/**/*.spec.js',
```
amitzur, line 5: aha! You use `node_modules` so this can be used from GitHub Actions!

packages/eyes-puppeteer/.mocharc.cvg.js
```diff
@@ -0,0 +1,13 @@
...
+  ],
+  parallel: true,
+  jobs: 10,
+  timeout: 0,
+  reporter: 'spec-xunit-file',
+  //require: ['../sdk-shared/coverage-tests/util/mocha-hooks.js'],
```
amitzur, line 11: can remove this comment

packages/eyes-puppeteer/package.json
```diff
@@ -0,0 +1,77 @@
...
+    "lib": "./src",
+    "test": "./test"
+  },
+  "main": "./index.js",
+  "dependencies": {
+    "@applitools/eyes-sdk-core": "12.5.7",
+    "@applitools/visual-grid-client": "15.1.1"
+  },
+  "peerDependencies": {
+    "puppeteer": ">=5.3.0"
```
amitzur, line 39: We haven't discussed this. Can we check what versions of Puppeteer we support?
How hard is it to use an earlier version of puppeteer in the docker container?

packages/eyes-puppeteer/package.json
```diff
@@ -0,0 +1,77 @@
...
+    "lint": "eslint '**/*.js'",
+    "render": "node ../sdk-shared/scripts/render.js",
+    "test": "yarn test:it",
+    "test:it": "yarn run:docker yarn mocha --exit --no-timeouts ./test/it/*.spec.js",
+    "test:coverage": "yarn run:docker yarn mocha --exit --config .mocharc.cvg.js",
+    "coverage:generate": "coverage-tests generate ../sdk-shared/coverage-tests/generic/index.js --name 'eyes-puppeteer'",
+    "coverage:report": "coverage-tests report ../sdk-shared/coverage-tests/generic/index.js --name 'eyes-puppeteer'",
+    "coverage:sandbox": "yarn coverage:generate && yarn test:coverage && yarn coverage:report --sandbox",
+    "coverage:prod": "yarn coverage:generate && yarn test:coverage && yarn coverage:report",
+    "preversion": "bongo preversion && yarn coverage:prod",
```
amitzur, line 65: not running `yarn test` on purpose?

amitzur (commented):
Very nice!
I like the docker-first approach. It makes it easier to make it run in GH actions.
You can delete `scripts/output` file.
I didn't go thoroughly into the spec-driver yet, only to the peripherals.
The main open point is knowing what versions of puppeteer we support.

packages/eyes-puppeteer/package.json
tourdedave: We can support different versions of Puppeteer. We just need to map it to the correct version of Chrome/Chromium.

https://github.com/puppeteer/puppeteer/blob/main/versions.js

packages/eyes-puppeteer/package.json
tourdedave: No reason. I'll add it.

tourdedave: FYI - that's how it is in eyes-playwright.

amitzur: thanks for the heads up. I'll fix it

packages/eyes-puppeteer/package.json
tourdedave: So, effectively, different containers running different version of Chrome and Puppeteer that are compatible. Running on GitHub actions means we'd be running latest Chrome, so we'd run latest Puppeteer.

Is this a use case we need to consider?

kyryloonufriiev (changes requested):
Looks pretty great. I'm really glad we can build a new sdk in that straightforward way.

packages/eyes-puppeteer/src/spec-driver.js
```diff
@@ -0,0 +1,272 @@
...
+  frame = extractContext(frame)
+  return frame.parentFrame()
+}
+async function childContext(_frame, element) {
+  return element.contentFrame()
+}
+async function findElement(frame, selector) {
+  const transformedSelector = transformSelector(selector)
+  return transformedSelector.startsWith('xpath=')
+    ? frame.$x(transformedSelector.replace(/^xpath=/, ''))
```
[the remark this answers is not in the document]
tourdedave: 👍

packages/eyes-puppeteer/src/spec-driver.js
```diff
@@ -0,0 +1,272 @@
...
+    : frame.$(transformedSelector)
+}
+async function findElements(frame, selector) {
+  return frame.$$(transformSelector(selector))
```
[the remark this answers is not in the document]
tourdedave: 👍

packages/eyes-puppeteer/src/spec-driver.js
```diff
@@ -0,0 +1,272 @@
...
+        const result = serializeArgs(arg, elements)
+        return result.argsWithElementMarkers
+      } else if (isElement(arg)) {
+        elements.push(arg)
+        return {isElement: true}
+      } else {
+        return arg
+      }
+    })
+  } else if (TypeUtils.isObject) {
```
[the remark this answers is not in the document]
tourdedave: 💯

packages/eyes-puppeteer/src/spec-driver.js
```diff
@@ -0,0 +1,272 @@
...
+      return deserializedArgs
+    }
+    return result
+  }
+  const args = Array.from(arguments)
+  const script = new Function(args[0].script.replace(/^function/, 'return function blah')) // needs a name and to be returned so it's usable
+  const deserializedArgs = deserializeArgs(args[0].argsWithElementMarkers, args.slice(1))
+  const result = args[0].script.startsWith('function')
+    ? await script()(deserializedArgs) // e.g., snippets
+    : await script(deserializedArgs) // e.g., dom-snapshot can be invoked directly
```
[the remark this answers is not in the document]
tourdedave: 👏

packages/eyes-puppeteer/src/spec-driver.js
```diff
@@ -0,0 +1,272 @@
...
+  }
+}
+
+function transformSelector(selector) {
+  if (TypeUtils.has(selector, ['type', 'selector'])) {
+    if (selector.type === 'css') return `${selector.selector}`
+    else if (selector.type === 'xpath') return `xpath=${selector.selector}`
+  }
+  return selector
+}
```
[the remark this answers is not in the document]
tourdedave: Something like this 840f22304f2bf8f3d3a0b0c1edc84b036bc1e3f4?

packages/eyes-puppeteer/src/spec-driver.js
```diff
@@ -0,0 +1,269 @@
...
+  return page.constructor.name === 'Page' ? page.mainFrame() : page
+}
+function isStaleElementError(err) {
+  return (
+    err &&
+    err.message &&
+    (err.message.includes('Execution context was destroyed') ||
+      err.message.includes('JSHandles can be evaluated only in the context they were created'))
+  )
+}
```
kyryloonufriiev, line 120: The only reason why we need to know `stale element error` is to try to re-find an element. In case of these errors, I don't think it will be possible. I would say even more `stale element error` is not something that might happen in cdp.
In playwright, I check in this function for some specific case when the page is reloaded and element reference doesn't exist anymore.

kyryloonufriiev (approved):
👍 
I left only one extra comment. But this is not a dealbreaker.

packages/eyes-puppeteer/src/spec-driver.js
kyryloonufriiev: Actually, I just get that this is (`Execution context was destroyed`) might be an error puppeteer throws in case of the page was reloaded. In this case first one makes sense, but the second still not.

tourdedave: The second error comes from execute script. If you remove it and run test/generic/ShouldNotFailIfScrollRootIsStale.spec.js you will see it.

kyryloonufriiev: Hmmm. I thought this is where the first error will be thrown. So in which case the first error is thrown?

tourdedave: On a call to the element that is no longer in scope is one case.

Something like this: https://github.com/applitools/eyes.sdk.javascript1/blob/puppeteer/packages/eyes-puppeteer/test/it/spec-driver.spec.js#L93

kyryloonufriiev: Ah. If so it's good. By the way, I don't understand what you mean under "scope". In CDP even if you entirely remove an element from the dom you still have a reference on the detached `Node` instance.

tourdedave: That's true, but it's not something you can interact with. By scope I mean it's in context or "is alive"

PR #294, opened 2020-11-09
Universal SDK

kyryloonufriiev:
- [x] universal sdk package design
- [x] communication interface between universal sdk and user api
- [ ] user api package
- [x] universal sdk binary


packages/eyes-api/src/utils/DateUtils.ts
```diff
@@ -0,0 +1,25 @@
...
+    return dateFormat(date, DateFormats.DATE_FORMAT_RFC1123, true);
+}
+
+export function toLogFileDateTime(date: Date = new Date(), utc: boolean = false): string {
+    return dateFormat(date, DateFormats.DATE_FORMAT_LOGFILE, utc);
+}
+
+export function fromISO8601DateTime(dateTime: string): Date {
+    return new Date(dateTime);
+}
```
kyryloonufriiev, line 25: @tool3 What the point of this whole file? You use only one function which is a long way to do `new Date(...)`.
Our main goal is to create a better core and not the same core with ts. So let's move only things we really need.

packages/eyes-api/src/enums/TestResultsStatus.ts
```diff
@@ -0,0 +1,5 @@
+export enum TestResultsStatus {
+    Passed = 'Passed',
+    Unresolved = 'Unresolved',
+    Failed = 'Failed',
+}
```
kyryloonufriiev: @tool3 Please use `export default` instead of `named export`

packages/eyes-api/src/output/StepInfo.ts
```diff
@@ -111,19 +111,19 @@ export default class StepInfoData implements Required<StepInfo> {
         this._hasCurrentImage = value;
     }
 
-    getAppUrls(): AppUrls {
-        return this._appUrls
+    getAppUrls(): AppUrlsData {
+        return this._appUrls as AppUrlsData;
```
kyryloonufriiev, line 115: Curious why something like this might be needed.

packages/eyes-api/src/enums/TestResultsStatuses.ts
```diff
@@ -1,4 +1,4 @@
-export enum TestResultsStatus {
+export enum TestResultsStatuses {
```
kyryloonufriiev, line 1: This should be singular

packages/eyes-api/src/utils/DateUtils.ts
tool3: I figured we would need these utility functions... will remove

packages/eyes-api/src/output/StepInfo.ts
tool3: since we say that `appUrls` can be `AppUrlsData` OR `object`

tool3: since we say that `appUrls` can be `AppUrlsData` OR `object`

kyryloonufriiev: I see in the constructor you make sure this is always the `AppUrlsData`

kyryloonufriiev:
After my latest investigation I find out that we need to extend the protocol to support logging.  We couldn't use stdout for this purposes. So I think we could fire an event from the server in case we want to log some information. We also could optimise it in case of logging in file by writing information by our selves.

packages/eyes-sdk-core/lib/sdk/makeSDK.js
```diff
@@ -20,7 +20,7 @@ function makeSDK({name, version, spec, VisualGridClient}) {
     }
 
     async function close() {
-      return eyes.check(false)
+      return eyes.close(false)
```
amitzur, line 23: oh :joy:

kyryloonufriiev:
I will merge this PR and create a separate one to continue working on the eyes-api package.

PR #296, opened 2020-11-16
fix: user config override

tool3:
fix for user config override.

if a user provides us with `applitools.config.js` we do not take properties which exist on the env.

there is this test:
```js
 it('loads config with env variables', () => {
      process.env.APPLITOOLS_BLA = 'env kuku'
      const config = getConfigAtConfigPath({configParams: ['bla']})
      const expectedConfig = {bla: 'env kuku', it: 'works'}
      assert.deepStrictEqual(config, expectedConfig)
    })
```

I am not sure if this was the author's intent - but this test basically explains the bug. 
if the user provides config property `bla` with value: `kuku` than we take `process.env.APPLITOOLS_BLA` instead of what the user provided.

this issue manifested when someone tried to provide a different `apiKey` than what existed on his env.

### DIFF
--- a/packages/eyes-sdk-core/lib/utils/ConfigUtils.js
+++ b/packages/eyes-sdk-core/lib/utils/ConfigUtils.js
@@ -29,7 +29,11 @@ function getConfig({
 
   const envConfig = {}
   for (const p of configParams) {
-    envConfig[p] = GeneralUtils.getEnvValue(toEnvVarName(p))
+    if (defaultConfig[p]) {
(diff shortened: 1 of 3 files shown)


amitzur:
@tool3 only now I realize what the issue really describes.
It **was** the author's intent. Env variables should override configuration. This let's the user have a base configuration which is committed to source control and shared across users, but override it on different machine, like CI, through env variables.
So this is a non issue.

PR #300, opened 2020-11-19
fix: cypress output

tool3:
fixing cypress output
ticket: https://trello.com/c/HtW8xfdb/471-cypress-test-result-reporting-issues

### improvements 
- **no empty tests**: I decided to not include empty tests in the test results - the use case of many test cases being defined with `eyes.open` and `eyes.close` in `beforeEach` and `afterEach` respectively, and not necessarily doing a `check` in the tests - sounds closer to reality and at which point we don't want to output many empty test results.
- **no empty tests results**:  do not print empty test results (e.g `Passed - 0`)

### possible improvements:
- **test output**: to better align with similar frameworks style outputs - I think the test name + viewport should be in gray when passed and yellow/red according to its status.
- **naming**: I think it would better reflect the naming in the applitools eyes UI if we change the name of 'Diffs detected' to 'Unresolved'.
- **in-app output**: use different output when running inside a browser so the cypress output doesn't have the color format string prefix (`[32m`).

## current shell output
image


packages/eyes-sdk-core/lib/sdk/EyesBase.js
```diff
@@ -1150,7 +1150,9 @@ class EyesBase {
       if (!this._runningSession) {
         this._logger.verbose('Server session was not started')
         this._logger.log('--- Empty test ended.')
-        return new TestResults()
+        const testResult = new TestResults({name: this._configuration._testName})
```
amitzur, line 1153: Worth noting that this will work only because:
1. In visual-grid mode this is actually an instance of EyesWrapper, and used only for one test then gets disposed.
2. in classic mode tests run serially per Eyes instance. so the test name is correct.

packages/eyes-cypress/test/unit/plugin/errorDigest.test.js
```diff
@@ -68,6 +76,51 @@ describe('errorDigest', () => {
        See details at: url1`;
 
     // console.log(_wrap(output)); // debugging
-    expect(output).to.equal(expectedOutput);
+    expect(output).to.deep.equal(expectedOutput);
```
amitzur, line 79: In mocha you can also do `.eql` instead of `.deep.equal`. But why do you need it? It's a string

packages/eyes-cypress/test/unit/plugin/errorDigest.test.js
```diff
@@ -68,6 +76,51 @@ describe('errorDigest', () => {
...
+  });
+
+  it('should only print existing results', () => {
```
amitzur, line 82: According to this test's name, I was expecting to see TestResults for empty tests. What's the purpose of this test?

packages/eyes-cypress/test/unit/plugin/errorDigest.test.js
```diff
@@ -31,19 +31,23 @@ describe('errorDigest', () => {
         name: 'test1',
         hostDisplaySize: {width: 100, height: 200},
         url: 'url1',
+        status: 'Unresolved'
```
amitzur, line 34: Why was it important to add this? The contract is that if this is part of the `diffs` array, it shows up under the Diffs category. Did anything change in that contract?

amitzur (approved):
Very nice change!

packages/eyes-cypress/test/unit/plugin/errorDigest.test.js
tool3: yes, contract had changed since now I am formatting the result based on the test result's status.

packages/eyes-cypress/test/unit/plugin/errorDigest.test.js
tool3: this was a great comment as it allowed me to understand that an empty test result in the passed array would still count it as a passing test - so this comment helped me resolve that bug :) thank you !

PR #302, opened 2020-11-22
chore: use branches and grouped stdout

tool3:
## improvements
- temp branch name with timestamp
- grouped stdout
- run `yarn docker:setup` if `use docker` is `true`
- update master with new version

## possible improvements
- change `use docker` to `prepublish` and have it be a string - so if you need to run `tsc` instead of `yarn docker:setup` you can just provide it from the action menu.
- update branch temp name to be the **intended** version - this is a bit tricky because we need to know the version prior to running `yarn publish`
- rename to `Publish` as now we can use the same workflow for dev\internal packages as well.

### DIFF
--- a/packages/eyes-common-legacy/package.json
+++ b/packages/eyes-common-legacy/package.json
@@ -58,7 +58,7 @@
     "preversion": "bongo preversion",
     "version": "bongo version",
     "postversion": "bongo postversion --skip-release-notification",
-    "setup": "echo 'no setup'"
+    "prepublish:setup": "echo 'no setup'"
   },
   "license": "SEE LICENSE IN LICENSE",
   "engines": {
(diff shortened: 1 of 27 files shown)


packages/eyes-common-legacy/package.json
```diff
@@ -57,7 +57,8 @@
     "typings": "tsc --declaration",
     "preversion": "bongo preversion",
     "version": "bongo version",
-    "postversion": "bongo postversion --skip-release-notification"
+    "postversion": "bongo postversion --skip-release-notification",
+    "setup": "echo 'no setup'"
```
amitzur, line 61: WDYT about `prepublish:setup`? Just `setup` seems to lack meaning

tool3: SGTM :)

PR #303, opened 2020-11-23
fix: bongo commit 

tool3:
- include `yarn.lock` in final commit
- run at the very end of the `preversion` and `deps` flow
- doesn't break CI when some files weren't changed (`git add` and `git commit` give out an exit code 1 if the provided files did not change)


tourdedave (approved):
Looks good. The only open question I have is where git pull is handled (a change not made in this PR). I'll follow up offline.

PR #304, opened 2020-11-24
Storybook using eyes-puppeteer (sort of)

tourdedave:
e.g.., just for calling takeDomSnapshot instead of eval'ing the dom-snapshot script directly. We didn't port eyes-storybook over completely because of performance concerns that came up when reviewing the initial POC.

This implementation runs, but the result from dom-snapshot does not contain blobs for some reason. It throws on `deserializeDomSnapshotResult` when trying to reduce the blobs.

It might be an issue in the serializer/deserializer used in executeScript in eyes-puppeteer. Need to investigate.


packages/eyes-storybook/src/eyesStorybook.js
```diff
@@ -54,9 +56,17 @@ async function eyesStorybook({
...
-  const processPageAndSerialize = `(${await getProcessPageAndSerialize()})(document, {useSessionCache: true, showLogs: ${
-    config.showLogs
-  }, dontFetchResources: ${config.disableBrowserFetching}})`;
+  const driver = new Driver(logger, page);
+
+  const processPageAndSerialize = async () => {
+    const domSnapshotOptions = {
+      useSessionCache: true,
+      showLogs: !!config.showLogs,
+      disableBrowserFetching: !!config.disableBrowserFetching,
```
amitzur, line 65: `dontFetchResources`

packages/eyes-storybook/src/eyesStorybook.js
```diff
@@ -54,9 +56,17 @@ async function eyesStorybook({
 
   const pagePool = createPagePool({initPage, logger});
 
-  const processPageAndSerialize = `(${await getProcessPageAndSerialize()})(document, {useSessionCache: true, showLogs: ${
-    config.showLogs
-  }, dontFetchResources: ${config.disableBrowserFetching}})`;
+  const driver = new Driver(logger, page);
```
amitzur, line 59: `page` in this scope is the master page from where we extract the list of stories.
`takeDomSnapshot` needs to run on a driver that wraps the specific page of the story. I'll comment also on the relevant place in `renderStory`

packages/eyes-storybook/src/eyesStorybook.js
```diff
@@ -54,9 +56,17 @@ async function eyesStorybook({
...
+
+  const processPageAndSerialize = async () => {
```
amitzur, line 61: You will need to pass in `driver` or `page` here. See comment in `renderStory`.

packages/eyes-storybook/src/getStoryData.js
```diff
@@ -52,9 +52,12 @@ function makeGetStoryData({
     }
 
     logger.log(`running processPageAndSerialize for story ${title}`);
-    const {resourceUrls, resourceContents, frames, cdt} = await page
-      .evaluate(processPageAndSerialize)
-      .then(deserializeDomSnapshotResult);
+    const {resourceUrls, resourceContents, frames, cdt} = await processPageAndSerialize().then(
```
amitzur, line 55: This is where you want to do
```
await processPageAndSerialize(page)
```

(when you finalize the PR I think `processPageAndSerialize` should be renamed)

amitzur (commented):
I think I know why you're not getting the expected results. You run `takeDomSnapshot` only on the master page, not on the stories themselves.

packages/eyes-storybook/src/eyesStorybook.js
tourdedave: Something along these lines?

https://github.com/applitools/eyes.sdk.javascript1/pull/304/commits/a9c1eb416c2be4907e12ca74683781cac033471a

Still getting the same result 😢

amitzur: Exactly, yes. Maybe you do have an issue in deserialization.

packages/eyes-storybook/src/eyesStorybook.js
tourdedave: I set this to `disableBrowserFetching` because of this: https://github.com/applitools/eyes.sdk.javascript1/blob/storybook+puppeteer/packages/eyes-sdk-core/lib/utils/takeDomSnapshot.js#L23

amitzur: haha :+1:

tourdedave:
Serialization/deserialization looks fine. Makes sense since all of the eyes-puppeteer VG tests pass and they use takeDomSnapshot.

I took a closer look at the CDT. Seeing things like this: `Sorry, but you either have no stories or none are selected somehow.` So the original theory still holds true. It looks like an issue of not accessing the story.

The URL in the DS output is `http://localhost:9000/iframe.html?eyes-storybook=true`, which leads to a page saying this:

```
No Preview
Sorry, but you either have no stories or none are selected somehow.

Please check the Storybook config.
Try reloading the page.
If the problem persists, check the browser console, or the terminal you've run Storybook from.
```

tourdedave:
I tried passing around the driver and using it to visit the stories (see https://github.com/applitools/eyes.sdk.javascript1/pull/304/commits/9ce34598fe08a8f7a9099bf00333676b9912444e). Still getting the same result.

amitzur:
@tourdedave the issue was that `takeDomSnapshot` already takes care of deserializing the result. So you just need to remove this extra call from storybook's code.
I also took the liberty to simplify (IMO) the definition of `doTakeDomSnapshot`.

tourdedave:
Set this as ready for review, but just realized there are some test failures. Looking.

tourdedave:
All existing tests pass now. Just need to add a new test with a cross-origin iframe.

tourdedave:
Alright, CORS test added.

amitzur (approved):
:100:

PR #305, opened 2020-11-27
Eyes API

kyryloonufriiev:
- [ ] Finalize ts work
- [ ] Create a generic test suite to test the functionality of the API
- [ ] Merge class/method descriptions with generated documentation
- [ ] Create an Applitools theme for typedoc


amitzur:
@kyryloonufriiev this PR is empty. Is it a placeholder?

kyryloonufriiev:
Today our SDKs are re-exporting some abstractions from the core which are not really useful for users.
Here is a short overview of something I think we don't need to export anymore.
1. `ImageProvider` - there is no legal way to provide or get an object of this class.
2. `FullPageCaptureAlgorithm` - there is no legal way to provide or get an object of this class. Also, this class is a part of internal logic that might change from version to version.
3. `EyesSimpleScreenshotFactory` - also not really useful. This class is not even used in our current implementation, only in `eyes-images`
4. `ScaleProvider`, `FixedScaleProvider`, `FixedScaleProviderFactory` - part of the internal implementation which couldn't be received from the outside.
5. `PositionProvider` - this is actually something the user could set now using `setPositionProvider`. But this is not documented, and when I did changes in this class I never preserve an API (since I treated it as internal), so probably no one really uses it. 
6. `PositionMemento` - something PositionProvider could return from one of the methods, doesn't make sense to have it without PositionProvider
7. `MutableImage` - this is our internal image wrapper for image manipulations, objects of this type are not reachable for users
8. `ImageDeltaCompressor` - internal implementation for image compression
9. `PropertyHandler`, `ReadOnlyPropertyHandler`, `SimplePropertyHandler` - internal implementation, not even natural for JS, want to rid of after refactoring our internal core.
10. `DebugScreenshotsProvider`, `FileDebugScreenshotsProvider`, `NullDebugScreenshotProvider` - internal data class
11. `RegionProvider`, `NullRegionProvider` - internal abstractions, with no reason for users to know about.

As we all can see today we are exporting a bunch of classes which are have nothing to do with user-facing API, the biggest part of them will be naturally eliminated once we will migrate to a new architecture of the core.

kyryloonufriiev:
> This is Playwright - I'd like to deliver to Yonatan Selenium first.

I will easily implement it for all sdks once I'm done with one.

> AccessibilityMatchSettings constructor(x,y,width,height,type) - this doesn't exist today. Why add it?

I add this constructor a long time ago just because I didn't notice an actual implementation in the core

> toString and toJSON on data classes should not be documented

I will mark them with `@internal` tag

> BatchInfo isGeneratedId should not be documented

Then I probably will remove this since we don't use it internally. What about `isCompleted`?

> CheckSettings all regions have multiple variations (ignore, ignores, ignoreRegion, ignoreRegions) - we should document only the ignoreRegion(s) variations

I will mark ignore(s) with `@deprecated` tag. 

> ConfigurationData data class should be renamed to Configuration

This is just a mistake

> Configuration data class has getters to check() general() classic() vg() - where are these used? they are internal

I will mark them with `@internal` tag

> RenderInfo type - I think we will get notified from Yonatan that other SDK's call this in a different way

I see in java documentation there are three classes `ChromeEmulationInfo`, `IosDeviceInfo` and `DesktopBrowserInfo` I can export something like this as well

> RegionPlain - I still need to figure out the confusion in our codebase and its proper user API

In our current codebase, we have only regions with `left` and `top`

> Region - Required<{left: number; top: number} & LocationPlain & RectangleSizePlain> it should be Required<({left: number; top: number} | LocationPlain) & RectangleSizePlain> no?

Nope. You are not able to extend unions

> VisualGridRunner - legacyConcurrency should not be documented

I will mark it with `@internal` tag

> Eyes - classic API methods shouldn't be documented (checkWindow etc.)

I will mark classic API with `@deprecated` tag. 

> Eyes - get config() should be renamed to the full name get configuration()

I will rename it.

> Eyes.setViewportSize static method still has <TDriver>

Still looking for an elegant solution

> VGConfigurationPlain - concurrentSessions should not be documented

Do we even need it?

kyryloonufriiev:
This PR looks pretty much done from my point of view. There might be some small fixes due to Yonatan's work on documentation. You can track documentation progress [here](https://docs.google.com/spreadsheets/d/1bsgz-3ImKJhxAOifGTsbltDsQVfkqeqcgzOqtRKNgi8/edit?ts=604f8804#gid=1187770102).

## What's new this PR introduces?
The biggest benefit of this PR I can notice is an introduction of a new part of the SDK architecture - the full-functional API layer. The API layer implementation is just a logical continuation of the very first universal SDK POC. Same as Eyes Universal package creates WebSocket based API on top of the remote core implementation, Eyes API package creates JS classes and functions on top of the same core.
## Why is this important?
Once we will have split API and the actual logic of the SDK it will unbind our hand to do more changes in the core. Also, it makes us simplify a core functionality, instead of implementation when anything could change at any time we will have a structure with a much simpler data-flow with a single entry point. It will simplify understanding, debugging, and testing a lot! I don't even say that our current implementation of the core is inherited from Java and some of the patterns there are not meant to be used in JS. Fortunately, after this PR, we will be free to implement core in a way we want it.
## What else?
### TypeScript
Not only brand new Eyes API package was written in typescript but also all (almost) SDK packages were rewritten using typescript to achieve the best integration with Eyes API.
### Better API
With this PR in production, we will support plain JS objects instead of data classes in every possible place. Also, all data classes now could be treated as a plain object because every `getXXX`/`setXXX` function will have `get xxx`/`set xxx` accessor equivalent.
I finish the transition of configuration properties from `Eyes` instance to `Configuration` object, which means after this PR everything could be configured with just one object.
### Logger
As a part of this PR, I had to solve a problem with logging. The problem is clear - we may want to log something on Eyes API level as well as on Core. It means that logger couldn't be a part of the core anymore. I create a separate logger package with Eyes Universal and Eyes API in mind. I also use a chance to improve something in the logger itself. I'm open to discussion about my implementation since I'm not really efficient in loggers and stuff.
### Beautiful `.d.ts` files and documentation
Another package that saw the light during the implementation of this PR is API Extractor which is just a wrapper around `typedoc` package but instead of building a website or JSON representation of our API, it generates ONE `.d.ts` files with everything. It allows Yonatan to generate documentation (api reference) for our SDKs.
## Open points
1. Eyes Images doesn't support the API layer and still depends on API implementations from the core. Solution: We will need to implement a separate package for communication with the Eyes server as one of the text steps. Then we will implement Eyes Images on top of this package instead of "core" which should be dedicated to working with "driver".  Then we could improve Eyes API to fit the needs of Eyes Images as well as all other sdks.
2....

I'm sure there is something else, but right now only that comes to my mind.

PR #306, opened 2020-11-30
chore: using snaptdout

tool3:
# snapshot testing
using a small utility I wrote, we can now use stdout snapshot testing that will validate output of `errorDigest` instead of intricate string comparison.

# benefits
- saves output to a json file line by line
- validates each snapshot by name
- per character diff errors
- no dependencies


tourdedave:
Is this handled by #330?

tool3:
> Is this handled by #330?

as we discussed this is the usage of `snaptdout` while #330 is the actual packages that will make this PR possible.

PR #311, opened 2020-12-02
Fix arguments serialization for puppeteer

kyryloonufriiev:


packages/eyes-puppeteer/src/spec-driver.js
```diff
@@ -62,19 +54,22 @@ function serializeArgs(args, elements = []) {
...
-    } else if (Array.isArray(args)) {
-      return args.map(arg => {
-        if (Array.isArray(arg)) return deserializeArgs(arg, elements)
-        return arg && arg.isElement ? elements.shift() : arg
-      })
-    } else if (typeof args === 'object') {
-      const deserializedArgs = {...args}
-      for (const [key, value] of Object.entries(args)) {
-        if (value.isElement) deserializedArgs[key] = elements.shift()
+    return args.map(deserializeArg)
```
amitzur, line 57: So let's not have `deserializeArgs` at all, have `deserializeArg` use `elements` from the closure (need to define `const elements = args.slice(1)`) and in line 80 we'll do:
```
const deserializedArgs = args[0].argsWithElementMarkers.map(deserializeArg)
```
@kyryloonufriiev @tourdedave wdyt?

kyryloonufriiev: Absolutely!

tourdedave: 👍

tourdedave (approved):

PR #316, opened 2020-12-08
Support dom-snapshot skip list

amitzur:
This PR adds support for DOM snapshot skip list.
It relies on addition API's from DS, from here: https://github.com/applitools/dom-scripts/pull/21

This feature is a performance optimization which causes resources necessary for UFG rendering not to be fetched in DOM snapshot more than once per runner.
Implementation is pretty simple: 
1. Send all the keys that exist in the SDK's resource cache to DS as the skip list (a.k.a `skipResources` argument).
2. Save the dependency graph of blobs received from DS into the SDK's resource cache.

A few points worth noting:
1. There's an existing assumption that breaks with the skip list, that resources fetched by DS will always be fetched by DS. The change in `getAllResources` shows exactly how this broken assumption is handled: the dependencies for each resource need to be saved to the cache, and this assures that even a blob received from DS can be taken from the cache with all its dependency graph.
2. Another assumption that breaks here, is that the same URL always serves the same content for the entire execution. This cannot be handled by us. Going forward we'll want to introduce a configuration parameter to disable the skip list to avoid customers' edge cases which break this assumption.
3. I needed to expose the resource cache from VGC to core. I made a conscious decision to not expose the entire cache, but only its keys as readonly. I thought it would be better for encapsulation if we don't let external modules manage the cache. Maybe there's a better way to do it than what I did (expose `getResourceUrlsInCache`), but the design principle is important IMO.
4. `ephemeral-middleware` serves the smurfs images only 2 times per image. I could not distinguish in the middleware which request comes from DS and which request is a built-in browser request. So this creates a situation that after reloading the page in the coverage test, the page is actually blank (!!) but the image turns out ok because of our cache. You could say it's a bug (see item 2 above) but it's the best test I could think of.


kyryloonufriiev (approved):
Simple and elegant. Looks totally great for me.

packages/sdk-shared/coverage-tests/util/ephemeral-middleware.js
```diff
@@ -0,0 +1,16 @@
+// this middleware causes smurfs.jpg to be served only twice - once for the page and once for dom snapshot
+const servedCount = {}
+module.exports = (req, res, next) => {
+    if (req.url.includes('smurfs')) {
+      servedCount[req.url] = servedCount[req.url] ? servedCount[req.url] + 1 : 1
+      if (servedCount[req.url] === 2) {
+        res.status(404).send('Not found')
+      } else {
+        wasServed = true
```
amitzur, line 9: whoops. can be removed

PR #318, opened 2020-12-09
Long running tasks enhancement

kyryloonufriiev:
Trello https://trello.com/c/ZX8NX2zg/665-gettaskstatus-enhancement

### DIFF
--- a/packages/eyes-sdk-core/CHANGELOG.md
+++ b/packages/eyes-sdk-core/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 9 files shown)


tool3 (commented):
looks great buddy 👏🏼

tourdedave (approved):
LGTM

I tested it out with a few coverage tests in eyes-selenium against production (e.g., what will soon be the "legacy" polling) 💯 

The real test will be when Eyes rolls it out, but this implementation matches the spec and is well tested. So, I'm confident we're in good shape.

PR #320, opened 2020-12-11
Mobile coverage tests

Artem-applitools:


Artem-applitools:
JS on native devices have differences in viewport sizes so it won't match the viewport size from the other sdks, so for now we need to run the mobile tests on different branch.

kyryloonufriiev (approved):

PR #322, opened 2020-12-11
fix: Appium viewport size extraction

kyryloonufriiev:
I find out that the capabilities of the appium drivers contain information that allows us to detect viewport size correctly without doing any extra calls to the driver. Before this PR the viewport size (and screenshot) never wasn't properly scaled and also include a status bar on the top.


amitzur:
@kyryloonufriiev i missed this PR. Does it have a Trello card?
Is there a way to test this?

kyryloonufriiev:
This PR was created because Artem and Yurii add mobile native generic tests which are not passing in js due to wrong scaling.

amitzur:
Can you reference the tests that they added?

kyryloonufriiev:
@amitzur https://github.com/applitools/sdk.coverage.tests/blob/master/coverage-tests.js#L1624-L1720

PR #325, opened 2020-12-17
fix: fixing storybook cli

tool3:
## storybook cli
added:
- usage of `yargs` defaults.
- `server-url` option that will run against a non-default eyes-server. (defaults to `https://eyes.applitools.com`.
- `config` with default value instead of the hardcoded `-f <some-path` in `package.json`.

## improvements that are not in this PR
- use `yargs` as god intended.
- for all sdks: use a unified reporting library for test results

### DIFF
--- a/packages/eyes-storybook/src/yargsOptions.js
+++ b/packages/eyes-storybook/src/yargsOptions.js
@@ -2,17 +2,23 @@ module.exports = {
   conf: {
     alias: 'f',
     description: 'Path to applitools.config.js config file',
+    default: 'test/fixtures/applitools.config.js',
     requiresArg: true,
     string: true,
   },
-
+  serverUrl: {
+    alias: ['server-url', 'srv'],
+    description: 'URL to eyes server',
+    default: 'https://eyes.applitools.com',
+    requiresArg: true,
+    string: true,
+  },
   storybookUrl: {
     alias: ['storybook-url', 'u'],
     description: 'URL to storybook',
     requiresArg: true,
     string: true,
   },
-
   // storybook options
   storybookPort: {
     alias: ['storybook-port', 'p'],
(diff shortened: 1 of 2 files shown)


packages/eyes-storybook/src/yargsOptions.js
```diff
@@ -2,17 +2,23 @@ module.exports = {
   conf: {
     alias: 'f',
     description: 'Path to applitools.config.js config file',
+    default: 'test/fixtures/applitools.config.js',
```
amitzur, line 5: This is the user facing cli args.

amitzur (changes requested):
What popped the need for this?
Changes here don't seem relevant, since (a) they treat user facing cli arguments with a default intended for internal tests, and (b) `server-url` from the CLI is a nice option, but where did the requirement come from? There are many other configuration parameters that are available in `applitools.config.js` and not in the cli, so I don't want to open this pandora box. Or in other words - handle it holistically and not with a one-off argument.

packages/eyes-storybook/src/yargsOptions.js
tool3: there was no need for this.
I added the serverUrl after testing something for Adam with the CLI which proved to be not smooth at all.
the default config passed to the cli in `package.json` was the thing that confused me, it was adding my config path to the existing config path passed to the cli - at which point `yargs` turned it to an array of paths.

we can create a separate PR if we ever want to improve it.

amitzur: oh, really? you mean it was like running `node bin/eyes-storybook.js -f test/fixtures/applitools.config.js -f your/applitools.config.js` ?

tool3: exactly ! 
`yargs` automatically coerces types to it's best understanding.
so argv looked like:
```javascript
{... ,  f: ['test/fixtures/applitools.config.js', 'your/applitools.config.js'] ,...}
```

but I understand that this is a usage bug that occurred to me and would never happen to customers.

PR #328, opened 2020-12-22
add streaming resource timeout

yarden-wieder:


packages/visual-grid-client/src/sdk/fetchResource.js
```diff
@@ -2,33 +2,65 @@
 const retryFetch = require('@applitools/http-commons/src/retryFetch')
 const createResourceCache = require('./createResourceCache')
 
-function makeFetchResource({logger, retries = 5, fetchCache = createResourceCache(), fetch}) {
+const psetTimeout = t =>
+  new Promise(res => {
+    setTimeout(res, t)
+  })
+
```
tourdedave, line 9: This function would be more readable wrapped in brackets (at first glance I was thrown off why it wasn't since it was a function w/ a multi-line body)

packages/visual-grid-client/src/sdk/fetchResource.js
```diff
@@ -2,33 +2,65 @@
...
+const noTimeout = new Promise(() => {})
+function bufferTimeout(response, url, timeout) {
+  const canBeStreaming =
```
tourdedave, line 12: I vote for more crisp wording, like `isStreaming`

packages/visual-grid-client/test/unit/sdk/fetchResources.test.js
```diff
@@ -83,4 +83,64 @@ describe('fetchResource', () => {
...
+      const url = 'http://something2'
+      nock(url)
+        .get('/')
+        .delayBody(200)
+        .reply(200, 'bla', {'content-type': 'audio/content-type', 'content-length': 3})
+
+      await fetchResource(url).then(
+        resource =>
+          expect(resource).to.eql({url, type: 'audio/content-type', value: Buffer.from('bla')}),
+        expect.fail,
```
tourdedave, line 127: I'm confused by this line. Why is it here? It seems to contradict the test.

packages/visual-grid-client/test/unit/sdk/fetchResources.test.js
```diff
@@ -83,4 +83,64 @@ describe('fetchResource', () => {
...
+      const url = 'http://something2'
+      nock(url)
+        .get('/')
+        .delayBody(200)
+        .reply(200, 'bla', {'content-type': 'some/content-type'})
+
+      await fetchResource(url).then(
+        resource =>
+          expect(resource).to.eql({url, type: 'some/content-type', value: Buffer.from('bla')}),
+        expect.fail,
```
tourdedave, line 142: Same here too. Why is it here? It seems to contradict the test.

tourdedave (changes requested):
Looks good! Just a couple of small things to address.

amitzur (commented):
WIP review

packages/visual-grid-client/src/sdk/fetchResource.js
amitzur: I think because it's a heuristic then it's more like `possiblyStreaming`, `mightBeStreaming`, `probablyStreaming`, `couldBeStreaming`, `maybeStreaming` or any variation of them.... I could go on :)

packages/visual-grid-client/test/unit/sdk/fetchResources.test.js
tourdedave: Then perhaps it's the name I find confusing. If it doesn't apply to requests with a content length, then why would we expect this test to fail?

packages/visual-grid-client/src/sdk/fetchResource.js
```diff
@@ -2,33 +2,65 @@
...
+    !response.headers.get('Content-Length') &&
+    ['audio/', 'video/'].some(prefix => response.headers.get('Content-Type').startsWith(prefix))
```
amitzur, line 14: @yarden-wieder I find this style more readable. what do you think?
```
/^(audio|video)\//.test(response.headers.get('Content-Type'))
```

packages/visual-grid-client/src/sdk/fetchResource.js
```diff
@@ -2,33 +2,65 @@
...
+
+    if (!resp.ok) {
+      logger.verbose(`failed to fetch ${url} status ${resp.status}, returning errorStatusCode`)
+      return {url, errorStatusCode: resp.status}
+    }
+
+    logger.verbose(`fetched ${url}`)
+
+    const bufferPromise = resp.buffer ? resp.buffer() : resp.arrayBuffer()
+    const buffer = await Promise.race([
```
amitzur, line 55: I think it makes more sense not to use `Promise.race` if this isn't streaming. Then you will not need the weird `noTimeout` promise above

packages/visual-grid-client/src/sdk/fetchResource.js
```diff
@@ -2,33 +2,65 @@
...
+  })
+
+const noTimeout = new Promise(() => {})
+function bufferTimeout(response, url, timeout) {
+  const canBeStreaming =
+    !response.headers.get('Content-Length') &&
+    ['audio/', 'video/'].some(prefix => response.headers.get('Content-Type').startsWith(prefix))
+
+  if (canBeStreaming) {
+    return psetTimeout(timeout).then(() =>
```
amitzur, line 17: why not use `await` here?

amitzur (changes requested):
There are 3 main issues here:
1. We are now retrying only on the initial response and not on the buffer. I think network conditions might cause the download to fail, and then we would like the retry. In case we time out on streaming we should not retry because it's not considered a failure.
2. We should return an `errorStatusCode` when timing out. Not throw an exception. Otherwise the URL will not be mapped. I suggest `599` so it's clear that it's something we made up and not coming from the webapp itself.
3. Using `Promise.race` when there isn't really a race seems like bad practice. The better way IMO is to use it conditionally if indeed streaming is suspected.

packages/visual-grid-client/test/unit/sdk/fetchResources.test.js
amitzur: The `expect.fail` is in the second parameter to `.then`. So essentially he's saying - fail the test if `fetchResource(url)` throws.
I agree this is redundant and confusing.

amitzur:
@yarden-wieder I fixed the 3 items mentioned above :point_up: https://github.com/applitools/eyes.sdk.javascript1/pull/328#pullrequestreview-562188010
And then when running it against the customer's page I saw that the request also needs to be aborted with AbortController. So I added it as well.
There is one flow I couldn't get to test properly, so I'm adding a draft PR for my attempt.
I believe this is ready for merge.

PR #329, opened 2020-12-27
feat(cypress): support js layouts

tool3:
### cypress js layouts
this PR introduces support for js layouts.

note that promises in `eyesCheckWindow` must be `Cypress` promises (not **real** promises!)

### general
- **config**: config is being handled in `commands.js` and passed down to `eyesCheckWindow` from the cypress command defined for it.
- **e2e test**: `test/e2e/cypress-run-breakpoints.test.js`
- **verbose props**: any property which is not sent in `args` to `eyesCheckWindow` will **NOT** be returned - this is different from the previous behavior where we would return `undefined` per property explicitly.
- **waits**: Cypress promises are not real promises. therefore, we added a `wait(0)` to all cypress calls - this was necessary due to the nature of how cypress queues commands.
the only place where we used `wait(300)` is in `cypress.viewport` where we needed this delay to ensure consistent and reliable viewport change **prior** to taking the snapshot.


packages/eyes-cypress/test/fixtures/testApp/cypress/integration-run/breakpoints.js
```diff
@@ -0,0 +1,18 @@
+/* global cy */
+describe('JS layout', () => {
+  it('should support js layouts in cypress', () => {
+    cy.visit('https://applitools.github.io/demo/TestPages/JsLayout/');
+    cy.eyesOpen({
+      appName: 'JS layout',
+      testName: 'testing js layout support in cypress',
+      browser: [
```
amitzur, line 8: The generic test we have uses different browsers. Let's be consistent with it:
https://github.com/applitools/sdk.coverage.tests/blob/master/coverage-tests.js#L112

packages/eyes-cypress/src/browser/eyesCheckWindow.js
```diff
@@ -1,124 +1,148 @@
...
+          snapshots,
+          ...sendArgs,
+        },
+      });
+    });
+
+    function takeDomSnapshots(options) {
+      if (!breakpoints) {
+        //console.log('no breakpoints, taking single dom snapshot');
+        return takeDomSnapshot(options).then(snapshot => Array(browsers.length).fill(snapshot));
```
amitzur, line 31: I think it's better to still support passing a single snapshot over the wire to the Node.js side. If you have 10 browser configuration and no layout breakpoints, and a large snapshot, then it would be serialized 10 times, which is a shame.

packages/eyes-cypress/src/plugin/handlers.js
```diff
@@ -93,7 +106,7 @@ function makeHandlers({
 
     checkWindow: async ({
       url,
-      snapshot = {},
+      snapshots = [],
```
amitzur, line 109: I think it still makes sense to support getting a single snapshot. See my comment in eyesCheckWindow line 31

amitzur (commented):
Initial pass at the PR. I will continue to review today/tomorrow

packages/eyes-cypress/src/browser/commands.js
```diff
@@ -6,12 +6,18 @@ const processPage = require('@applitools/dom-snapshot/dist/processPageCjs');
 const domSnapshotOptions = {dontFetchResources: Cypress.config('eyesDisableBrowserFetching')};
 const send = makeSend(Cypress.config('eyesPort'), window.fetch);
 const makeSendRequest = require('./sendRequest');
-const makeEyesCheckWindow = require('./eyesCheckWindow');
+const makeEyesCheckWindow = require('./eyesCheckWindow.js');
```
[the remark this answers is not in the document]
tool3: 😱 sorry about that

packages/eyes-cypress/src/browser/commands.js
```diff
@@ -20,22 +26,19 @@ if (!Cypress.config('eyesIsDisabled')) {
...
 
   after(() => {
     cy.then({timeout: 86400000}, () => {
-      return batchEnd({timeout: Cypress.config('eyesTimeout')}).catch(e => {
-        if (
-          Cypress.config('eyesFailCypressOnDiff') ||
-          Cypress.config('eyesFailCypressOnDiff') === undefined
-        ) {
+      return batchEnd({timeout: getGlobalConfigProperty('eyesTimeout')}).catch(e => {
+        if (!!getGlobalConfigProperty('eyesFailCypressOnDiff')) {
```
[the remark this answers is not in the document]
tool3: yes. previously it was set to be `true` if user provided it or if it's `undefined`.
today I set `eyesFailCypressOnDiff` explicitly to `true` in `pluginExport` unless provided by user.
even if user sets `eyesCypressOnDiff: undefined` this clause will work.

packages/eyes-cypress/src/browser/commands.js
```diff
@@ -67,7 +71,28 @@ Cypress.Commands.add('eyesOpen', function(args = {}) {
...
+
+  const browser = eyesOpenArgs.browser || globalBrowser || defaultBrowser;
+  const layoutBreakpoints =
+    (args && args.layoutBreakpoints) ||
+    (eyesOpenArgs && eyesOpenArgs.layoutBreakpoints) ||
+    globalBreakpoints;
+  const checkArgs = {};
+
+  if (typeof args === 'object') {
+    Object.assign(checkArgs, {...args, eyesOpenArgs, browser, layoutBreakpoints});
```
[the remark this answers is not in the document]
tool3: `args` do take precedence over `eyesOpenArgs` - but only for layout breakpoints.
for `browser` - you cannot set `browser` in `check` as far as I understood.

packages/eyes-cypress/test/fixtures/testApp/cypress/integration-run/breakpoints.js
tool3: 👍🏼

packages/eyes-cypress/src/plugin/handlers.js
```diff
@@ -78,7 +86,12 @@ function makeHandlers({
       );
       return client;
     },
-
+    getIosDevicesSizes: async () => {
```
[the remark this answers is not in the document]
tool3: 👍🏼

packages/eyes-cypress/src/browser/commands.js
```diff
@@ -67,7 +71,28 @@ Cypress.Commands.add('eyesOpen', function(args = {}) {
 Cypress.Commands.add('eyesCheckWindow', args => {
   Cypress.log({name: 'Eyes: check window'});
   if (isCurrentTestDisabled) return;
-  return cy.document({log: false}).then({timeout: 60000}, doc => eyesCheckWindow(doc, args));
+
+  const defaultBrowser = {width: 800, height: 600, name: 'chrome'};
```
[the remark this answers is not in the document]
tool3: to be perfectly honest I wasn't sure what I should use.
in cypress we take the default width and height which is `1000x660`, but in every other sdk we use the default from vgc which is `800x600`.

packages/eyes-cypress/src/browser/eyesCheckWindow.js
tool3: this was something I thought about but preferred concentrating on the core concept. now that we have that down I agree that this optimization is required.

I wasn't sure about the prop name so I left it as plural: `snapshots` - even though it can be both an array and a single object.

packages/eyes-cypress/src/browser/commands.js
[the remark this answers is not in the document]
amitzur: But `args` are first, and then `eyesOpenArgs` overrides it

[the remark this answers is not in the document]
amitzur: ah, sorry. I see you fixed that

packages/eyes-cypress/src/browser/commands.js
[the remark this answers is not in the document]
amitzur: 1. vgc default is 1024x768, not 800x600.
2. in Cypress if `browser` is not specified by the user, we use the viewport size from Cypress' configuration. It's not necessarily 1000x660. This makes sense because then the visual test will be at the same viewport size of the local browser.

packages/eyes-cypress/src/browser/eyesCheckWindow.js
```diff
@@ -1,124 +1,148 @@
...
+              .viewport(Number(requiredWidth), height, {log: false})
+              .wait(300, {log: false})
+              .then(() => {
+                takeDomSnapshot(options).then(snapshot => {
+                  browsersInfo.forEach(({index}) => (snapshots[index] = snapshot));
+                });
+              });
+          }
+          return cypress
+            .viewport(width, height)
```
amitzur, line 65: missing `{log: false}`

packages/eyes-cypress/src/browser/eyesCheckWindow.js
```diff
@@ -1,124 +1,148 @@
...
+          snapshots,
+          ...sendArgs,
+        },
+      });
+    });
+
+    function takeDomSnapshots(options) {
+      if (!breakpoints) {
+        //console.log('no breakpoints, taking single dom snapshot');
+        return takeDomSnapshot(options).then(snapshot => snapshot);
```
amitzur, line 31: I think it's not needed to do `.then(snapshot => snapshot)`

packages/eyes-cypress/src/browser/commands.js
```diff
@@ -67,7 +71,33 @@ Cypress.Commands.add('eyesOpen', function(args = {}) {
...
+    layoutBreakpoints: getGlobalConfigProperty('eyesLayoutBreakpoints'),
+  };
+
+  const browser = eyesOpenArgs.browser || globalArgs.browser || defaultBrowser;
+  const layoutBreakpoints =
+    (args && args.layoutBreakpoints) ||
+    (eyesOpenArgs && eyesOpenArgs.layoutBreakpoints) ||
+    globalArgs.layoutBreakpoints;
+
+  const checkArgs = Object.assign(globalArgs, eyesOpenArgs, args);
```
amitzur, line 91: What will happen now is that because you removed the white list from `eyesCheckWindow`, and you mass assign all `eyesOpenArgs` onto the `checkArgs` - then all the arguments from open will be sent to the handler of `checkWindow`. I don't think that's a good thing.

I think maybe your initial implementation was closer to correct. You only need `browser` and `layoutBreakpoints` from open/global args. So I think you should just remove this line.

packages/eyes-cypress/test/it/browser/eyesCheckWindow.it.test.js
```diff
@@ -16,20 +26,23 @@ describe('eyesCheckWindow', () => {
...
     const eyesCheckWindow = makeEyesCheckWindow({
       sendRequest,
       processPage,
+      cypress,
     });
 
     const tag = 'some tag';
 
-    await eyesCheckWindow('bla doc', tag);
+    await eyesCheckWindow(fakeDoc, tag, cypress);
```
amitzur, line 38: If you used `doc.defaultView` only when it's needed, then you could still leave this test untouched.

packages/eyes-cypress/test/it/browser/eyesCheckWindow.it.test.js
```diff
@@ -5,6 +5,16 @@ const {expect} = require('chai');
 const makeEyesCheckWindow = require('../../../src/browser/eyesCheckWindow');
 
 describe('eyesCheckWindow', () => {
+  const fakeDoc = {defaultView: {innerWidth: 800, innerHeight: 600}};
+  const fakeCypressFunction = async data => {
+    Promise.resolve(data);
```
amitzur, line 10: What is this line for?

packages/eyes-cypress/test/fixtures/testApp/cypress/integration-run/breakpoints.js
```diff
@@ -0,0 +1,18 @@
+/* global cy */
```
amitzur, line 1: makes sense to call this file `layout-breakpoints.js`

amitzur (changes requested):
1. Config changes in pluginExports feel a bit risky and unrelated.

2. Tests for handlers shouldn't be changed to receive array of snapshots, since this scenario still happens and is actually the happy path. So in order to have tests for multiple snapshots you can add to the existing test suite. In general - **tests should not be removed or altered unless their use case is no longer relevant**. Especially if they are very fast unit tests.

3. saving `openEyesArgs` on the Config is essentially putting them in global scope. I think this would turn out ok just because Cypress handles each file in a separate process, and each file runs the tests serially. I think it would be better if we found a way to scope it, like having a `this` object from Cypress which has context of the current test. But then we need to account for open commands being inside before/beforeEach. So I think the global scope solution simplifies it a lot, and as I said I believe it should be fine. But we should be aware of it, since polluting this global scope might be dangerous. We also don't do any cleanup. I hope Cypress doesn't pass configuration defined in one file to another file.

4. No support for `layoutBreakpoints: true` (crash on exception in `.sort`). Maybe this requirement wasn't communicated clearly. Need to add e2e test to this (see (6) below). In core this is unit tested, but here in Cypress this feature is only tested e2e.

5. Passing args in `eyesCheckWindow` still isn't correct. See comment below.

6. e2e tests should have also:
  a. define `layoutBreakpoints` in global config
  b. define `layoutBreakpoints` in check args
  c. define `layoutBreakpoints` as `true` (in one of the 3 methods)

7. Readme. See what I did for Storybook.

packages/eyes-cypress/src/browser/eyesCheckWindow.js
```diff
@@ -1,124 +1,148 @@
...
+function makeEyesCheckWindow({sendRequest, processPage, domSnapshotOptions, cypress = cy}) {
+  return function eyesCheckWindow(doc, args = {}) {
+    const browser = args.browser;
+    const layoutBreakpoints = args.layoutBreakpoints;
+    const browsers = Array.isArray(browser) ? browser : [browser];
+    const {innerWidth: width, innerHeight: height} = doc.defaultView;
+    const breakpoints = layoutBreakpoints
+      ? layoutBreakpoints.sort((a, b) => (a > b ? -1 : 1))
+      : undefined;
+    const sendArgs = typeof args === 'string' ? {tag: args} : args;
```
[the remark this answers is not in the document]
tool3: removing this.
this creates a problem with a test that checks for string input.
essentially I am no longer accepting string input because I handle all of that in `commands`

packages/eyes-cypress/src/browser/eyesCheckWindow.js
```diff
@@ -1,124 +1,148 @@
...
-      sendDom = args.sendDom;
-      useDom = args.useDom;
-      enablePatterns = args.enablePatterns;
-      ignoreDisplacements = args.ignoreDisplacements;
-      accessibility = args.accessibility;
-      matchLevel = args.matchLevel;
-      visualGridOptions = args.visualGridOptions;
+function makeEyesCheckWindow({sendRequest, processPage, domSnapshotOptions, cypress = cy}) {
+  return function eyesCheckWindow(doc, args = {}) {
+    const browser = args.browser;
```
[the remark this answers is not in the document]
tool3: 👍🏼

packages/eyes-cypress/src/browser/eyesCheckWindow.js
```diff
@@ -1,124 +1,148 @@
...
-          method: 'PUT',
-          headers: {'Content-Type': 'application/octet-stream'},
-        }).catch(_e => {
-          snapshot.blobData.splice(
-            snapshot.blobData.findIndex(({url: blobUrl}) => blobUrl === url),
-            1,
-          );
-          snapshot.resourceUrls.push(url);
+    function getBrowserInfo(browser) {
+      const browserInfo = !!browser.name;
```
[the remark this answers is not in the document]
tool3: 👍🏼

packages/eyes-cypress/src/browser/eyesCheckWindow.js
```diff
@@ -1,124 +1,148 @@
...
+      if (browserInfo) {
+        const {name, width} = browser;
+        return Promise.resolve({name, width});
+      } else {
+        const iosDevice = !!browser.iosDeviceInfo;
+        const {deviceName, screenOrientation = 'portrait'} =
+          browser.iosDeviceInfo || browser.chromeEmulationInfo || browser;
+        const command = iosDevice ? 'getIosDevicesSizes' : 'getEmulatedDevicesSizes';
```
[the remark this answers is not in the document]
tool3: 👍🏼

packages/eyes-cypress/src/browser/eyesCheckWindow.js
tool3: 👍🏼

packages/eyes-cypress/src/browser/eyesCheckWindow.js
tool3: yes, remnants from battling cypress promises

packages/eyes-cypress/test/fixtures/testApp/cypress/integration-run/breakpoints.js
tool3: 👍🏼

packages/eyes-cypress/test/it/browser/eyesCheckWindow.it.test.js
tool3: I was trying to mimic cypress but I understand this is not needed. removing.

packages/eyes-cypress/test/it/browser/eyesCheckWindow.it.test.js
tool3: I have completely reverted the test and no longer using fake doc. 
however, the string input test was altered to `await eyesCheckWindow('bla doc', {tag});`  since `eyesCheckWindow` does not accept a string **directly** (handled in `commands.js`)

packages/eyes-cypress/test/unit/plugin/handlers.test.js
```diff
@@ -229,8 +228,8 @@ describe('handlers', () => {
       id3: {url: 'id3', type: 'type3', value: 'buff3'},
     };
 
-    const result = await handlers.checkWindow({snapshot: {blobData}});
-    expect(result.snapshot.resourceContents).to.eql(resourceContents);
+    const result = await handlers.checkWindow({snapshots: [{blobData}]});
```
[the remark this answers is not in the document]
tool3: 👍🏼

packages/eyes-cypress/src/browser/commands.js
tool3: I want to make sure this code is maintainable and readable.. but i'm afraid "maybe closer to correct" doesn't sound like it's there...
is there anything else I can do here that will make this code more robust and understandable ? 

this came from trying to follow your suggestion in a previous comment:
https://github.com/applitools/eyes.sdk.javascript1/pull/329#discussion_r549356268

currently, I have reverted this to my original implementation:
```javascript
const checkArgs = {};
```

tool3:
1. config in `pluginExport` - reverted.
2. all altered tests were reverted and only slightly tweaked to account for some new changes.
3. verified cypress is not passing configurations between tests  (https://docs.cypress.io/guides/references/configuration.html#Test-Configuration)
4. in progress.
5. fixed.
6. in progress.

packages/eyes-cypress/src/browser/eyesCheckWindow.js
```diff
@@ -5,30 +5,28 @@ const getAllBlobs = require('./getAllBlobs');
...
+          ...args,
         },
       });
     });
 
     function takeDomSnapshots(options) {
+      const browser = args.browser;
+      const breakpoints = args.layoutBreakpoints;
+      const browsers = Array.isArray(browser) ? browser : [browser];
+      const width = doc.defaultView && doc.defaultView.innerWidth;
```
amitzur, line 24: no need for this check. You want to require `doc` to have the `defaultView` property. But to avoid the problem in tests, use `doc` only when you need it, meaning in line 50.

amitzur (approved):
:100:

amitzur:
This looks great.
There's one small thing about when the user specifies a browser without a name. We currently detect that as device emulation or iOS. But this is wrong. Probably we should use the same logic as `isEmulation` from vgc. This is an issue that I spotted only now, in all SDK's not only in this PR.

packages/eyes-cypress/src/browser/commands.js
```diff
@@ -67,7 +71,32 @@ Cypress.Commands.add('eyesOpen', function(args = {}) {
...
+    layoutBreakpoints: getGlobalConfigProperty('eyesLayoutBreakpoints'),
+  };
+
+  const browser = eyesOpenArgs.browser || globalArgs.browser || defaultBrowser;
+  const layoutBreakpoints =
+    (args && args.layoutBreakpoints) ||
+    (eyesOpenArgs && eyesOpenArgs.layoutBreakpoints) ||
+    globalArgs.layoutBreakpoints;
+
+  const checkArgs = {layoutBreakpoints, browser};
```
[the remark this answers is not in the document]
tool3: that's one great shorthand but I feel like the intent is clearer using the more verbose structure.

packages/eyes-cypress/src/browser/commands.js
```diff
@@ -67,7 +71,32 @@ Cypress.Commands.add('eyesOpen', function(args = {}) {
...
+  if (typeof args === 'object') {
+    Object.assign(checkArgs, args);
+  } else {
+    Object.assign(checkArgs, {tag: args});
+  }
+  console.log(checkArgs);
```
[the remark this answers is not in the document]
tool3: 🙈

packages/eyes-cypress/src/browser/eyesCheckWindow.js
tool3: but that would mean `cannot read property innerWidth of undefined` when you pass in a string like we do in the tests right?  or am I missing something here...?

tool3:
> This looks great.
> There's one small thing about when the user specifies a browser without a name. We currently detect that as device emulation or iOS. But this is wrong. Probably we should use the same logic as `isEmulation` from vgc. This is an issue that I spotted only now, in all SDK's not only in this PR.

just pointing out here that in cypress I was able to resolve the default browser name in `eyesOpen`, by verifying that no browser name was provided, nor any ios or emulated deviceName were given -> use `chrome`.

PR #330, opened 2020-12-28
feat: adding fancy and snaptdout

tool3:
# snaptdout
`snaptdout` is a utility that helps test string output.
it does this by comparing strings line by line and saving the result to a `.json` file with the name of the test file.

this helps save much of the tedious and taxing work of building a string in a test in order to validate that the results is exactly the same.

### current limitations
- global config only: currently `snaptdout` accepts global config via `package.json` - this needs to change to accept all possible config as a second argument as well.

### features
- drop-in and start testing. 
- **should** support basically all testing frameworks (not tested).
- great diffs.

### example
```javascript
const output = myFunctionWhichReturnsAFormattedOutput(someArg);
await snaptdout(output, 'snapshot name');
```
example diff:
image

# fancy
`fancy` is a string formatting tool that will allow us to keep a consistent style across all of our sdks.

### features
- template based formatting for structural flexibility.
- DIY formatting - `fancy` accepts a formatting function which will receive the formatter as it's first argument (currently defaults to the cypress format output. see `errorDigest` in `eyes-cypress`) - the formatter object has many methods and colors to format the output.

### example
```javascript
const { output } = fancy({
      header: 'My Header',
      footer: 'My Footer',
      body: {
        passed,
        failed,
        unresolved: diffs,
      },
      indent: 4
});
```
output:
image


packages/fancy/index.js
```diff
@@ -0,0 +1,21 @@
+const formatter = require('./lib/formatter');
+
+function fancy({
+	header = '',
+	indent = 0,
+	body = {},
+	dull = false,
+	footer = '',
+	template = '',
+	formatterFunction = require('./lib/default/formatResults')
```
tourdedave, line 10: `formatterFunction` feels uncomfortable as part of this destructured object. It stands out as different than the rest of the params and the name is both too generic and too specific (it doesn't convey what it does and contains an implementation detail - which matters because the user needs to know this key to use it).

I think it would work better as its own argument. Then the separation is clear and the name becomes less critical.

e.g.,

```js
function formatter({ header, footer, body, indent, dull, template } = {}, formatterFunction) {
```

As for the name, something that's more crisp like `customReportFormatter` would be better about conveying the intent w/o leaking details about the implementation.

wdyt?

packages/fancy/index.js
```diff
@@ -0,0 +1,21 @@
...
+}) {
+
+	const formatFunction = formatter({ header, footer, body, formatterFunction, indent, dull, template });
+
+	return {
+		formatter: formatFunction,
```
tourdedave, line 16: What's the case where you need this?

packages/fancy/index.js
```diff
@@ -0,0 +1,21 @@
...
+		output: formatFunction.report()
```
tourdedave, line 17: Since `formatterFunction` is only used in `report`, you could just pass it there instead of passing it at the call-site for `formatter`. 

`formatFunction.report(formatterFunction)` - then following the bread crumbs would be a bit easier

packages/fancy/lib/formatter.js
```diff
@@ -0,0 +1,106 @@
...
+        print() {
+            console.log(this.report())
+        },
+        getTemplate() {
+            return defaultTemplate;
+        },
+        report() {
+            // call the provided format function
+            const templateObj = { ...templateObject, ...formatted };
+            formatterFunction(this, templateObj);
```
tourdedave, line 81: This will throw if it's not specified. I know that a default formatter is passed in if one is not provided, so I suspect this is an exceptional case that we would want to throw on, yeah?

tourdedave (changes requested):
Great work!

I'm still grokking everything so I'm sure there will be more feedback. But here's my first read of it. I offered some suggestions around the API design, and wanted to discuss some of the uses cases about the information we're returning to the user.

Other than that, there are a couple of points I didn't mention, but will note here:

1. Use of `_` for local variables

Stylistically I don't think we've come to a consensus on our preferred code style when it comes to the use of `_` across our packages. But I do know that we use it for linting to note that a variable is unused. So when I read your code, I kept thinking "why are there so many unused variables?". But on the second pass, it was clear that they were local variables and this was likely the shortest path to avoiding naming collisions with functions of the same name. I don't have anything concrete here, just wanted to surface my initial take when reading your code and to open up the discussion about this.

2. Mutable state

In the `report` function, when the `formatterFunction` is used, it mutates the state of what was passed in. Normally I try to stay away from side-effects, but maybe this mutation is by design to make it easier to implement a custom function? I'm not in love with it, but it is passable since the state is internal to this function. Worth a discussion though.

packages/fancy/README.md
```diff
@@ -0,0 +1,229 @@
...
+  });
+  expect(output).to.be(someOtherOutPut);
+  // or better yet
+  // await snap(output, 'charm')
+  // see: applitools/snaptdout
+});
+```
+
+outputs:  
+<img src="https://github.com/applitools/fancy/blob/master/example.png?raw=true" />
```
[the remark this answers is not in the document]
tool3: yeah i've readded the images and they will be resolved once this is merged back to master

packages/fancy/index.js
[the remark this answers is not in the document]
tool3: agreed. makes sense

tool3: I agree with you that, in this case, it makes more sense to put it as a second argument.

but in general, I think it's fine if you use `key`ed arguments but provide a good documentation and typings.

i'm not sure I follow your conflicts about the name... what implementation detail does it leak ? 
and also, how `customReportFormatter` addresses that point ?

I totally agree that it makes sense to put it as a second argument - so naming **isn't** a thing.

packages/fancy/index.js
tool3: yes ! great catch thank you !

packages/fancy/lib/formatter.js
tool3: so are you saying we **should** `trycatch` this ?

packages/fancy/index.js
tool3: I wanted to provide a way to use the formatting functions individually.
so you could:
```js
const {formatter} = require('@applitools/fancy')();
console.log(formatter.red('hello'))
```

packages/fancy/index.js
tourdedave: I was referring to "function" being in the name. If it's documented/typed that it is a function, then it doesn't need to be in the name.

packages/fancy/index.js
tourdedave: Makes sense.

packages/fancy/index.js
tool3: `customFormatter` ?

tourdedave: Perfect.

tool3:
once we complete the reviews, prior to merging, i'll add our release boilerplate to each of the packages in this PR (`bongo`, `CHANGELOG`)

PR #332, opened 2021-01-05
Redo image location calculation

kyryloonufriiev:
I focused more on the screenshoter since this should replace everything else in some point. It's funny how good this logic fits there.

### DIFF
--- a/packages/eyes-sdk-core/lib/sdk/EyesBase.js
+++ b/packages/eyes-sdk-core/lib/sdk/EyesBase.js
@@ -2130,18 +2130,6 @@ class EyesBase {
     this._logger.verbose('getting screenshot...')
     let screenshot, screenshotUrl, screenshotBuffer
 
-    // START NOTE (amit): the following is something I'm not proud nor confident of. I copied it from EyesSelenium::getScreenshot, and it is to solve https://trello.com/c/O5sTPAU1. This should be rewritten asap.
-    let isMobileDevice, originalPosition
-    const positionProvider = this.getPositionProvider()
-    if (this._driver) {
-      isMobileDevice = await this._driver.isNative
-
-      if (!isMobileDevice && positionProvider) {
-        originalPosition = await positionProvider.getState()
-      }
-    }
-    // END NOTE
-
     // Getting the screenshot (abstract function implemented by each SDK).
     screenshot = await this.getScreenshot()
     this._logger.verbose('Done getting screenshot!')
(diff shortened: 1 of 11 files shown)


packages/eyes-sdk-core/lib/sdk/EyesBase.js
```diff
@@ -2187,28 +2178,16 @@ class EyesBase {
     const title = await this.getTitle()
     let domUrl = await this.getDomUrl()
     const imageLocation = await this.getImageLocation()
+    console.log(imageLocation)
```
amitzur, line 2181: can remove this

packages/eyes-sdk-core/lib/sdk/EyesBase.js
```diff
@@ -2130,25 +2130,16 @@ class EyesBase {
...
-    // END NOTE
-
     // Getting the screenshot (abstract function implemented by each SDK).
     screenshot = await this.getScreenshot()
     this._logger.verbose('Done getting screenshot!')
 
     if (screenshot) {
       // Cropping by region if necessary
       if (!region.isSizeEmpty()) {
+        this._imageLocation = screenshot
```
amitzur, line 2140: this is a copy of the inside of `getSubScreenshot`, right?

packages/screenshoter/src/takeStitchedScreenshot.js
```diff
@@ -124,7 +124,7 @@ async function takeStitchedImage({
   }
 
   await saveScreenshot(composition, {path: debug.path, name: 'stitched', logger})
-  return composition
+  return {image: composition, region: cropRegion}
```
amitzur, line 127: shouldn't it be `{image, location}`. What does width/height add here?

packages/eyes-sdk-core/lib/sdk/EyesBase.js
kyryloonufriiev: That's right

packages/screenshoter/src/takeStitchedScreenshot.js
kyryloonufriiev: I want to return the whole effective region, we already have this information, so it doesn't hurt to return it.

kyryloonufriiev:
@amitzur that's correct after taking a screenshot we move the whole page back to its positions before screenshots were taken. And then we taking a dom snapshot.

PR #338, opened 2021-01-06
Try to test abort controller

amitzur:
I tried to test that `fetchResource` aborts the streaming request, but for some reason testServer isn't exiting when I use `streaming-middleware`. I couldn't get this to work properly.


kyryloonufriiev:
Closing this since it is already tested in `ufg-client`

PR #349, opened 2021-01-12
chore: prepush hook for lint and changelog

tool3:
# husky
adding husky to enable a prepush git hook that will `lint` and `verify-changelog` per push.

## excluded
- eyes-api
- eyes-common-legacy
- eyes-images-legacy
- eyes-leanft
- eyes-sdk-core-legacy
- eyes-universal
- eyes-universal-poc
- eyes-webdriverio-4-service


tourdedave (changes requested):
Based on the conversation we had in today's daily, we're dropping `yarn bongo verify-changelog` as part of pre-push (to be done in a GH action). Other than that, looks good.

PR #350, opened 2021-01-12
chore(logger): add log levels and log formatting

tool3:
this PR aims to open the discussion and bring about improvements to our current logs.

 a possible result will look like this in a `tty`:
image

# things to discuss
- log filtering - choose what module to log
- timing - seconds/milliseconds using the first line as reference
- structured logs - every line is an object
- consider this and the above as post process scripts (log -> pretty log)


tourdedave:
Definitely worth a discussion. IIRC we don't offer log levels at the moment (basically verbose or nothing), and there is a much bigger opportunity here to change the format/schema of our logging into something standardized that would open us up to a much more rich debugging/diagnostic setup when troubleshoot user issues.

kyryloonufriiev:
Wow, I didn't saw this PR @tool3 please have a look at my logger package in Eyes API PR. I actually also add coloring and log levels and all other goodies.

PR #351, opened 2021-01-12
feat(eyes-storybook): filter a single story

tool3:
# description
give the option to run a single eyes test against a given storybook url.
sometimes, for troubleshooting purposes, you want to run a test against a single story.

# usage
```bash
eyes-storybook --storybookUrl="https://some.storybookUrl.com" --storyId="story_id_from_url"
```


amitzur:
@tool3 this is possible with the `include` configuration parameter

PR #352, opened 2021-01-13
fix(takeDomSnapshot): fix CORS frames with same src attribute

tool3:
when we encounter cross origin frames with the same `src` attribute - the resulting image will have inappropriately positioned iframes and therefore will result in a bad image.

[Trello card](https://trello.com/c/qn5glrAD)

### DIFF
--- a/packages/eyes-sdk-core/lib/utils/takeDomSnapshot.js
+++ b/packages/eyes-sdk-core/lib/utils/takeDomSnapshot.js
@@ -10,6 +10,8 @@ const ArgumentGuard = require('./ArgumentGuard')
 const EyesUtils = require('../sdk/EyesUtils')
 const deserializeDomSnapshotResult = require('./deserializeDomSnapshotResult')
 const createFramesPaths = require('./createFramesPaths')
+const GeneralUtils = require('./GeneralUtils')
+const {URL} = require('url')
 
 const EXECUTION_TIMEOUT = 5 * 60 * 1000
 const POLL_TIMEOUT = 200
@@ -88,6 +90,10 @@ async function takeDomSnapshot(logger, driver, options = {}) {
         })
       if (frameContext) {
         const frameSnapshot = await takeContextDomSnapshot(frameContext)
+        const uniqueId = GeneralUtils.guid()
+        const url = new URL(frameSnapshot.url)
+        url.searchParams.append('applitools-iframe', uniqueId)
+        frameSnapshot.url = url.href
         parentSnapshot.frames.push(frameSnapshot)
         cdtNode.attributes.push({name: 'data-applitools-src', value: frameSnapshot.url})
       }
--- a/packages/eyes-sdk-core/lib/utils/takeDomSnapshot.js
+++ b/packages/eyes-sdk-core/lib/utils/takeDomSnapshot.js
@@ -90,9 +90,7 @@ async function takeDomSnapshot(logger, driver, options = {}) {
         })
       if (frameContext) {
         const frameSnapshot = await takeContextDomSnapshot(frameContext)
-        if (frameSnapshot.url) {
-          frameSnapshot.url = generateUniqueUrl(frameSnapshot.url, 'applitools-iframe')
-        }
+        frameSnapshot.url = generateUniqueUrl(frameSnapshot.url, 'applitools-iframe')
         parentSnapshot.frames.push(frameSnapshot)
         cdtNode.attributes.push({name: 'data-applitools-src', value: frameSnapshot.url})
       }
--- a/packages/eyes-sdk-core/lib/utils/takeDomSnapshot.js
+++ b/packages/eyes-sdk-core/lib/utils/takeDomSnapshot.js
@@ -90,10 +90,12 @@ async function takeDomSnapshot(logger, driver, options = {}) {
         })
       if (frameContext) {
         const frameSnapshot = await takeContextDomSnapshot(frameContext)
-        const uniqueId = GeneralUtils.guid()
-        const url = new URL(frameSnapshot.url)
-        url.searchParams.append('applitools-iframe', uniqueId)
-        frameSnapshot.url = url.href
+        if (frameSnapshot.url) {
+          const uniqueId = GeneralUtils.guid()
+          const url = new URL(frameSnapshot.url)
+          url.searchParams.append('applitools-iframe', uniqueId)
+          frameSnapshot.url = url.href
+        }
         parentSnapshot.frames.push(frameSnapshot)
         cdtNode.attributes.push({name: 'data-applitools-src', value: frameSnapshot.url})
       }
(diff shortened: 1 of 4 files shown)


tourdedave (changes requested):
This needs a test.

amitzur:
@tool3 can't this be done in dom-snapshot? why isn't this covered by the following: https://github.com/applitools/dom-scripts/blob/master/packages/dom-snapshot/src/browser/domNodesToCdt.js#L129 ?

Can you do some order and explain the bug? Because I know you already fixed this previously in the line of code I referenced.

tool3:
> @tool3 can't this be done in dom-snapshot? why isn't this covered by the following: https://github.com/applitools/dom-scripts/blob/master/packages/dom-snapshot/src/browser/domNodesToCdt.js#L129 ?
> 
> Can you do some order and explain the bug? Because I know you already fixed this previously in the line of code I referenced.

as we discussed, since this is handling cross origin frames - and historically we found out that we cannot rely on a CORS frame's src attribute - we do this in core.

packages/eyes-sdk-core/test/it/takeDomSnapshot.spec.js
```diff
@@ -341,3 +446,29 @@ function generateSnapshotObject(overrides) {
     ...overrides,
   }
 }
+// TODO move to sdk-shared
+function derandomize(subjects, name, prop, counter = 0) {
```
amitzur, line 450: These 2 functions confused me a bit. I see it as an over abstraction.
Although it's very generic, it's also doing a lot of implicit (or unexpected) things. Like checking the `value` property on the cdt node, en route to the attribute object. But that property is not part of the shape of the object.

For this sort of operation I prefer something like the following, which:
(A) leverages the API of receiving a snapshot and derandomizing it, since the nested frames have the same shape of the top level snapshot.
(B) has a granular function for doing the actual work, applying the [SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle):
(C) Doesn't have the array checks - I think it's safe to assume cdt and frames are arrays.

```js
function derandomizeUrls(snapshot) {
  snapshot.url = derandomizeValue(snapshot.url)
  for (const node of snapshot.cdt) {
    if (node.attributes) {
      for (const attr of node.attributes) {
        attr.value = derandomizeValue(attr.value)
      }
    }
  }

  for (const frame of snapshot.frames) {
    derandomizeUrls(frame)
  }
}

function derandomizeValue(value) {
  if (value && value.includes('?applitools-iframe')) {
    const randomNumber = value.match(/[^?=]*$/)[0]
    return value.replace(randomNumber, counter++)
  } else {
    return value
  }
}
```

tool3: the array checks were added only to accommodate for the fact that we sometimes pass a string instead of an array.

this is cleaner, but I had to make some modifications to make it work:

```js
function derandomizeUrls(snapshot) {
  let counter = 0
  for (const node of snapshot.cdt) {
    if (node.attributes) {
      for (const attr of node.attributes) {
        attr.value = derandomizeValue(attr.value, counter++)
      }
    }
  }
  // reset counter for frames
  counter = 0

  for (const frame of snapshot.frames) {
    frame.url = derandomizeValue(frame.url, counter++)
    derandomizeUrls(frame)
  }
}

function derandomizeValue(value, counter) {
  if (value && value.includes('?applitools-iframe')) {
    const randomNumber = value.match(/[^?=]*$/)[0]
    return value.replace(randomNumber, counter)
  } else {
    return value
  }
}
```
also note that I am resetting the counter before looping through the frames, just to keep the counter start at 0 for each category - for me this is more intuitive and expectable when writing tests

amitzur (approved):
This looks great. Another option is to change it as @kyryloonufriiev suggested - to accept a function that generates the unique ID.

tool3:
> This looks great. Another option is to change it as @kyryloonufriiev suggested - to accept a function that generates the unique ID.

provided the option to add a `generateUniqueUrl` function that overrides the default one.
### api
but of course you can also decide on behavior:
```js
generateUniqueUrl: ('https://cors.com', 'applitools-iframe') // => https://cors.com?applitools-iframe=123
```
### usage
```js
const snapshot = await takeDomSnapshot(logger, eyesDriver, {
    generateUniqueUrl: () => 'someString'
})
```

tool3:
> This looks great. Another option is to change it as @kyryloonufriiev suggested - to accept a function that generates the unique ID.

provided the option to add a `uniqueUrl` function that overrides the default one.
### api
```js
uniqueUrl('https://cors.com', 'applitools-iframe') // => https://cors.com?applitools-iframe=123
```
### usage
```js
const snapshot = await takeDomSnapshot(logger, eyesDriver, {uniqueUrl: () => 'someString'})
```

packages/eyes-sdk-core/lib/utils/takeDomSnapshot.js
```diff
@@ -28,6 +29,7 @@ async function takeDomSnapshot(logger, driver, options = {}) {
     skipResources,
     removeReverseProxyURLPrefixes = !!process.env
       .APPLITOOLS_SCRIPT_REMOVE_REVERSE_PROXY_URL_PREFIXES,
+    uniqueUrl,
```
amitzur, line 32: you are passing a function as an argument to dom snapshot? I don't understand

packages/eyes-sdk-core/test/unit/utils/GeneralUtils.spec.js
```diff
@@ -66,6 +67,30 @@ describe('GeneralUtils', () => {
     })
   })
 
+  describe('uniqueUrl()', () => {
+    const validGuid = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-5][0-9a-f]{3}-[089ab][0-9a-f]{3}-[0-9a-f]{12}$/i
```
amitzur, line 71: look for uses of `chai-uuid`

amitzur (commented):
WIP review

packages/eyes-sdk-core/lib/utils/takeDomSnapshot.js
tool3: yes this is what we decided in order to avoid derandomizing the urls.
we simply pass a different function in the tests.
or are you talking about something else ?

packages/eyes-sdk-core/test/unit/utils/GeneralUtils.spec.js
tool3: updated to use `chai-uuid`

packages/eyes-sdk-core/lib/utils/takeDomSnapshot.js
amitzur: oh, sorry. I confused this with `const arg` a few lines later. I thought you were sending this to the browser.

amitzur (changes requested):
I think the tests should be different. See this PR https://github.com/applitools/eyes.sdk.javascript1/pull/361

amitzur (approved):

PR #356, opened 2021-01-14
OCR extract text regions

kyryloonufriiev:


packages/eyes-sdk-core/lib/sdk/EyesSDK.js
```diff
@@ -51,6 +51,9 @@ function EyesSDK({name, version, spec, VisualGridClient}) {
       newCheckSettings(...args) {
         return new SDKCheckSettings(...args)
       },
+      newTextRegionSettings(...args) {
```
amitzur, line 54: This is not needed anymore, I think

amitzur (commented):
Looks good. Missing escaping for extractTextRegions

PR #359, opened 2021-01-18
chore(cypress): don't format output when running in interactive mode

tool3:
we have the ability to get a config property from `cypress` which indicates whether the run is interactive or not, meaning was the test run started with `cypress open` or `cypress run` respectively.

when running in interactive mode - we do not want to format the output with color codes - as it makes the output unreadable.
we could not rely on `process.stdout.isTTY` because it's always `true`.

### DIFF
--- a/packages/eyes-cypress/test/unit/plugin/handlers.test.js
+++ b/packages/eyes-cypress/test/unit/plugin/handlers.test.js
@@ -41,7 +41,7 @@ describe('handlers', () => {
   });
 
   it('handles "open"', async () => {
-    handlers.batchStart({isInteractive: true});
+    handlers.batchStart({});
     const {checkWindow} = await handlers.open({__test: 123, accessibilityValidation: 'bla'});
     const checkResult = await checkWindow();
     expect(checkResult.__test).to.equal('checkWindow_123');
@@ -50,7 +50,7 @@ describe('handlers', () => {
   });
 
   it('throws when calling "checkWindow" before "open"', async () => {
-    handlers.batchStart({isInteractive: true});
+    handlers.batchStart({});
     expect(
       await handlers.checkWindow({}).then(
         x => x,
@@ -70,7 +70,7 @@ describe('handlers', () => {
         openEyes: openEyesWithCloseRejection,
       }),
     });
-    handlers.batchStart({isInteractive: true});
+    handlers.batchStart({});
     expect(
       await handlers.checkWindow({}).then(
         x => x,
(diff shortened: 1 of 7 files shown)


amitzur (changes requested):
almost :100: - code looks great but missing tests for `waitForBatch` and `errorDigest`. And maybe also `handlers` but I'm not sure that's testable.

tool3:
as a side note, #306 and #330  would have really helped with the `errorDigest` tests

packages/eyes-cypress/test/unit/plugin/handlers.test.js
```diff
@@ -41,7 +41,7 @@ describe('handlers', () => {
   });
 
   it('handles "open"', async () => {
-    handlers.batchStart();
+    handlers.batchStart({isInteractive: true});
```
amitzur, line 44: what are you testing? Or is this now required to avoid an exception?

amitzur, line 44: what are you testing? Or is this now required to avoid an exception?

amitzur (approved):
:100:
Only one question - can you explain the changes to `handlers` tests?

packages/eyes-cypress/test/unit/plugin/handlers.test.js
tool3: yeah it's to avoid the exception, I didn't want to set a default value because we expect this value to always exist in Cypress' config

amitzur: So 2 things:
1. You can see that in the lines before `GeneralUtils.getPropertyByPath` is used. So you can do the same. But better than that:
2. Do `handlers.batchStart({})` (pass empty object) in the test. Since you're not testing `isInteractive` anywhere it will be less confusing, and also serve the purpose of not throwing an exception.

Because right now, it's confusing to the reader of the test why specifically this property needs to be passed.

tool3: absolutely agree. updated to use `{}`.
but I am not sure what you mean by `GeneralUtils.getPropertyByPath` ? where is this used? and what are you suggesting that i'll replace with it ?

tool3: oh are you referring to the way I access `data.isInteractive` in `handlers.js:81` ?

tool3: fixed to use `GeneralUtils.getPropertyByPath`

PR #360, opened 2021-01-18
chore(cypress): add device error handling

tool3:
adding a clause that will check if a given device name exists in the list of devices we got from the server.

there is room for confusion when it comes to device configuration, therefore I added an indicative error message that points to the device list (similar to the readme) and details whether it's an emulated device or iOS.

image

### DIFF
--- a/packages/eyes-cypress/src/browser/eyesCheckWindow.js
+++ b/packages/eyes-cypress/src/browser/eyesCheckWindow.js
@@ -100,6 +100,14 @@ function makeEyesCheckWindow({sendRequest, processPage, domSnapshotOptions, cypr
           browser.iosDeviceInfo || browser.chromeEmulationInfo || browser;
         const command = browser.iosDeviceInfo ? 'getIosDevicesSizes' : 'getEmulatedDevicesSizes';
         return sendRequest({command}).then(devicesSizes => {
+          if (!devicesSizes.hasOwnProperty(deviceName)) {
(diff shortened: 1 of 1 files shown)


packages/eyes-cypress/src/browser/eyesCheckWindow.js
```diff
@@ -100,6 +100,14 @@ function makeEyesCheckWindow({sendRequest, processPage, domSnapshotOptions, cypr
           browser.iosDeviceInfo || browser.chromeEmulationInfo || browser;
         const command = browser.iosDeviceInfo ? 'getIosDevicesSizes' : 'getEmulatedDevicesSizes';
         return sendRequest({command}).then(devicesSizes => {
+          if (!devicesSizes.hasOwnProperty(deviceName)) {
+            const category = browser.iosDeviceInfo ? 'iOS' : 'emulated';
+            const deviceListUrl =
+              'https://github.com/applitools/eyes.sdk.javascript1/blob/master/packages/eyes-sdk-core/lib/config/DeviceName.js';
```
amitzur, line 106: there is a different enum for iOS: `IosDeviceName.js`

amitzur (changes requested):
almost :100: - see comment below

packages/eyes-cypress/src/browser/eyesCheckWindow.js
tool3: my bad !

amitzur (approved):

PR #361, opened 2021-01-19
Clearer tests for takeDomSnapshot

amitzur:
This change intends to make the tests clearer and more correct in the following aspects:

1. Correct usage of `uniqueUrl` parameter, which actually tests that the arguments are sent correctly, and that the output is used correctly. It is important to return different values with `counter` when needed, and in other cases just return some manipulation of the input arguments, `url` and `query`.
2. Marking frame nodes in the mocked CDT with attributes like `cross-origin-iframe` so that when looking at the expected output it's clear how to connect the mock with the result.


tool3 (approved):
This makes perfect sense, I like the verbosity and how explicit the expectation is.

PR #365, opened 2021-01-26
Fix: Integrate dom-capture with screenshoter 

kyryloonufriiev:


kyryloonufriiev:
Publishing order:
1. `@applitools/snippets`
2. `@applitools/screenshoter`
3. `@applitools/eyes-sdk-core`

amitzur (approved):
1. I verified that `ShouldExtractTextFromRegions` coverage test works now.
2. I noticed the diff in frame-fully.png with the frame's border - very nice.
3. `getElementContentSize` - the change was needed because if it's translated then scrollbars are influencing calculation?
4. `dom` parameter is not mandatory. `takeDomCapture` could act both as the operation and as a flag. If it's not passed, then no dom.
5. Need to add tests for dom in screenshoter.
6. Tests fail locally for me for `screenshoter`. I will nonetheless merge and release it and if needed I will skip the tests and you will fix them later. Might be Mac/Linux differences.

amitzur:
@kyryloonufriiev FYI -
https://github.com/applitools/eyes.sdk.javascript1/runs/1772570785?check_suite_focus=true
https://github.com/applitools/eyes.sdk.javascript1/commit/cc12ac3c8b98b6a3a6e15ffe5870d187bae28aa6

kyryloonufriiev:
3. Not exactly. Just if your `html` size is not viewport related (lets say 300x300) and you will do `transform: translate(-Xpx, -Ypx)`, then `scrollWidth` will be 300 - X and `scrollHeight` will be 300 - Y
4. I see `takeDomCapture` as a temporary solution. After refactoring, we will not pass this function and everything will happen in the package itself. So I implemented desired structure from the beginning. And after we will just remove this function from arguments.
6. I run it against chrome 87 in a docker container. So it is probably not Mac/Linux problem.

PR #366, opened 2021-01-26
fix documentation about Applitools configuration

silversonicaxel:
added the configuration of the Server url in the environment variables for setting up Applitools in your project

### DIFF
--- a/packages/eyes-cypress/CHANGELOG.md
+++ b/packages/eyes-cypress/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


amitzur:
Thanks @silversonicaxel!
I think for onboarding experience there should be the simplest steps possible. I would add a dedicated section about server URL after the API key section, instead of doing them together.
We'll take it from here, and I'll post the result here to get your feedback.

amitzur:
@silversonicaxel https://github.com/applitools/eyes.sdk.javascript1/pull/369

silversonicaxel:
Thanks @amitzur !

PR #367, opened 2021-01-26
Fix: Firefox screenshots

kyryloonufriiev:
Providing a new way to overcome firefox weird behavior in versions [48 - 72]


amitzur:
@kyryloonufriiev this looks like a different solution than what was implemented. Curious to know more about the nature of the change. And a generic coverage test makes sense here.

amitzur (approved):

PR #368, opened 2021-01-27
feat(eyes-sdk-core): don't upload screenshot to server if it matches the previous one

tool3:
[trello](https://trello.com/c/VAsOZIWA)
this PR introduces an optimization that will create a `SHA256` of the current image, save it for reference, and avoid uploading the same image to the server if it matches the sha.


tool3:
just to make sure we understand this feature properly, I will illustrated the way things were, and the way they are now.
this is using classic mode only:

### legacy
1. we check window and send a screenshot to the server.
2. server replies with mismatch.
3. we take a new screenshot.
4. upload screenshot to the server again -> we use `ignoreMismatch` to instruct the server to keep trying to compare the screenshots.
5. as long as we are in the limits of the provided `retryTimeout` -> we will keep doing 3 and 4 until timeout is reached.
6. the final screenshot at the server is unchanged.

### present
1. we check window and send a screenshot to the server.
2. server replies with mismatch.
3. we take a new screenshot.
4. compare screenshot's `sha256` representation to the previous screenshot we took.
5. - shas are identical -> we don't upload the screenshot to the server.
    - shas are different -> upload to the server again with `replaceLast: true`.
7. as long as we are in the limits of the provided `retryTimeout` AND shas are identical-> we will keep doing 3 through 5 until timeout is reached.
8. the final screenshot at the server is dependent of the result of comparing the shas. if the shas at any point were different -> the server will have the latest screenshot we took, otherwise it will remain unchanged.

kyryloonufriiev:
I'm closing this PR because it is too outdated. Here is a Trello card to cover this functionality at some point - https://trello.com/c/1eHRjGKw/50-do-not-upload-retried-screenshot-to-the-server-if-it-mutches-the-original-one

PR #372, opened 2021-02-01
feat: custom batch properties

tool3:
this PR introduces support for sending custom batch properties.

[trello](https://trello.com/c/IKTydXLv)
[coverage tests PR](https://github.com/applitools/sdk.coverage.tests/pull/50)


packages/eyes-sdk-core/test/unit/config/BatchInfo.spec.js
```diff
@@ -55,6 +55,19 @@ describe('BatchInfo', () => {
...
+  it('supports custom properties', () => {
+    const properties = [{name: 'custom_prop', value: 'custom value'}]
+    const batch = new BatchInfo({
+      id: 'id',
+      name: 'hello',
+      properties,
+      sequenceName: 'bla',
+    })
+    const anotherBatch = new BatchInfo().setProperties(properties)
+    assert.strictEqual(batch.getProperties(), properties)
```
amitzur, line 67: how come you don't need `deepStrictEqual` here?

packages/eyes-selenium/CHANGELOG.md
```diff
@@ -1,7 +1,11 @@
 # Change Log
 
 ## Unreleased
-
+- adding support for custom batch properties
+- updated to @applitools/eyes-sdk-core@12.14.6 (from 12.14.1)
```
amitzur, line 5: remove the following 4 lines

packages/eyes-selenium/package.json
```diff
@@ -36,14 +36,14 @@
   },
   "main": "./index.js",
   "dependencies": {
-    "@applitools/eyes-sdk-core": "12.14.1",
-    "@applitools/visual-grid-client": "15.5.13"
+    "@applitools/eyes-sdk-core": "12.14.7",
```
amitzur, line 39: these changes should not be part of the PR

packages/sdk-shared/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- add support for custom batch properties
```
amitzur, line 5: Not the correct text.
`add support for getting batch info`

packages/sdk-shared/src/get-batch-info.js
```diff
@@ -0,0 +1,16 @@
+const fetch = require('node-fetch')
+const {URL} = require('url');
+
+async function getBatchInfo(
+    testResults,
+    apiKey = process.env.APPLITOOLS_API_KEY_SDK || process.env.APPLITOOLS_API_KEY,
+  ) {
+    const sessionUrl = new URL(testResults.getAppUrls().getSession())
+    const accountId = sessionUrl.searchParams.get('accountId')
+    const url = `${sessionUrl.origin}/api/sessions/batches/${testResults.getBatchId()}/batch?accountId=${accountId}&apiKey=${apiKey}`
```
amitzur, line 10: I think `accountId` is not needed

packages/sdk-shared/src/test-setup.js
```diff
@@ -295,9 +295,11 @@ function Env(
 }
 
 const batchName = process.env.APPLITOOLS_BATCH_NAME || 'JS Coverage Tests'
-const batch = typeof BatchInfo === 'undefined' ? batchName : new BatchInfo(batchName)
+const customBatch = typeof BatchInfo === 'undefined';
```
amitzur, line 298: `isCustomBatch`. But that's not even the meaning of it. I think this check is because of a bug that existed ages ago. Anyhow, it's not a "custom batch". If any, it's `isBatchInfoDefined`

packages/sdk-shared/src/test-setup.js
```diff
@@ -295,9 +295,11 @@ function Env(
...
+const batch =  customBatch ? batchName : new BatchInfo(batchName)
 
-function getEyes({vg, showLogs, saveLogs, saveDebugScreenshots, runner, ...config} = {}) {
+function getEyes({vg, showLogs, saveLogs, saveDebugScreenshots, runner, batch: batchInfo, ...config} = {}) {
+  if (customBatch && batchInfo) batch.setProperties(batchInfo.properties);
```
amitzur, line 302: This feels a little dirty. What did you want to support?

amitzur (commented):
Looks great

packages/eyes-sdk-core/test/unit/config/BatchInfo.spec.js
tool3: I do need it. updated to `deepStrictEqual`

packages/eyes-selenium/CHANGELOG.md
tool3: removed

packages/eyes-selenium/package.json
tool3: reverted

packages/sdk-shared/CHANGELOG.md
tool3: done

packages/sdk-shared/src/get-batch-info.js
tool3: I used `getTestInfo` as reference, which uses it. why wouldn't it be needed?

packages/sdk-shared/src/test-setup.js
tool3: I agree on the name, but I feel like `isSomething` and `something` are stylistic choices.
I updated to `isBatchInfoDefined` but I think `batchInfoDefined` would also work just as well.

packages/sdk-shared/src/test-setup.js
tool3: refactored

packages/sdk-shared/src/get-batch-info.js
amitzur: Well, maybe it does. I think the link I sent you doesn't have accountId and still works. Doesn't matter, we can leave it here.

packages/sdk-shared/src/test-setup.js
```diff
@@ -294,15 +294,14 @@ function Env(
   return env
 }
 
-const batchName = process.env.APPLITOOLS_BATCH_NAME || 'JS Coverage Tests'
-const batch = typeof BatchInfo === 'undefined' ? batchName : new BatchInfo(batchName)
-
-function getEyes({vg, showLogs, saveLogs, saveDebugScreenshots, runner, ...config} = {}) {
+function getEyes({vg, showLogs, saveLogs, saveDebugScreenshots, runner, batch, ...config} = {}) {
+  const batchName = process.env.APPLITOOLS_BATCH_NAME || 'JS Coverage Tests'
+  const batchInfo =  batch ? new BatchInfo({ name: batchName, ...batch }) : batchName;
```
amitzur, line 299: The point of specifying the batch outside of `getEyes` is so that all tests are batched together. We could have also just defined the batch ID outside with a uuid. If that makes things simpler for you, then you can change it.
I think the export of `batch` is used, but I don't remember for what and if it's still true.

amitzur (changes requested):
just the batch in test-setup needs to be reverted. Other than that it's all :100:

packages/sdk-shared/src/test-setup.js
tool3: updated to a combination of global batch and custom props within `getEyes`

amitzur (approved):

PR #375, opened 2021-02-10
fix(eyes-storybook): use testConcurrency instead of concurrency

tool3:
[trello](https://trello.com/c/JHAjBl2q)

the term `concurrency` was used across `eyes-storybook`, but I noticed that we updated it recently (with the latest concurrency model changes) to `testConcurrency`.

unfortunately not all occurrences were updated.

I made this check explicit by checking that the current `testConcurrency` value is **exactly** the same as the default config's `testConcurrency`.


tourdedave (approved):

amitzur:
@tool3 weird that no test fails because of this change. We don't have any tests for this message?

tool3:
> @tool3 weird that no test fails because of this change. We don't have any tests for this message?

we have e2es that verifies the entire output, but nothing specific to concurrency.

PR #377, opened 2021-02-11
fix(eyes-cypress): validate browser/device configuration

tool3:
[trello](https://trello.com/c/onmEQzNx)

`eyesCheckWindow` was checking for the `name` property of a provided browser configuration.

unfortunately, we (I) also added a default browser name, that explicitly adds the `name` property to any browser configuration.

### DIFF
--- a/packages/eyes-cypress/src/browser/eyesCheckWindow.js
+++ b/packages/eyes-cypress/src/browser/eyesCheckWindow.js
@@ -2,6 +2,7 @@
 'use strict';
 
 const getAllBlobs = require('./getAllBlobs');
+const getBrowserInfo = require('./getBrowserInfo');
 
 function makeEyesCheckWindow({sendRequest, processPage, domSnapshotOptions, cypress = cy}) {
   return function eyesCheckWindow(doc, args = {}) {
@@ -30,7 +31,7 @@ function makeEyesCheckWindow({sendRequest, processPage, domSnapshotOptions, cypr
       return browsers
         .reduce((widths, browser, index) => {
           return widths.then(widthsMap => {
-            return getBrowserInfo(browser).then(({name, width}) => {
+            return getBrowserInfo(browser, sendRequest).then(({name, width}) => {
               const requiredWidth = getBreakpointWidth(breakpoints, width);
               let groupedBrowsers = widthsMap[requiredWidth];
               if (!groupedBrowsers) {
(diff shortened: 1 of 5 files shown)


packages/eyes-cypress/src/browser/eyesCheckWindow.js
```diff
@@ -92,7 +92,7 @@ function makeEyesCheckWindow({sendRequest, processPage, domSnapshotOptions, cypr
     }
 
     function getBrowserInfo(browser) {
-      if (browser.name) {
+      if (!browser.deviceName) {
```
tourdedave, line 95: If we follow what's listed in [the VG protocol](https://github.com/applitools/mono/blob/master/docs/visual-grid-protocol.md#rendering-api-app-api), it would be possible to hit this code path with a browser object for an iOS simulator. `deviceName` works but it's not what is spec'd.

tourdedave: Also, doing gating on positive conditions would be easier to reason about and likely lead to fewer bugs since you'd be whitelisting instead of blacklisting.

e.g.,

```
if (browser.deviceName) {
  // do the mobile stuff
} else {
  // do the browser stuff
}

tourdedave (changes requested):
See my comments on the conditional branch.

Also:
- Can you provide a reference to the default name setting that you mention in the PR description?

- Does this PR resolve the issue listed in the ticket? If so, is there already a test reference in the SDK's test set that demonstrates this? If not, we should add one.

packages/eyes-cypress/src/browser/eyesCheckWindow.js
tool3: since we overload any browser object with a default `name` (if it doesn't already have it) than we created a situation where we would treat devices as browsers... 

this negation makes sense because you can't be sure by only checking the name.
I am in favor of non-negated code when possible, but in this case it makes things a little more readable.

tool3: and sure! i'll add a test !

tool3: okay so following your comment and our conversation about it I made the following changes:
- extracted the function and its dependencies to a new module
- reversed if for positive condition
- extended the condition to include all known permutations
- added an integration test that checks all of this

tourdedave: Does this PR address the issue referenced in the Trello card? Put another way - do we have a breakpointLayout test that surfaces the issue and does this PR fix it?

tool3: not sure I understand the question...
the problem was the customer's usage of `name` inside emulated configuration objects - which we state that we do support (it's even in the examples).
 
this PR fixes the issue and enables us to support all said permutations.

I also added a test that verifies this.

tourdedave:
It was unclear from reading the Trello ticket that that was the problem. Thanks for clarifying. It might be helpful to include that kind of detail in future Trello tickets (I read both the Trello ticket and the PR when reviewing).

For context into why I asked... the implicit questions I bring to PRs that address a Trello ticket are:
- does it fix the issue that the customer reported? and,
- has it been verified at their level?

We often find and fix things at a lower level. Sometimes we don't verify them at a higher level (e.g., e2e) which can hide the fact that we didn't fully fix their issue, which can lead to more back and forth. It sounds like you're thinking of things in a similar way and did everything you needed to in this PR - it just wasn't clear to me at first glance.

tourdedave (approved):

PR #378, opened 2021-02-11
sb: fix config validator so it doesn't set the batch name automatically

tourdedave:


packages/eyes-storybook/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- fix for the cli runner so it doesn't auto-populate the batch name when not specified ([Trello](https://trello.com/c/MGGJ3Rh8))
```
amitzur, line 5: Can we rephrase to "don't auto-populate the batch name when not specified ([Trello](https://trello.com/c/MGGJ3Rh8))" ?

amitzur (approved):
great catch, @tourdedave !

packages/eyes-storybook/CHANGELOG.md
tourdedave: Sure

PR #381, opened 2021-02-12
SIDE Eyes: add iOS simulator support

tourdedave:
screenshot


packages/side-eyes/src/app/components/VisualGrid/options.js
```diff
@@ -71,6 +71,38 @@ export const DeviceName = {
...
+  iPad_Pro_3: 'iPad Pro (12.9-inch) (3rd generation)',
+  iPad_7: 'iPad (7th generation)',
+  iPad_Air_2: 'iPad Air (2nd generation)',
+  iPhone_12_Pro_Max: 'iPhone 12 Pro Max',
+  iPhone_12_Pro: 'iPhone 12 Pro',
+  iPhone_12: 'iPhone 12',
+  iPhone_12_mini: 'iPhone 12 mini',
+}
+
+function makeDeviceList() {
```
tool3, line 92: I think we can make this function more readable, and a bit more performant. but I think the question is why are we transforming to an array here ?

If I understand correctly, the only place you need it as an array is in the react app. 

but for example in, `parsers.js`:
```js
devices.forEach(device => {
    orientations.forEach(orientation => {
      orientation = orientation.toLowerCase()
      if (device.type === 'emulator') {
        matrix.push({
          screenOrientation: orientation,
          deviceName: device.name,
          deviceId: DeviceList.find(entry => entry.name === device.name).id,      // <--- here
        })
...
```
you iterate over this array for every device and for every device's orientation, but you can achieve a much faster lookup if you keep it as an object.

one thought I had was:

```js
const DeviceList = { emulator: DeviceName, simulator: IosDeviceNames }
const emulators = Object.keys(DeviceList.emulators).map(key => ({ name: DeviceList.emulators[key], type: 'emulator', id: key }))
const simulators = Object.keys(DeviceList.simulators).map(key => ({ name: DeviceList.simulators[key], type: 'simulator', id: key }))
```
or if you really want to be dry:
```js
const combined = Object.keys(DeviceList).reduce((acc, type) => {
   const category = DeviceList[type];
    Object.keys(category).forEach(key => acc.push({ name: category[key], type, id: key }))
   return acc;
 }, [])
```

than it makes lookup O(1):
```js
devices.forEach(device => {
    orientations.forEach(orientation => {
      orientation = orientation.toLowerCase()
      if (device.type === 'emulator') {
        matrix.push({
          screenOrientation: orientation,
          deviceName: device.name,
          deviceId: DeviceList[device.type][device.name].id,      // <--- here
        })
```

regardless of the time complexity (which isn't really meaningful in small datasets), I think keeping with objects as long as possible is ultimately beneficial, and only transform it to an array when needed.

packages/side-eyes/src/app/components/VisualGrid/options.js
```diff
@@ -71,6 +71,38 @@ export const DeviceName = {
...
+  const emulators = Object.values(DeviceName).map(entry => {
+    const id = Object.keys(DeviceName).find(key => DeviceName[key] === entry)
+    return { name: entry, type: 'emulator', id }
+  })
+  const simulators = Object.values(IosDeviceNames).map(entry => {
+    const id = Object.keys(DeviceName).find(key => DeviceName[key] === entry)
```
tool3, line 98: note that you are running on `IosDeviceNames` but still trying to get the ID from `DeviceName` so it ends up as `undefined`.

packages/side-eyes/src/app/components/VisualGrid/options.js
tourdedave: Thanks!

The primitive data type used in the view state is an array (you're right this is the only place it's needed - but it's a big piece). So the lifecycle of the data is 1) device enum copy/pasted out of core & stripped down to a simple collection, 2) shaped into an array of objects (for use in the view state), 3) user selected options from the view are transformed into an object structure for use in visual-grid-client when running the tests.

The parser.js function can be improved, as you mentioned. I'll look at that.

I was thinking that a more influential change might be making it so we can just import the device enums out of core instead of copying them. Thoughts?

packages/side-eyes/src/app/components/VisualGrid/options.js
tourdedave: Nice catch!

packages/side-eyes/src/app/components/VisualGrid/options.js
tourdedave: We can actually just use the `devices` array that is passed into the `parseBrowsers` function, instead of requiring it externally and performing a find on it. The lookup is fairly trivial and easy to reason about (even if it is "expensive", since it's such a small dataset). That, and the end result of DeviceList *needs* to be an array in order to be used in the UI and ultimately `parseBrowsers` - since the inputs to this function are coming from the user's specified settings (which are a series of arrays). Also, the need to look up an id is actually just needed for device emulation (not iOS sims). So there's not a lot of value in optimizing this. Thanks for the recommendation though!

The only other open thoughts I have for this PR are:

1. optimizing eyes-sdk-core so we can require specific parts of it in the browser (e.g., the device enums)
2. a limitation in the the UX - if you select options that appear in both Chrome Emulation and iOS Simulators (e.g., `iPhone 11 Pro`), you won't know which is which in the selection options view. You'll only know if you go back into the options picker and look at which check boxes are selected.

tourdedave:
Fishing this out of a comment thread so it's top-level.

There is a limitation in the the UX - if a user selects options that appear in both Chrome Emulation and iOS Simulators (e.g., iPhone 11 Pro), they won't know which is which in the selected options view. They'll only know if they go back into the options picker and look at which check boxes are selected.

Not sure how common this configuration would be for users - but didn't feel it was my call to make. Passed it onto product for feedback.

PR #382, opened 2021-02-18
core: fix retry in EyesUtils.setViewportSize

tourdedave:
- add unit test coverage for EyesUtils.setViewportSize
- improve logging in it, and
- fix bug in its retry logic that prevented the full number of retries and
  incorrect error message from bubbling when the window/viewport is unable to
  be resized
  
  re: [Trello](https://trello.com/c/xNCZNfPi)

### DIFF
--- a/packages/eyes-sdk-core/lib/sdk/EyesUtils.js
+++ b/packages/eyes-sdk-core/lib/sdk/EyesUtils.js
@@ -23,7 +23,6 @@ const EyesDriverOperationError = require('../errors/EyesDriverOperationError')
  * @return {RectangleSize} viewport size
  */
 async function getViewportSize(logger, context) {
-  logger.verbose('EyesUtils.getViewportSize()')
   let size
   if (!context.driver.isNative) {
     size = await context.execute(snippets.getViewportSize)
@@ -37,7 +36,6 @@ async function getViewportSize(logger, context) {
       }
     }
   }
-  logger.verbose('Done! Viewport size: ', size)
   return new RectangleSize(size)
 }
 /**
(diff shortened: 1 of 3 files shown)


packages/eyes-sdk-core/lib/sdk/EyesUtils.js
```diff
@@ -50,47 +48,67 @@ async function setViewportSize(logger, context, requiredViewportSize) {
...
+    `EyesUtils.setViewportSize(): actualWindowSize: ${JSON.stringify(
+      actualWindowSize,
+    )}, actualViewportSize: ${JSON.stringify(actualViewportSize)}`,
+  )
 
   const sleep = 3000
   let retries = 3
 
-  while (retries >= 0) {
+  while (retries >= 1) {
```
kyryloonufriiev, line 84: What does it solve? It looks like it decreases a possible number of retries.

tourdedave: It looked like the intention was to do 3 retries, not 4.

tourdedave: I updated the test to make this more explicit (asserting the number of retries)

amitzur:
@tourdedave from what I can see the changes here are only to logs and the decreasing the number of retries by 1. How does it solve the issue for the customer, and how does it cause the correct error message to be thrown?

packages/eyes-sdk-core/lib/sdk/EyesUtils.js
```diff
@@ -50,47 +48,67 @@ async function setViewportSize(logger, context, requiredViewportSize) {
...
     await GeneralUtils.sleep(sleep)
     actualViewportSize = await getViewportSize(logger, context)
     if (requiredViewportSize.equals(actualViewportSize)) return true
     logger.verbose(
-      `Attempt to set viewport size to ${requiredViewportSize} failed. actualViewportSize=${actualViewportSize}`,
+      `EyesUtils.setViewportSize(): Attempt to set viewport size to ${JSON.stringify(
+        requiredViewportSize,
+      )} failed. actualViewportSize=${JSON.stringify(actualViewportSize)}`,
     )
-    actualWindowSize = requiredWindowSize
```
tourdedave, line 88: @amitzur this was breaking retries

amitzur: ahhh so carefully hidden :flushed: 
i can't wrap my head around it to understand, but happy to know we found and fixed a bug here

tourdedave:
> @tourdedave from what I can see the changes here are only to logs and the decreasing the number of retries by 1. How does it solve the issue for the customer, and how does it cause the correct error message to be thrown?

It doesn't solve the issue for the customer (they still can't resize their browser because of something in their environment) - this PR fixes retries and makes it more clear what the issue is because it now returns the correct error message.

Before this PR, if the driver was unable to resize the window, a second attempt would be made and then error with: `Error: IllegalArgument: width < 0` (coming from the argument guard in `RectangleSize`) because the retry was mutating its inputs.

With this PR, all retries are attempted now, and if unsuccessful the correct error message will be reached: `Failed to set viewport size!`.

PR #384, opened 2021-02-19
feat(eyes-cypress): global hooks

tool3:
# cypress global hooks 
this PR introduces support for cypress' new global hooks `before:run` and `after:run`
- [cypress github issue](https://github.com/cypress-io/cypress/issues/6665)
- [trello](https://trello.com/c/2tluhwWN/748-cypress-issue-with-batches-and-github-integration-vonage)

## constraints
- `cypress@6.2.0` or higher
- `"experimentalRunEvents": true` in `cypress.json`
- **non**-interactive mode only

## terms
- **legacy hooks** - `before` and `after` hooks - they run before and after all **tests in a single file**, which is executed by cypress in its own process.
- **global hooks** - `before:run` and `after:run` - runs before and after **all tests run**.

### today
in the current architecture, we use `before` and `after` legacy hooks **in browserland** (`commands.js`) to open and close the batch.

cypress refactor (6)

in this structure `makeVisualGridClient` is being passed and called inside `handlers.js` and all of the methods from it are assigned to global variables for later use.

### tomorrow
the proposed design utilizes the new global hooks and enables us to perform open and close batch while still **in nodeland**.

after

in this structure, `makeVisualGrid` is initiated at the entry point `startPlugin.js` and passed down to `handlers.js`.
this way we can use visual grid client outside of `handlers.js` and still retain the same behavior.

## support
this feature supports all variations of the new cypress hooks and their constraints, with the opt-in approach: 

| cypress | @applitools/eyes-cypress | flag | file | hooks |
| -------- | -------- | -------- | -------- | -------- |
| `>=6.2.0`     | `latest`     | `experimentalRunEvents`    | `cypress.json` | global |
| `>=6.2.0`     | `latest`     | - | - | legacy |
| `<6.2.0`     | `latest`     | - | - | legacy |

if it's possible to use global hooks, then we use them. Otherwise we fallback to legacy hooks.
if you want to run with the new global hooks, you will need to add the `experimentalRunEvents` flag to `cypress.json`

## results  
the new global hooks usage has some implications in terms of when we return to the user and how we display the results.
for example, the mocha test will return almost immediately with success, and then wait for **all** tests results, where previously we waited on every **test file** and returned when we got a response from eyes.

there's an additional change in the way we present the error digest, since cypress wraps the errors of `after:run` differently:
image


packages/eyes-cypress/src/plugin/handlers.js
```diff
@@ -52,43 +45,24 @@ function makeHandlers({
...
-        }),
-      );
-      openEyes = client.openEyes;
-      getEmulatedDevicesSizes = client.getEmulatedDevicesSizes;
-      getIosDevicesSizes = client.getIosDevicesSizes;
+      runningTests.reset();
       const waitForBatch = makeWaitForBatch({
         logger: (logger.extend && logger.extend('waitForBatch')) || logger,
-        concurrency: config.concurrency,
+        concurrency: config.testConcurrency,
```
amitzur, line 51: I think we should call the property `testConcurrency`, and then inside `waitForBatch` check `testConcurrency === 5`

packages/eyes-cypress/src/plugin/hooks.js
```diff
@@ -0,0 +1,42 @@
...
+const errorDigest = require('./errorDigest');
+const {tests} = require('./runningTests');
+
+function setGlobalRunHooks(on, {visualGridClient, logger}) {
+  let waitForBatch;
+
+  on('before:run', ({config}) => {
+    waitForBatch = makeWaitForBatch({
+      logger: (logger.extend && logger.extend('waitForBatch')) || console,
+      concurrency: config.concurrency,
```
amitzur, line 17: see comment about `testConcurrency`

packages/eyes-cypress/src/plugin/hooks.js
```diff
@@ -0,0 +1,42 @@
...
+      errorDigest,
+      isInteractive: !config.isTextTerminal,
+      handleBatchResultsFile: makeHandleBatchResultsFile(config),
+    });
+  });
+
+  on('after:run', async ({config}) => {
+    try {
+      const pollBatchEnd = pollingHandler(
+        waitForBatch.bind(null, tests, visualGridClient.closeBatch),
```
amitzur, line 29: potential bug here - the reference to `tests` is the initial array in `runningTests.js`. But you assume `runningTests.reset` was not called (which is correct in this case, but if that becomes untrue you will have a hard time spotting the bug).
If `reset` was called, the reference would not be correct. This is why in `commands.js` the function `waitForBatch` is binded to the array that was created inside `batchStart`.

packages/eyes-cypress/src/plugin/hooks.js
```diff
@@ -0,0 +1,42 @@
...
+      getErrorsAndDiffs,
+      errorDigest,
+      isInteractive: !config.isTextTerminal,
+      handleBatchResultsFile: makeHandleBatchResultsFile(config),
+    });
+  });
+
+  on('after:run', async ({config}) => {
+    try {
+      const pollBatchEnd = pollingHandler(
```
amitzur, line 28: No need to poll here! This is inside the plugin. Just await `waitForBatch`

packages/eyes-cypress/test/it/plugin/handlers.test.js
```diff
@@ -8,97 +8,65 @@ const errorDigest = require('../../../src/plugin/errorDigest');
...
-
-      handlers.batchStart({viewport: {width: 200}});
-      expect(_vgcConfig).to.deep.eq({});
-
-      handlers.batchStart({viewport: {width: null, height: null}});
-      expect(_vgcConfig).to.deep.eq({});
+      await handlers.open({});
+      await handlers.checkWindow({useDom: true, enablePatterns: true, ignoreDisplacements: true});
+      expect(_args.useDom).to.be.true;
+      expect(_args.enablePatterns).to.be.true;
```
amitzur, line 30: what is this checking? `enablePatterns` was passed to checkWindow

packages/eyes-cypress/test/it/plugin/handlers.test.js
```diff
@@ -8,97 +8,65 @@ const errorDigest = require('../../../src/plugin/errorDigest');
...
-        },
-        config: {useDom: true, enablePatterns: true, ignoreDisplacements: true},
+        visualGridClient: {openEyes: args => (_args = args)},
+        config: {enablePatterns: false},
         logger: console,
         processCloseAndAbort,
         getErrorsAndDiffs,
         errorDigest,
       });
       handlers.batchStart({});
```
amitzur, line 46: No need to call `handlers.batchStart` anymore, since it's not instantiating VGC anymore.

packages/eyes-cypress/test/it/plugin/handlers.test.js
```diff
@@ -8,97 +8,65 @@ const errorDigest = require('../../../src/plugin/errorDigest');
...
         processCloseAndAbort,
         getErrorsAndDiffs,
         errorDigest,
       });
-      handlers.batchStart({viewport: {width: 200, height: 300}});
-      expect(_vgcConfig).to.deep.eq({browser: {width: 400, height: 400}});
-    });
 
-    it('does not set VGC with a wrong viewport', async () => {
       handlers.batchStart({});
```
amitzur, line 26: No need to call `handlers.batchStart` anymore, since it's not instantiating VGC anymore.

amitzur (changes requested):
Almost done with the review. But in the meantime there are many points to address.

packages/eyes-cypress/test/fixtures/testApp/cypress/integration-run/legacy-hooks.js
```diff
@@ -0,0 +1,13 @@
+/* global cy */
+describe('legacy hooks', () => {
```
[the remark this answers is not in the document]
tool3: went with `hooks.js` and both calling it

packages/eyes-cypress/src/plugin/waitForBatch.js
```diff
@@ -18,7 +18,7 @@ function makeWaitForBatch({
     const {failed, diffs, passed} = getErrorsAndDiffs(testResultsArr);
     await handleBatchResultsFile(testResultsArr);
 
-    if (concurrency == 1) {
+    if (Number(concurrency) === 1) {
```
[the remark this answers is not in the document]
tool3: changed to `Number(testConcurrency) === 5` - we need `Number` to support string numbers (env vars)

packages/eyes-cypress/src/plugin/config.js
```diff
@@ -0,0 +1,56 @@
...
+    baseConfig.failCypressOnDiff = false;
+  }
+  if (TypeUtils.isString(vgConfig.showLogs)) {
+    baseConfig.showLogs = vgConfig.showLogs === 'true' || vgConfig.showLogs === '1';
+  }
+
+  if (getProp(vgConfig, 'viewport.height') && getProp(vgConfig, 'viewport.width')) {
+    baseConfig.browser = vgConfig.viewport;
+  }
+  if (getProp(vgConfig, 'userAgent')) {
```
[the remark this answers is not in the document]
tool3: correct, removed.

packages/eyes-cypress/src/plugin/config.js
```diff
@@ -0,0 +1,56 @@
...
+  configParams,
+  ConfigUtils,
+  GeneralUtils,
+  TypeUtils,
+} = require('@applitools/visual-grid-client');
+const {version: packageVersion} = require('../../package.json');
+const agentId = `eyes-cypress/${packageVersion}`;
+const getProp = GeneralUtils.getPropertyByPath;
+
+function makeConfig(baseConfig = {}) {
```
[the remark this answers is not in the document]
tool3: agreed and removed

packages/eyes-cypress/src/browser/commands.js
```diff
@@ -17,36 +17,30 @@ function getGlobalConfigProperty(prop) {
   const shouldParse = ['eyesBrowser', 'eyesLayoutBreakpoints'];
   return property ? (shouldParse.includes(prop) ? JSON.parse(property) : property) : undefined;
 }
+if (!getGlobalConfigProperty('eyesIsDisabled')) {
+  if (getGlobalConfigProperty('eyesLegacyHooks') || getGlobalConfigProperty('isInteractive')) {
```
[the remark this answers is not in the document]
tool3: made it more readable to me - if `eyesIsDisabled` no need to check for other conditions.
changed to a single if

packages/eyes-cypress/src/plugin/pluginExport.js
```diff
@@ -1,24 +1,22 @@
...
-          config.failCypressOnDiff === undefined ? true : !!config.failCypressOnDiff,
-        eyesTimeout: config.eyesTimeout,
-        eyesDisableBrowserFetching: !!config.disableBrowserFetching,
-      };
-      return Object.assign(eyesConfig, {eyesPort}, moduleExportsResult);
+      const [on, config] = args;
+
+      if (config.experimentalRunEvents) {
+        setGlobalRunHooks(on, {visualGridClient, logger});
+        eyesConfig.eyesLegacyHooks = false;
```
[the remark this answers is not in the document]
tool3: added cypress version check

packages/eyes-cypress/src/plugin/handlers.js
tool3: changed to `Number(testConcurrency) === 5`

packages/eyes-cypress/src/plugin/hooks.js
tool3: thanks ! done !

packages/eyes-cypress/src/plugin/config.js
```diff
@@ -0,0 +1,56 @@
...
+        'failCypressOnDiff',
+        'tapDirPath',
+        'eyesTimeout',
+        'disableBrowserFetching',
+      ],
+    }),
+  );
+
+  if (vgConfig.failCypressOnDiff === '0') {
+    baseConfig.failCypressOnDiff = false;
```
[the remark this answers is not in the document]
tool3: right, removed.

packages/eyes-cypress/src/plugin/config.js
```diff
@@ -0,0 +1,56 @@
...
+  }
+  if (TypeUtils.isString(vgConfig.showLogs)) {
+    baseConfig.showLogs = vgConfig.showLogs === 'true' || vgConfig.showLogs === '1';
+  }
+
+  if (getProp(vgConfig, 'viewport.height') && getProp(vgConfig, 'viewport.width')) {
```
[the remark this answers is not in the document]
tool3: I used the same config code that was in `startPlugin` with small modifications - this is the way it was written.
so I should remove this `if` completely?
what should we send in `eyesOpen` ?

packages/eyes-cypress/src/browser/commands.js
```diff
@@ -55,7 +49,16 @@ Cypress.Commands.add('eyesOpen', function(args = {}) {
   Cypress.config('eyesOpenArgs', args);
   Cypress.log({name: 'Eyes: open'});
   const {title: testName} = this.currentTest || this.test;
-  const {browser, isDisabled} = args;
+  const {browser: eyesOpenBrowser, isDisabled} = args;
+  const globalBrowser = getGlobalConfigProperty('eyesBrowser');
+  const defaultBrowser = {
+    width: getGlobalConfigProperty('viewportWidth'),
```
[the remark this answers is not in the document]
tool3: so I am not sure I fully understand.. you want to remove the `defaultBrowser` ? or update it to `innerWidth` and `innerHeight` ?
`viewportWidth` and `viewportHeight` are coming from cypress.. why is this a mistake ? mind you that this is only in the case where customer didn't provide **any** browser in global config or `open`

packages/eyes-cypress/src/browser/commands.js
```diff
@@ -123,3 +121,16 @@ Cypress.Commands.add('eyesClose', () => {
   }
   return sendRequest({command: 'close'});
 });
+
+function fillDefaultBrowserName(browser) {
+  if (!browser.name && !browser.iosDeviceInfo && !browser.chromeEmulationInfo) {
```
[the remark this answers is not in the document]
tool3: no this is just for filling in the `name` - in #377 I validate the browser in `getBrowserInfo` which is in `eyesCheckWindow` while this runs in `open`

packages/eyes-cypress/test/it/plugin/handlers.test.js
tool3: that `eyesCheckWinow` is being run from `visualGridClient`

packages/eyes-cypress/src/browser/commands.js
```diff
@@ -73,15 +76,10 @@ Cypress.Commands.add('eyesOpen', function(args = {}) {
...
-  function fillDefaultBrowserName(browser) {
-    if (!browser.name && !browser.iosDeviceInfo && !browser.chromeEmulationInfo) {
-      browser.name = 'chrome';
-    }
-  }
   return handleCypressViewport(browser).then({timeout: 15000}, () =>
     sendRequest({
       command: 'open',
-      data: Object.assign({testName}, args),
+      data: Object.assign({testName}, args, {browser}),
```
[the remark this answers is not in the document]
tool3: updated to be in `eyesOpen`. will think of a test for it.

packages/eyes-cypress/src/plugin/hooks.js
tool3: I needed a way for the plugin to access the running tests so we can call `waitForBatch` with them.

I rely on the fact that `reset` wasn't called, do you feel like I shouldn't make that assumption ?

packages/eyes-cypress/src/plugin/config.js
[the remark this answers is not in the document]
amitzur: `startPlugin` doesn't have any reference to the word `viewport`. vgConfig is what gets loaded from applitools.config.js. There should not be an entry called `viewport` in the config file.
In `eyesOpen`, the viewport height and width should be passed (window.innerHeight, window.innerWidth), in order to use it in the plugin in case the user doesn't provide *any* browser configuration.

packages/eyes-cypress/src/browser/commands.js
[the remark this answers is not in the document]
amitzur: "coming from cypress" ===> we read it from our configuration file. But nowhere is it documented that there are these properties in the configuration file. Just look them up in the source code - this is the only place you'll find them.
So, of course NOT remove the default browser. Pass the window's innerHeight and innerWidth from the browser to the plugin, then if there is no globalBrowser or eyeOpenBrowser use those values as the default browser.

packages/eyes-cypress/src/plugin/hooks.js
amitzur: I think it's a fair assumption, but I feel it's fragile. So at least mention this assumption in a comment here. For future us.

packages/eyes-cypress/src/plugin/config.js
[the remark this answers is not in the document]
tool3: gotcha ! thanks !

amitzur (commented):
Another iteration of WIP review

packages/eyes-cypress/src/plugin/pluginExport.js
```diff
@@ -1,23 +1,23 @@
...
-        eyesBrowser: JSON.stringify(config.browser),
-        eyesLayoutBreakpoints: JSON.stringify(config.layoutBreakpoints),
-        eyesFailCypressOnDiff:
-          config.failCypressOnDiff === undefined ? true : !!config.failCypressOnDiff,
-        eyesDisableBrowserFetching: !!config.disableBrowserFetching,
-      };
-      return Object.assign(eyesConfig, {eyesPort}, moduleExportsResult);
+      const [on, config] = args;
+
+      if (config.version >= CYPRESS_SUPPORTED_VERSION && config.experimentalRunEvents) {
```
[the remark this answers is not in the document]
tool3: yeah. I realized that I can transform it to a number and validate that... but in javascript, this works perfectly:
`console.assert('1.2.0' < '1.3.0')`

packages/eyes-cypress/src/plugin/handlers.js
```diff
@@ -49,42 +42,25 @@ function makeHandlers({
...
+        waitForBatch.bind(null, runningTests.tests, visualGridClient.closeBatch),
+      );
+      return visualGridClient;
     },
-    getIosDevicesSizes: () => getIosDevicesSizes(),
-    getEmulatedDevicesSizes: () => getEmulatedDevicesSizes(),
+    getIosDevicesSizes: () => visualGridClient.getIosDevicesSizes(),
+    getEmulatedDevicesSizes: () => visualGridClient.getEmulatedDevicesSizes(),
     batchEnd: async () => {
-      logger.log(`[handlers] batchEnd`);
```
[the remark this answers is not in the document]
tool3: by mistake, thanks for catching

packages/eyes-cypress/src/plugin/hooks.js
tool3: cool thanks

packages/eyes-cypress/src/browser/commands.js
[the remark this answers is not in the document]
tool3: `viewportWidth` and `viewportHeight` are cypress configuration values - not ours.
we're basically saying: set up a default browser with the default `viewportWidth` and `viewportHeight` as reported by cypress. 
the reason I say that is because these values can be changed by the user - so using these props we will always respect the "true" default.
https://docs.cypress.io/api/commands/viewport.html#Defaults

does that make sense?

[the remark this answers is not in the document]
amitzur: ohhhhhh. forgive me, i missed this point

packages/eyes-cypress/src/plugin/hooks.js
```diff
@@ -0,0 +1,36 @@
...
+
+  on('before:run', ({config}) => {
+    const {isTextTerminal, testConcurrency} = config;
+    waitForBatch = makeWaitForBatch({
+      logger: (logger.extend && logger.extend('waitForBatch')) || console,
+      testConcurrency,
+      processCloseAndAbort,
+      getErrorsAndDiffs,
+      errorDigest,
+      isInteractive: !isTextTerminal,
```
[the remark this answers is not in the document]
tool3: unfortunately, at the point where I `setGlobalHooks` - the config from cypress doesn't have this property yet.
only inside the events I get the config with `isTextTerminal`.

since I wanted to keep `setGlobalHooks` as minimal as possible, I feel like this will complicate things a bit.

do you agree? or do you still think it's worth it ?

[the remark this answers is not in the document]
amitzur: ah, I understand what you're saying. Then I think you chose right

tool3:
waiting for the final review iteration.
things I haven't done yet:
1. documentation - I want us to discuss this because we need to have a very clear feature description
1. test for `navigator.userAgent` in `commands.js` - in general we don't have **any** tests for `commands.js`

packages/eyes-cypress/test/it/plugin/pluginExport.test.js
```diff
@@ -15,128 +17,182 @@ describe('pluginExport', () => {
...
+  });
+
   it('works with eyes disableBrowserFetching', async () => {
-    const pluginExport = makePluginExport({startServer, config: {disableBrowserFetching: true}});
+    const pluginExport = makePluginExport({startServer, eyesConfig});
     const __module = {
-      exports: () => ({bla: 'ret'}),
+      exports: (_on, config) => {
+        config.version = '6.3.0';
+        return {bla: 'ret', eyesDisableBrowserFetching: true};
```
amitzur, line 184: You mutate `config` but then return another object. Is that on purpose?
Also, returning `eyesDisableBrowserFetching` from here doesn't make sense - this is the original module, it is the "user's code".

packages/eyes-cypress/test/it/plugin/pluginExport.test.js
```diff
@@ -15,128 +17,182 @@ describe('pluginExport', () => {
...
       eyesDisableBrowserFetching: false,
       eyesLayoutBreakpoints: undefined,
+      eyesLegacyHooks: true,
       eyesIsDisabled: false,
       eyesFailCypressOnDiff: false,
       eyesBrowser: undefined,
     });
   });
 
+  it('works with eyes timeout', async () => {
```
amitzur, line 152: eyes timeout? wasn't this removed?

packages/eyes-cypress/test/it/plugin/pluginExport.test.js
```diff
@@ -15,128 +17,182 @@ describe('pluginExport', () => {
   }
 
   beforeEach(() => {
+    logger = new Logger(process.env.APPLITOOLS_SHOW_LOGS, 'eyes');
+    on = (_event, _callback) => {};
+    visualGridClient = makeVisualGridClient({logger});
     prevEnv = process.env;
     process.env = {};
+    eyesConfig = makeConfig().eyesConfig;
+    __module = {
```
amitzur, line 26: Every test defines its own `__module`. It should move from here

packages/eyes-cypress/test/it/plugin/pluginExport.test.js
```diff
@@ -15,128 +17,182 @@ describe('pluginExport', () => {
   }
 
   beforeEach(() => {
+    logger = new Logger(process.env.APPLITOOLS_SHOW_LOGS, 'eyes');
+    on = (_event, _callback) => {};
```
amitzur, line 21: also here the same comment about `noop`

packages/eyes-cypress/test/it/plugin/startPlugin.test.js
```diff
@@ -5,9 +5,10 @@ const fetch = require('../../util/fetchWithNoCAVerify');
 let startPlugin = require('../../../src/plugin/startPlugin');
 
 describe('start plugin', () => {
-  let getCloseServer, __module;
+  let getCloseServer, __module, on;
```
amitzur, line 8: `on` doesn't need to be defined separately in `beforeEach` - it doesn't hold data. You can just define a `noop` function in the file and use that.

packages/eyes-cypress/test/e2e/cypress-run-global-hooks.js
```diff
@@ -0,0 +1,58 @@
...
+      console.error('Error during test!', ex.stdout);
+      throw ex;
+    }
+  });
+
+  it('does not fail without experimentalRunEvents flag', async () => {
+    try {
+      const config = {...cypressConfig, experimentalRunEvents: false};
+      fs.writeFileSync(`${targetTestAppPath}/cypress.json`, JSON.stringify(config, 2, null));
+      await pexec(
```
amitzur, line 47: You could use `simple.js` here. Not sure we need the `hooks.js` file

packages/eyes-cypress/test/e2e/cypress-run-global-hooks.js
```diff
@@ -0,0 +1,58 @@
...
+    await pexec(`npm install`, {
+      maxBuffer: 1000000,
+    });
+  });
+
+  after(async () => {
+    fs.rmdirSync(targetTestAppPath, {recursive: true});
+  });
+
+  it('works with experimentalRunEvents flag', async () => {
```
amitzur, line 29: I think this test is not different in any way from the other e2e tests. Why should it exist?

packages/eyes-cypress/src/plugin/hooks.js
```diff
@@ -0,0 +1,38 @@
...
+const getErrorsAndDiffs = require('./getErrorsAndDiffs');
+const processCloseAndAbort = require('./processCloseAndAbort');
+const errorDigest = require('./errorDigest');
+const {tests} = require('./runningTests');
+
+function setGlobalRunHooks(on, {visualGridClient, logger}) {
+  let waitForBatch;
+
+  on('before:run', ({config}) => {
+    const {isTextTerminal, testConcurrency} = config;
```
amitzur, line 13: How come `testConcurrency` is provided here? Isn't it a Cypress object?

packages/eyes-cypress/test/it/plugin/pluginExport.test.js
```diff
@@ -15,128 +17,182 @@ describe('pluginExport', () => {
...
       eyesBrowser: undefined,
+      version: '6.3.0',
     });
 
-    const ret2 = await __module.exports(on, 'second');
-    expect(x).to.equal('second');
+    __module = {
+      exports: (_on, config) => {
+        config.version = '6.0.0';
+        return {bla: 'blah'};
```
amitzur, line 57: You mutate `config` but then return another object. Is that on purpose?

packages/eyes-cypress/test/it/plugin/pluginExport.test.js
```diff
@@ -15,128 +17,182 @@ describe('pluginExport', () => {
...
   });
 
   it('handles async module.exports', async () => {
-    const pluginExport = makePluginExport({startServer, config: {}});
+    const pluginExport = makePluginExport({startServer, eyesConfig, visualGridClient});
     const __module = {
-      exports: async () => {
+      exports: async (_on, config) => {
         await psetTimeout(0);
+        config.version = '6.3.0';
```
amitzur, line 79: You mutate `config` but then return another object. Is that on purpose?

packages/eyes-cypress/test/it/plugin/pluginExport.test.js
```diff
@@ -15,128 +17,182 @@ describe('pluginExport', () => {
...
+    const pluginExport = makePluginExport({
+      startServer,
+      eyesConfig,
+      visualGridClient,
+    });
     const __module = {
-      exports: () => ({bla: 'ret'}),
+      exports: (_on, config) => {
+        config.version = '6.3.0';
+        return {bla: 'ret', eyesIsDisabled: true};
```
amitzur, line 107: You mutate `config` but then return another object. Is that on purpose?
Also, returning `eyesIsDisabled` from here doesn't make sense - this is the original module, it is the "user's code".

packages/eyes-cypress/test/it/plugin/pluginExport.test.js
```diff
@@ -15,128 +17,182 @@ describe('pluginExport', () => {
...
     });
   });
 
   it('works with dont fail cypress on diff', async () => {
-    const pluginExport = makePluginExport({startServer, config: {failCypressOnDiff: false}});
     const __module = {
-      exports: () => ({bla: 'ret'}),
+      exports: (_on, config) => {
+        config.version = '6.3.0';
+        return {bla: 'ret', eyesFailCypressOnDiff: false};
```
amitzur, line 129: You mutate `config` but then return another object. Is that on purpose?
Also, returning `eyesFailCypressOnDiff` from here doesn't make sense - this is the original module, it is the "user's code".

packages/eyes-cypress/test/unit/plugin/waitForBatch.test.js
```diff
@@ -30,13 +31,18 @@ function processCloseAndAbort({runningTests}) {
...
-    getErrorsAndDiffs,
-    errorDigest,
-    handleBatchResultsFile: results => results,
+  let visualGridClient, logger, waitForBatch;
+  beforeEach(() => {
+    logger = new Logger(process.env.APPLITOOLS_SHOW_LOGS, 'eyes');
+    visualGridClient = makeVisualGridClient({logger});
+    waitForBatch = makeWaitForBatch({
+      logger,
+      visualGridClient,
```
amitzur, line 40: But `makeWaitForBatch` doesn't get `visualGridClient` as an argument.

packages/eyes-cypress/test/unit/plugin/handlers.test.js
```diff
@@ -402,15 +393,6 @@ describe('handlers', () => {
     expect((await close()).__test).to.equal('close_123');
   });
 
-  it('handles "batchStart"', () => {
```
amitzur, line 405: Why remove this test? Is there no logic at all in `batchStart`? What about `runningTests.reset()`?

amitzur (commented):
Where is the test in `pluginExport.test.js` that checks the condition when `eyesLegacyHooks` is false?

packages/eyes-cypress/src/plugin/config.js
```diff
@@ -0,0 +1,38 @@
...
+
+  if (vgConfig.failCypressOnDiff === '0') {
+    baseConfig.failCypressOnDiff = false;
+  }
+
+  if (TypeUtils.isString(vgConfig.showLogs)) {
+    baseConfig.showLogs = vgConfig.showLogs === 'true' || vgConfig.showLogs === '1';
+  }
+
+  const config = Object.assign(vgConfig, baseConfig);
```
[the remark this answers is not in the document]
tool3: 👍🏼

packages/eyes-cypress/test/it/plugin/pluginExport.test.js
tool3: 👍🏼

packages/eyes-cypress/test/it/plugin/pluginExport.test.js
tool3: merge conflicts - will merge from master soon and this will be removed

packages/eyes-cypress/test/it/plugin/pluginExport.test.js
tool3: 👍🏼

packages/eyes-cypress/src/plugin/hooks.js
tool3: no, it is a combined object of both cypress and what we return from the plugin.

since these events run asynchronously, the config here already has what we returned from `pluginExport`.

but this comment made me realize that we didn't set `testConcurrency` properly like other properties (for example `eyesDisableBrowserFetching`) - fixed.

packages/eyes-cypress/test/it/plugin/pluginExport.test.js
tool3: no it's a mistake - I only do this once in the file and that is for testing `eyesLegacyHooks`.

fixed all occurrences of this.

packages/eyes-cypress/test/unit/plugin/waitForBatch.test.js
tool3: my bad. no other test in this file used it.

packages/eyes-cypress/test/unit/plugin/handlers.test.js
tool3: readded it to test `runningTests.reset()` is happening as expected

packages/eyes-cypress/test/e2e/cypress-run-global-hooks.js
tool3: used `simple.js`

packages/eyes-cypress/test/e2e/cypress-run-global-hooks.js
tool3: I updated this test to have 2 steps:
1. add the `experimentalRunEvents` flag in the test and verify it's working.
2. remove the `experimentalRunEvents` in the test and verify it's working.

so now, all tests are still running with legacy hooks, and only this one verifies the flag.

tool3:
> Where is the test in `pluginExport.test.js` that checks the condition when `eyesLegacyHooks` is false?

`sets eyesLegacyHooks` in `pluginExport.test.js:31`

packages/eyes-cypress/test/e2e/cypress-run-global-hooks.js
amitzur: Where is part (2) of what you wrote? (remove `experimentalRunEvents`)

packages/eyes-cypress/test/unit/plugin/handlers.test.js
amitzur: where is it?

packages/eyes-cypress/test/e2e/cypress-run-global-hooks.js
tool3: the 2nd `it` in `packages/eyes-cypress/test/e2e/cypress-run-global-hooks.js`:

```javascript
it('does not fail without experimentalRunEvents flag', async () => {
```

packages/eyes-cypress/test/unit/plugin/handlers.test.js
tool3: `packages/eyes-cypress/test/unit/plugin/handlers.test.js` line 389

packages/eyes-cypress/test/unit/plugin/config.test.js
```diff
@@ -0,0 +1,34 @@
...
+      eyesLegacyHooks: true,
+      eyesDisableBrowserFetching: false,
+      eyesTestConcurrency: 5,
+    });
+  });
+
+  it('should work with env variables', () => {
+    process.env.APPLITOOLS_IS_DISABLED = true;
+    const {config, eyesConfig} = makeConfig();
+    expect(config.isDisabled).to.be.true;
```
amitzur, line 23: why is Cypress configuration affected by `APPLITOOLS_IS_DISABLED` ?

amitzur (approved):
We're there!
Only need to merge https://github.com/applitools/eyes.sdk.javascript1/pull/393

packages/eyes-cypress/test/unit/plugin/config.test.js
tool3: the `config` comes from VG - which takes care of the env side of things.

so the `eyesConfig` is just a reflection of what we got on the `config` object:
```javascript
 eyesIsDisabled: !!config.isDisabled,
```

it's part of the default `configParams` from `visual-grid-client`:
https://github.com/applitools/eyes.sdk.javascript1/blob/master/packages/visual-grid-client/src/sdk/configParams.js

PR #385, opened 2021-02-19
adding param to eyesGetAllTestResults to not throw when we have errors

zalohcin:
This function does not seem to be in the documentation.

Calling eyesGetTestResults only returns one test, even when you have multiple browsers enabled.

Calling eyesGetAllTestResults when a check failed causes the function to throw, which also prevents me from getting the full list of checks.

### DIFF
--- a/packages/eyes-webdriverio-5/src/eyes-service.js
+++ b/packages/eyes-webdriverio-5/src/eyes-service.js
@@ -94,8 +94,8 @@ class EyesService {
       return this._eyes.getConfiguration()
     })
 
-    global.browser.addCommand('eyesGetAllTestResults', async () => {
-      return this._eyes.getRunner().getAllTestResults()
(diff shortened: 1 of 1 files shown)


packages/eyes-webdriverio-5/src/eyes-service.js
```diff
@@ -94,8 +94,8 @@ class EyesService {
       return this._eyes.getConfiguration()
     })
 
-    global.browser.addCommand('eyesGetAllTestResults', async () => {
-      return this._eyes.getRunner().getAllTestResults()
+    global.browser.addCommand('eyesGetAllTestResults', async (throwEx = true) => {
```
amitzur, line 97: No need for the default value here. It's handled in `runner.getAllTestResults`.

amitzur (commented):
Thanks @zalohcin !
We'll merge this PR and release the service with the changes.

PR #386, opened 2021-02-22
sdk-test-kit-mocha

tourdedave:
We rely heavily on a regex to parse and run our tests in Mocha. Let's abstract it into a central place and add unit tests for it!


amitzur:
This was the actual bug: https://github.com/applitools/eyes.sdk.javascript1/pull/387
So this PR is not needed

tourdedave:
I still think we should do it. We rely on this regex and have had troubles with it in the past. Also, this would be a good place to put a few things that are in sdk-shared at the moment (e.g., the mocha hooks)

tourdedave:
Also, this PR already shows a bug (feature?) in the regex. If we specify a tag in a test name but it's not in the .mocha.cvg.js file then the test will not be run. Even if there are other tags in the test name which do match what is in the .mocha.cvs.js file.

amitzur:
@tourdedave If I'm not mistaken, it's a feature. All tags need to be matched. Otherwise you're running a test with a feature that isn't supported.
For example, there is no native-selector support in any SDK but wdio. So a test that has native-selector in its tags shouldn't be run **unless** native-selector is present in the regex.
I hope this helps resolve your confusion. Does it?

amitzur:
But regardless I'm all for breaking down sdk-shared into dedicated packages!
@tool3 @kyryloonufriiev please provide feedback

tourdedave:
Yes, thanks for explaining, and tests that demonstrate this would make this clear for everyone 😁

tool3:
I think this is great !

The one comment I have is about the name, I think `sdk-test-kit` is a bit confusing.

`sdk-test-utils` makes more sense for me, as the name implies that it's a wrapper for common things we use for testing, but not something we use for running tests.

tourdedave:
> I think this is great !
> 
> The one comment I have is about the name, I think `sdk-test-kit` is a bit confusing.
> 
> `sdk-test-utils` makes more sense for me, as the name implies that it's a wrapper for common things we use for testing, but not something we use for running tests.

👍 I was envisioning pulling in other util-like things (e.g., the mocha hooks) from sdk-shared too, so the name change makes sense

amitzur:
There was a moment I thought, `sdk-mocha-utils`, but that might prove to be short-sighted. `sdk-test-utils` it is?

tourdedave:
> There was a moment I thought, `sdk-mocha-utils`, but that might prove to be short-sighted. `sdk-test-utils` it is?

I'm worried that a more general name like `sdk-test-utils` could turn this into a bit of a junk drawer like `sdk-shared` was. But going too granular could be annoying if we need to share functionality between more than one test-util package. Thoughts?

amitzur:
@tourdedave let's go with `sdk-test-utils` and **see what happens** :tm:

amitzur:
@tourdedave so?

kyryloonufriiev:
Implemented https://github.com/applitools/eyes.sdk.javascript1/commit/d1b1e8efe0bcd7b906b534e9a4b2b0e08be86aa4

PR #387, opened 2021-02-22
Fix bug with unexecuted and unskipped cvg tests

amitzur:
There was a bug when a coverage test is emitted, but it's excluded from the test suite because the framework doesn't support this feature (e.g. native-selectors in Selenium). The bug is that this test is reported as `isSkipped: false` (because it's not emitted to be skipped) but with no test result status (since it was not executed).

This PR fixes the issue by filtering those cases out of the report.

### DIFF
--- a/packages/sdk-coverage-tests/CHANGELOG.md
+++ b/packages/sdk-coverage-tests/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- fix bug with unskipped and unexecuted tests
 
 ## 2.3.7 - 2021/2/6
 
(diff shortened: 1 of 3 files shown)


tourdedave (approved):

tourdedave:
For the case of the native selectors test, we'll need to make sure to add the `@native-selector` tag to the list of tags in `.mocha.cvg.js` - or else the test will not be run. This issue surfaced in PR #386

amitzur:
@tourdedave this isn't supported in Selenium. It's intentionally not run.

tourdedave:
Then I misunderstand how the regex is supposed to work

PR #389, opened 2021-02-24
chore(eyes-cypress): global cypress test

tool3:
following the work of #384, this PR is an attempt to create a wrapper around the `Cypress` and `cy` objects for testing purposes.

the `fakeCypress` method accepts a relative path to a module and `require`s it with overwritten globals, such as `window`, `navigator`, `cy` and `Cypress`.

we can validate access to global variables and `Cypress`/`cy` methods, as well as arguments to `sendRequest` - which is ultimately what the commands in `commands.js` are doing.

an example test can look like this:
```javascript
it('should call handleViewport with correct viewport', async () => {
  const {context, calls} = fakeCypress(path.resolve(__dirname, '../../../src/browser/commands'));
  const eyesOpen = calls['eyesOpen'];
  const browser = {width: 800, height: 600};
  eyesOpen.call({currentTest: {title: 'test' }}, {browser});
  const viewport = context.journal.cy.viewport;
  expect(viewport.called).to.be.true;
  expect(viewport.callCount).to.equal(1);
  expect(viewport.args).to.deep.equal([800, 600]);
});
```

breakdown:
- we call `fakeCypress` with a path to our module (`commands.js` in this case)
- we destructure `calls` and `context`:
  - `calls` is an object with the keys being the `eyes` commands (e.g `eyesOpen`, `eyesCheckWindow`etc...)
  - `context` is the object which overrides `Cypress`, `cy`, `window` and `navigator`
- call the function `eyesOpen` with an object that will replace `this` (because of usage such as `this.currentTest` in `eyesOpen`)
- use the `journal` object on `context` which records the arguments and calls metadata about a given function.
- validate args and calls metadata

at this point we need to extend `fakeCypress` to include more methods for `cy` (currently only `viewport` implemented)

example of args validation:
```javascript
it('should validate args to sendRequest', async () => {
  const {context, calls} = fakeCypress(path.resolve(__dirname, '../../../src/browser/commands'));
  const eyesOpen = calls['eyesOpen'];
  const browser = {width: 800, height: 600};
  eyesOpen.call({currentTest: {title: 'test' }}, {browser});
  const viewport = context.journal.cy.viewport;
  const fetch = context.journal.window.fetch;
  viewport.cb();
  const [, args] = fetch.args;
  expect(fetch.called).to.be.true;
  expect(JSON.parse(args.body)).to.deep.equal({
    testName: 'test',
    browser: {width: 800, height: 600, name: 'chrome'},
    userAgent: 'fake_userAgent',
  });
});
```

### DIFF
--- a/packages/eyes-cypress/test/it/browser/commands.it.test.js
+++ b/packages/eyes-cypress/test/it/browser/commands.it.test.js
@@ -19,8 +19,13 @@ function override(target, method, name) {
 }
 
 function fakeCypress(modulePath, calls = {}) {
-  const stub = {called: false, callCount: 0, args: []};
-  this.state = {viewport: {...stub}, fetch: {...stub}};
+  function setFake(name, args) {
+    this.state[name].called = true;
+    this.state[name].args = args;
+    this.state[name].callCount++;
+  }
+  const fake = {called: false, callCount: 0, args: []};
+  this.state = {viewport: {...fake}, fetch: {...fake}};
   this.Cypress = {
     Commands: {add: (name, func) => (calls[name] = func)},
     config: () => {},
@@ -29,18 +34,14 @@ function fakeCypress(modulePath, calls = {}) {
   };
   this.cy = {
     viewport: (...args) => {
-      this.state.viewport.called = true;
-      this.state.viewport.args = args;
-      this.state.viewport.callCount++;
+      setFake('viewport', args);
       return {then: (_args, cb) => (this.state.viewport.cb = cb)};
     },
   };
 
   this.window = {
     fetch: (...args) => {
-      this.state.fetch.called = true;
-      this.state.fetch.callCount++;
-      this.state.fetch.args = args;
+      setFake('fetch', args);
       return {
         then: () => ({
           json: () => {},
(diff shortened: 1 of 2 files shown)


amitzur:
To me this is confusing, both in API and implementation.
I looked at it thoroughly and even tried to come up with a different approach (actually run Cypress, but mock the server) but without much success.
I think this is a hard endeavor. There's a need here to mock entire Cypress, and I fear that it won't be maintainable.

A couple of notes about implementation:
1) `createContext` pollutes the global namespace, instead of actually, well, creating a context. I think you meant to use `vm.runInContext` and `vm.createContext` maybe?
2) `fakeCypress` - To ease readability, please adhere to the code style guidelines of writing the inner functions after the code.
3) `this.journal` inits `cy` and `window` to `{...fake}` - but the reference to `args` is the same. I think there's a bug hiding there.

If you think this is a direction worth pursuing, I'd gladly brainstorm about it more.

tool3:
I think faking cypress is inevitable and will definitely be a maintenance task - not a hard one - but it will require maintenance for sure. I don't think we even want another way because we want to control the cypress side of things.

at this point, I think we'll sunset this and revisit it if our need to test browser-side javascript in cypress becomes more serious.

PR #390, opened 2021-02-25
Fix/testcafe viewport size

tourdedave:
This PR is for a more robust implementation of `getWindowRect` & `setWindowRect` which resolves the issues noted in [Trello 736](https://trello.com/c/xNCZNfPi)


packages/eyes-testcafe/test/it/legacy-api.spec.js
```diff
@@ -248,6 +248,7 @@ test('eyes failTestcafeOnDiff true', async t => {
 test('should load applitools.config.js', async t => {
   const configPath = path.join(__dirname, 'applitools.config.js')
   const eyes = new Eyes({configPath})
+  await t.resizeWindow(2000, 1000)
```
amitzur, line 251: what difference does it make in this specific test?

packages/eyes-testcafe/test/it/spec-driver.spec.js
```diff
@@ -274,21 +274,18 @@ test('getWindowRect', async driver => {
...
-test.skip('setWindowRect (width, height)', async driver => {
-  const expectedRect = {x: 0, y: 0, width: 500, height: 500}
+test('getWindowRect after resize', async driver => {
+  await driver.resizeWindow(2000, 1000)
+  const rect = await spec.getWindowRect(driver)
+  assert.deepStrictEqual(rect.width, 2000)
+  assert.deepStrictEqual(rect.height, 1000)
+})
+test('setWindowRect (width, height)', async driver => {
+  const expectedRect = {width: 500, height: 500}
```
amitzur, line 284: we don't care about x,y ?

packages/eyes-testcafe/src/spec-driver.js
```diff
@@ -306,31 +306,57 @@ async function waitUntilDisplayed(_driver, element, timeout) {
...
+// window.outerWidth & window.outerHeight is not reliable since these
+// values don't get updated when resizing the window, and there are cases
+// where it's not populated (e.g., headless Chrome on Mac). And using a
+// third-party library like testcafe-browser-tools isn't sufficient either
+// since the native binary for linux runs into issues out-of-the-box when
+// running in Github Actions.
+//
+// As a result, we opt for a less efficient implementation in getWindowRect
+// but it is based on one fact which seems to hold true -- window.innerHeight
+// and window.innerWidth are a trustworthy source of truth. From this starting
```
amitzur, line 319: We also assume
a) `setWindowRect(w, h)` sets it correctly to `w` and `h`
b) the diff between the window and the viewport doesn't change when changing the window size

To remove assumption (b) we do the compensation/verification loop that you mention below.

packages/eyes-testcafe/src/spec-driver.js
```diff
@@ -306,31 +306,57 @@ async function waitUntilDisplayed(_driver, element, timeout) {
...
+  await setWindowRect(driver, {width: 600, height: 600})
+  const anchorViewportSize = await executeScript(
+    driver,
+    `return {
+      width: window.innerWidth,
+      height: window.innerHeight
+    }`,
+  )
+  const diff = {
+    width: initialViewportSize.width - anchorViewportSize.width,
```
amitzur, line 348: This is not the correct calculation. The diff should be `600 - anchorViewportSize.width`. Then the `windowRect` you set back to should logically be `diff.width = initialViewportSize.width + diff.width`.

amitzur (changes requested):
It's the wrong calculation - see comment inline

packages/eyes-testcafe/test/it/legacy-api.spec.js
tourdedave: It made sense with the old implementation (it surfaced the bug) - this test can be removed I think

packages/eyes-testcafe/test/it/spec-driver.spec.js
tourdedave: I think we do (based on what's in other specs) but the SDK is not able to change it w/ setWindowRect so I took it out to see how it behaved - all tests pass w/o it (including coverage)

packages/eyes-testcafe/src/spec-driver.js
tourdedave: 👍

packages/eyes-testcafe/src/spec-driver.js
tourdedave: Okay, I felt confident in the current calculation since it was getting to the correct size and all of the tests passed with it (including coverage). But I realize now it's creating a window size based on just the viewport sizes - so it would fall down when running a headed browser. I'll see about adding a test to verify, and update the calculation.

amitzur: Exactly.

packages/eyes-testcafe/test/it/spec-driver.spec.js
tourdedave: It looks like screenX and screenY don't get updated as part of a window resize in TestCafe, so there might not be a reliable way to track it

tourdedave: added x, y back to getWindowRect

packages/eyes-testcafe/src/spec-driver.js
tourdedave: Running with a headed browser still worked with the existing calculation 🤔. Regardless, I've updated it.

amitzur: that is so weird. did you debug it to see what intermediate values the code produces?

tourdedave: No, I didn't. Do you think it's worth investigating?

amitzur: I think when we get results like this that we cannot explain - we must investigate, especially given how little effort it requires. 

Actually, I was so curious how this could happen that I did the investigation. I added a couple of console logs to `getWindowRect`, and reached a conclusion that TestCafe's `driver.resizeWindow` method actually resizes the browser to the desired **viewport size**, and not window size.
So the reason that tests always passed before your change is because the diff was always 0.
I opened a PR for this: https://github.com/applitools/eyes.sdk.javascript1/pull/392

tourdedave:
## Results

| browser | result |
| - | - |
| Chrome | ✔ |
| Firefox | ✔|
| Safari | ✔|
| IE 11 | ✘ |
| Edge classic | TBD |
| Edge latest | TBD |

## Stack traces

### IE 11

```sh
 ✖ setViewportSize (width, height)

   1) AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
      + actual - expected

      {
      +   height: 683,
      +   width: 1024
      -   height: 500,
      -   width: 500
      }

      + expected - actual

       {
      -  height: 683
      -  width: 1024
      +  height: 500
      +  width: 500
       }
```

amitzur (approved):

amitzur:
@tourdedave :flushed: what a discovery about remote browsers!

PR #392, opened 2021-02-27
Fix setViewportSize in testcafe

amitzur:
It appears that TestCafe's `resizeWindow` actually interprets the input as the desired **viewport size**.
This PR validates it on chrome/firefox headless/headful. Needs to be validated for IE/Safari/Edge as well.
I couldn't get the Sauce Labs connector (testcafe-browser-provider-saucelabs) to work


tourdedave:
Nice find! Confirmed in the TestCafe source code (e.g., [test](https://github.com/DevExpress/testcafe/blob/ec6a44313d91d82b5d51aff466fef4e4c7d636e0/test/functional/fixtures/multiple-windows/testcafe-fixtures/api/api-test.js#L352), [window size util used in test](https://github.com/DevExpress/testcafe/blob/master/test/functional/window-helpers.js#L6-L7))

tourdedave (approved):

PR #393, opened 2021-02-28
Reset color in Cypress error message

amitzur:
I'm not sure it looks better, but I wanted to point this -
Cypress paints the error in yellow. So anything we don't reset becomes yellow:
Screenshot from 2021-02-28 09-49-26

WDYT about this change?
Screenshot from 2021-02-28 09-49-04

### DIFF
--- a/packages/eyes-cypress/package.json
+++ b/packages/eyes-cypress/package.json
@@ -40,6 +40,7 @@
   "dependencies": {
     "@applitools/dom-snapshot": "4.4.11",
     "@applitools/functional-commons": "1.6.0",
+    "@applitools/snaptdout": "^1.0.0",
     "@applitools/visual-grid-client": "15.5.22",
     "body-parser": "1.19.0",
     "chalk": "3.0.0",
--- a/packages/eyes-cypress/src/plugin/errorDigest.js
+++ b/packages/eyes-cypress/src/plugin/errorDigest.js
@@ -14,6 +14,7 @@ const formatByStatus = {
   },
   Unresolved: {
     color: 'yellow',
+    chalkFunction: chalk.ansi256(214),
     symbol: '\u26A0',
     title: tests => `Diffs detected - ${tests} tests`,
   },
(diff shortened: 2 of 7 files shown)


tool3:
I think it's a good idea. the only thing I see here that could use color is the url.

I think the url should be colored in a different color (teal will make it applitoolsy) to make it distinct amongst all of the other information the user is seeing. for example:

image

tool3 (approved):

amitzur:
@tool3 agreed. Can you make it `teal`? While we're at it, I suddenly realize that diffs show up in yellow. They should be **orange**.

tool3:
okay so now it looks like this:
image


now regarding the failing tests in `errorDigest` - I feel like this is a perfect time to introduce `snaptdout` because I realize just how simple it makes these types of fixes, where string comparison is precise and shouldn't be done by hand.

I initially started to fix the tests in `errorDigest` but I noticed how brittle it is, and then I remembered we had a PR for that #330  :)

amitzur:
@tool3 is the `Error:` and stack trace in your screenshots **orange**?
In mine they are yellow. And I expect them to stay yellow, but only the diffs should appear orange.
Also, the url in the first screenshot you shared is not the same color as in the second screenshot you shared.
Do you mind committing what you wish to add?

amitzur:
@tool3 I will review snaptdout and let's ship this

tool3:
the screenshot I posted earlier was a mistake. this is how it looks now:
image


I am happy with the url color and diff's orange. WDTY?

tool3:
added `snaptdout`. if you want to see the snapshots, copy the `print` npm script from `snaptdout`, or simply:
```bash
node -e 'Object.values(require(`./tests/snapshots/stdout.test.json`)).forEach(i => console.log(`\n${i.join(`\n`)}\n`))'
```

amitzur:
@tool3 the snapshot printing is great! This is what I ran to see it:
```
node -e 'Object.values(require(`./test/unit/plugin/snapshots/errorDigest.test.json`)).forEach(i => console.log(i.join(`\n`)))'
```

I spoke with product - let's add the following:
1) Add `See details at <batch url>` line also before the test details. So something like:
```
Eyes-Cypress detected diffs or errors during execution of visual tests. See details at <batch url>.
    Passed - 4 tests
    ...
```

2) Add 2 line feeds at the end of the error digest, to make a clear separation between the URL and the stack trace.

amitzur:
@tool3 Also in order to review the PR it needs to be cleaned up from commits from master.

tool3:
PR cleaned. formatting added. 
please pull and re-run:
```bash
node -e 'Object.values(require(`./test/unit/plugin/snapshots/errorDigest.test.json`)).forEach(i => console.log(i.join(`\n`)))'
```

amitzur:
@tool3 notes:
1. URL needs to be colored in the "header" the same way it's colored in the "footer".
2. The `\n\n` is not at the end

PR #376, opened 2021-02-11
fix(eyes-cypress): remove timeout

tool3:
[trello](https://trello.com/c/6UBivn4V)

remove `eyesTimeout` completely.

PR #394, opened 2021-02-28
fix(eyes-cypress): eyes check window timeout

tool3:
[trello](https://trello.com/c/5Thz1QFs/808-experian-cypress-timeout-when-setlayoutbreakpoint-is-enabled)

while #376 removed eyesTimeout, we notice customers that are hitting the default cypress command timeout (4 seconds) when using js-layouts.

the fix is to set explicit timeout for the `cy.then` that wrap our `takeDomSnapshot` call.

I set it to 900,000 ms which is 15 minutes.

### DIFF
--- a/packages/eyes-cypress/src/browser/eyesCheckWindow.js
+++ b/packages/eyes-cypress/src/browser/eyesCheckWindow.js
@@ -53,7 +53,7 @@ function makeEyesCheckWindow({sendRequest, processPage, domSnapshotOptions, cypr
             cypress
               .viewport(Number(requiredWidth), height, {log: false})
               .wait(300, {log: false})
-              .then(() => {
+              .then({log: false, timeout: 900000}, () => {
                 return takeDomSnapshot(options).then(snapshot => {
                   browsersInfo.forEach(({index}) => (snapshots[index] = snapshot));
                 });
--- a/packages/eyes-cypress/CHANGELOG.md
+++ b/packages/eyes-cypress/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- set explicit timeout for eyesCheckWindow [trello](https://trello.com/c/5Thz1QFs)
 
 ## 3.19.1 - 2021/2/24
 
--- a/packages/eyes-cypress/test/e2e/cypress-run-breakpoints.test.js
+++ b/packages/eyes-cypress/test/e2e/cypress-run-breakpoints.test.js
@@ -4,13 +4,26 @@ const {exec} = require('child_process');
 const {promisify: p} = require('util');
 const path = require('path');
 const pexec = p(exec);
+const testServer = require('@applitools/sdk-shared/src/run-test-server');
 const fs = require('fs');
 
 const sourceTestAppPath = path.resolve(__dirname, '../fixtures/testApp');
 const targetTestAppPath = path.resolve(__dirname, '../fixtures/testAppCopies/testApp-breakpoint');
 
 describe('layout breakpoints', () => {
+  let closeServer;
   before(async () => {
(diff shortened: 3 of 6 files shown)


amitzur (commented):
great, need to add tests :)

tool3:
a couple of notes about the latest commit:
1. we will have to release `sdk-shared` and modify the test to include the middleware file from there (rather than the local filesystem).
2. this made me realize that we should update `testServer` with the ability to provide a middleware inline.

about the test:
initially, I added an `img` tag (`smurgs.jpg`) and I didn't understand why it kept working even if I configured some obscene value for the resource timeout in the middleware (`8000ms`).
then, I realized that `cy.visit` will wait for the entire page to load, so `eyesCheckWindow` still ran on an already loaded page.

the solution was to append the `img` in a `setTimeout` using javascript, and keep the resource timeout at `5000ms`- this simulates a "slow" dom-snapshot.

packages/eyes-cypress/src/browser/eyesCheckWindow.js
```diff
@@ -53,7 +53,7 @@ function makeEyesCheckWindow({sendRequest, processPage, domSnapshotOptions, cypr
             cypress
               .viewport(Number(requiredWidth), height, {log: false})
               .wait(300, {log: false})
-              .then(() => {
+              .then({log: false, timeout: 900000}, () => {
```
tourdedave, line 56: Do we want to make this configurable in the off-chance that a user bumps up against it?

tourdedave (approved):
LGTM! 💯 

I'm surprised Cypress doesn't offer the option to disable timeouts.

packages/eyes-cypress/src/browser/eyesCheckWindow.js
amitzur: what's also surprising to me is that this is the first and only place we had to do something like this. Why is that really? can you tell?

tool3: actually, it's not the only place -  please see `commands.js`. 

I have just confirmed that this timeout takes precedence and you cannot override it with global cypress config.

we need to make sure that the value we are setting here, is not too long, and not too short.

too short - obviously not good because customers will hit it all the time.
too long - also not good because if dom-snapshot hangs - the test will only fail after the amount of time we specify here.

15 minutes felt like a good middle point - do you think we should increase it to say 30 minutes ?

tourdedave: Magic numbers are concerning. Picking a sensible default but making it user configurable is a pretty safe play. Especially since it sounds like a house of cards waiting to fall over 😂

tool3: these are not magic numbers such as ones that you cannot explain in the code, but rather something cypress is forcing us to do, and I don't think the user should be concerned with this.

this value - is strictly to avoid a timeout when capturing snapshots that could take a while - such as js-layouts snapshots.

if we make this user-configurable - than it makes more sense to just drop it, and have the customer control it with the regular cypress timeouts. WDYT?

tourdedave: I'm fine with whatever you decide.

amitzur: I think dom-snapshot can be assumed to not hang, so the timeout we specify here might as well be infinity.
dom-snapshot sometimes has bugs which cause it to hang (like when there was a streaming resource). That's pretty rare. But when it happens then the Cypress execution will hang and we will find a solution to fix it. The timeout will not save us there, it will just cause a different output, but we will still be prompted to fix the issue.

PR #398, opened 2021-03-08
Add runAfter hook

SophieDePaula:
Added runAfter hook to allow the user to clean up the browser before the next story renders.
This was requested by Wix ([trello card](https://trello.com/c/Y9JyMiIt/653-storybook-add-runafter-method?menu=filter&filter=due:notdue))


packages/eyes-storybook/CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 # Changelog
 
 ## Unreleased
-
+ - added after hook
```
amitzur, line 4: after --> runAfter

packages/eyes-storybook/README.md
```diff
@@ -659,6 +660,24 @@ storiesOf('UI components', module)
...
+
+An asynchronous function that is evaluated after the story's screenshot is taken. This is the place to perform any clean ups that could change the way the next story renders.
+
+For example, removing a background color that was left from a previouse component:
+
+```js
+.add('background color', () => (
+  <div>Component with background color</div>
+  ), {
+  eyes: {
```
amitzur, line 673: I think the better example is what you have in the e2e test - a `runBefore` function that creates dirty state, and a `runAfter` function that cleans it up.
And the text inside the div could say this - `Component with runBefore hook that leaves a background-color style on the html element`.

packages/eyes-storybook/src/renderStories.js
```diff
@@ -103,6 +105,13 @@ function makeRenderStories({
             throw new Error(errMsg);
           }
 
+          if (eyesParameters && eyesParameters.runAfter) {
```
amitzur, line 108: I think this should be in `getStoryData`, after `takeDomSnapshots`.

packages/eyes-storybook/test/fixtures/appWithStorybook/index.js
```diff
@@ -15,6 +15,18 @@ const circular = {param: true}
 circular.inner = circular
 
 storiesOf('Button', module)
+.add('background color', () => (
+  <div style={{fontSize: 'xxx-large'}}>PINK!</div>
```
amitzur, line 19: `xxx-large` is not really a valid value for fontSize. Even though it doesn't matter, maybe it's confusing? Not sure.

packages/eyes-storybook/test/fixtures/appWithStorybook/index.js
```diff
@@ -15,6 +15,18 @@ const circular = {param: true}
...
+  ), {
+  eyes: {
+    runBefore({rootEl, story}) {
+     document.querySelector("html").style.backgroundColor = 'pink';
```
amitzur, line 23: Just FYI that `fuchsia` is also an option :)

packages/eyes-storybook/test/fixtures/appWithStorybook/index.js
```diff
@@ -15,6 +15,18 @@ const circular = {param: true}
...
+    },
+    runAfter({rootEl, story}){
+     document.querySelector("html").style.backgroundColor = '';
```
amitzur, line 26: If you wanted to be 100% purist, you'd save the original value from the `runBefore` function. Maybe it was red before it was changed to pink?
I don't think it's necessary here, but I'm wondering if it makes sense to add this to the documentation, or as a recipe.

packages/eyes-storybook/src/getStoryIndex.js
```diff
@@ -0,0 +1,18 @@
+const getClientAPI = require('./browser/getClientAPI');
+
+function getStoryIndex(index) {
```
amitzur, line 3: `getStoryByIndex`, not `getStoryIndex`.

packages/eyes-storybook/src/getStoryIndex.js
```diff
@@ -0,0 +1,18 @@
+const getClientAPI = require('./browser/getClientAPI');
```
amitzur, line 1: This file should be under `src/browser`, where all files which are intended to be run in the browser are located.

packages/eyes-storybook/src/browser/runRunBeforeScript.js
```diff
@@ -1,17 +1,11 @@
...
-    const story = api.getStories()[index];
-    if (!story) {
-      console.log('error cannot get story', index);
-      return;
-    }
+    let story = getStoryIndex(index);
     return story.parameters.eyes.runBefore({rootEl: document.getElementById('root'), story});
   } catch (ex) {
-    return {message: ex.message, version: api ? api.version : undefined};
+    return {message: ex.message};
```
amitzur, line 8: returning the api version from this function is important for troubleshooting

packages/eyes-storybook/src/browser/runRunBeforeScript.js
```diff
@@ -1,17 +1,11 @@
...
 function runRunBeforeScript(index) {
-  let api;
   try {
-    api = getClientAPI();
-    const story = api.getStories()[index];
-    if (!story) {
-      console.log('error cannot get story', index);
-      return;
-    }
+    let story = getStoryIndex(index);
```
amitzur, line 5: I think this is not preserving existing behavior. It's a problem that we don't have tests for it. Since this is code that runs in the browser and has quite a few dependencies on Storybook, then it's a bit hard to test.
So either we keep it untested or we add tests, but we want to make sure that the same behavior is preserved. Before these changes, this was the behavior:
1. `getClientAPI` may throw, in which case we return `{message: ex.message}`
2. `getClientAPI` may return undefined, in which case we return `{message: ex.message}` but the error is actually `cannot call method getStories of undefined` or something like that.
3. `api.getStories()[index]` may return undefined, in which case we log it and return undefined.
4. Otherwise - run the runBefore function (btw, it's available since the Node.js side verified that it is defined).
5. if `runBefore` throws, we return `{message, version}`

But now, `getStoryIndex` doesn't behave like this at all:
1. if `getClientAPI` throws, `getStoryIndex` returns {message} and runRunBeforeScript will throw `cannot call eyes of undefined`. Which is an irrelevant exception.
2. if `getClientAPI` returns undefined, same problem as (1).
3. if `api.getStories()[index]` returns undefined, similar problem as (1) and (2)

amitzur (changes requested):
Several changes needed, but most importantly:
1) Behavior is not preserved in error handling of client API.
2) Execution of `runAfter` is done from the wrong place.

packages/eyes-storybook/test/fixtures/appWithStorybook/index.js
SophieDePaula: Done ^_^

packages/eyes-storybook/test/fixtures/appWithStorybook/index.js
SophieDePaula: Make sense, I can keep it under the window object. Is there a better way?

```
  eyes: {
    runBefore({rootEl, story}) {
     window.originalBackgoundColor = document.querySelector("html").style.backgroundColor;
     document.querySelector("html").style.backgroundColor = 'fuchsia';
    },
    runAfter({rootEl, story}){
     document.querySelector("html").style.backgroundColor = window.originalBackgoundColor;
    }
  }
```

packages/eyes-storybook/src/browser/runRunBeforeScript.js
SophieDePaula: To resolve this (and the previous comment), I throw the exception from `getStoryByIndex.js` instead of returning so that runBefore/After is not called and the relevant info is returned. 
I also added a check in both before and after hooks for #3.

`if (!story) return;`

packages/eyes-storybook/README.md
```diff
@@ -659,6 +660,28 @@ storiesOf('UI components', module)
...
+.add('background color', () => (
+  <div style={{fontSize: '30px'}}>Component with runBefore hook that modifies the background color</div>
+  ), {
+  eyes: {
+    runBefore({rootEl, story}) {
+     window.originalBackgoundColor = document.querySelector("html").style.backgroundColor;
+     document.querySelector("html").style.backgroundColor = 'fuchsia';
+    },
+    runAfter({rootEl, story}){
+     document.querySelector("html").style.backgroundColor = window.originalBackgoundColor;
```
amitzur, line 679: and `delete window.originalBackgroundColor`

amitzur (approved):
:100:

PR #399, opened 2021-03-09
fix(snippets): don't throw when no matching translate offset

tool3:
[trello](https://trello.com/c/duAwaupv/816-fanatics-js-selenium-4-test-stuck-in-running)

this test is running on iOS simulator on Sauce Labs.
we forgot to wrap `getInnerOffset` in try catch in `EyesScreenshotNew`, and so when we encounter something like: `transform: none` - we quit the test unexpectedly with exception: 
```text
WebDriverError: An unknown server-side error occurred while processing the command. Original error: Can't parse CSS transition: none!
```

I wasn't sure how to test this, since it's a very niche feature, and going out to sauce is wasteful.

### DIFF
--- a/packages/eyes-sdk-core/CHANGELOG.md
+++ b/packages/eyes-sdk-core/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- fix exception thrown in getInnerOffset [Trello](https://trello.com/c/duAwaupv)
 
 ## 12.15.0 - 2021/2/26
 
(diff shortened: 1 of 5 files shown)


kyryloonufriiev:
`getInnerOffset` is a combined function of `getTranslateOffset` and `getScrollOffset`. I think we need to wrap only the translate part. Since scrolling still could be there.

tool3:
> `getInnerOffset` is a combined function of `getTranslateOffset` and `getScrollOffset`. I think we need to wrap only the translate part. Since scrolling still could be there.

I think it makes more sense to simply not throw, and instead push `{x: 0, y: 0}` to the resulting array.
this will relieve the need to wrap every call to it in `trycatch` and reset to  o.

kyryloonufriiev (approved):
I think this is much better.

PR #400, opened 2021-03-10
eyes-images: expose and fix error in getScreenshot

tourdedave:
SIDE Eyes uses eyes-images under the hood for classic execution. In this scenario `getScreenshot` is ultimately calling `FullPageCaptureAlgorithm` which was recently changed to return a different object structure ([link](https://github.com/applitools/eyes.sdk.javascript1/commit/6412cea720605cf56c4ff0e4896f6b9350064a47)). This change was not propagated to eyes-images, which causes it to throw. There was no test to surface it which is why it went unnoticed when releasing eyes-images with the change. Currently, this error causes classic execution in SIDE Eyes to break (__in production__). I've confirmed that this patch fixes it.

re: [Trello 822](https://trello.com/c/6oVADkSf)

The error throws inside of the constructor in `eyes-sdk-core/lib/capture/EyesSimpleScreenshot`. Arguably, a better fix might be to make the change there (then it would be seamless to eyes-images). Open to that direction as well, but didn't look into the potential impacts of it.

### DIFF
--- a/packages/eyes-images/CHANGELOG.md
+++ b/packages/eyes-images/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- fix: update getScreenshot to support the correct return type when performing a full page stitch ([Trello](https://trello.com/c/6oVADkSf))
 
 ## 4.19.0 - 2021/2/22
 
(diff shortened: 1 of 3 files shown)


amitzur:
How come eyes-images tests didn't catch this? What was it that didn't propagate and to where?

amitzur:
Oh only now I understand that SIDE is using eyes-images to create a full page screenshot. That's so backwards. Of course eyes-images doesn't have tests for FPS, it's not part of functionality.

tourdedave:
Right. But, there is a function in eyes-images that passes the message to get the screenshot. It's part of its API, so it should be tested (even if only unit tested).

Assuming all goes well with running core in the browser we'll be able to drop SIDE Eyes' dependency on eyes-images and remove this functionality from eyes-images.

tourdedave:
Merging since this fix is needed to fix SIDE Eyes which is currently broken in production.

PR #404, opened 2021-03-16
Handle story parameters with circular reference better

amitzur:
Today, when a story has a circular reference in parameters (see below example) we show an error to the console that we ignore parameters for this story.
But there's no need to do this if the circular reference is unrelated to the eyes section of the parameters. We only care about the eyes section.
You could argue that we don't need to return the entire parameters object from the browser at all. That might be right. But it's both a big change in code, and also we don't lose anything from it, and we might want to use it in the future. We just need to not let it fail us.

```
const circular = {param: true}
circular.inner = circular

storiesOf('bla bla', module)
  .add('testing circular parameters', () => <div>nothing to see here</div>, {circular, eyes: {include: false}})
```

### DIFF
--- a/packages/eyes-storybook/CHANGELOG.md
+++ b/packages/eyes-storybook/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Handle story parameters with circular reference better
 
   ## 3.20.0 - 2021/3/17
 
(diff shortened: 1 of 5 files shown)


SophieDePaula:
Amit, the logic and changes looks good to me. The only change that I am not sure why it's necessary is adding `filename` and `framework` parameters separately for each component rather then doing that at the end

PR #405, opened 2021-03-18
fix(eyes-sdk-core): round element position

tool3:
[trello](https://trello.com/c/ZPnysSGY/826-wdio5-mobile-web-bottom-of-screenshot-blacked-out)

when using classic mode on mobile devices (for example: android running chrome) the resulting stitched image (when using scroll stitching specifically) was missing the final stitched region, and showed up as a black box on the bottom of the screenshot.

the solution was to properly `Math.round` the position of the element so that we piece the final stitched region - it does not contain fractions.

a note on the added test:
I had a challenge to test only `copyRasterData` - as it doesn't return anything and the only way to validate that it did work was to use the resulting image as a reference.

the test - creates a `MutableImage` from a local `.png` file and copies it to a new image created in the test.
since I needed to use fractions when creating the image, the stitched image was positioned a bit off, naturally, so I cropped it and compared the stitched image to the cropped image. if we remove `Math.round` from `copyRasterData` this test will fail.


kyryloonufriiev:
Looks good.
But I don't think you need to use `crop` there in order to compare images. To be honest I didn't even get how this test work. To understand it better I would need to understand how `MutableImage` works. Even if we put aside fact that I don't understand this test, this test depends on a crop function to work properly, and I would love to avoid as many dependencies as we can.

tool3:
@kyryloonufriiev was so kind to explain to me why I wasn't testing anything.

the test had a nice intention, but it lacked the fact that the original image wasn't cropped - so I essentially compared the same image to itself.

the test now reads from a local file (the cropped stitched image) - and compares it to the cropped image in the test - which will fail without this change.

PR #413, opened 2021-03-22
chore(deps): adding dependabot

tool3:
# adding dependabot

[trello](https://trello.com/c/e3cea3jO/833-fix-all-dependabot-and-audit-alerts)

this PR was created to handle the security and `npm audit` alerts that are triggered for our repo.
it introduces 2 files: 

1. `dependabot.yml` - initial dependabot config (under the `.github` directory).
2.  `generate-dependabot-config` - dependabot config generator in `sdk-shared` so we can better control what packages we want to continuously update.

since the dependabot UI doesn't offer much flexibility - and their monorepo detection is not working so well, I decided to create an entry per package (not only SDKs) in the `dependabot.yml` file, and have a single script that can generate this file with a skip list.

## ⚠️ warnings ⚠️

- it is important to understand that once we enable this, we should expect a short period of many PRs as our initial "ramp-up".

- all PRs opened by dependabot will initially fail - if we keep the `changelog` action running on every PR


amitzur:
@tool3 what workflow do you expect to have with the generator script?

packages/sdk-shared/src/generate-dependabot-config.js
```diff
@@ -0,0 +1,28 @@
...
+    - package-ecosystem: "npm"
+      directory: "/packages/_PACKAGE_"
+      schedule:
+        interval: "daily"`
+
+let template = `  version: 2\n  updates:`
+
+;(async () => {
+  try {
+    const {stdout: items} = await exec('ls packages')
```
amitzur, line 17: I expect to use `fs.readdir` here

packages/sdk-shared/src/generate-dependabot-config.js
```diff
@@ -0,0 +1,28 @@
...
+    const packages = items.split('\n')
+    for (const package of packages) {
```
amitzur, line 19: A note on coding style. I prefer we choose more functional style. This is important for readability and consistency.
In this case it is very clear:
1. filter out skip list from packages array
2. map filtered array with inline string template (no need for string replace)
3. join with `\n`

packages/sdk-shared/src/generate-dependabot-config.js
```diff
@@ -0,0 +1,28 @@
...
+      if (!skipList.includes(package)) {
+        template += `  ${packageTemplate.replace('_PACKAGE_', package)}`
+      }
+    }
+    await fs.writeFile(path.join(__dirname,'../../../.github/dependabot.yml'), template)
+  } catch (error) {
```
amitzur, line 25: why try and catch an error if you just throw it?

tool3:
> @tool3 what workflow do you expect to have with the generator script?

it was created so we don't forget to add new packages to the audit/dependency checks.
so we can run it periodically (and update master if there were changes) - possibly with a scheduled job - or simply run it as a part of our release procedure to make sure it is always up to date.

amitzur (changes requested):
See my question about workflow.
I want to work on coding style. Mainly, be more functional. But also not do things like try/catch and just rethrow. Or use `exec` for a standard filesystem operation. This is important. When reading code, a standard way of thinking makes it easy to understand the author's intent.

amitzur:
> it was created so we don't forget to add new packages to the audit/dependency checks.
so we can run it periodically (and update master if there were changes) - possibly with a scheduled job - or simply run it as a part of our release procedure to make sure it is always up to date.

@tool3 @kyryloonufriiev @tourdedave do you think we should start having a generator for packages? If so, I think that generator can add the new package to the yml file.

packages/sdk-shared/src/generate-dependabot-config.js
```diff
@@ -0,0 +1,18 @@
+const fs = require('fs').promises;
+const path = require('path');
+
+const skipList = ['eyes-universal-poc', 'eyes-leanft', 'eyes-images-legacy', 'eyes-sdk-core-legacy'];
+let template = `  version: 2\n  updates:`;
+
+;(async () => {
+    const items = await fs.readdir(path.join(__dirname, '../../'));
+    const packages = items.filter(package => !skipList.includes(package) && !package.startsWith('.'));
+    for (const package of packages) {
```
amitzur, line 10: What do you think about the following?
```
const packagesStr = items
  .filter(package => !skipList.includes(package) && !package.startsWith('.'))
  .map(package => `- package-ecosystem: "npm"
    directory: "/packages/${package}"
    schedule:
      interval: "daily"`)
  .join('\n')
const fileContent = `version: 2\n updates:\n${packagesStr}`
```

kyryloonufriiev:
> @tool3 @kyryloonufriiev @tourdedave do you think we should start having a generator for packages? If so, I think that generator can add the new package to the yml file.

@amitzur I absolutely agree that package generation script is a nice thing to have. Because we have a really similar structure of packages and all our tools rely on we keeping this structure. I just felt it by publishing a bunch of new packages.

packages/sdk-shared/src/generate-dependabot-config.js
tool3: this was changed to reflect something similar.
I say similar, because we don't really need to `map` it again and `join`.

since after `filter` I already have a list of packages to run on - I don't need to `join` it or do anything special with it, just run something per item :)

packages/sdk-shared/src/generate-dependabot-config.js
tool3: removed

packages/sdk-shared/src/generate-dependabot-config.js
amitzur: This `for` loop is exactly what I suggested to get rid of :)
It doesn't convey any meaning. There's no perf overhead in this case for doing `map` and `join`. And it passes a clear message what you are doing.
The `for` loop is imperative code. `map` and `join` is functional. In most cases we want to favor functional style since it's more semantic.

packages/sdk-shared/src/generate-dependabot-config.js
tool3: done!

packages/sdk-shared/src/generate-dependabot-config.js
tool3: done :)

packages/sdk-shared/src/generate-dependabot-config.js
```diff
@@ -0,0 +1,18 @@
+const fs = require('fs').promises;
+const path = require('path');
+
+const skipList = ['eyes-universal-poc', 'eyes-leanft', 'eyes-images-legacy', 'eyes-sdk-core-legacy'];
```
amitzur, line 4: eyes-common-legacy and eyes-selenium-3 should be skipped as well

tool3: done

PR #414, opened 2021-03-22
chore(release): save coverage test report

tool3:
[trello](https://trello.com/c/hL7aN2J2/828-save-coverage-test-results-in-github-actions)

this PR introduces saving the `coverage-test-report.xml` and `coverage-test-metadata.json` files for each release.

this was implemented in `Coverage test a package` action, with default retention.

since we run `release` more often than `Coverage test a package` - we need to decide on the retention period for these release runs.
`retention-days` - defaults to 90 days, if not set.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -62,6 +62,7 @@ jobs:
       - uses: actions/upload-artifact@v2
         with:
           name: coverage-tests-report
+          retention-days: 30
           path: |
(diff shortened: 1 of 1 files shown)


tool3:
since this has 0 impact on releasing and our workflow, I am merging this PR with a 30 day retention period.

if we feel this is too short, we can always increase/decrease according to our needs.

PR #418, opened 2021-04-01
Add api parameter to test object in bongo report

amitzur:
Currently tests with fluent vs. classic API are reported with exactly the same parameters to bongo, causing them to not be counted in the report.
This PR adds `api` entry to the metadata file, then after tests are executed it reads from the metadata file and populates it in the test object in the report.

### DIFF
--- a/packages/sdk-coverage-tests/CHANGELOG.md
+++ b/packages/sdk-coverage-tests/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


kyryloonufriiev (approved):
Looks great! Does bongo know about this parameter?

amitzur:
@kyryloonufriiev good question. bongo isn't aware of any parameter specifically, it (he?) does a general comparison (maybe with JSON.stringify)

PR #419, opened 2021-04-01
fix(sdk-coverage-tests): fixing custom coverage tests in report

tool3:
our coverage report was consistently missing the custom coverage tests.

the `convertJunitXmlToResultSchema` function in `sdk-coverage-tests` used to iterate over the json metadata file, which by definition does not contain any information about custom tests.

with this PR, we are running on *all* test results in the xml file - mapping any generic test to it's own object as before, but this time, including the custom coverage tests as well.

### DIFF
--- a/packages/sdk-coverage-tests/src/report/xml.js
+++ b/packages/sdk-coverage-tests/src/report/xml.js
@@ -4,11 +4,12 @@ const {logDebug} = require('../log')
 function convertJunitXmlToResultSchema({junit, browser, metadata}) {
   const tests = parseJunitXmlForTests(junit)
   logDebug(tests)
-  return Object.entries(metadata)
-    .map(([testName, testMeta]) => {
-      const testResult = tests.find(t => testName === parseBareTestName(t._attributes.name))
+  return tests
+    .map(testResults => {
+      const testName = parseBareTestName(testResults._attributes.name)
+      const testMeta = metadata[testName] || {name: testName, isGeneric: false}
       const isSkipped = testMeta.skip || testMeta.skipEmit || false // we explicitly set false to preserve backwards compatibility
-      if (!testResult && !isSkipped) return
+      if (!testMeta && !isSkipped) return
 
       return {
         test_name: testMeta.name || testName,
@@ -16,7 +17,7 @@ function convertJunitXmlToResultSchema({junit, browser, metadata}) {
           browser: browser || 'chrome',
           mode: testMeta.executionMode,
         },
-        passed: testResult && !isSkipped ? !testResult.failure : undefined,
+        passed: testResults && !isSkipped ? !testResults.failure : undefined,
         isGeneric: testMeta.isGeneric,
         isSkipped,
       }
--- a/packages/sdk-coverage-tests/src/report/xml.js
+++ b/packages/sdk-coverage-tests/src/report/xml.js
@@ -10,9 +10,9 @@ function convertJunitXmlToResultSchema({junit, browser, metadata}) {
     return acc
   }, metadata)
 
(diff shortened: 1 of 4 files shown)


kyryloonufriiev (changes requested):
I think now we will miss tests that exist in metadata, but not in xunit xml.

packages/sdk-coverage-tests/src/report/xml.js
```diff
@@ -4,11 +4,12 @@ const {logDebug} = require('../log')
...
-  return Object.entries(metadata)
-    .map(([testName, testMeta]) => {
-      const testResult = tests.find(t => testName === parseBareTestName(t._attributes.name))
+  return tests
+    .map(testResults => {
+      const testName = parseBareTestName(testResults._attributes.name)
+      const testMeta = metadata[testName] || {name: testName, isGeneric: false}
       const isSkipped = testMeta.skip || testMeta.skipEmit || false // we explicitly set false to preserve backwards compatibility
-      if (!testResult && !isSkipped) return
+      if (!testMeta && !isSkipped) return
```
kyryloonufriiev, line 12: How `testMeta` could be falsy if you explicitly set it two lines before.

packages/sdk-coverage-tests/src/report/xml.js
```diff
@@ -4,11 +4,12 @@ const {logDebug} = require('../log')
...
   const tests = parseJunitXmlForTests(junit)
   logDebug(tests)
-  return Object.entries(metadata)
-    .map(([testName, testMeta]) => {
-      const testResult = tests.find(t => testName === parseBareTestName(t._attributes.name))
+  return tests
+    .map(testResults => {
+      const testName = parseBareTestName(testResults._attributes.name)
+      const testMeta = metadata[testName] || {name: testName, isGeneric: false}
       const isSkipped = testMeta.skip || testMeta.skipEmit || false // we explicitly set false to preserve backwards compatibility
```
kyryloonufriiev, line 11: I think here we need to try to extract `skip` from XML as well. I know that our reporter doesn't add it, but I have a creative solution to understand that test was skipped. Skipped test in XML result has `time` property equals to `0`.

packages/sdk-coverage-tests/src/report/xml.js
```diff
@@ -4,25 +4,28 @@ const {logDebug} = require('../log')
...
-        parameters: {
-          browser: browser || 'chrome',
-          mode: testMeta.executionMode,
-          api: testMeta.api,
-        },
-        passed: testResult && !isSkipped ? !testResult.failure : undefined,
-        isGeneric: testMeta.isGeneric,
-        isSkipped,
-      }
+  return Object.entries(allTests).reduce((acc, [testName, testMeta]) => {
```
kyryloonufriiev, line 13: You don't really need `reduce` here, just use `map` instead.

packages/sdk-coverage-tests/src/report/xml.js
```diff
@@ -4,25 +4,28 @@ const {logDebug} = require('../log')
...
+    const isSkipped = testMeta.skip || testMeta.skipEmit || false // we explicitly set false to preserve backwards compatibility
+    acc.push({
+      test_name: testMeta.name || testName,
+      parameters: {
+        browser: browser || 'chrome',
+        mode: testMeta.executionMode,
```
kyryloonufriiev, line 19: what will be here for custom tests? `undefined`? Is it allowed?

packages/sdk-coverage-tests/src/report/xml.js
tool3: done thanks !

PR #421, opened 2021-04-13
feat(eyes-storybook): fake ie

tool3:
# Fake IE in storybook

[trello](https://trello.com/c/P0SeWKRO)

this PR introduces the ability to fake IE browsers in storybook.
we do this by setting a custom `documentMode` and `userAgent`.

### `pagePool.drain`
this method was added to loop over the existing pages in the pool and fake them as IE.
a comment was added to clarify that this API should only be used at a point where you know that no pages are currently rendering.

### `pagePool.removeAndAddPage`
this method was added to handle the logic of removing and readding a page from the pool after it has been closed.

### `initPage`
extracted `page.close` event listener to pagePool to be used both in `initPage` and in the `pagePool.drain` method.

## config changes
a new flag called `fakeIE` was added to opt-in to the fake IE feature.
if no IE browsers are in your configuration AND you set this flag to `true` - a warning message will be shown.   

image


packages/eyes-storybook/test/e2e/snapshots/eyes-storybook.e2e.test.json
```diff
@@ -0,0 +1,98 @@
+{
+  "stdout": [
+    "Using @applitools/eyes-storybook version 3.20.2.",
```
amitzur, line 3: so with every release we will need to update this file?
Maybe snaptdout should support not only `.json` but also dynamic `.js`?

packages/eyes-storybook/test/e2e/TestAccessibility.e2e.test.js
```diff
@@ -41,9 +42,11 @@ describe('eyes-storybook accessibility', () => {
     const globalConfig = require(configPath);
     const defaultConfig = {waitBeforeScreenshots: 50};
     const config = generateConfig({argv: {conf: configPath}, defaultConfig, externalConfigParams});
+
     let results = await eyesStorybook({
       config: {
         storybookUrl: 'http://localhost:9001',
+        browser: [{name: 'chrome', width: 800, height: 600}],
```
amitzur, line 49: is this added because it's now mandatory? I don't mind it being here, just making sure no existing user API got broken

packages/eyes-storybook/src/validateAndPopulateConfig.js
```diff
@@ -55,6 +56,16 @@ async function validateAndPopulateConfig({config, packagePath, logger}) {
       config.puppeteerOptions.args.push('--disable-dev-shm-usage');
     }
   }
+
+  if (config.fakeIE) {
```
amitzur, line 60: no need for nested `if`. Better to use `&&`.

packages/eyes-storybook/src/renderStories.js
```diff
@@ -129,8 +133,12 @@ function makeRenderStories({
       allTestResults.every(didTestPass) ? spinner.succeed() : spinner.fail();
     }
 
+    function updateSpinnerText(number, length) {
+      return `Done ${number} stories out of ${length} ${hasIE(config) ? '(IE)' : ''}`;
```
amitzur, line 137: this is not correct logic. `hasIE` doesn't check if config has `fakeIE`. I don't think you need `hasIE` at all. Only `shouldRenderIE`, which is what should be used in this line as well.
How come no test failed because of this?

amitzur (commented):
Review is WIP.
I went over most of the things. There are some issues that need addressing.
I expect the second part of the review to contain more requests for change than this one (I started from the easier things :) ).

packages/eyes-storybook/src/fakeIE.js
```diff
@@ -0,0 +1,16 @@
+async function fakeIE({logger, page}) {
+  logger.verbose('[fakeIE] - faking IE');
```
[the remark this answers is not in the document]
tool3: I don't have access to the full page object from `initPage`, but I can pass in the `pageId`

packages/eyes-storybook/src/initPage.js
```diff
@@ -18,29 +21,37 @@ function makeInitPage({iframeUrl, config, browser, logger}) {
         },
       });
     }
+
+    if (!getRenderIE()) {
```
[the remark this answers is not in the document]
tool3: no, it only means that for IE it's handled elsewhere ... in `pagePool.drain` as explained the in PR description.

packages/eyes-storybook/src/renderStories.js
tool3: I am not sure I understand. `renderStories` always gets called with separated configs.
`hasIE` checks for the existence of `IE` in your config browsers.
`shouldRenderIE` also validates the flag. which is being handled in `executeRenders`

packages/eyes-storybook/src/validateAndPopulateConfig.js
tool3: done

packages/eyes-storybook/test/e2e/snapshots/eyes-storybook.e2e.test.json
tool3: I will replace it with current version from `package.json` - maybe worth adding this to `snaptdout` later on

packages/eyes-storybook/test/e2e/TestAccessibility.e2e.test.js
tool3: yup, it's mandatory

packages/eyes-storybook/src/renderStories.js
amitzur: Consider just a regular case of executing the SDK with IE in the browser configuration, but without the fakeIE flag. What will the output be?

tool3: gotcha! fixing !

packages/eyes-storybook/src/initPage.js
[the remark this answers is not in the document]
amitzur: but in `pagePool.drain` you are not listening on `close` event of the page.

packages/eyes-storybook/test/e2e/TestAccessibility.e2e.test.js
amitzur: but until now it wasn't mandatory. So an API broke

packages/eyes-storybook/src/initPage.js
```diff
@@ -18,29 +21,33 @@ function makeInitPage({iframeUrl, config, browser, logger}) {
...
-    page.on('error', async err => {
-      logger.log(`Puppeteer error for page ${pageId}:`, err);
-      pagePool.removePage(pageId);
-      const {pageId: newPageId} = await pagePool.createPage();
-      pagePool.addToPool(newPageId);
-    });
+
     page.on('close', async () => {
-      if (pagePool.isInPool(pageId)) {
+      if (!getRenderIE() && pagePool.isInPool(pageId)) {
```
amitzur, line 26: In continuation to another thread on this line - can you explain why this is needed?

packages/eyes-storybook/src/pagePool.js
```diff
@@ -2,11 +2,12 @@
 
 function createPagePool({initPage, logger}) {
   let counter = 0;
-  const fullPageObjs = [];
+  let fullPageObjs = [];
```
amitzur, line 5: where do you reassign this variable that it needs to be `let`?

packages/eyes-storybook/src/shouldRenderIE.js
```diff
@@ -0,0 +1,43 @@
...
+
+function validateBrowsers(config) {
+  if (config.browser) {
+    return Array.isArray(config.browser) ? config.browser : [config.browser];
+  } else {
+    return [];
+  }
+}
+
+module.exports = {splitConfigsByBrowser, hasIE, shouldRenderIE};
```
amitzur, line 43: no need to export `hasIE`. it's not used outside so it's confusing

packages/eyes-storybook/test/e2e/TestAccessibility.e2e.test.js
```diff
@@ -23,15 +24,13 @@ describe('eyes-storybook accessibility', () => {
...
-  after(async () => {
-    await closeStorybook();
-  });
 
   let closeTestServer;
   before(async () => {
     closeTestServer = (await testServer({port: 7272})).close;
   });
   after(async () => {
+    await closeStorybook();
```
amitzur, line 33: there were 2 before/after blocks, each one responsible for a different thing. If you think it makes sense to merge them - please do it symmetrically also to the `before` block.

packages/eyes-storybook/test/unit/shouldRenderIE.test.js
```diff
@@ -0,0 +1,81 @@
+'use strict';
+const {describe, it} = require('mocha');
+const {expect} = require('chai');
+const {splitConfigsByBrowser, hasIE, shouldRenderIE} = require('../../src/shouldRenderIE');
+
+describe('shouldRenderIE', () => {
+  describe('splitConfigsByBrowser', () => {
```
amitzur, line 7: I think it's also worth having a test for a config that has only IE. I was looking for this test to see what the output would be.

packages/eyes-storybook/test/unit/shouldRenderIE.test.js
```diff
@@ -0,0 +1,81 @@
...
+    it('should work with object', () => {
+      const config = {
+        browser: {name: 'chrome'},
+      };
+
+      expect(splitConfigsByBrowser(config)).to.eql([{browser: [{name: 'chrome'}]}]);
+    });
+  });
+
+  describe('hasIE', () => {
```
amitzur, line 46: if you stop exporting this function then you can remove its tests as well (see comment above)

packages/eyes-storybook/test/e2e/snapshots/fake-ie.e2e.test.json
```diff
@@ -0,0 +1,6 @@
+{
+  "fake ie": [
+    "Fake IE: Fake IE Page [Chrome 90.0] [800x600] - Passed",
```
amitzur, line 3: is this order deterministic?

packages/eyes-storybook/src/eyesStorybook.js
```diff
@@ -120,45 +123,42 @@ async function eyesStorybook({
...
+      stories: storiesIncludingVariations,
     });
 
     logger.log('finished creating functions');
 
-    const [error, results] = await presult(
-      timeItAsync('renderStories', () => renderStories(storiesIncludingVariations)),
-    );
+    const configs = splitConfigsByBrowser(config);
+    const results = await executeRenders(configs);
```
amitzur, line 147: there is a big bug here. if there is an error you throw it from `executeRenders`, and will not reach the `closeBatch` call below. Better return the result of `presult` from `executeRenders` and handle the exception throwing here like it was before.

packages/eyes-storybook/src/executeRenders.js
```diff
@@ -0,0 +1,35 @@
+const {refineErrorMessage} = require('./errMessages');
+const {presult} = require('@applitools/functional-commons');
+const {shouldRenderIE} = require('./shouldRenderIE');
+
+function makeExecuteRenders({timeItAsync, renderStories, pagePool, stories, logger, setRenderIE}) {
```
amitzur, line 5: I don't think the `makeExecuteRenders` function should receive `stories`. It makes more sense to be passed to `executeRenders`.
Actually, why do you need to have a `makeExecuteRenders` at all? You "make" the function just one line before calling it. You might as well have just a `executeRenders` function with the union of arguments.

packages/eyes-storybook/test/unit/executeRenders.test.js
```diff
@@ -0,0 +1,141 @@
...
+      timeItAsync: (_a, b) => b(),
+      renderStories: async function(stories, config) {
+        Object.assign(results[counter], {stories, config});
+        return stories;
+      },
+      pagePool: {},
+      stories: [{hello: 'world'}],
+      logger: {
+        verbose: function(log) {
+          counter++;
```
amitzur, line 19: this is super fragile. your test will brake if a log entry is added to `executeRenders`!

packages/eyes-storybook/test/unit/executeRenders.test.js
```diff
@@ -0,0 +1,141 @@
...
+          results[counter] = {log};
+        },
+      },
+      setRenderIE: {},
```
amitzur, line 23: this should be a function, for clarity

amitzur (changes requested):
Only thing left in the review is a deeper look at the tests for `executeRenders`.

packages/eyes-storybook/src/pagePool.js
tool3: at some point we thought about simply overriding the `fullPageObjs` array, but I never set it back to `const` - fixing.

packages/eyes-storybook/src/shouldRenderIE.js
tool3: removing thanks

packages/eyes-storybook/test/e2e/TestAccessibility.e2e.test.js
tool3: sorry ! didn't notice that other before at all (temporary blindness - is that a thing ?) - fixing

packages/eyes-storybook/test/unit/shouldRenderIE.test.js
tool3: added

packages/eyes-storybook/test/unit/shouldRenderIE.test.js
tool3: removed this test case

packages/eyes-storybook/test/e2e/snapshots/fake-ie.e2e.test.json
tool3: yes - we separate the config specifically to have the fake IE pages last - because of what we discussed of avoiding the need to reset the useragent and document mode.

packages/eyes-storybook/src/eyesStorybook.js
tool3: extracted to `eyesStorybook` - great catch ! thank you !

packages/eyes-storybook/test/unit/executeRenders.test.js
tool3: updated

packages/eyes-storybook/src/executeRenders.js
tool3: I removed the `makeExecuteRender` and consolidated it to a single function

packages/eyes-storybook/test/unit/executeRenders.test.js
tool3: updated the test to contain an array of logs, and updated the test to match the new single function design

packages/eyes-storybook/src/initPage.js
tool3: we want to retain the option to still run the logic for when a page closes unexpectedly - but at the same time - we don't want to do that when running the IE stories - because we handle it differently when it's a fake IE page.

when the page is set to be IE - we do `page.close` before running the logic of removing and readding the page to the pool - otherwise we will only create new pages and the run will just hang.

packages/eyes-storybook/src/eyesStorybook.js
amitzur: I don't see it extracted back there

packages/eyes-storybook/src/initPage.js
amitzur: we **do** want to have the listener when running IE stories. We just want to remove it (or just not perform the recovery) while draining the pool.

amitzur (changes requested):
1. Still not fixed calling `closeBatch` event in case of error in `renderStories`.
2. Need to listen to `close` event on page also while running IE stories.

PR #422, opened 2021-04-19
WDIO 7 support

kyryloonufriiev:


kyryloonufriiev:
Support for WDIO with Puppeteer will be added once this [PR](https://github.com/webdriverio/webdriverio/pull/6725) will be released. Unfortunately, it means that users could use WDIO + Puppeteer + VG only with the latest version of WDIO.

kyryloonufriiev:
I have found another issue that should be solved in this [PR](https://github.com/webdriverio/webdriverio/pull/6751). This was the cause of one failed test on our side, but we probably didn't have enough coverage to catch this. I have added a spec-driver test which will fail until it will be fixed on the WDIO side since it is really hard to overcome on our side.

PR #423, opened 2021-04-22
fix: add iosVersion optional type

tool3:
[trello](https://trello.com/c/B71uljbi)

Warner are using typescript with `testcafe`, and I found a missing type in the eyes open options (in `visual-grid-client`) which caused a type error for the customer.

I also turned `accessibilityValidation` (in `eyes-testcafe`) to optional - since it's not required and will create a type error.


kyryloonufriiev:
💯

PR #424, opened 2021-04-22
fix(eyes-cypress): remove need for `experimentalRunEvents` flag in later cypress versions

tool3:
[trello](https://trello.com/c/2tluhwWN)

in #384 we provided a way to use cypress' global hooks which was introduced in version `6.2.0` using a flag - `experimentalRunEvents`.

this PR addresses the fact that this flag is no longer needed in cypress >= `6.7.0`

### conditions table
| version        | mode/flag           | hooks  |
| ------------- |:-------------:| -----:|
| `>=6.2.0`      | `experimentalRunEvents` flag | global |
| `>=6.7.0`      | run mode      |   global |
| any                | interactive mode   | non-global

### DIFF
--- a/packages/eyes-cypress/CHANGELOG.md
+++ b/packages/eyes-cypress/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- remove the need for `experimentalRunEvents` in cypress >= `6.7.0`
 
 ## 3.21.0 - 2021/4/22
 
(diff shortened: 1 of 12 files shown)


tool3:
I have moved the check for interactive mode inside the `before:run` hook.

that is the only place where cypress exposes information about its run mode (`isTextTerminal`).

this will ensure that when we get to `commands` in the browser - we don't do a double batch.

PR #585, opened 2021-05-02
chore(visual-grid-client): support multiple bcs hooks

tool3:
[trello](https://trello.com/c/JYn99gKJ)
[trello](https://trello.com/c/GMWV2GAw)

this PR introduces the ability to run a before capture screenshot hook per browser.

previously, you could only specify a single bcs hook for all browsers in the customer's configuration.

### API
before:
```javascript
{
  scriptHooks: {
    beforeCaptureScreenshot: 'script that will run before each browser render'
  }
}
```
after:
```javascript
{
  scriptHooks: {
    beforeCaptureScreenshot: { 
      chrome: 'script that will run before rendering on chrome',
      firefox: 'script that will run before rendering on firefox',
      default: 'script that will run before rendering all other browsers'
    } 
  }
}
```

### DIFF
--- a/packages/visual-grid-client/CHANGELOG.md
+++ b/packages/visual-grid-client/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-
+- adding multiple before capture screenshot hooks support
 ## 15.8.2 - 2021/4/26
 
 - rename `variantId` to `variationGroupId`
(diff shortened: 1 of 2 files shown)


kyryloonufriiev:
What about API for core-based SDKs?

tool3:
> What about API for core-based SDKs?

we will wait for `eyes-api` package and implement this logic there. 
the other option is to implement it now in core, but since this is not an urgent feature, we can wait.

amitzur:
@tool3 @kyryloonufriiev there were also different implementation details that I spoke with Tal regarding the API that vgc exposes. We can talk about it f2f2

kyryloonufriiev:
Closing this PR, and moving discussion on the API to the [Trello](https://trello.com/c/dhdxaODs/40-support-multiple-bcs-hooks)

PR #593, opened 2021-05-14
Get and set renderInfo separately from eyesOpen

SophieDePaula:
Add the option to get and set rendering info separately from eyesOpen. This will also validate the response from getRenderInfo and will throw the appropriate exception.
This is in preparation to [gracefully fail Storybook tests when using wrong API key ](https://trello.com/c/NFmgtYtl/50-storybook-there-should-be-a-proper-output-in-case-of-401-authentication-error?menu=filter&filter=due:notdue)


amitzur (approved):
I added a commit which addresses some points:
1) puppeteer browser was not closed in case of invalid API key. This is the reason that the new test in eyesStorybook hangs. I moved it down to be inside the try/catch block.
2) `_req` renamed to `req`. Only reason to name a variable with underscore is if it's not used, to avoid the eslint warning.
3) `testServer` was not really needed in the vgc test

PR #594, opened 2021-05-17
Universal client reference

amitzur:
Placeholder PR, I'll add notes later.


packages/eyes-universal-poc/reference/src/spec-driver.js
```diff
@@ -0,0 +1,142 @@
...
+
+    // #region HELPERS
+    function transformSelector(selector) {
+      // if (TypeUtils.has(selector, ['type', 'selector'])) {
+      if (typeof selector === 'object' && 'type' in selector) {
+        if (selector.type === 'css') return `css=${selector.selector}`
+        else if (selector.type === 'xpath') return `xpath=${selector.selector}`
+      }
+      return selector
+    }
```
kyryloonufriiev, line 28: This looks a bit confusing to me. A constructor with functions inside... I think a better example would be to use instance methods since otherwise the whole point of using class is evaporated.

kyryloonufriiev:
I see you decided to go with `Agent` instead of `Runner` or `Client` and I like it. But do you still want to have `Agent.new` or rather place it under `Core` namespace (e.g. `Core.newAgent` or `Core.makeAgent`). I think this will be correlated with the way how `Eyes` are created from an `Agent` by calling `Agent.openEyes` and also it will be a better reflection of our internal core API. Also, the chain of abstractions will be pretty clear `Core` -> `Agent` -> `Eyes`

packages/eyes-universal-poc/reference/src/spec-driver.js
amitzur: You're right. This was just a shortcut for me to get it working.

amitzur:
@kyryloonufriiev The reason I skipped `Core` was that it seemed like an unnecessary level of indirection. I may be wrong. But the API I ended up with when trying it out, looked like this:
```
const client = UniversalClient.connectToServer()
const core = client.makeCore()
const agent = await core.makeAgent()
const eyes = await agent.makeEyes()
await eyes.open()
```

But why is this better than this?
```
const client = UniversalClient.connectToServer()
const agent = await client.makeAgent()
const eyes = await agent.makeEyes()
await eyes.open()
```
Note 2 things:
1. `client.makeCore()` isn't async, as opposed to `core.makeAgent()` and `core.makeEyes()`.
2. Each of these classes has a constructor that is passed the specDriver, refer, socket and a server ref. I need to create yet another class, which didn't make sense because it doesn't have a state of its own (no server ref).

kyryloonufriiev:
You have been understand me wrong . I don't say you that we need to have `makeCore` method. In my opinion `UniversalClient` is the `Core` here. But when you do `universalClient.makeAgent`, ws command should be `Core.makeAgent` and not `Agent.new`. It will preserve consistency across all of the layers.

amitzur:
Aha, yes. I totally agree

PR #638, opened 2021-05-23
fix(eyes-api): add properties key to config

tool3:
[trello](https://trello.com/c/5UA5T3uI)

the `Configuration` class in eyes-api had a method `addProperty` which pushed a property to the `properties` array. unfortunately, the `_config` local private variable was initiated as an empty object - without first creating an empty `properties` array.

### DIFF
--- a/packages/eyes-api/src/input/Configuration.ts
+++ b/packages/eyes-api/src/input/Configuration.ts
@@ -104,7 +104,8 @@ export type Configuration<TElement = unknown, TSelector = unknown> = GeneralConf
   VGConfiguration
 
 export class ConfigurationData<TElement = unknown, TSelector = unknown>
-  implements Required<Configuration<TElement, TSelector>> {
+  implements Required<Configuration<TElement, TSelector>>
+{
   protected readonly _spec: ConfigurationSpec<TElement, TSelector>
 
   private _config: Configuration<TElement, TSelector> = {}
@@ -512,7 +513,7 @@ export class ConfigurationData<TElement = unknown, TSelector = unknown>
     this._config.properties = properties
   }
   getProperties(): PropertyDataData[] {
-    return this.properties?.map((property) => new PropertyDataData(property)) ?? []
+    return this.properties?.map(property => new PropertyDataData(property)) ?? []
   }
   setProperties(properties: PropertyData[]): this {
     this.properties = properties
@@ -920,15 +921,11 @@ export class ConfigurationData<TElement = unknown, TSelector = unknown>
     return this._config.scrollRootElement
   }
   set scrollRootElement(scrollRootElement: TElement | TSelector) {
-    utils.guard.custom(
-      scrollRootElement,
-      (value) => this._spec.isElement(value) || this._spec.isSelector(value),
-      {
-        name: 'scrollRootElement',
-        message: 'must be element or selector',
-        strict: false,
-      },
-    )
+    utils.guard.custom(scrollRootElement, value => this._spec.isElement(value) || this._spec.isSelector(value), {
+      name: 'scrollRootElement',
+      message: 'must be element or selector',
+      strict: false,
+    })
     this._config.scrollRootElement = scrollRootElement
   }
   getScrollRootElement(): TElement | TSelector {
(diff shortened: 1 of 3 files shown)


packages/eyes-api/src/input/Configuration.ts
```diff
@@ -107,7 +107,7 @@ export class ConfigurationData<TElement = unknown, TSelector = unknown>
   implements Required<Configuration<TElement, TSelector>> {
   protected readonly _spec: ConfigurationSpec<TElement, TSelector>
 
-  private _config: Configuration<TElement, TSelector> = {}
+  private _config: Configuration<TElement, TSelector> = {properties: []}
```
kyryloonufriiev, line 110: I don't think handling it like this is the clearest solution. I just don't want to see here a bunch of empty objects and arrays. I would rather solve it in the same way as it is already solved for `browserInfo` property. Check it out.

kyryloonufriiev (changes requested):
You can fix the one thing I have mentioned and merge it. Also please update your prettier or something, because I see that you have fixed some linting problems, but those are not problems if you use the latest prettier configured from the `.eslintrc` file.

packages/eyes-api/src/input/Configuration.ts
tool3: I already made that change but haven't pushed it apparently

tool3:
this opened up a bigger discussion on things like `BrowserName` which only accepts enum values, and therefore will give a type error for customers who are using strings like `chrome`.

there is a solution to this, but unfortunately, it doesn't translates the type definition correctly.

for example:
```typescript
type valueof<T> = T[keyof T]
```

translates to `undefined` when running through our type definition generator for some reason.

packages/eyes-api/src/input/Configuration.ts
```diff
@@ -512,7 +512,7 @@ export class ConfigurationData<TElement = unknown, TSelector = unknown>
     this._config.properties = properties
   }
   getProperties(): PropertyDataData[] {
-    return this.properties.map(property => new PropertyDataData(property))
+    return this.properties.map((property) => new PropertyDataData(property))
```
kyryloonufriiev, line 515:
```suggestion
    return this.properties?.map(property => new PropertyDataData(property)) ?? []
```

kyryloonufriiev:
@tool3 Don't worry about it. I have already fixed it. In `feat/universal` I'm planning to fix more things tomorrow and release them all together.

packages/eyes-api/src/input/Configuration.ts
```diff
@@ -522,7 +522,6 @@ export class ConfigurationData<TElement = unknown, TSelector = unknown>
   addProperty(prop: PropertyData): this
   addProperty(propOrName: PropertyData | string, value?: string): this {
     const property = utils.types.isString(propOrName) ? {name: propOrName, value} : propOrName
-    if (!this.properties) this.properties = []
```
kyryloonufriiev: Why?

tool3: I think it makes more sense to handle all initializations in the constructor and avoid multiple instantiations for different functions/cases.

kyryloonufriiev: But you are not initializing anything in the constructor. I also not agree about handling something like this in the constructor. I understand that it might be a little more optimal this way, but I don't really like this idea that we will have a big piece of code in constructor which will affect all getters and setters in this large file.
I also would like the class to work just the same as the plain object, and in the plain object, we don't have pre-initialized fields.
I will try to explain why it seems important to me. The `Configuration` class is just a legacy variant of the `ConfigurationPlain` type. The class implements the plain type, to allow us to forget about class stuff and work with an instance of the class just the same way as we work with the plain object. And when we work with the plain object we always have to assume that field could be missed, so it doesn't make a lot of sense to add an empty array in the class constructor, because later we anyway will assume and behave like this field could be empty.

tool3: I didn't perform the changes yet, I wanted to explain my point of view.

Maybe I misunderstood you, but I disagree that we should always assume a property is empty, it makes no difference if it's a class or a plain object.

If you always treated props as possibly non-existant - wouldn't that bloat the code with endless ternary expressions and if statements? 

What if we have some new method that performs some action on `properties`? Will we also check if it doesn't exist there? That would probably bloat the code more than checking for the few properies that needs initilizations no?

kyryloonufriiev: We don't have other choices, we have to assume that any property in configuration object could be missed. Because the user can pass it just as an object and not a data class. We obviously don't want to make user pass empty array in a property he/she doesn't need. We need to keep in mind that normally there should not be more methods in this class because we want to make plain objects a recommended way to use our sdks.

I understand your point. It totally makes sense for me. My only motivation is to keep this ugly code (all of the methods in this class and the data class itself) in a single place where it will be easy to read, without scrolling this file up and down to understand what this going on. And the most important is consistency, if we decide to go with other solution I would love to see it consistent at least with this class.

tool3: I absolutely agree with you buddy. Theae classes are indeed legacy in our minds, and as you said, I was mainly concerned with missing these kinds of things, and to make workflow more easier. But I agree that we strive for objects and what you say totally makes sense.

I did want to bring up type assertions, but that's for tomorrow :)

PR #647, opened 2021-05-25
feat(scripts): eyes compare

tool3:
this PR introduces some changes to `yarn render`, primarily - the ability to run with the `--compare` flag which will run a visual test on a given url - create a baseline using a classic runner, and then run another visual test using the ultrafast grid against that same baseline.

### output improvements
image

### aliases
I have added some aliases for the most used options (such as `b` for `browser` and `v` for `vg` etc...)

### new flags
the `sdk` flag now allows you to provide an sdk name to use in the render.

### DIFF
--- a/packages/scripts/cli/render.js
+++ b/packages/scripts/cli/render.js
@@ -14,10 +14,12 @@ const checkConfig = {
   name: {
     describe: 'tag for checkpoint',
     type: 'string',
+    alias: 'n',
     default: 'render script',
   },
   region: {
     describe: 'region to check',
+    alias: 'r',
     type: 'string',
     coerce(str) {
       const region = utils.parseRegion(str)
@@ -88,6 +90,7 @@ const checkConfig = {
 const buildConfig = {
   browser: {
     describe: 'preset name of browser. (e.g. "edge-18", "ie-11", "safari-11", "firefox")',
+    alias: 'b',
     type: 'string',
     default: 'chrome',
   },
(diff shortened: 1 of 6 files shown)


kyryloonufriiev:
💯 🎉

packages/test-utils/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- fix require for `yarn render`
```
kyryloonufriiev, line 5: I was forced to do just the same change yesterday, so this is already done and published.
```suggestion
```

packages/scripts/cli/render.js
```diff
@@ -436,34 +473,37 @@ function formatUrl(url) {
...
+  const {width, height} = testResults.getHostDisplaySize()
+  const outputTable = new table({ style: { head: [], border: [] }, chars: { 'mid': '', 'left-mid': '', 'mid-mid': '', 'right-mid': '' } })
+  outputTable.push(['name', testResults.getName()])
+  outputTable.push(['status', testResults.getStatus()])
+  outputTable.push(['url', testResults.getUrl()])
+  outputTable.push(['total steps', testResults.getSteps()])
+  outputTable.push(['matches', testResults.getMatches()])
+  outputTable.push(['diffs', testResults.getMismatches()])
+  outputTable.push(['missing', testResults.getMissing()])
+  outputTable.push(['viewport', `${width}x${height}`])
```
kyryloonufriiev, line 500: no need to do this since I have added `.toString` impl to `RectangleSize` class

tool3: nice ! will change it to `toString`

PR #648, opened 2021-05-25
fix(eyes-sdk-core): browser info

tool3:
[trello](https://trello.com/c/EX9VYn1s)

fixing a situation where customers provide the following browser configuration object:
```javascript
 {
      deviceName: 'iPhone X',
      screenOrientation: 'landscape',
      name: 'chrome' 
  }
```

we give this example in our documentation (cypress, testcafe, storybook) to illustrate that this is a chrome emulation and not an ios simulation.

### DIFF
--- a/packages/eyes-sdk-core/CHANGELOG.md
+++ b/packages/eyes-sdk-core/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- fix browser configuration for emulated device with `name` property
 
 ## 12.20.0 - 2021/5/24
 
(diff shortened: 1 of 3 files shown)


amitzur:
@tool3 worth thinking how to consolidate this with a function called `fillDefaultBrowserName` in Cypress and a function called `isEmulation` in visual-grid-client.
There are also `mapChromeEmulationInfo` and `getDeviceInfoFromBrowserConfig` in visual-grid-client, which are also related but I'm not sure they should be consolidated.

PR #650, opened 2021-06-02
docs(eyes-storybook): documenting runBefore and runAfter limitations

tool3:
[trello](https://trello.com/c/A160yfTX)

this PR adds a warning about `runBefore` and `runAfter` functions - emphasizing side-effects.

### DIFF
--- a/packages/eyes-storybook/README.md
+++ b/packages/eyes-storybook/README.md
@@ -671,7 +671,7 @@ storiesOf('Components with ignoreDisplacements', module)
 
 The `runBefore` and `runAfter` functions serve as ways to interact with the page prior to taking the story screenshot.
 > ⚠️ Warning ⚠️   
-> the `rootEl` argument being passed to this function might be the same for several stories - this means that if your function creates side effects - it may affect other, non-related stories.
+> the `rootEl` argument being passed to these function might be the same for several stories - this means that if your function creates side effects - it may affect other, non-related stories.
 
 #### `runBefore`
 
--- a/packages/eyes-storybook/README.md
+++ b/packages/eyes-storybook/README.md
@@ -669,9 +669,10 @@ storiesOf('Components with ignoreDisplacements', module)
 
 ### `runBefore` and `runAfter` functions
 
-The `runBefore` and `runAfter` functions serve as ways to interact with the page prior to taking the story screenshot.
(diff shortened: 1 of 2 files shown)


packages/eyes-storybook/README.md
```diff
@@ -665,9 +665,15 @@ storiesOf('Components with ignoreDisplacements', module)
...
 ```
 
-## Parameters that cannot be set as an [Advanced configuration](#advanced-configuration)
+## Parameters that cannot be set as an [advanced configuration](#advanced-configuration)
 
-### `runBefore`
+### `runBefore` and `runAfter` functions
+
+The `runBefore` and `runAfter` functions serve as ways to interact with the page prior to taking the story screenshot.
+> ⚠️ Warning ⚠️
```
amitzur, line 673: That's a bit harsh. We provide the `runAfter` API exactly for addressing this culprit. So it's not exactly a warning, but rather guidance for using the API correctly.

packages/eyes-storybook/README.md
```diff
@@ -665,9 +665,15 @@ storiesOf('Components with ignoreDisplacements', module)
...
 
-## Parameters that cannot be set as an [Advanced configuration](#advanced-configuration)
+## Parameters that cannot be set as an [advanced configuration](#advanced-configuration)
 
-### `runBefore`
+### `runBefore` and `runAfter` functions
+
+The `runBefore` and `runAfter` functions serve as ways to interact with the page prior to taking the story screenshot.
+> ⚠️ Warning ⚠️   
+> the `rootEl` argument being passed to these function might be the same for several stories - this means that if your function creates side effects - it may affect other, non-related stories.
```
amitzur, line 674: The problem is not only with `rootEl`. It's any global side-effect that doesn't get cleaned when the next story is rendered. I suggest the following (for this entire section):

---

The `runBefore` function can be used to perform any action prior to taking the story snapshot. The `runAfter` method should be used to cleanup any side-effect that `runBefore` creates, if there is one. For example, if `runBefore` addsa class to the `body` element, this class should be removed in `runAfter`. This is because the browser tab's window is not reloaded between stories.

_Note: `rootEl` also needs to be cleaned up, so any modification that was done on this element in `runBefore` should be reverted in `runAfter`._

PR #651, opened 2021-06-06
fix(eyes-sdk-core): use `require.resolve` instead of cosmiconfig

tool3:
[trello](https://trello.com/c/e6wSt1au)

this PR removes the dependency of `cosmiconfig` and uses `require.resolve` to get the config file from the customer.

due to the nature of how `cosmiconfig` works - it will return `null` if the config wasn't found/corrupted - which we don't currently handle.

I have preserved all the same config file names (except `package.json`) and will `throw` if a config file wasn't found.

`require.resolve` is the same API that node itself uses to resolve modules.

### DIFF
--- a/packages/eyes-sdk-core/lib/utils/ConfigUtils.js
+++ b/packages/eyes-sdk-core/lib/utils/ConfigUtils.js
@@ -59,6 +59,8 @@ function findConfigFile(possibleConfigs, errors = []) {
     try {
       return require.resolve(`./${configFile}`, {paths: [process.cwd()]})
     } catch (error) {
+      // save errors to continue checking for other config files
+      // throw the earliest encountered error
       errors.push(error)
       if (index === possibleConfigs.length - 1) {
         throw errors[0]
--- a/packages/eyes-sdk-core/lib/utils/ConfigUtils.js
+++ b/packages/eyes-sdk-core/lib/utils/ConfigUtils.js
@@ -9,20 +9,20 @@ function getConfig({
   logger = new Logger(!!process.env.APPLITOOLS_SHOW_LOGS),
 } = {}) {
   let defaultConfig = {}
+  const possibleConfigs = ['applitools.config.js', 'package.json', 'eyes.config.js', 'eyes.json']
   try {
-    const fullConfigPath = configPath ? `${configPath}/applitools.config.js` : undefined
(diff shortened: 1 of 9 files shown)


packages/eyes-sdk-core/lib/utils/ConfigUtils.js
```diff
@@ -9,24 +8,58 @@ function getConfig({
...
+      logger.log('Loading configuration from', customConfigPath)
+      const config = require(customConfigPath)
+      return Object.assign(config, envConfig)
+    } else {
+      const config = require(findConfigFile(possibleConfigs))
+      return Object.assign(config, envConfig)
     }
   } catch (ex) {
-    logger.log(`An error occurred while loading configuration. configPath=${configPath}\n`, ex)
+    if (Object.keys(envConfig).length) {
```
amitzur, line 26: I think this is not good behavior to throw depending on the env state.
How I expect it to work is that if a `configPath` is provided, then an error should be thrown. Otherwise, write to log the fact that you searched for this and this files and you didn't find it, and return what's in the env.

The assumption here is that the `configPath` was provided from the user. And I looked for where `ConfigUtils.getConfig` is used, and it's just Storybook and Cypress (I ignore troubleshooting for now). And only Storybook exposes this parameter to the user (actually, also Cypress from APPLITOOLS_CONFIG_PATH env var). So we can define this formally, that `ConfigUtils.getConfig` expects a `configPath` parameter **only from the user** and not from another internal function, and so it can decide to throw an exception if a configuration file is not found in that case.

packages/eyes-sdk-core/lib/utils/ConfigUtils.js
```diff
@@ -9,24 +8,58 @@ function getConfig({
...
 
+  for (const configFile of possibleConfigs) {
+    try {
+      return require.resolve(`./${configFile}`, {paths: [process.cwd()]})
+    } catch (error) {
+      // save errors to continue checking for other config files
+      // throw the earliest encountered error
+      errors.push(error)
+      if (index === possibleConfigs.length - 1) {
+        throw errors[0]
```
amitzur, line 54: That's a shame. You went all this trouble to save all the errors, and now you lose this information. I suggest throwing and error from this function which has the `errors` array. This way you can catch it and output a nice message to the user that says what files we looked for and didn't find.

amitzur (changes requested):

PR #654, opened 2021-06-24
fix: (eyes-nightwatch): support nightwatch-api with cucumber

tool3:
[trello](https://trello.com/c/wnolMIVh)

this PR introduces fixes and support for nightwatch-api with cucumber.

the majority of this PR is fixes to things that we didn't yet fix after the transition to eyes-api.


packages/eyes-nightwatch/src/spec-driver.ts
```diff
@@ -208,6 +211,10 @@ export async function waitUntilDisplayed(driver: Driver, selector: Selector, tim
 
 // #endregion
 
+export async function getCookies(_driver: Driver): Promise<any> {
+  return {cookies: [], all: false}
+}
+
```
kyryloonufriiev, line 217: Do you want to keep it?

packages/eyes-nightwatch/test/nightwatch.conf.js
```diff
@@ -47,22 +48,16 @@ module.exports = {
       desiredCapabilities: {
         browserName: 'chrome',
         'goog:chromeOptions': {
-          args: ['--headless'],
+          args: ['--headless=false'],
```
kyryloonufriiev, line 51: why `headless=false`?

packages/eyes-nightwatch/test/nightwatch.conf.js
```diff
@@ -47,22 +48,16 @@ module.exports = {
...
-      //    w3c: true,
-      //    args: ['--headless'],
-      //  },
-      //},
 
       webdriver: {
         port: 4444,
+        start_process: true,
         default_path_prefix: '/wd/hub',
+        server_path: chromedriver.path,
```
kyryloonufriiev, line 60: I think you have to use docker here and not a local browser

packages/eyes-nightwatch/src/spec-driver.ts
tool3: nope, removed

packages/eyes-nightwatch/test/nightwatch.conf.js
tool3: removed

packages/eyes-nightwatch/test/nightwatch.conf.js
tool3: this is actually using docker, it just needs the `chromedriver` path from the local `node_modules` but it will still connect to chrome running in docker.

tool3: removed

PR #661, opened 2021-07-08
Add browser environment only once

SophieDePaula:
When calling `mergeConfig`, we add all the browser environments to the config regardless if the browser env is already there.  This caused unnecessary calls to `createRGridDOMAndGetResourceMapping`.
I added a check to add an environment to the configuration only if it's not already included in the `browserInfo` array.

### DIFF
--- a/packages/eyes-sdk-core/CHANGELOG.md
+++ b/packages/eyes-sdk-core/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Change Log
 
 ## Unreleased
-
+ - make sure that we are not adding the same browser environment more than once when marging configs
 
 ## 12.21.3 - 2021/6/30
 
--- a/packages/eyes-sdk-core/lib/config/Configuration.js
+++ b/packages/eyes-sdk-core/lib/config/Configuration.js
@@ -1237,7 +1237,7 @@ class Configuration {
     ArgumentGuard.isArray(value, 'properties')
 
     for (const data of value) {
-      this._browsersInfo.push(data)
+      if (!this._browsersInfo.includes(data)) this._browsersInfo.push(data)
     }
     return this
   }
--- a/packages/eyes-sdk-core/lib/config/Configuration.js
+++ b/packages/eyes-sdk-core/lib/config/Configuration.js
@@ -1236,9 +1236,8 @@ class Configuration {
     }
     ArgumentGuard.isArray(value, 'properties')
 
-    for (const data of value) {
-      if (!this._browsersInfo.includes(data)) this._browsersInfo.push(data)
-    }
+    this._browsersInfo = value
+
     return this
   }
 
--- a/packages/eyes-sdk-core/test/unit/config/Configuration.spec.js
+++ b/packages/eyes-sdk-core/test/unit/config/Configuration.spec.js
@@ -15,6 +15,7 @@ const {
   SessionType,
   ProxySettings,
   PropertyData,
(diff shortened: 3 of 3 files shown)


kyryloonufriiev:
That's nice that you add a test here. Unfortunately, you did not handle the most important part of your test.
I will try to explain, in your code, you use `.includes(value)` method of an array, this method will return `true` only if `value` is *equal* to the array item, but since objects are compared by reference they are will not be equal even if they are looks the same. So `{a: 1} !== {a: 1}`. But in your test, you are merging configuration object with itself, so all of the `browserInfo` items are indeed equal because these are exactly the same objects.

I would recommend making it even simpler, I don't think we need to merge `browserInfo` at all, better just override them with the new value. And the same for other arrays, for example, if we would merge `layoutBreakpoints` what should we get?

kyryloonufriiev (changes requested):
That's nice that you add a test here. Unfortunately, you did not handle the most important part of your test.
I will try to explain, in your code, you use `.includes(value)` method of an array, this method will return true only if `value` is equal to the array item, but since objects are compared by reference they are will not be equal even if they are looks the same. So `{a: 1} !== {a: 1}`. But in your test, you are merging configuration object with itself, so all of the `browserInfo` items are indeed equal because these are exactly the same objects.

I would recommend making it even simpler, I don't think we need to merge `browserInfo` at all, better just override them with the new value. And the same for other arrays, for example, if we would merge `layoutBreakpoints` what should we get?

SophieDePaula:
@kyryloonufriiev I appreciate the detailed explanation. I am aware of the by ref comparison and actually I relay on it. I noticed that when we call `mergeConfig` from `check`, we always try to merge the config with itself:
`eyes-api` package, `Eyes.js`, `check` method:

`const result = await this._eyes.check({ settings, config: this._config.toJSON() });`

Therefore, the only time the condition will not be `false`, is when we add a new browser environment between `eyes.open` and `eyes.check` which is what I was going for.

I agree that replacing the browser config makes more sense, but since `mergeConfig` is called from more places I didn't want to inherently change the logic.

You are familiar with the overall code a lot better than me for sure, if you think this should not affect badly any other scenario, then I am all for it.

Let me know what you think.

kyryloonufriiev:
@SophieDePaula sorry that my answer took so long time. Actually, merge config called only in the code I wrote (the same code as you mentioned here like `check`, `locate`, `extractText` etc.). And for all of these places, it doesn't make sense to merge arrays, however, what you did is also makes sense just doesn't feel that solid for me, relay on a fact that the user will not provide a new object with the same parameters is a bit too optimistic on my opinion. Also, I think the solution I propose is just simpler you treat arrays as a primitive value and just override the old one with a new one.

SophieDePaula:
@kyryloonufriiev I made the change you suggested.
Notice, that I left my test unchanged, as the bug that was fixed here was that we added each browser environment the user added to the configuration as the multiplication of steps in the test.
Lmk if you have any other comments.

@amitzur fyi.

amitzur (approved):
@SophieDePaula this is good.
I would like to merge it since it does the job. But there's one feedback point:
IMO the test that needs to be here is the one that validates the actual issue, meaning `eyes-sdk-core/lib/new/check.js`. I know there isn't any test for that yet, but this would be the best validation that core actually performs the right thing as a black box. It would be considered an integration test, which is more valuable than the unit test here.
But still, let's merge this :)

PR #667, opened 2021-08-02
Cypress - fix hanging tests when calling open before visit

SophieDePaula:
Fixing the issue described in the following Trello cards: [trello1](https://trello.com/c/e9VdjegT/993-cypress-tests-hanging-batchend-on-eyes-close-in-aftereach-hook-multiple-customers) and [trello2](https://trello.com/c/6tApvbec/984-intuit-cypress-hangs-and-never-returns).

**Breakdown of the issue**:
When calling open we add tests to `runningTests`, when calling `batchStart` we reset `runningTests` var.
Since `batchStart` is called in Cypress before hook and cypress calls this hook when calling visit it resets this variable which causes the test to hang in some cases when calling open before visit.


packages/eyes-cypress/test/e2e/cypress-run-open-before-visit.test.js
```diff
@@ -0,0 +1,41 @@
...
+  });
+
+  after(async () => {
+    fs.rmdirSync(targetTestAppPath, {recursive: true});
+  });
+
+  it('works for openBeforeVisit.js', async () => {
+    try {
+      await pexec(
+        './node_modules/.bin/cypress run --headless --config testFiles=openBeforeVisit.js,integrationFolder=cypress/integration-run,pluginsFile=cypress/plugins/index-run.js,supportFile=cypress/support/index-run.js ',
```
SophieDePaula, line 31: @amitzur, @tool3 if this test runs prior to the fix I made, it will just hand and not throw an error. I would prefer to throw an error after a certain amount of time, but we run our tests with no timeouts parameter. Do you have any suggestions on how to handle this better?

tool3 (approved):
I have been meaning to fix this for a while ! great job ! 🎉

packages/eyes-cypress/test/e2e/cypress-run-open-before-visit.test.js
tool3: sure ! 
since we are spawning `cypress` on it's own process - notice that we are providing the config as command line arguments.

therefore, we can simply add: `pageLoadTimeout=timeoutInMs` or any other [timeout](https://docs.cypress.io/guides/references/configuration#Timeouts) - just be sure to give it enough so the test passes without hitting it.

Alternatively, you could do this in-code and simply modify the `cypress.json` file within the e2e, add the relevant timeout config and save it to the test directory (like we do in `./test/e2e/cypress-run-global-hooks` test).

amitzur (commented):
@SophieDePaula @tool3 I'd like to understand better what the behavior here. This fix feels hacky, and although it might be the only/best solution, but I'd like to understand better what state we reach, and why exactly is Cypress hanging. What promise is it that never resolves? What's the order of things before and after the fix?

SophieDePaula:
@amitzur the line that hangs is under `processCloseAndAboart.js`:
`const [closeErr, closeResult] = await closePromise; `

**The order of things**:

`pollingHandler` calls `doWork()` which actually calls `waitForBatch`. `waitForBatch` awaits on `processCloseAndAboart` which handgs on `closePromise` as mentioned above.

`processCloseAndAboart` loops over runningTests , each entry holds`abort` and `closePromise`.  `closePromise` is being overwritten due to the reset we call in `startBetch`.

packages/eyes-cypress/test/e2e/cypress-run-open-before-visit.test.js
SophieDePaula: Thanks @tool3, unfortunately, neither worked. A timeout error is not thrown when I set the timeout neither in `cypress.json` or in the cli.
Any other thoughts?

SophieDePaula:
@amitzur I made the adjustments we discussed and added an appropriate test to `handlers.test.js`.
I also deleted one unnecessary test (`handles "batchStart"`) that checks if the runningTests array is queal to `[]` since this is not the logic.
I added an `abort` method to one of @tool3 tests as the test failed because this method was missing after adding the new logic.

packages/eyes-cypress/src/plugin/runningTests.js
```diff
@@ -6,7 +6,10 @@ const runningTests = {
     this.tests.push(test);
   },
   reset() {
-    this.tests = [];
+    this.tests.forEach(test => { 
+      if(typeof(test.abort) === 'function')
```
amitzur, line 10: 1. `typeof test.abort` (no need for parens).
2. there's no need to check the typeof abort. It will always be there. Need to check if `closePromise` is undefined.

packages/eyes-cypress/test/unit/plugin/handlers.test.js
```diff
@@ -390,15 +390,9 @@ describe('handlers', () => {
     });
   });
 
-  it('handles "batchStart"', async () => {
```
amitzur, line 393: This test does make sense, but the expected behavior is that `abort` is called on every running test that doesn't have a closePromise

packages/eyes-cypress/test/e2e/cypress-run-open-before-visit.test.js
amitzur: As discussed, ptimeoutWithError from `@applitools/functional-commons` is your aid

amitzur (changes requested):

amitzur:
@SophieDePaula @tool3 documenting here that we found the true reason for the bug:
This reproduces only when we hit the concurrency limit. And then there is a promise for **open** which will never be countered by either **close** or **abort**, since the runningTests are reset.
So while "not resetting" runningTests works, it's a more explicit solution to actually abort all existing runningTests before cleaning them up.

packages/eyes-cypress/test/unit/plugin/handlers.test.js
SophieDePaula: @amitzur since we are not setting an empty array to `runningTests`, when this test is executed `runningTests` array will contain all the tests from all the tests that were already executed so far in `handlers.test.js`.  What indication do I have to make sure that abort was called on each and every one of these tests?
I added 'abort tests when calling "open" before batchStart' test just for this case and I built that test to give me an indication that abort was actually called.
lmk what you think.

packages/eyes-cypress/src/plugin/runningTests.js
amitzur: @SophieDePaula it does make more sense! Put in other words - the change in code creates a requirement for the type of `test`. This means it's a "breaking API change" of the `runningTests` module, and requires all consumers to align to it.

SophieDePaula: Gotcha!
We changed it again, so now no additional changes are needed.
Resolving this conversation.

packages/eyes-cypress/test/unit/plugin/handlers.test.js
SophieDePaula: We spoke about this one as well and agreed that the test I added was sufficient. Resolving this thread as well.

amitzur (approved):
Looks great!
Just to make sure - if we don't call `runningTests.abortTests()` in handlers, the test `works for openBeforeVisit.js` fails, right? And the failure is because of the 30000ms timeout, and Cypress continues to run in the background, which is undesired but we can leave it as a technical debt.
Is what I wrote correct?
If so, let's just add a comment in the test about the Cypress process still running in the background, and merge it.

SophieDePaula:
@amitzur that is exactly correct. I added a comment on that test.

PR #678, opened 2021-08-23
Region inside shadow

SophieDePaula:
This PR is meant to add the ability to capture a region within shadow DOM and frames when using the UFG **only**.
This is still WIP and we do have some limitations when working with this feature.


packages/eyes-webdriverio-5/src/spec-driver.ts
```diff
@@ -156,8 +156,16 @@ export async function findElement(browser: Driver, selector: Selector): Promise<
...
-  const elements = await browser.$$(transformSelector(selector))
+export async function findElements(browser: Driver, selector: Selector, element?: Element): Promise<Applitools.WebdriverIO.Element[]> {
+  let elements
+  if(element){
+    // the return element from getShadowDomContext snippet is of form element-6066-11e4-a52e-4f735466cecf, 
+    //therefore I need to first retrive the wdio element to take advantage of shadow method.
+     let shadowDoc = await browser.$(element as any)
+     elements = await shadowDoc.shadow$$(selector.toString())
+  } 
+  else
```
amitzur, line 167: code style:
```
} else {
  ...
}

File: packages/eyes-api/src/input/CheckSettings.ts [hunk 2]
```diff
@@ -180,6 +187,12 @@ export class CheckSettingsFluent<TElement = unknown, TSelector = unknown> {
     return this
   }
 
+  shadow(shadow: ElementReference<TElement, TSelector>): this {
+    utils.guard.custom(shadow, value => this._isRegionReference(value), {name: 'shadow'})
```
amitzur, line 191: `this._isElementReference`

amitzur: Also throw an error if there are frames

packages/eyes-api/src/input/CheckSettings.ts
```diff
@@ -189,6 +202,7 @@ export class CheckSettingsFluent<TElement = unknown, TSelector = unknown> {
     const context = utils.types.has(contextOrFrame, 'frame')
       ? contextOrFrame
       : {frame: contextOrFrame, scrollRootElement}
+    if (this._settings.shadow != undefined) throw new EyesError('frame cannot follow shadow')
```
amitzur, line 205: why EyesError and not Error?

packages/eyes-api/src/input/CheckSettings.ts
```diff
@@ -180,6 +187,12 @@ export class CheckSettingsFluent<TElement = unknown, TSelector = unknown> {
     return this
   }
 
+  shadow(shadow: ElementReference<TElement, TSelector>): this {
+    utils.guard.custom(shadow, value => this._isRegionReference(value), {name: 'shadow'})
+    this._settings.shadow = shadow
```
amitzur, line 192: You don't support shadow inside shadow?

packages/driver/src/context.ts
```diff
@@ -459,6 +459,30 @@ export class Context<TDriver, TContext, TElement, TSelector> {
     return region
   }
 
+  async switchToFrame(frame: string) {
```
amitzur, line 462: why is this addition needed?

packages/driver/src/context.ts
```diff
@@ -459,6 +459,30 @@ export class Context<TDriver, TContext, TElement, TSelector> {
...
+    this._target = await this._spec.childContext(this.target, await this._spec.findElement(this.target, frame))
+    await this._driver.updateCurrentContext(this)
+  }
+
+  async switchToDefaultContext() {
```
amitzur, line 467: why is this addition needed?

packages/driver/src/context.ts
```diff
@@ -459,6 +459,30 @@ export class Context<TDriver, TContext, TElement, TSelector> {
...
+    await this._driver.switchToMainContext()
+  }
+
+  async getRegionWithInShadowElement(shadowElement: TElement, targetSelector: string) {
```
amitzur, line 471: I think instead of this method you could just use the `elements` method on `context`. Just need to add a `rootElement` parameter to it. Correct me if I'm wrong

packages/eyes-sdk-core/lib/sdk/CheckSettingsUtils.js
```diff
@@ -5,9 +5,14 @@ const ImageMatchSettings = require('../config/ImageMatchSettings')
 
 async function toPersistedCheckSettings({checkSettings, context, logger}) {
   const elementsById = {}
+  const elementsByType = []
+  let shadowElement = undefined
+  let isFrameOrShadow = checkSettings.frame || checkSettings.shadow ? true : false
```
amitzur, line 10: no need for ternary expresssion

amitzur (commented):
1. Add Cypress e2e test
2. Cypress doc refinement
3. `getShadowContext` --> `getShadowRoot`

packages/driver/src/context.ts
SophieDePaula: I removed it as part of removing all frames support for now. Also, Kyrylo mentioned that this is not how it should be done so we can add support in the future.

packages/driver/src/context.ts
SophieDePaula: removed it as well. I needed it to switch back to the default context when we call `cleanupPersistance` to start from the top frame and drill down until I get to the region the user sent.

packages/eyes-api/src/input/CheckSettings.ts
SophieDePaula: This was changed to support a nested structure that relies on SpecSelector

packages/driver/src/context.ts
SophieDePaula: done

packages/eyes-sdk-core/lib/sdk/CheckSettingsUtils.js
SophieDePaula: This is logic was changed completely

packages/driver/src/context.ts
```diff
@@ -247,7 +247,18 @@ export class Context<TDriver, TContext, TElement, TSelector> {
         return new Element({spec: this._spec, context: this, selector: selectorOrElement, logger: this._logger})
       }
       await this.focus()
-      const element = await this._spec.findElement(this.target, selectorOrElement)
+
+      let rootElement = null
```
amitzur, line 251: when does this piece of code get executed?

packages/driver/src/context.ts
```diff
@@ -266,7 +277,18 @@ export class Context<TDriver, TContext, TElement, TSelector> {
         return [new Element({spec: this._spec, context: this, selector: selectorOrElement, logger: this._logger})]
       }
       await this.focus()
-      const elements = await this._spec.findElements(this.target, selectorOrElement)
+
+      let rootElement = null
```
amitzur, line 281: when does this piece of code get executed?

packages/driver/test/fixtures/spec-driver.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
     return (
       utils.types.isString(selector) ||
       utils.types.has(selector, ['using', 'value']) ||
-      utils.types.has(selector, ['type', 'selector'])
+      utils.types.has(selector, ['selector'])
```
amitzur, line 14: if there is no `type`, then the `selector` should also be checked for `isSelector`

packages/driver/src/context.ts
amitzur: nvm, found it

packages/driver/src/context.ts
```diff
@@ -266,7 +277,18 @@ export class Context<TDriver, TContext, TElement, TSelector> {
...
+      let selector = selectorOrElement
+      while (utils.types.has(selector, ['selector', 'shadow']) && this._spec.isSelector(selector.shadow)) {
+        const element: TElement = await this._spec.findElement(this.target, selector, rootElement)
+        if (!element) return null
```
amitzur, line 285: do we handle returning `null` from `.elements`?

packages/driver/src/context.ts
amitzur: nvm, found it

packages/eyes-webdriverio-5/src/spec-driver.ts
```diff
@@ -17,13 +21,12 @@ function extractElementId(element: Element): string {
...
-function transformSelector(
-  selector: Selector,
-): string | ((element: HTMLElement) => HTMLElement) | ((element: HTMLElement) => HTMLElement[]) {
+function transformSelector(selector: Selector): Applitools.WebdriverIO.Selector {
   if (selector instanceof legacy.By) {
     return selector.toString()
-  } else if (utils.types.has(selector, ['type', 'selector'])) {
-    if (selector.type === 'css') return `css selector:${selector.selector}`
+  } else if (utils.types.has(selector, ['selector'])) {
+    if (!selector.type) return selector.selector.toString()
```
amitzur, line 28: what do you expect to get in the `selector.selector` field?

packages/driver/src/context.ts
kyryloonufriiev: fixed. Now we will return an empty array

packages/driver/src/context.ts
kyryloonufriiev: This was implemented only for consistency.

packages/eyes-sdk-core/lib/sdk/CheckSettingsUtils.js
```diff
@@ -19,40 +20,53 @@ async function toPersistedCheckSettings({checkSettings, context, logger}) {
...
-      if (utils.types.has(region, ['width', 'height'])) {
-        referenceRegions = [region]
-      } else {
-        const elements = await context.elements(region)
-        referenceRegions = elements.map(element => {
+      if (!reference) continue
+      const referenceRegion = reference.region ? reference.region : reference
+      if (utils.types.has(referenceRegion, ['width', 'height'])) {
+        persistedRegions.push(persistReference(reference, referenceRegion))
+      } else if (referenceRegion) {
```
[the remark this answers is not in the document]
kyryloonufriiev: because if `region` was not specified in check settings we still will wrap it in an array, like that `[checkSettings.region]` so it is better to filter it out here than use it to find an element.

packages/eyes-webdriverio-5/src/spec-driver.ts
```diff
@@ -97,6 +100,9 @@ export function isSelector(selector: any): selector is Selector {
     utils.types.isString(selector) ||
     utils.types.isFunction(selector) ||
     utils.types.has(selector, ['type', 'selector']) ||
+    (utils.types.has(selector, ['selector']) &&
```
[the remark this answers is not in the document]
kyryloonufriiev: Spec drivers are not only about frameworks but also about compatibility with applitools driver. And our driver makes every framework support this kind of selectors.

packages/eyes-webdriverio-5/src/spec-driver.ts
```diff
@@ -17,13 +21,12 @@ function extractElementId(element: Element): string {
   else if (utils.types.has(element, ELEMENT_ID)) return element[ELEMENT_ID] as string
   else if (utils.types.has(element, LEGACY_ELEMENT_ID)) return element[LEGACY_ELEMENT_ID] as string
 }
-function transformSelector(
-  selector: Selector,
-): string | ((element: HTMLElement) => HTMLElement) | ((element: HTMLElement) => HTMLElement[]) {
+function transformSelector(selector: Selector): Applitools.WebdriverIO.Selector {
```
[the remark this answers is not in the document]
kyryloonufriiev: I don't know why I did it like that before. I had just copy-pasted wdio selector type instead of using the reference on it.

PR #679, opened 2021-08-23
Avoid the need for chalkify and STDOUT_LEVEL

amitzur:
The issue with `chalk` was actually because we spawn `/bin/bash` in the tests, and that is not detected as a tty (terminal). So the solution is to set FORCE_COLOR env var, which is read by `chalk` and brings colors back to output of e2e tests.
This removes the need for the incomplete solution of STDOUT_LEVEL, which stopped working for `ora`'s stderr for some reason.


amitzur:
Judging by the results here https://github.com/applitools/eyes.sdk.javascript1/runs/3404314632?check_suite_focus=true I may be wrong... this needs more investigation

PR #680, opened 2021-08-30
add pixel 5 to DeviceName

SophieDePaula:
Added Pixel 5 to core, eyes-api and types packages. 
[This was the original request](https://trello.com/c/YcYNF39r/1046-add-pixel-5-to-chromeemulation-devicename-enum)

### DIFF
--- a/packages/eyes-api/CHANGELOG.md
+++ b/packages/eyes-api/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## Unreleased
-
+- add Pixel 5 to DeviceName 
 
 ## 1.0.12 - 2021/8/13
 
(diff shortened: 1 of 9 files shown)


amitzur:
I'm curious if it's really necessary to have this in 3 different places (api, core, types). What can we do to avoid it?

kyryloonufriiev:
Oh. I thought this PR was already merged and this is the reason tests are broken in the core. I have implemented the same thing in my commit https://github.com/applitools/eyes.sdk.javascript1/commit/7e8f47e8eef34fb6fb0a56fdae5d5b4c855f6b57.

I also moved enum test to eyes-api package, since we don't really use `DeviceName` from core anymore. It also helped me realise that we are missing Samsung Galaxy 20. I have added a similar test for `IosDeviceName` and I have found that we are missing all of the iPhone 12 variations. Now it is fixed so I think we can close a PR without merging.

@amitzur answering your question. It is required only in types and api, we don't use enum from core. In api, we declare how users can use it, and in types we declare how we will use it. This is what we have to do in order to separate user api and internal implementation.

PR #685, opened 2021-09-10
Prevent trimming CDT 

SophieDePaula:
This is to fix trimmed CDT regarding [this trello card](https://trello.com/c/D8kvSPGI/1010-cypress-ufg-failed-to-render-screenshots)
Note: I also increased the body maximum size we allow to send to the UFG from 16MB to 30MB as I saw that Tomer mentioned the grid can handle this body size.
However, we will not get to a successful test run for the issue mentioned in the card above as the content size for the page CDT is 98MB.


packages/eyes-sdk-core/lib/server/ServerConnector.js
```diff
@@ -637,7 +637,7 @@ class ServerConnector {
         'X-Auth-Token': this._renderingInfo.getAccessToken(),
         'Content-Type': resource.getContentType(),
       },
-      maxBodyLength: 15.5 * 1024 * 1024, // 15.5 MB  (VG limit is 16MB)
+      maxBodyLength: 29.5 * 1024 * 1024, // 29.5 MB  (VG limit is 30MB)
```
amitzur, line 640: let's do `35.5 * 1024 * 1024`

packages/eyes-sdk-core/lib/renderer/RGridResource.js
```diff
@@ -4,7 +4,7 @@ const crypto = require('crypto')
 const ArgumentGuard = require('../utils/ArgumentGuard')
 const GeneralUtils = require('../utils/GeneralUtils')
 
-const VISUAL_GRID_MAX_BUFFER_SIZE = 15 * 1000000
+const VISUAL_GRID_MAX_BUFFER_SIZE = 30 * 1000000
```
amitzur, line 7: `35.5 * 1024 * 1024`

amitzur (commented):
small changes, then you can merge it

packages/eyes-sdk-core/lib/renderer/RGridResource.js
SophieDePaula: @amitzur I changed it to 34.5 since I see that previously we left a buffer for other things (I set it to 34.5 MB).

PR #689, opened 2021-09-29
Universal logging

kyryloonufriiev:
What was done:
1. Replace logger implementation from the core with logger package.
2. Implement rolling file handler for logger
3. Add support for passing logger directly to `openEyes` method

What should be done:
1. Logger configuration on the server-side (if needed)
2. Decide what we want to do with warnings that are currently showing up in the console (such as one during window resize for layout breakpoints)

What I propose:
1. Maybe instead of configuring logging on the server we can just add some endpoint to the server which will return server-internal information, such as a path to the logging directory (right now it is always a temp directory with datetime in name)
2. To pass some server-side warnings to the client console we can add another event to the server and instead of logging to the server console, we will send the event on the client-side, where they could decide how they want to show it.


packages/eyes-universal/src/universal-server.ts
```diff
@@ -28,6 +36,8 @@ export async function makeServer({debug = false, idleTimeout = IDLE_TIMEOUT, ...
       socket: makeSocket(client) as Socket & types.ServerSocket<Driver, Context, Element, Selector>,
     })
 
+    const logger = baseLogger.extend()
```
amitzur, line 39: why is this needed?

packages/eyes-sdk-core/lib/sdk/EyesBase.js
```diff
@@ -1404,13 +1390,11 @@ class EyesBase {
    * @return {Promise}
    */
   async openBase(appName, testName, viewportSize, sessionType = SessionType.SEQUENTIAL, skipStartingSession = false) {
-    this._logger.getLogHandler().open()
```
amitzur, line 1407: So the `.open` is removed but the `.close` is still kept? I think we can remove the open/close API altogether

packages/eyes-sdk-core/lib/sdk/EyesBase.js
```diff
@@ -1837,7 +1821,7 @@ class EyesBase {
         this._isViewportSizeSet = true
         await this._sessionEventHandlers.setSizeEnded()
       } catch (err) {
-        this._logger.verbose('Can not ensure ViewportSize', err)
+        this._logger.error('Can not ensure ViewportSize', err)
```
amitzur, line 1824: maybe `.warn`?

PR #702, opened 2021-10-31
Fix backwards compatibility deprecation warnings

amitzur:
I learned that we output the deprecation warning backwards.
The correct logic should be:
batchSequence → batchSequenceName
branch → branchName
parentBranch → parentBranchName
baselineBranch → baselineBranchName
batchNotify → notifyOnCompletion

This change doesn't affect any user who specified either old OR new param value. Because whatever they specify was applied. Only the deprecation warning is now changed.
If a user specified both, then we used to take the old value and show a (wrong) deprecation warning, but now we will take the new value and show no deprecation warning.


SophieDePaula:
Looks good ✅

SophieDePaula (approved):

PR #703, opened 2021-11-02
Side eyes cors iframes

SophieDePaula:
This PR was created to support CORS iframes in side-eyes. This is the [trello card](https://trello.com/c/2tOMIn78/740-research-support-for-cross-origin-frames-for-ufg-in-selenium-ide)

The objective is to use `takeDomSnapshot` from `core` in order to capture the snapshot.


packages/side-eyes/src/content/index.js
```diff
@@ -63,6 +66,49 @@ window.addEventListener('message', event => {
   }
 })
 
+const apiScript = document.createElement('script')
+apiScript.src = browser.runtime.getURL('api.js')
```
amitzur, line 70: but there is no `api.js` file. I don't think it's needed. And since it's not there, nobody is using `mark`, so no need for `unmark` below

amitzur (commented):
Looks good. See comments inline

packages/side-eyes/src/manifest.json
```diff
@@ -26,7 +26,10 @@
     "<all_urls>"
   ],
   "web_accessible_resources": [
-    "assets/pageScripts.js"
+    "assets/pageScripts.js",
+    "assets/dom-capture/*",
```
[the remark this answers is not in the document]
SophieDePaula: it is not, I removed it

packages/side-eyes/src/content/index.js
SophieDePaula: Yes, we don't need that, I removed it

PR #705, opened 2021-11-03
Fix cypress tests not being cleaned up

amitzur:
[Trello](https://trello.com/c/Mlf6U6k8/1099-cypress-test-results-combining-on-consecutive-runs)

We currently abort tests when `batchStart` is fired, but we don't reset the state of `runningTests`, so they accumulate across executions.


SophieDePaula:
💯

SophieDePaula (approved):

PR #708, opened 2021-11-09
Fix cypress global hook not called

amitzur:
2 issues happened because of #706 and #704 :
1. We stopped calling the global hooks `before:run` and `after:run` at all, unless the user also called them in his plugins file.
2. We stopped calling the browser hooks for Cypress versions < 6.2.0 (where global hooks are not supported).

This was not caught by tests because our e2e tests only executed Cypress with Eyes, but didn't check the output. Since a passing output doesn't have any indication that Eyes ran, I changed those e2e files to fail on diffs, and check the output for the diff error message.


packages/eyes-cypress/src/plugin/pluginExport.js
```diff
@@ -1,16 +1,26 @@
...
       const moduleExportsResult = await pluginModuleExports(onThatCallsUserDefinedHandler, config);
+      if (isGlobalHooksSupported(config)) {
+        if (!isGlobalHookCalledFromUserHandler) {
+          for (const [eventName, eventHandler] of Object.entries(globalHooks)) {
+            origOn.call(this, eventName, eventHandler);
+          }
+        }
+      } else {
+        eyesConfig.eyesIsGlobalHooksNotSupported = true;
+      }
```
kyryloonufriiev, line 23: Isn't it better to calculate `eyesIsGlobalHooksNotSupported` once before `if` statement and just use it here and also in `onThatCallsUserDefinedHandler` ?

kyryloonufriiev (approved):
I do not fully understand cypress. But PR change itself seems pretty straightforward to me.
This problem makes me think that to provide the simplest setup for cypress with eyes, we lost some flexibility, so now we have to deal with wrapping user code and preserving the behavior.

PR #709, opened 2021-11-10
Cypress support wait before capture

SophieDePaula:
This PR is meant to support waitBeforeCapture on `eyes-cypress`, as requested in this [trello card](https://trello.com/c/Bh7fu8al/1125-support-waitbeforecapture-in-cypress)


packages/eyes-cypress/README.md
```diff
@@ -449,6 +450,20 @@ cy.eyesCheckWindow({variationGroupId: 'Login screen variation #1'})
 
 For more information, visit our documentation page: https://applitools.com/docs/features/baseline-variations-groups.html
 
+##### `waitBeforeCapture`
+
+A parameter that is set to wait a certain amount of milliseconds between the page resize when using `layoutBreakpoints`
```
amitzur, line 455: I suggest:
```
A parameter that is set to wait a certain amount of milliseconds before capturing the pages snapshot. This will also apply between page resizes when using `layoutBreakpoints`.
```

packages/eyes-cypress/src/browser/commands.js
```diff
@@ -15,7 +15,7 @@ const eyesCheckWindow = makeEyesCheckWindow({sendRequest, processPage, domSnapsh
 
 function getGlobalConfigProperty(prop) {
   const property = Cypress.config(prop);
-  const shouldParse = ['eyesBrowser', 'eyesLayoutBreakpoints'];
+  const shouldParse = ['eyesBrowser', 'eyesLayoutBreakpoints', 'eyesWaitBeforeCapture'];
```
amitzur, line 18: why? it's just a number

packages/eyes-cypress/src/browser/eyesCheckWindow.js
```diff
@@ -21,11 +21,17 @@ function makeEyesCheckWindow({sendRequest, processPage, domSnapshotOptions, cypr
     function takeDomSnapshots(options) {
       const browser = args.browser;
       const breakpoints = args.layoutBreakpoints;
+      const waitBeforeCapture = args.waitBeforeCapture ? args.waitBeforeCapture : 0;
```
amitzur, line 24: the default in core is 100ms. I see that in Cypress it's 300. So let's change it back to 100. And you can just set it here instead of setting it to `0` and then checking if it's truthy later

packages/eyes-cypress/src/browser/eyesCheckWindow.js
```diff
@@ -52,7 +58,7 @@ function makeEyesCheckWindow({sendRequest, processPage, domSnapshotOptions, cypr
             // console.log(`taking dom snapshot for width ${requiredWidth}`);
             cypress
               .viewport(Number(requiredWidth), height, {log: false})
-              .wait(300, {log: false})
+              .wait(waitBeforeCapture ? waitBeforeCapture : 300, {log: false})
```
amitzur, line 61: see my earlier comment about setting this to 100, and here you could already trust that it's defined

packages/eyes-cypress/src/plugin/config.js
```diff
@@ -43,6 +43,7 @@ function makeConfig() {
       config.failCypressOnDiff === undefined ? true : !!config.failCypressOnDiff,
     eyesDisableBrowserFetching: !!config.disableBrowserFetching,
     eyesTestConcurrency: config.testConcurrency || DEFAULT_TEST_CONCURRENCY,
+    eyesWaitBeforeCapture: JSON.stringify(config.waitBeforeCapture),
```
amitzur, line 46: Why stringify it? the user is supposed to pass a number

packages/eyes-cypress/test/fixtures/testApp/cypress/integration-run/waitBeforeCapture.js
```diff
@@ -0,0 +1,39 @@
+/* global cy */
+describe('works with waitBeforeCapture', () => {
+    it('test waitBeforeCapture in open', () => {
+    cy.viewport(600, 600)
+    cy.visit('https://applitools.github.io/demo/TestPages/waitBeforeCapture/');
```
amitzur, line 5: This is something that existed already and is used to test the feature in all other SDK's, right?

amitzur (approved):
Awesome! :100: 
Very small notes regarding the stringify/parse, default values, and documentation

packages/eyes-cypress/test/fixtures/testApp/cypress/integration-run/waitBeforeCapture.js
SophieDePaula: yes

PR #711, opened 2021-11-12
fix(eyes-cypress): use `Cypress.currentTest` as a fallback to get `testName` (#710)

### LINKED ISSUE #710: [eyes-cypress] eyesOpen command can't be overwritten
There is an issue in Cypress currently that prevents overwriting commands that reference `this`: cypress-io/cypress#18892

`eyesOpen()` currently references the title like this:

```js
const {title: testName} = this.currentTest || this.test;
```

The fix would be to change it to `Cypress.currentTest`:

```js
const {title: testName} = Cypress.currentTest;
```

slikts:
Replaces reference to `this.test` with `Cypress.currentTest` and fixes #710.

### DIFF
--- a/packages/eyes-cypress/src/browser/commands.js
+++ b/packages/eyes-cypress/src/browser/commands.js
@@ -53,7 +53,7 @@ Cypress.Commands.add('eyesOpen', function(args = {}) {
   Cypress.config('eyesOpenArgs', args);
   Cypress.log({name: 'Eyes: open'});
   const userAgent = navigator.userAgent;
-  const {title: testName} = this.currentTest || this.test;
+  const {title: testName} = Cypress.currentTest;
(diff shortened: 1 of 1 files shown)


kyryloonufriiev:
@slikts I believe `Cypress.currentTest` is relatively new api ([docs](https://docs.cypress.io/api/cypress-api/currenttest#Get-full-path-of-current-test-title). But our SDK should also support older versions as well. So feel free to fix your PR with this in mind.

slikts:
@kyryloonufriiev I've added `Cypress.currentTest` as a fallback to the previous behavior. Also, since Cypress v9.1 the issue is fixed and Cypress sets the `this` context for overwritten commands correctly.

kyryloonufriiev:
Thank you! Happy to see that Cypress has resolved your issue faster. Anyway, I think the change is valid so I'm merging this PR.

PR #713, opened 2021-11-12
Cypress, fix types for region inside shadow

SophieDePaula:
[Trello card](https://trello.com/c/x7Sxk8qa/1077-fix-types-for-cypress-region-inside-shadow)


kyryloonufriiev (approved):
💯

PR #721, opened 2021-12-07
update generic tests configs 

Artem-applitools:
to point to 'universal-sdk' branch of the sdk.coverage.tests repo
draft PR of the sdk.coverage.tests changes: https://github.com/applitools/sdk.coverage.tests/pull/107

Branch "universal-sdk" was created for baselines and for coverage tests to point all of the tests for an sdk that migrated to the universal sdk to that branch. Once all sdks migrated to universal sdk, we could merge this branch to master. As all sdks migrating to the universal JS suite going to be the main one. 

In the JS repo I have updated all configs and tests generation scripts to point to the "universal-sdk" branch


amitzur:
@Artem-applitools @Yarden-Ingber this seems ok. The only point I'm confused about is that there are many changes for JS in the `sdk.coverage.tests` repo `universal-sdk` branch. Shouldn't JS configuration stay more or less the same there?

amitzur (commented):
@Artem-applitools @Yarden-Ingber this seems ok. The only point I'm confused about is that there are many changes for JS in the `sdk.coverage.tests` repo `universal-sdk` branch. Shouldn't JS configuration stay more or less the same there?

PR #722, opened 2021-12-08
VGC - Add Logs

SophieDePaula:
In order to better understand why we get missing resources in the resource mapping, I added some logs.
This is the [Trello card](https://trello.com/c/9rJGqrLD/1147-next-insurance-cypress-error-fetching-font-files) that reported the problem


packages/visual-grid-client/src/sdk/getAllResources.js
```diff
@@ -87,6 +87,12 @@ function makeGetAllResources({resourceCache, fetchResource, extractCssResources,
             resources,
             mapKeys(mapValues(cacheEntry, fromCacheToRGridResource), value => value.getUrl()),
           )
+          const cacheEntryKeys = Object.keys(cacheEntry)
+          logger.log(
+            `resource retrieved from cache, with dependencies (${
+              cacheEntryKeys.length
+            }): ${rGridResource.getUrl()} with dependencies --> ${cacheEntryKeys}`,
```
amitzur, line 94: @SophieDePaula why write `with dependencies` twice?

SophieDePaula: I fixed this in PR #729

PR #723, opened 2021-12-11
Support Storybook 6.4

SophieDePaula:
`Storybook` changed the client API we are using. `setSelection` was removed from storybook version 6.4.0 onwards.
Also, when `Storybook` starts, it outputs a different message, so we need to update our `Regex`
[Trello card](https://trello.com/c/mMmMfPMY/1170-storybook-slack-eyes-storybook-timing-out-after-upgrading-storybook-dependency-to-640)


amitzur:
@SophieDePaula duplicate of #720 ?

SophieDePaula:
duplicate of PR 720

PR #727, opened 2021-12-20
Adjust eyes-storybook tests to align with storybook 6.4

SophieDePaula:
1) Skip zero stories test for now, as the option to load a story after storybook started was removed from storybook API
2) Add a delay option for a render request in `fakeEyesServer`

### DIFF
--- a/packages/eyes-storybook/test/e2e/zero-stories-retry.e2e.test.js
+++ b/packages/eyes-storybook/test/e2e/zero-stories-retry.e2e.test.js
@@ -9,7 +9,8 @@ const envWithColor = {...process.env, FORCE_COLOR: true};
 const spawnOptions = {stdio: 'pipe', env: envWithColor};
 
 describe('zero stories retry', () => {
-  it('zero stories retry', async () => {
+  //skipped as storybook removed the option to load a story after storybook starts
+  it.skip('zero stories retry', async () => {
     const [err, result] = await presult(
       utils.process.sh(
         `node ${path.resolve(__dirname, '../../bin/eyes-storybook')} -f ${path.resolve(
--- a/packages/eyes-storybook/test/e2e/zero-stories-retry.e2e.test.js
+++ b/packages/eyes-storybook/test/e2e/zero-stories-retry.e2e.test.js
@@ -10,6 +10,7 @@ const spawnOptions = {stdio: 'pipe', env: envWithColor};
 
 describe('zero stories retry', () => {
   //skipped as storybook removed the option to load a story after storybook starts
+  // going forward we would like to re-add this test by running it with Storybook version 5
   it.skip('zero stories retry', async () => {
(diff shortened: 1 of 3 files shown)


packages/eyes-storybook/test/e2e/zero-stories-retry.e2e.test.js
```diff
@@ -9,7 +9,8 @@ const envWithColor = {...process.env, FORCE_COLOR: true};
 const spawnOptions = {stdio: 'pipe', env: envWithColor};
 
 describe('zero stories retry', () => {
-  it('zero stories retry', async () => {
+  //skipped as storybook removed the option to load a story after storybook starts
```
amitzur, line 12: // going forward we would like to re-add this test by running it with Storybook version 5

packages/eyes-storybook/test/it/eyesStorybook.it.test.js
```diff
@@ -29,8 +29,11 @@ describe('eyesStorybook', () => {
   });
 
   let serverUrl, closeEyesServer;
-  beforeEach(async () => {
-    const {port, close} = await fakeEyesServer();
+  beforeEach(async function() {
+    let fakeEyesConfig = {}
+    if(this.test.title.includes('enforces legacy concurrency'))
```
amitzur, line 34: I would like to avoid this. So I would propose either moving the `fakeEyesServer` call inside each test (since there are not a lot of tests) or create a separate `describe` for the legacy concurrency test.

packages/eyes-storybook/test/util/fakeEyesServer.js
```diff
@@ -8,7 +8,7 @@ const path = require('path');
 const filenamify = require('filenamify');
 const {TestResultsStatus} = require('@applitools/eyes-sdk-core');
 
-function fakeEyesServer({expectedFolder, updateFixtures, port, hangUp: _hangUp} = {}) {
+function fakeEyesServer({expectedFolder, updateFixtures, port, hangUp: _hangUp, delay=0} = {}) {
```
amitzur, line 11: I would call the parameter `renderDelay` to be extra clear

packages/eyes-storybook/test/it/eyesStorybook.it.test.js
```diff
@@ -29,8 +29,11 @@ describe('eyesStorybook', () => {
...
+      fakeEyesConfig = {delay: 500}
```
amitzur, line 35: eslint fails for this. Please add `{}` in the `if`

amitzur, line 35: For me `500` was not enough. I still got 9 instead of 10. `1000` was enough.

amitzur (approved):
It's great that this fixes the test!
See comments inline

PR #729, opened 2021-12-23
vgc get all dependent resources

SophieDePaula:
This PR should fix the issue that was raised in this [trello card](https://trello.com/c/hAyfn0zB/1129-lytx-surfsight-prospect-poc-cypress-ufg-dom-snapshot-has-an-absent-resource-from-time-to-time)

**Problem**:
When we get a resource with dependencies (such as a `CSS` file), we add it to the `resourceCache` with it's `dependencies`, but we don't add the dependencies themselves to the `resourceCache` only until a later stage. This cause a problem when another step access that `CSS` file from the `resourceCache` as the only dependencies that will be returned are of types `CSS` and `SVG`

**Solution**:
In `processResource` we add ALL the dependencies to the `resourceCache` with a pending `Promise`, this promise will be resolved in `putResources` when we set the correct value to `resourceCache`.
This mechanism allows us to wait on these promises to make sure we get all the dependent resource.


packages/visual-grid-client/src/sdk/putResources.js
```diff
@@ -57,6 +57,14 @@ function makePutResources({
       logger.verbose('setting resource to cache: ', cacheKey)
       fetchCache.remove(cacheKey)
       const doesRequireProcessing = Boolean(resourceType(resource.getContentType()))
+      // resolve the dummy promise as we now set the real value of the resource to the cacheResouce
+      if (
```
amitzur, line 61: Would be more explicit and robust to save the promise in a variable and resolve it only AFTER setting the actual value in the cache. Otherwise you rely on Node's microtask mechanism that runs promise resolutions at the end of the current synchronous call stack.

packages/visual-grid-client/src/sdk/putResources.js
```diff
@@ -57,7 +57,16 @@ function makePutResources({
...
+        !doesRequireProcessing &&
+        resourceCache.getValue(cacheKey) &&
+        resourceCache.getValue(cacheKey).resolve
+      ) {
+        const promise = resourceCache.getValue(cacheKey)
+        resourceCache.setValue(cacheKey, toCacheEntry(resource, doesRequireProcessing))
+        promise.resolve()
+      } else resourceCache.setValue(cacheKey, toCacheEntry(resource, doesRequireProcessing))
```
amitzur, line 69: I would write this in a cleaner way:
```
// resolve the dummy promise as we now set the real value of the resource to the cacheResouce
const resourceValue = resourceCache.getValue(cacheKey)
resourceCache.setValue(cacheKey, toCacheEntry(resource, doesRequireProcessing))
if (resourceValue && resourceValue.resolve) {
  resourceValue.resolve(resourceCache.getValue(cacheKey))
}
```

Note that here we resolve the promise with the value that's needed in `getAllResources` so we don't need to do `if typeof... === 'string'` and then get the value from the cache there.

packages/visual-grid-client/test/unit/sdk/getAllResources.test.js
```diff
@@ -935,4 +944,26 @@ describe('getAllResources', () => {
...
+    const server = await testServerInProcess()
+    const baseUrl = `http://localhost:${server.port}`
+    closeServer = server.close
+    let resourcesOne, resourcesTwo
+
+    try {
+      resourcesOne = await getAllResources({
+        resourceUrls: [`${baseUrl}/mediumCss.css`],
+      })
+      putResources(Object.values(resourcesOne))
```
amitzur, line 958: You intentionally not `await` here, right? In order to create the race condition?

packages/visual-grid-client/test/unit/sdk/getAllResources.test.js
```diff
@@ -935,4 +944,26 @@ describe('getAllResources', () => {
       getSha256Hash(mainUserAgent),
     )
   })
+
+  it('get all css resources from resourceCache', async () => {
+    const server = await testServerInProcess()
```
amitzur, line 949: I believe you never get to actually await a promise in the line `const cacheEntries = await Promise.all(Object.values(cacheEntry))` because the server is responding very quickly. I suggest to pass `middlewares: ['slow']`

packages/visual-grid-client/src/sdk/getAllResources.js
```diff
@@ -130,6 +138,16 @@ function makeGetAllResources({resourceCache, fetchResource, extractCssResources,
       const doesRequireProcessing = !!resourceType(rGridResource.getContentType())
       if (doesRequireProcessing) {
         resourceCache.setValue(rGridResource.getCacheKey(), toCacheEntry(rGridResource))
+      } else {
+        // Create a dummy promise so when we call getWithDependencies we will wait for ALL the resource dependencies
+        // to be added to the resourceCache. This promise is resolved in putResources
+        let resolve
+        const promise = new Promise(thisResolve => {
```
amitzur, line 145: I think this can be reduced to:
```
let resolve
const promise = new Promise(thisResolve => resolve = thisResolve)
promise.resolve = resolve
```
This works in correlation to a change I suggested for line 90, and in `putResources`

packages/visual-grid-client/src/sdk/getAllResources.js
```diff
@@ -83,15 +83,23 @@ function makeGetAllResources({resourceCache, fetchResource, extractCssResources,
         handledResourceUrls.add(rGridResource.getUrl())
         const cacheEntry = resourceCache.getWithDependencies(rGridResource.getCacheKey())
         if (cacheEntry) {
+          const cacheEntries = await Promise.all(Object.values(cacheEntry))
+          cacheEntries.forEach((entry, index) => {
+            // we need to get the real value from the resourceCache after the dummy promise is resolved
+            // in putResource and the real value is set to the resourceCache
+            if (typeof entry == 'string') {
```
amitzur, line 90: I don't think this duck typing is a good idea. We can take advantage of having the actual value we need in `putResources` in order to resolve the promise with the value that we actually need. See the change proposed in `putResources`.

packages/visual-grid-client/test/unit/sdk/getAllResources.test.js
```diff
@@ -935,4 +944,26 @@ describe('getAllResources', () => {
...
+      putResources(Object.values(resourcesOne))
+      resourcesTwo = await getAllResources({
+        resourceUrls: [`${baseUrl}/mediumCss.css`],
+      })
+    } catch (ex) {
+      throw ex
+    } finally {
+      closeServer()
+    }
+    expect(Object.keys(resourcesOne).length).to.eql(Object.keys(resourcesTwo).length)
```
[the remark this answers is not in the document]
amitzur: I would compare the actual content of these objects: `expect(resourcesOne).to.eql(resourcesTwo)`
The test even fails in this comparison, so worth investigating why.

amitzur (changes requested):

packages/visual-grid-client/src/sdk/getAllResources.js
```diff
@@ -141,12 +134,11 @@ function makeGetAllResources({resourceCache, fetchResource, extractCssResources,
...
         let resolve
-        const promise = new Promise(thisResolve => {
-          resolve = function() {
-            thisResolve(rGridResource.getCacheKey())
-          }
-        })
+        const promise = new Promise(thisResolve => (resolve = thisResolve))
+        promise.resolve = resolve
+
         resourceCache.setValue(rGridResource.getCacheKey(), Object.assign(promise, {resolve}))
```
amitzur, line 142: no need to do `Object.assign` anymore since you do `promise.resolve = resolve`

packages/visual-grid-client/test/unit/sdk/getAllResources.test.js
```diff
@@ -54,6 +58,15 @@ describe('getAllResources', () => {
       }
       return r
     }
+    getAllResourcesWithSha = async args => {
```
amitzur, line 61: if it's just an alias, then do `getAllResourcesWithSha = getAllResourcesOrig`.
Or better yet, just rename `getAllResourcesOrig` to `getAllResourcesWithSha`

packages/visual-grid-client/test/unit/sdk/getAllResources.test.js
```diff
@@ -22,7 +22,11 @@ require('@applitools/isomorphic-fetch')
 
 describe('getAllResources', () => {
   let closeServer
-  let getAllResources, resourceCache, putResources
+  let getAllResources,
+    resourceCache,
+    putResources,
+    getAllResourcesWithSha,
+    removeContentFromResouces
```
amitzur, line 29: typo - `Resouces` should be `Resources`

packages/visual-grid-client/test/unit/sdk/getAllResources.test.js
```diff
@@ -54,6 +58,15 @@ describe('getAllResources', () => {
       }
       return r
     }
+    getAllResourcesWithSha = async args => {
+      return await getAllResourcesOrig(args)
+    }
+
+    removeContentFromResouces = resouces => {
+      for (const resouce of resouces) {
```
amitzur, line 66: typo - `reouce` and `resouces` should be `resource` and `resources`

amitzur (approved):

PR #730, opened 2021-12-31
testCafe - set failTestcafeOnDiff default value to true

SophieDePaula:
Set the default value of `failTestcafeOnDiff` to true and fix `eyes failTestcafeOnDiff true` test as before this test would always pass due to the fact that if the test fails the exception would be caught and pass the test.

Trello card: https://trello.com/c/lSau8Xfy/1152-testcafe-eyes-not-throwing-any-diffsfoundexception


packages/eyes-testcafe/src/legacy.ts
```diff
@@ -54,6 +54,8 @@ export function LegacyTestCafeEyesMixin<TDriver extends Driver, TElement extends
           runnerOrConfigOrOptions?.runner ??
           new api.VisualGridRunner({testConcurrency: testcafeConfig.concurrency ?? testcafeConfig.testConcurrency ?? 1})
         super(runner, transformConfig(testcafeConfig))
+        testcafeConfig.failTestcafeOnDiff =
+          testcafeConfig.failTestcafeOnDiff === undefined ? true : testcafeConfig.failTestcafeOnDiff
```
kyryloonufriiev, line 58:
```suggestion
        testcafeConfig.failTestcafeOnDiff ??= true
```

PR #731, opened 2022-01-08
separate local server to it's own process

SophieDePaula:
In order to better understand what happens when presumably our local server crashes, we now start the local server in a separate process (relates to this [trello card](https://trello.com/b/ECya7n3i/js-sdks?filter=due:notdue)).

In `pluginExport.js` we now `fork` a new process that gets `server.js` file and does:

1) Create `logger`
2) Create `visualGridClient`
3) Create `handlers`
4) Create the `app`
5) Start the local server
6) Add a new endpoint to `app`

`server.js` now listen to messages from `pluginExport` to perform the functionalty mentioned above.
In addition, a new endpoint was created for `app`. This endpoint (`eyes/waitForBatch`) will be called from the `after:run` of the global hooks instead of calling `waitForBatch` directly.

TODO:
Add a test to crash the server and make sure we are getting an informative error message


packages/eyes-cypress/test/it/plugin/server.test.js
```diff
@@ -6,7 +6,7 @@ const makeStartServer = require('../../../src/plugin/server');
 const express = require('express');
 
 describe('plugin server', () => {
-  it('starts at a random port', async () => {
+  it.skip('starts at a random port', async () => {
```
SophieDePaula, line 9: I skipped this test as we start the server from other places, we get the port and are able to connect. Therefore, I don't believe doing the same thing here will benefit us.

packages/eyes-cypress/src/plugin/hooks.js
```diff
@@ -1,39 +1,26 @@
 'use strict';
-const makeWaitForBatch = require('./waitForBatch');
-const makeHandleBatchResultsFile = require('./makeHandleBatchResultsFile');
-const getErrorsAndDiffs = require('./getErrorsAndDiffs');
-const processCloseAndAbort = require('./processCloseAndAbort');
-const errorDigest = require('./errorDigest');
 const runningTests = require('./runningTests');
+const axios = require('axios');
 
 function makeGlobalRunHooks({visualGridClient, logger}) {
```
amitzur, line 5: No more need to `makeGlobalRunHooks` since both `visualGridClient` and `logger` are no longer used

packages/eyes-cypress/src/plugin/hooks.js
```diff
@@ -1,39 +1,26 @@
...
 
-      try {
-        await waitForBatch(runningTests.tests, visualGridClient.closeBatch);
-      } catch (e) {
-        if (!!config.eyesFailCypressOnDiff) {
-          throw e;
-        }
+      const params = {params: config};
+      const headers = {headers: {'Content-Type': 'application/json'}};
+      const response = await axios.get(
```
amitzur, line 17: I think the right way to do it is `axios.get(url, {params, headers})`. But are you sure that `config` is serializable? You are sending a `GET` request here, and `params` will become query params on the URL. Is that the intention?

packages/eyes-cypress/src/plugin/hooks.js
```diff
@@ -1,39 +1,26 @@
...
+        `https://localhost:${config.eyesPort}/eyes/waitForBatch`,
+        params,
+        headers,
+      );
+      if (response.data.message && response.data.message.includes('Eyes-Cypress detected diffs')) {
```
amitzur, line 22: Throwing an error here should be done based on the `success` value of the response. The server should respond with a `{success, error}` object, and the logic of the message should be in the server. The code here should just check if `success` is `false` then throw the `error`

packages/eyes-cypress/src/plugin/pluginExport.js
```diff
@@ -1,15 +1,29 @@
 'use strict';
 const isGlobalHooksSupported = require('./isGlobalHooksSupported');
 const {presult} = require('@applitools/functional-commons');
+const childProcess = require('child_process');
+const path = require('path');
 
-function makePluginExport({startServer, eyesConfig, globalHooks}) {
+function makePluginExport({eyesConfig, globalHooks, basicConfig = ''}) {
```
amitzur, line 7: why is the default value for `baseConfig` an empty string? Its "type" is actually an object, according to what is passed from `startPlugin`

packages/eyes-cypress/src/plugin/server.js
```diff
@@ -2,45 +2,98 @@
...
-          logger.log(`plugin server running at port: ${server.address().port}`);
-          resolve({
-            eyesPort: server.address().port,
-            closeServer: server.close.bind(server),
-          });
-        }
-      });
+process.on('message', message => {
+  if (message === 'close server') {
+    closeServer();
```
amitzur, line 20: missing `await`

packages/eyes-cypress/src/plugin/server.js
```diff
@@ -2,45 +2,98 @@
...
+        console.log('error in plugin server:', err);
+        logger.log('error in plugin server:', err);
+      }
+      reject(err);
     });
-  };
-}
 
-module.exports = makeStartServer;
+    app.get('/eyes/waitForBatch', express.json({limit: '100mb'}), async (req, res) => {
```
amitzur, line 72: Could this not be done in `handlers`? We already listen to `/eyes/:command`, I would expect it to work normally just like any other handler. What you will get from doing that is the ability to just throw an error from the handler (or return a value from it), and the server will respond with a response of the shape `{success, result}` or `{success, error}`.

packages/eyes-cypress/src/plugin/server.js
```diff
@@ -2,45 +2,98 @@
...
+          errorDigest,
+          isInteractive: !req.query.isTextTerminal,
+          handleBatchResultsFile: makeHandleBatchResultsFile(req.query),
+        });
+        await waitForBatch(runningTests.tests, visualGridClient.closeBatch);
+        res.sendStatus(200);
+      } catch (ex) {
+        res.status(200).send({message: ex.message});
+      } finally {
+        process.exit(0);
```
amitzur, line 88: I don't think it's good to stop the process from here. The fact that the batch ended isn't coupled with the logic to end the process.

packages/eyes-cypress/src/plugin/server.js
```diff
@@ -2,45 +2,98 @@
...
+      }
+    });
+    process.on('disconnect', function() {
+      process.exit();
+    });
+  });
+  createAppAndStartServer.then(res => {
```
amitzur, line 95: I'm not sure this is more readable than using async/await.
What about doing something like `const {closeServer, eyesPort} = await new Promise(...)` ?

amitzur (changes requested):
Review is WIP. My main concerns are around when and how to stop the process, how to pass the config to it, and the way `waitForBatch` is implemented.

packages/eyes-cypress/src/plugin/hooks.js
SophieDePaula: @amitzur we still pass the hooks to `makePluginExport` so I kept the `function` just removed the parameter it's getting. I support that's what you meant :)

packages/eyes-cypress/src/plugin/hooks.js
SophieDePaula: Good point. I replaced `params` with `data` so that the `config` passes in the `body`.
When I send the `body` and `header` in an object like you suggested, I don't see the `headers` and `body` in the `request`

SophieDePaula: Please ignore my previous comment, I switched to `post` to take advantage of the endpoint we already have as you mentioned below

PR #733, opened 2022-01-11
side-eyes: fix check element command

tourdedave:
When running tests on the Visual Grid with SIDE Eyes, `eyes check element` commands error.

The error `IllegalArgument: logger is null or undefined` is briefly presented in the GUI briefly before it is overwritten by the output from the test ending (e.g., `Failed to run visual test`). The error is caught in `background/index.js` and not logged, so it doesn't show up in the background page.


SophieDePaula (approved):
Looks good to me 💯

tourdedave:
Merging. No need to add a test since this is covered by `side-eyes/tests/applitools-ext.side`. It needs to be run both in classic and VG mode (manually, unfortunately).

PR #738, opened 2022-01-13
Support lazy loading as part of CheckSettings

tourdedave:
[Trello](https://trello.com/c/BIHFMhVk)

- [x] eyes-api
- [x] types
- [x] snippets
- [x] eyes-sdk-core (classic)
- [x] eyes-sdk-core (VG)
- [x] coverage tests
- [x] update readmes
- [x] resolve SDK build error ([link](https://github.com/applitools/eyes.sdk.javascript1/pull/738#issuecomment-1036492078))
- [x] finalize decision on whether to keep a boolean as a valid option in eyes-api (waiting on input from @amitzur)
- [x] decide on when to run the script as part of grid execution (waiting on input from @amitzur) ([link](https://github.com/applitools/eyes.sdk.javascript1/pull/738#discussion_r797897528))
- [x] finalize text update for the SDK readmes ([link](https://github.com/applitools/eyes.sdk.javascript1/pull/738#discussion_r813049404))
- [x] decide whether to keep `lazyLoadBeforeCapture` or change it to something more elegant ([link](https://github.com/applitools/eyes.sdk.javascript1/pull/738#discussion_r813141966))
- [x] Implement once-before-all execution of lazy loading
- [x] decide when to perform input validation ([link](https://github.com/applitools/eyes.sdk.javascript1/pull/738#discussion_r812743437))
- [ ] README entries for eyes-cypress, eyes-testcafe, and eyes-storybook (if needed)

Some things to note:

- We're not updating `eyes-cypress` or `eyes-storybook` as part of this. `eyes-cypress` will get this when ported over to the universal SDK and use the core API with an alternative option for specifying sensible defaults for lazy loading (e.g., `eyes.check({lazyLoad: true})`) which has been implemented as part of this PR

- Coverage tests have been added along with JS emitters in the following PRs ([here](https://github.com/applitools/sdk.coverage.tests/pull/117) and [here](https://github.com/applitools/sdk.coverage.tests/pull/118)). One based off of master and one based off of the `universal-sdk` branch

- Baseline images have been saved for the `eyes-sdk-core` E2E tests (e.g., in JS SDK in Eyes) and the coverage tests (e.g., in Eyes SDK in Eyes)

- Coverage tests have been run and passed against `eyes-selenium` ([link](https://eyes.applitools.com/app/test-results/00000251758905687428?accountId=xIpd7EWjhU6cjJzDGrMcUw~~&display=details&top=00000251758905687428%2830%29))


packages/eyes-api/src/input/CheckSettings.ts
```diff
@@ -596,6 +602,15 @@ export class CheckSettingsFluent<TElement = unknown, TSelector = unknown> {
     return this
   }
 
+  lazyLoad(options: LazyLoadOptions): this {
+    this._settings.lazyLoad = {
+      scrollLength: options && options.scrollLength || 300,
+      waitingTime: options && options.waitingTime || 2000,
+      pageHeight: options && options.pageHeight || 15000
```
kyryloonufriiev, line 609: Just a suggestion. You can use modern syntax here since this is a type script and we will compile it anyway.
```suggestion
      scrollLength: options?.scrollLength ?? 300,
      waitingTime: options?.waitingTime ?? 2000,
      pageHeight: options?.pageHeight ?? 15000
```

kyryloonufriiev:
You also should update `types` package, it declares types for internal API, at the moment it looks like a copypasting, but it allows us completely unbind internal API from user-facing API.

packages/snippets/src/browser/lazyLoad.js
```diff
@@ -0,0 +1,36 @@
+const getDocumentSize = require('./getDocumentSize')
+
+function lazyLoad([{scrollLength, waitingTime, pageHeight} = {}] = []) {
```
kyryloonufriiev, line 3: I think this function should accept also an element to scroll. But pay attention that this should be a separate item in the array and not the one in an object, because of some sdks limitations.

packages/snippets/src/browser/lazyLoad.js
```diff
@@ -0,0 +1,36 @@
...
+  const preferredPageHeight = pageHeight < actualPageHeight ? pageHeight : actualPageHeight
+  const scrollsToAttempt = Math.ceil(preferredPageHeight / scrollLength)
+  const log = [
+    {
+      preferredPageHeight,
+      scrollsToAttempt,
+      params: {scrollLength, waitingTime, pageHeight},
+    },
+  ]
+  let p = Promise.resolve()
```
kyryloonufriiev, line 16: Looking at this I have understood that if we wanna call this function and wait until it is really done, we will have to implement a new method in all of the spec drivers (!). Because `executeAsyncScript` is a different method in wd protocol. But we can avoid it with a polling approach.

packages/snippets/src/browser/lazyLoad.js
```diff
@@ -0,0 +1,36 @@
...
+
+  for (let i = 0; i < scrollsToAttempt; i++) {
+    p = p.then(() => {
+      window.scrollTo({top: scrollLength * (i + 1)})
```
kyryloonufriiev, line 20: This function signature is not supported in IE. But do we need to support it in the first place?

packages/snippets/src/browser/lazyLoad.js
tourdedave: The initial requirements for this feature didn't say anything about an element. But I'll ask for clarification. Understood about the array arguments/limitations.

packages/snippets/src/browser/lazyLoad.js
tourdedave: Ah, good point. I just thought it was more declarative. I'll switch it.

packages/snippets/src/browser/lazyLoad.js
tourdedave: 🤯 okay.

packages/snippets/src/browser/lazyLoad.js
kyryloonufriiev: You can use our scrollTo snippet which accepts object with x and y, supports element scrolling, and handles different behavior in browsers.

packages/snippets/src/browser/lazyLoad.js
kyryloonufriiev: I don't see a reason to limit functionality on a default scrolling element only. There is no overhead to do so, because when you will execute this script you will already have scrollingElement in our driver.

tourdedave: Sure, but I checked with the business owner and they said it is not required right now.

tourdedave:
@kyryloonufriiev I'm having trouble getting a bundle that works with IE and promises. Maybe we can sync up on this tomorrow?

packages/snippets/src/browser/lazyLoad.js
tourdedave: I misunderstood your initial comment. I thought you meant accept an element to scroll *to*. Not accept an element to scroll *with*. That seems important to include. I'll follow up with the business.

tourdedave:
> @kyryloonufriiev I'm having trouble getting a bundle that works with IE and promises. Maybe we can sync up on this tomorrow?

Going to switch the implementation to polling/sync, which removes the need to polyfill to support promises or regeneratorRuntime (which also side-steps the challenges I've been running into getting a build that works for IE)

kyryloonufriiev (commented):
I still have a concern about the fact that we are going to support only window scrolling and don't have support for `scrollRootElement`. If a user is going to specify an SRE + lazeLoad, then he definitely wants an SRE to be the element we gonna scroll.

Also, Amit has convinced me that polling is the correct way to go, but I still think that we should adjust `executePollScript`, because it was initially designed for dom scripts only, so the default value for `pollTimeout` was chosen depend on that. I don't believe that `200ms` makes any sense if `waitingTimeout` is set to `5s`.

packages/snippets/rollup/commands.js
```diff
@@ -16,6 +17,7 @@ function configure() {
       snippet: './dist/index.js',
     },
     plugins: [
+      resolve(),
```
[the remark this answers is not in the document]
tourdedave: To use dom-shared. See `lazyLoadPollResult`

packages/eyes-api/test/unit/CheckSettings.spec.ts
```diff
@@ -36,4 +36,59 @@ describe('CheckSettings', () => {
...
+      scrollLength: 300,
+      waitingTime: 2000,
+      pageHeight: 15000
+    }
+
+    it('set lazyLoad with sensible defaults', () => {
+      const checkSettings = new CheckSettings().lazyLoad()
+      const lazyLoadOptions = checkSettings.toJSON().lazyLoad
+      assert.deepStrictEqual(lazyLoadOptions, defaultOptions)
+    })
```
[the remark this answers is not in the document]
tourdedave: I'm glad you say so. The duplication between eyes-api and core felt like a mis-step. And supporting a boolean felt like a natural direction to take things.

packages/eyes-sdk-core/lib/sdk/EyesVisualGrid.js
```diff
@@ -169,7 +172,21 @@ class EyesVisualGrid extends EyesCore {
...
+                  script: lazyLoad,
+                  args: [[lazyLoadOptions]],
+                },
+                poll: {
+                  script: lazyLoadPollResult,
+                },
+              }
+              await EyesUtils.executePollScript(this._logger, this._driver, scripts)
+            }
+          },
```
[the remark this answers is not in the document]
tourdedave: Where would you recommend putting it?

packages/snippets/src/browser/lazyLoad.js
```diff
@@ -0,0 +1,72 @@
...
+const LAZY_LOAD_KEY = 'lazyLoadResult'
+window[EYES_NAMESPACE] = window[EYES_NAMESPACE] || {}
+const lazyLoadPollResult = require('./lazyLoadPollResult')
+
+function currentScrollPosition() {
+  return {
+    x: window.pageXOffset,
+    y: document.documentElement.scrollTop,
+  }
+}
```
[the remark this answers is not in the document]
tourdedave: Great idea! Didn't know we already had this specified elsewhere.

tourdedave:
> I still have a concern about the fact that we are going to support only window scrolling and don't have support for `scrollRootElement`. If a user is going to specify an SRE + lazeLoad, then he definitely wants an SRE to be the element we gonna scroll.
> 
> Also, Amit has convinced me that polling is the correct way to go, but I still think that we should adjust `executePollScript`, because it was initially designed for dom scripts only, so the default value for `pollTimeout` was chosen depend on that. I don't believe that `200ms` makes any sense if `waitingTimeout` is set to `5s`.

I don't know what to tell you about the SRE. I agree with you, but as far as I know we presented this to the business and they said it's a good idea but it's not something we're going to do right now. As for polling time, makes sense. What do you recommend?

packages/snippets/rollup/commands.js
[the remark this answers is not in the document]
kyryloonufriiev: Ah. Didn't notice that you use it. This is overhead because the biggest part of this script is supposed to splint the response into chunks, this is never gonna be needed for this script. I also don't really see a reason to create a dedicated poll script. We have it for dom-snapshot because the script itself is so big that it was nice to have a small one for polling.

packages/eyes-sdk-core/lib/sdk/EyesVisualGrid.js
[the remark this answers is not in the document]
kyryloonufriiev: I would do it just before taking dom snapshot.

kyryloonufriiev:
Regarding SRE, okay if someone told you that we don't have to do it then it is fine. I think `pollTimeout` should be not less than `waitingTime`. Maybe you have a better juristic in mind.

packages/eyes-sdk-core/lib/sdk/EyesVisualGrid.js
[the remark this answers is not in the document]
tourdedave: Can you elaborate? `prepareBeforeCapture` is called just before `takeDomSnapshot` in `takeDomSnapshots`.

packages/snippets/rollup/commands.js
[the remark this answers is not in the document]
tourdedave: If we're mapping the snippet to a result object structure used throughout the code base then I figured it would be easier to leverage the existing polling script which handles this (even though there's functionality re: chunking that we'd never use here). I was initially hoping to use both `pollify` and `poll` from dom-shared (which is why I initially implemented it), but couldn't use `pollify` because it's async, . Once I realized I could only use 1/2 of the solution I still thought its was good enough to keep. I'm not opposed to backing it out if you feel strongly about this.

tourdedave:
> I think `pollTimeout` should be not less than `waitingTime`. Maybe you have a better juristic in mind.
> 
> Regarding SRE, okay if someone told you that we don't have to do it then it is fine.

A `pollTimeout` of `waitingTime` seems like a good place to start.

packages/eyes-sdk-core/lib/sdk/EyesVisualGrid.js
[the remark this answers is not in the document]
kyryloonufriiev: Ah. You are right. But should we do it before each of the snapshots? or only once before all of them. Also by changing `waitBeforeCapture` to `prepareBeforeCapture` you broke the storybook, just something to keep in mind.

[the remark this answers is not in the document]
tourdedave: That's good to know about storybook since we aren't planning to update it as part of this. Probably better to separate them into two functions then. I'm not sure which is better (before each snapshot or once before all). Which do you think makes more sense?

packages/snippets/rollup/commands.js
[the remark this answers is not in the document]
tourdedave: Also, re: having a dedicated polling script, I see value in it.

Specifically:
1. it follows a precedent that's used elsewhere (e.g., dom-snapshot)
2. it's easier to reason about as a result
3. it maintains encapsulation (e.g., we don't need to reference window variables in core)

packages/eyes-sdk-core/lib/sdk/EyesVisualGrid.js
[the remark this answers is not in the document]
kyryloonufriiev: If you use `layoutBreakpoints` with `lazyLoad`, it doesn't make a lot of sense for you to load content again and again once you resize a window, because the content is already loaded after the first time, I believe this is the case in 99%. This is why I think we should load content once before all dom snapshots. Also, it might be a reason for screenshots with a different number of elements loaded on different screen sizes. So I'm voting for the "once before all" option.

packages/snippets/rollup/commands.js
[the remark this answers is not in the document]
kyryloonufriiev: I see your point and it makes sense. It just doesn't feel right to me to use dom-shared in snippets. But maybe this is just because they are located in different repositories and dom-shared mostly associated with dom-scripts.
But my other point is that having a separate script to poll results doesn't really serve the purpose it serves in dom-scripts. We have a `pollResult` script there to make requests to the browser smaller. But in this case, we make them bigger with no good reason for that.

| Script | Chars |
|-------|------------|
| `lazyLoad` | 2948 |
| `lazyLoadPollResult` | 1760 |

`lazyLoadPollResult` script is more than half the size of the `lazyLoad` script, in fact, it is even bigger than the actual business logic of `lazyLoad` because the `lazyLoadPollResult` script is called inside `lazyLoad`.

So I propose to smush together `lazyLoad` and `lazyLoadPollResult`, avoid using `poll` from `dom-shared` because it is too fat for such a small script. But you can still use the same protocol as well as the `executePollScript` utility from the core.

packages/snippets/src/browser/lazyLoad.js
```diff
@@ -0,0 +1,67 @@
...
+const LAZY_LOAD_KEY = 'lazyLoadResult'
+window[EYES_NAMESPACE] = window[EYES_NAMESPACE] || {}
+const lazyLoadPollResult = require('./lazyLoadPollResult')
+const currentScrollPosition = require('./getElementScrollOffset')
+
+function lazyLoad([{scrollLength, waitingTime, pageHeight} = {}] = []) {
+  if (!scrollLength && !waitingTime && !pageHeight)
+    throw new Error(
+      'Incomplete set of arguments provided. Please provide scrollLength, waitingTime, and pageHeight',
+    )
```
kyryloonufriiev, line 11: Hmm. I understand the motivation, but I think this contradicts the idea of snippets, they should be minimalistic and pure logical units we execute to do a specific task. I don't think that input validation is something it should include, if you need to validate such a thing it should be done before you execute a snippet.

packages/snippets/src/browser/lazyLoad.js
```diff
@@ -0,0 +1,67 @@
...
+        scrollAndWait(scrollAttempt + 1)
+      }, waitingTime)
+    }
+
+    scrollAndWait()
+    return lazyLoadPollResult()
+  } catch (error) {
+    window[EYES_NAMESPACE][LAZY_LOAD_KEY] = {status: 'ERROR', error}
+    return error
+  }
```
kyryloonufriiev, line 64: If you are going to handle errors here you also should catch them inside the `setTimeout` callback, because this is where all of the logic happens.

packages/snippets/src/browser/lazyLoad.js
```diff
@@ -0,0 +1,67 @@
...
+            scrollAttempt,
+            x,
+            y,
+            msSinceStart: Date.now() - start,
+          })
+          window[EYES_NAMESPACE][LAZY_LOAD_KEY] = {value: log}
+          return
+        }
+        window.scrollTo(startingScrollPosition.x, scrollLength * scrollAttempt)
+        const {x, y} = currentScrollPosition()
```
kyryloonufriiev, line 46: consider using the `scrollTo` snippet, because it was already written to return the current scroll position after scroll. Also, it handles `scroll-behavior: smooth` style property (it will break your logic here).

packages/snippets/src/browser/lazyLoad.js
tourdedave: Okay. Will do. I added error handling like this so it matched what is expected by the polling script in EyesUtils in core

packages/eyes-sdk-core/lib/sdk/EyesVisualGrid.js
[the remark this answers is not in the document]
tourdedave: @amitzur thoughts? ☝️

tourdedave:
Updated the snippet and readme. Need to revisit the infinite scroll test in IE since the test is failing. I added details as to what's happening in the comments of the test.

tourdedave:
I'm having trouble getting a build on an SDK with everything wired up.

I've gone to each package (in the following order) `types`, `eyes-api`, `eyes-sdk-core`, `eyes-selenium` and done the following.

1. Linked deps (with `file:../path/to/package`) when appropriate
2. `rm -rf node_modules; yarn`
3. `yarn build` when available

Everything works until I get to `eyes-selenium`. I get the following when trying to build.

```
/Users/me/Documents/_dev/applitools/eyes.sdk.javascript1/packages/eyes-selenium/node_modules/typescript/lib/typescript.js:114713
                throw e;
                ^

TypeError: Cannot read properties of undefined (reading 'length')
    at createTypeMembers (/Users/me/Documents/_dev/applitools/eyes.sdk.javascript1/packages/eyes-selenium/node_modules/@applitools/api-extractor/dist/transformer.js:446:122)
    at createTypeLiteralNode (/Users/me/Documents/_dev/applitools/eyes.sdk.javascript1/packages/eyes-selenium/node_modules/@applitools/api-extractor/dist/transformer.js:339:25)
    at createTypeNode (/Users/me/Documents/_dev/applitools/eyes.sdk.javascript1/packages/eyes-selenium/node_modules/@applitools/api-extractor/dist/transformer.js:311:20)
    at createPropertyDeclaration (/Users/me/Documents/_dev/applitools/eyes.sdk.javascript1/packages/eyes-selenium/node_modules/@applitools/api-extractor/dist/transformer.js:575:26)
    at createTypeMembers (/Users/me/Documents/_dev/applitools/eyes.sdk.javascript1/packages/eyes-selenium/node_modules/@applitools/api-extractor/dist/transformer.js:453:33)
    at createTypeLiteralNode (/Users/me/Documents/_dev/applitools/eyes.sdk.javascript1/packages/eyes-selenium/node_modules/@applitools/api-extractor/dist/transformer.js:339:25)
    at createTypeNode (/Users/me/Documents/_dev/applitools/eyes.sdk.javascript1/packages/eyes-selenium/node_modules/@applitools/api-extractor/dist/transformer.js:311:20)
    at createPropertyDeclaration (/Users/me/Documents/_dev/applitools/eyes.sdk.javascript1/packages/eyes-selenium/node_modules/@applitools/api-extractor/dist/transformer.js:575:26)
    at createTypeMembers (/Users/me/Documents/_dev/applitools/eyes.sdk.javascript1/packages/eyes-selenium/node_modules/@applitools/api-extractor/dist/transformer.js:453:33)
    at createTypeLiteralNode (/Users/me/Documents/_dev/applitools/eyes.sdk.javascript1/packages/eyes-selenium/node_modules/@applitools/api-extractor/dist/transformer.js:339:25)
error Command failed with exit code 1.
```

I'm running Node 16.13.1.

tourdedave:
Actually, I'm getting the same error when trying to build `eyes-selenium` with `eyes-api@1.1.7`. It works with `eyes-api@1.1.6`.

kyryloonufriiev:
@tourdedave Works for me. Verify if you have the latest version of the `api-extractor` in this package.

tourdedave:
> @tourdedave Works for me. Verify if you have the latest version of the `api-extractor` in this package.

That works, thanks!

packages/eyes-api/src/input/CheckSettings.ts
```diff
@@ -596,6 +598,11 @@ export class CheckSettingsFluent<TElement = unknown, TSelector = unknown> {
     return this
   }
 
+  lazyLoad(options: boolean | LazyLoadOptions = true): this {
```
amitzur, line 601: I spoke with Ben. Neither of us think that it makes sense to support `boolean` at the user API level.
So the default value IMO should just remain undefined (i.e. `options` is optional), and if it's undefined then `this._settings.lazyLoad = true`

packages/eyes-sdk-core/lib/sdk/EyesVisualGrid.js
```diff
@@ -169,7 +172,26 @@ class EyesVisualGrid extends EyesCore {
           getEmulatedDevicesSizes: this._getEmulatedDevicesSizes,
           getIosDevicesSizes: this._getIosDevicesSizes,
           showLogs,
-          waitBeforeCapture: () => utils.general.sleep(waitBeforeCapture),
+          waitBeforeCapture: async () => {
```
amitzur, line 175: was it not readable the way it was?

packages/eyes-sdk-core/lib/utils/takeDomSnapshots.js
```diff
@@ -15,13 +15,15 @@ async function takeDomSnapshots({
   getEmulatedDevicesSizes,
   getIosDevicesSizes,
   waitBeforeCapture,
+  lazyLoadBeforeCapture,
```
amitzur, line 18: I think I would not create another option. It's defined and used exactly like `waitBeforeCapture`. So I think a cleaner API for `takeDomSnapshots` would be to change `waitBeforeCapture` option to `beforeCapture` and the consumer should provide the functionality that they want to run before each capture. WDYT?

packages/eyes-sdk-core/lib/utils/takeDomSnapshots.js
```diff
@@ -65,7 +68,9 @@ async function takeDomSnapshots({
     try {
       await driver.setViewportSize({width: requiredWidth, height: viewportSize.height})
       if (waitBeforeCapture) await waitBeforeCapture()
+      if (lazyLoadBeforeCapture) await lazyLoadBeforeCapture()
     } catch (err) {
+      console.log(err)
```
amitzur, line 73: probably you want to use `logger` here

packages/types/src/setting.ts
```diff
@@ -74,6 +75,7 @@ export type CheckSettings<TElement, TSelector> = MatchSettings<RegionReference<T
     renderId?: string
     variationGroupId?: string
     timeout?: number
+    lazyLoad?: boolean | LazyLoadOptions
```
amitzur, line 78: As mentioned in the comment in `CheckSettings.ts`, need to remove `boolean`

packages/snippets/src/browser/lazyLoad.js
```diff
@@ -0,0 +1,70 @@
...
+        maxAmountToScroll,
+        scrollLength,
+        waitingTime,
+        startingScrollPositionX: startingScrollPosition.x,
+        startingScrollPositionY: startingScrollPosition.y,
+      })
+      const start = Date.now()
+
+      function scrollAndWait({doneScrolling, previousScrollResult = {}} = {}) {
+        setTimeout(() => {
```
amitzur, line 27: do you intentionally wait before the first iteration? It's no biggy. I was just expecting the function to immediately perform the operation, and then call `setTimeout(() => scrollAndWait(...), waitingTime)`.

packages/snippets/src/browser/lazyLoad.js
```diff
@@ -0,0 +1,70 @@
...
+          try {
+            if (doneScrolling) {
+              const {x, y} = scrollTo([undefined, startingScrollPosition])
```
amitzur, line 30: This means setting `scroll-behavior` to 'auto' and then reverting it on each iteration. Of course, if it was already `auto` then that's a non-issue.
Ok, I'm fine with it. But let's add a comment that states that this is the behavior.

packages/eyes-sdk-core/test/e2e/lazyLoad.spec.js
```diff
@@ -0,0 +1,72 @@
...
+    name: 'eyes-core',
+    version: require('../../package.json').version,
+    spec,
+    VisualGridClient,
+  })
+
+  afterEach(async () => {
+    if (destroyDriver) await destroyDriver()
+    if (eyes) await eyes.abort()
+    manager.closeAllEyes()
```
amitzur, line 18: `await` ?

packages/eyes-sdk-core/test/e2e/lazyLoad.spec.js
```diff
@@ -0,0 +1,72 @@
...
+  })
+
+  describe('vg', () => {
+    beforeEach(async () => {
+      ;[driver, destroyDriver] = await spec.build({browser: 'chrome'})
```
amitzur, line 23: I would move this to a `beforeEach` outside of `describe`. Actually, since each describe has only 1 test, I thinks all `it`s can be flat under the main `describe`. I don't like nested suites unless it's absolutely necessary :)

amitzur (changes requested):
Great and clean!
I love the readme, it's very well structured.
What happens if the driver is already focused on an inner iframe? Should we focus on the main context before the snippet? And if we do, do we need to explicitly focus it back? Or will that be managed by core+driver? cc @kyryloonufriiev 

Requested changes mainly because of the boolean parameter in eyes-api.

Dave, what did you mean by the following question?
> decide on when to run the snippet as part of grid execution in eyes-sdk-core

packages/eyes-api/CHANGELOG.md
```diff
@@ -151,4 +153,4 @@
 
 ## 0.0.2 - 2021/3/24
 
-- initial implementation of API layer for js eyes SDKs
\ No newline at end of file
+- initial implementation of API layer for js eyes SDKs
```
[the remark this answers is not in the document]
tourdedave: I removed a trailing new line

```
-- initial implementation of API layer for js eyes SDKs
\ No newline at end of file
+- initial implementation of API layer for js eyes SDKs
```

packages/eyes-nightwatch/README.md
```diff
@@ -969,4 +995,4 @@ Possible values for `iosVersion` are:
 
 - `IosVersion.LATEST` - the latest iOS version that's supported by the UFG
 - `IosVersion.LATEST_ONE_VERSION_BACK'` - one version prior to the latest version
-- `undefined` - the UFG's default
\ No newline at end of file
+- `undefined` - the UFG's default
```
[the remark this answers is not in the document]
tourdedave: Looks like same as the other one. Removed newline at the end of the file.

```
-- `undefined` - the UFG's default
\ No newline at end of file
+- `undefined` - the UFG's default
```

packages/eyes-sdk-core/lib/utils/takeDomSnapshots.js
tourdedave: This was the original direction taken, but Kyrylo pointed out that this would have been a breaking change for eyes-storybook. Since we weren't planning to update eyes-storybook as part of this work, I reverted the change and implemented it as a new option.

packages/snippets/src/browser/lazyLoad.js
tourdedave: I guess I interpreted it differently and didn't give it a whole lot of thought.

packages/eyes-sdk-core/lib/utils/takeDomSnapshots.js
tourdedave: A choose-your-own-adventure moment:

a) leave it as-is and update it later or b) change it now and update eyes-storybook

WDYT?

packages/eyes-sdk-core/lib/sdk/EyesClassic.js
```diff
@@ -120,6 +122,22 @@ class EyesClassic extends EyesCore {
       },
     }
 
+    const lazyLoadOptions = CheckSettingsUtils.makeLazyLoadOptions(this._checkSettings.lazyLoad)
```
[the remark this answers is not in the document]
kyryloonufriiev: I don't think input validation should happen in `makeLazyLoadOptions` at all. You are right that this happens very deep in our business logic. This is why I believe that validation should happen once we get data from the user, which means in `check` function (the one in `new` folder). There is also another direction - validate on api level.

packages/eyes-sdk-core/lib/utils/takeDomSnapshots.js
kyryloonufriiev: @amitzur @tourdedave I don't think my point was about the storybook. The update would cost us just to rename the function in the storybook SDK. 
I was concerned about the fact we will run it before each `takeDomSnapshot` because in the case of `layoutBreakpoints` it could be lots of snapshots, so lots of scrolling and waiting through the same content again and again.

packages/eyes-sdk-core/lib/sdk/EyesClassic.js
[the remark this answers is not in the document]
tourdedave: I'll take a look at both. Thanks!

packages/eyes-sdk-core/lib/utils/takeDomSnapshots.js
tourdedave: I discussed this offline with Kyrylo. Since the behavior for `waitBeforeCapture` and lazy load need to happen at different times, we will keep them as separate functions.

pending-changes.yaml
```diff
@@ -10,17 +10,17 @@ Apollo SDK:
       An example of how to inline `markdown code highlights`
 "@applitools/eyes-universal":
   feature:
-    - 
+    - Add the ability for the SDK to lazy load the page prior to performing a check window
```
tourdedave, line 13: @benbabayoff8

tourdedave:
@kyryloonufriiev Please review when you get a chance. Thanks!

PR #753, opened 2022-01-18
Fix global hooks override in Cypress

amitzur:
[Trello](https://trello.com/c/AK2jwzoj)
The bug was happening when the user only calls **one** global hook. We didn't have the granularity to detect this, and it led to our own global hook not being called.
I changed the boolean `isGlobalHookCalledFromUserHandler` to be a map from `eventName` to `boolean`.


SophieDePaula (approved):
Looks good :+1:

PR #758, opened 2022-01-24
Storybook

chaim-aharonson-applitools:


packages/eyes-storybook/src/processResults.js
```diff
@@ -52,27 +56,43 @@ function processResults({results = [], totalTime, testConcurrency}) {
...
+      `\nA total of ${unresolved.length} difference${_toMany(unresolved, [
+        's were',
+        ' was',
+      ])} found and ${errors.length} stor${_toMany(errors, [
+        'ies',
+        'y',
+      ])} failed for ${_toMany(errors, ['', 'an '])}unexpected error${_toMany(errors)}.`,
     );
-  } else if (passedOrNew.length) {
+  } else if (passedOrNew.length && !warnForUnsavedNewTests) {
```
amitzur, line 81: I would simplify the condition a bit:

```
} else if (warnForUnsavedNewTests) {
...
} else if (passedOrNew.length) {
...
}
```

packages/eyes-storybook/src/processResults.js
```diff
@@ -52,27 +56,43 @@ function processResults({results = [], totalTime, testConcurrency}) {
...
     outputStr += chalk.green(`\nNo differences were found!`);
   }
+  if (warnForUnsavedNewTests) {
+    const countText =
+      newTestsSize > 1
+        ? `are ${newTestsSize} new tests`
+        : `is a new test: '${newTests[0].getName()}'`;
+    outputStr += chalk.red(
+      `\n'SaveNewTests' was set as false and there ${countText}. Please approve ${_toMany(
```
amitzur, line 90: Nice output! Text correction:
```
`\n'saveNewTests' was set to false and there ${countText}. Please approve ${_toMany(
```

packages/eyes-storybook/src/processResults.js
```diff
@@ -52,27 +56,43 @@ function processResults({results = [], totalTime, testConcurrency}) {
...
+        newTestsSize,
+        ['their', 'its'],
+      )} baseline${_toMany(newTestsSize)} in Eyes.\n`,
```
amitzur, line 93: I would say `in Eyes` --> `in the Eyes dashboard`. Actually, could you consult with Ben about it? We use the term "Test manager" in our formal documentation (but nowhere in the SDK readme's. There we use the term "Eyes dashboard")

packages/eyes-storybook/src/processResults.js
```diff
@@ -2,17 +2,21 @@
...
+const utils = require('@applitools/utils');
 const uniq = require('./uniq');
 const concurrencyMsg = require('./concurrencyMsg');
 
-function processResults({results = [], totalTime, testConcurrency}) {
+function processResults({results = [], totalTime, testConcurrency, saveNewTests = true}) {
   let outputStr = '\n';
   const formatter = new TestResultsFormatter();
-
+  const _toMany = utils.general.toMany;
```
amitzur, line 12: variable names that start with `_` are considered unused. This is how to avoid an eslint error for intentional cases of not using a variable. In this case it's not needed because the variable is used.

packages/utils/src/general.ts
```diff
@@ -73,3 +73,14 @@ export function toJSON(object: Record<PropertyKey, any>, props?: string[] | Reco
 export function toString(object: Record<PropertyKey, any>): string {
   return `${this.constructor.name} ${JSON.stringify(object, null, 2)}`
 }
+
+export function toMany(object: [] | number, config?: [manyCase: string, singleCase: string]): string {
```
amitzur, line 77: I initially thought this isn't really needed. And I thought `pluralize` would be a better name.
TBH it's a bit weird to return only the suffix. But after seeing how it's used in `processResults` I'm fine with it.

Regarding the configuration, I also thought a better practice would be an object `{manyCase, singleCase}`. It would be better both for understanding, and for specifying only a `singleCase` if that's really a use case. But it's very verbose for the caller, especially since it's likely to be used inside a template literal.

@kyryloonufriiev @SophieDePaula @tourdedave let's hear your opinion as well?

packages/eyes-storybook/test/it/processResults.test.js
```diff
@@ -274,4 +274,95 @@ describe('processResults', () => {
...
+            appUrls: {batch: 'https://eyes.com/results'},
+          }),
+        ],
+      },
+    ];
+    const {outputStr, exitCode} = processResults({
+      results,
+      totalTime: 10000,
+      concurrency: 1,
+      saveNewTests: true,
```
amitzur, line 363: you can also not send anything and this way you will also check the default value

amitzur (changes requested):
Great job!
There are just a few minor comments, and I asked for more opinions on `isMany`. But the code is good and to the point.

packages/utils/src/general.ts
```diff
@@ -73,3 +73,14 @@ export function toJSON(object: Record<PropertyKey, any>, props?: string[] | Reco
...
+  const count = types.isArray(object) ? object.length : types.isNumber(object) ? object : 0;
```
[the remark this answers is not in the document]
chaim-aharonson-applitools: the '0' case is a fallback, maybe in ts we don't need that.
I think it's more 'developer friendly' and shorter to avoid ```length``` and send to the method 'what u have'

packages/eyes-storybook/src/processResults.js
chaim-aharonson-applitools: added 'dashboard' seams clearer (will update the ticket)

packages/utils/src/general.ts
chaim-aharonson-applitools: ```pluralize``` it is from now.
Config was defined as an Array for shorter code and due to the fact that if 'config' is used it **must** send the two values. additionally, ```manyCase``` is first because I saw that all places which this was used, the first case handled many, aka: ```err.length>0?'s':''```

packages/eyes-storybook/src/processResults.js
```diff
@@ -52,25 +56,40 @@ function processResults({results = [], totalTime, testConcurrency}) {
...
+        ? `are ${newTestsSize} new tests`
+        : `is a new test: '${newTests[0].getName()}'`;
+    outputStr += chalk.red(
+      `\n'saveNewTests' was set to false and there ${countText}. Please approve ${pluralize(
+        newTestsSize,
+        ['their', 'its'],
+      )} baseline${pluralize(newTestsSize)} in Eyes dashboard.\n`,
     );
-  } else if (passedOrNew.length) {
+  } else if (passedOrNew.length && !warnForUnsavedNewTests) {
```
amitzur, line 92: no need for `!warnForUnsavedNewTests` since it's already in the previous `if/else`

amitzur (approved):
Great! See only 1 comment inline, and you can publish:
1) utils
2) storybook

PR #759, opened 2022-01-26
Convert Cypress to work on top of universal server

SophieDePaula:


packages/eyes-cypress-universal/src/plugin/pluginExport.js
```diff
@@ -6,25 +6,30 @@ const path = require('path');
...
+    let closeEyesServer;
     const pluginModuleExports = pluginModule.exports;
     pluginModule.exports = async function(...args) {
       const {localServerPort, closeServer} = await startServer();
 
       // fork a new process that start universal server
-      const server = childProcess.fork(path.resolve(__dirname, 'makeUniversalServer.js'));
-      server.send('start server');
-
+      const server = childProcess.spawn('node', [path.resolve(__dirname,'../../node_modules/@applitools/eyes-universal/dist/cli.js')], {
```
amitzur, line 15: executable path is `path.resolve(require.resolve('@applitools/eyes-universal'), 'dist/cli.js')`

amitzur: maybe even `require.resolve('@applitools/eyes-universal/dist/cli.js')`? should be checked

packages/eyes-cypress-universal/src/browser/spec-driver.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export function setViewportSize(vs: any) {
+  cy.viewport(vs.size.width, vs.size.height);
+}
+
+export function findElement(element: Selector) {
+  // need to check if the element is selector or HTMLElement
+
+  return new Promise(resolve => { 
+    cy.get(element.selector).then((el) => {
```
amitzur, line 94: `el` in this case is the jQuery object, and I think the return value should be `el[0]`. Also, to show it's the jQuery object do this:
```
cy.get(element.selector).then($el => resolve($el[0]))
```

packages/eyes-cypress-universal/src/browser/spec-driver.ts
```diff
@@ -0,0 +1,104 @@
...
+export function isElement(element: Element) {
+  // return typeof element === 'object ' && element.hasOwnProperty('specWindow');
+  return Cypress.dom.isElement(element);
+}
+
+export function isSelector(selector: Selector) {
+  return selector.hasOwnProperty('selector');
+}
+
+export function getViewportSize(driver: Driver) {
```
amitzur, line 78: you can remove this function. If spec driver doesn't support this, then core will run a js snippet to get it:
https://github.com/applitools/eyes.sdk.javascript1/blob/9c6cb79c7eb048a04e64a09c5ec1c7173aaeaa9a/packages/driver/src/driver.ts#L350

packages/eyes-cypress-universal/src/browser/commands.js
```diff
@@ -133,66 +139,69 @@ Cypress.Commands.add('eyesOpen', function(args = {}) {
...
-  socket.request('Eyes.check', {
-    eyes,
-    settings: checkArgs,
-    config: config,
-  });
+  })
 });
 
-Cypress.Commands.add('eyesClose', () => {
+Cypress.Commands.add('eyesClose', async () => {
```
amitzur, line 197: I believe you should get a warning or even error in the browser's console, since a custom command should not return a native promise.

packages/eyes-cypress-universal/src/plugin/pluginExport.js
SophieDePaula: That worked!

packages/eyes-cypress-universal/src/browser/spec-driver.ts
SophieDePaula: keeping this as Kyrylo mentioned that it's better to have it if possible.

SophieDePaula:
@amitzur I added support for most functions in `spec-driver` and now I use the `context` I am getting from `universal`.
Feel free to start going over `spec-driver`.

packages/eyes-cypress-universal/src/browser/spec-driver.ts
SophieDePaula: Using `document` and not `cy` object, therefore, is not relevant anymore.

packages/eyes-cypress-universal/test/e2e/cypress-run-global-hooks.test.js
```diff
@@ -62,8 +62,8 @@ describe('global hooks', () => {
     expect(err.stdout).to.contain('Eyes-Cypress detected diffs or errors');
   });
 
-  it('works with cypress version 3 (< 6.2.0 no global hooks available)', async () => {
-    await pexec(`npm install cypress@3`);
+  it('works with cypress version 4 (< 6.2.0 no global hooks available)', async () => {
```
SophieDePaula, line 65: @amitzur fyi I change this run on `Cypress` version `4` as it seems that `Cypress` version `3` has a bug where when we set an `object` as a value of a property in the return configuration to the `Cypress` object it crashes.

amitzur: Interesting. This means that we dont support Cypess 3 in this branch. It explains why things were done the way they are (eyesIsDisabled etc.). We dont really know if this is used by customers :(

packages/eyes-cypress-universal/src/browser/commands.js
```diff
@@ -107,21 +110,23 @@ Cypress.Commands.add('eyesOpen', function(args = {}) {
...
       manager,
       driver: driverRef,
-      config: Object.assign({testName}, config, {
-        dontCloseBatches:
-          !shouldUseBrowserHooks || Cypress.config('appliConfFile').dontCloseBatches,
-      }),
+      config: Object.assign(
+        {testName, dontCloseBatchesconfig: !shouldUseBrowserHooks},
+        args,
+        Cypress.config('appliConfFile')
```
SophieDePaula: @amitzur what about all the configuration from `applitools.config.js` file , the mapping of `browser` to `browserInfo` and setting a default browser if the user did not specify any?

packages/eyes-cypress-universal/src/browser/eyesCheckMapping.js
```diff
@@ -1,40 +1,5 @@
...
-    Object.assign(checkArgs, {tag: args});
-  }
-
-  for (const prop of propsToDelete) {
-    delete appliConfFile[prop];
-  }
-
-  return Object.assign({}, appliConfFile, toCheckWindowConfiguration(checkArgs));
+function eyesCheckMapValues({args}) {
+  return toCheckWindowConfiguration(args);
```
SophieDePaula, line 2: @amitzur the new change does not take `applitools.config.file` into account.

packages/eyes-cypress-universal/src/browser/commands.js
```diff
@@ -91,7 +94,7 @@ Cypress.Commands.add('eyesOpen', function(args = {}) {
       socket.emit('Core.makeSDK', {
         name: 'eyes.cypress',
         version: require('../../package.json').version,
-        commands: Object.keys(spec),
+        commands: Object.keys(spec).concat(['isSelector', 'isDriver', 'isElement']), // TODO fix spec.isSelector and spec.isDriver and spec.isElement in driver utils
```
SophieDePaula, line 98: @amitzur why do we need to `concat` this list of methods? we get everything from `spec`

PR #762, opened 2022-01-27
Cypress set limitation of concurrency on local server

SophieDePaula:
This PR attempts to fix the issue described [here](https://trello.com/c/RJnRpGLj/1140-capital-one-i-cypress-i-failures-blocking-nightly-build)

In order to prevent our local server from crashing due to multiple parallel requests from the browser, we now limit the parallel requests to 100.

Without this change, the test I added `overloadServer.test.js`, fails on `ECONNRESET`


kyryloonufriiev (approved):
👍🏼

PR #764, opened 2022-02-02
Set cut provider

chaim-aharonson-applitools:


packages/eyes-sdk-core/test/unit/new/MakeCheck.spec.js
```diff
@@ -0,0 +1,37 @@
...
+            await eyes.check()
+            const result = await eyes.getCut()
+            assert.strictEqual(result, initialCut)
+        })
+        it('setCutViaCheck', async () => {
+            await eyes.check({ config: { cut: updatedCut } })
+            const result = await eyes.getCut()
+            assert.strictEqual(result, updatedCut)
+        })
+    })
```
kyryloonufriiev, line 36: This is not exactly what I was envisioning. Why would you need `getCut` and `setCut` in the core api. What you really need is to read the cut value directly from your "fake eyes" object that you should pass to the `makeCheck` function. I also don't think that `EyesMockWithCutFunc` in `testUtils` makes sense, since this kind of mock object is valid only for this test.

kyryloonufriiev (changes requested):

packages/eyes-sdk-core/test/unit/new/MakeCheck.spec.js
chaim-aharonson-applitools: get and set cut as functions and not directly: IMHO it is more 'elegant', readable and consist with real **eyes** API, but, if you think I should remove and call directly I will do so.
Moved the relevant functions from ```testUtils```

amitzur (approved):
:100: @chaim-aharonson-applitools you can merge this into master and publish:
1) eyes-sdk-core
2) eyes-webdriverio

(basically all SDK's should be released with this fix, but only 1 customer is waiting for it, and they are using webdriverio)

PR #765, opened 2022-02-04
wdio: Add error handling to element lookup on non-standard driver implementations

tourdedave:
[Trello](https://trello.com/c/DX8UlrTk)


packages/spec-driver-webdriverio/src/spec-driver.ts
```diff
@@ -176,7 +176,19 @@ export async function findElement(
...
+    if (
+      /element could not be located/i.test(error.message) ||
+      /cannot locate an element/i.test(error.message) ||
+      /wasn\'t found/i.test(error.message)
+    ) {
+      return null
+    }
+    throw error
+  }
   return !utils.types.has(element, 'error') ? element : null
```
kyryloonufriiev, line 192: I would prefer something like this, becasue it is a little easy to read, on my opinion.
```suggestion
  try {
    const element = await root.$(selector)
    return !utils.types.has(element, 'error') ? element : null
  } catch (error) {
    if (
      /element could not be located/i.test(error.message) ||
      /cannot locate an element/i.test(error.message) ||
      /wasn\'t found/i.test(error.message)
    ) {
      return null
    }
    throw error
  }
```

packages/spec-driver-webdriverio/test/e2e/spec-driver.spec.ts
```diff
@@ -375,6 +375,35 @@ describe('spec driver', async () => {
...
+    it('findElement(non-existent-errors)', async () => {
+      await findElement({input: {selector: 'non-existent-generic-error'}, expected: null})
+      await findElement({input: {selector: 'non-existent-specific-error-v1'}, expected: null})
+      await findElement({input: {selector: 'non-existent-specific-error-v2'}, expected: null})
+    })
+
+    it('findElement(valid-error)', async () => {
+      assert.rejects(async () => await spec.findElement(browser, 'valid-error'), Error)
+    })
+  })
```
kyryloonufriiev, line 405: Don't you think this kind of test should be implemented under the unit folder?

kyryloonufriiev (commented):
Looks reasonable. But I also think that we also should verify how perfecto works with other frameworks. I always try to keep consistency across all of the spec drivers.

packages/spec-driver-webdriverio/test/e2e/spec-driver.spec.ts
tourdedave: Sure

tourdedave:
I spent some time trying to add support for Perfecto into `test-utils`, but wasn't able to get a successful session start. I'll circle back to this later.

tourdedave:
> I spent some time trying to add support for Perfecto into `test-utils`, but wasn't able to get a successful session start. I'll circle back to this later.

Got it working.

tourdedave:
@kyryloonufriiev`spec-driver-selenium` is good. `findElement` returns `null` correctly on element not found when running native Android on Perfecto.

tourdedave:
Merging this. I'll work on a more cohesive e2e solution outside of this PR.

tourdedave:
(Also, Kyrylo added the new env var to our Github Actions)

PR #769, opened 2022-02-07
wdio: fix mis-classification of mobile browser tests as native

tourdedave:
[Trello](https://trello.com/c/xpTOZqWN)


tourdedave:
Discussed this with @kyryloonufriiev offline. While this fixes the acute issue, the correct fix is...

- to remove `getDriverInfo` from `spec-driver-webdriver` (relying instead on the work that `driver` does to determine `isNative` from the driver's capabilities), and
- to add a coverage test (since all SDKs could be suffer from this kind of misclassification - e.g., mobile web being classified as native)

PR #788, opened 2022-02-15
Improved SDK changelogs

tourdedave:
[Trello](https://trello.com/c/SgCznEud)

### TODOs

- [x] add a templated/central file where feature/bug-fix blurbs can be added for each SDK
- [x] verify pending changes added to central file as part of an SDK's release
- [x] create a changelog entry using pending changes file and a detailed list of updated dependencies
- [x] remove SDK feature/bug-fix entries from pending changes file after changelog entry created
- [x] drop commits from new changelog entries in favor of utility script (built into bongo)
- [x] verify no manual changelog entries made to an SDK as part of the release (it's not in the spec, but something we might want to do)
- [x] drop changelog support from internal packages
- [x] update scripts in SDKs to opt-in to the new changelog flow
- [ ] ~update GH actions workflow for verifying changelogs~ nothing to update since `bongo verify-changelog` was repurposed
- [ ] ~create a changelog entry using pending changes file and the latest universal server info (for non-JS universal SDKs)~ (will be handled later)

### Some relevant details

The new changelog functionality is opt-in through args for the relevant bongo commands:

```
yarn bongo deps --includeCommitHistoryInChangelog
yarn bongo preversion --verifyPendingChanges
yarn bongo version --withPendingChanges
```

The central file is YAML, not JSON like initially spec'd since it is more approachable for non-devs to work with. It lives in the root of the repo. You can see it [here](https://github.com/applitools/eyes.sdk.javascript1/blob/feat/changelogs%2B%2B/pending-changes.yaml).

A sample of changelog output can be seen [here](https://github.com/applitools/eyes.sdk.javascript1/blob/feat/changelogs%2B%2B/packages/eyes-selenium/sample-CHANGELOG.md#4583---2022218).

The updated dependencies sections is built using commit history. Auto commits from dependency bumps are expanded recursively to include the commits for the packages accessible in our mono that are listed as dependencies. So, for example, if we look at `@applitools/eyes-sdk-core` getting updated from `12.24.10` to `12.24.11` we would see the following in the commit history:

```sh
ff7e580c6 [auto commit] @applitools/eyes-sdk-core: upgrade deps
930d4e1df more info in RunnerStartedEvent
```

In the changelog we currently see this:

```md
- more info in RunnerStartedEvent
- updated to @applitools/dom-snapshot@4.5.12 (from 4.5.11)
- updated to @applitools/logger@1.0.9 (from 1.0.8)
```

But with this new functionality, we get the following commit output instead:

```sh
b7df0cb39 [utils] use git bush shell on windows
3935cb48f [utils] use default shell on windows
33bddbe4b [utils] use default shell on windows
0e9f2cc58 [utils] use default shell on windows
d10a69cba [utils] use shell options
4c5249da5 [utils] add support for windows bash command execution
f65c478f2 [logger] fix colors in extended logger
ff7e580c6 updated to @applitools/dom-snapshot@4.5.12 (from 4.5.11)
930d4e1df more info in RunnerStartedEvent
```

This includes the specific dependency bump info (for external deps) and all commits for the internal packages listed as dependencies for their given version bumps (e.g., core has logger as a dependency, and logger has utils as a dependency). What this looks like in the final output for an SDK bump would be something like this:

```md
### Updated dependencies
- updated to @applitools/eyes-sdk-core@12.24.11 (from 12.24.10)
  - b7df0cb39 [utils] use git bush shell on windows
  - 3935cb48f [utils] use default shell on windows
  - 33bddbe4b [utils] use default shell on windows
  - 0e9f2cc58 [utils] use default shell on windows
  - d10a69cba [utils] use shell options
  - 4c5249da5 [utils] add support for windows bash command execution
  - f65c478f2 [logger] fix colors in extended logger
  - ff7e580c6 updated to @applitools/dom-snapshot@4.5.12 (from 4.5.11)
  - 930d4e1df more info in RunnerStartedEvent
```

Also, when written to a changelog file, the shas get linked. You can see the end result in the sample changelog output [here](https://github.com/applitools/eyes.sdk.javascript1/blob/feat/changelogs%2B%2B/packages/eyes-selenium/sample-CHANGELOG.md#4583---2022218).


pending-changes.yaml
```diff
@@ -0,0 +1,45 @@
+"@applitools/eyes-cypress":
```
amitzur, line 1: I see value in having the first entry an example:
```
"Apollo SDK":
  feature:
    - Support including humans when launching to the moon
    - Better probability of successful landing
  bug-fix:
    - Avoid consuming all oxygen during takeoff
    - Remove kitchen to accommodate space for PlayStation
```

packages/sdk-release-kit/src/changelog/index.js
```diff
@@ -1,9 +1,19 @@
 const {getLatestReleaseEntries, verifyChangelog} = require('./query')
 const {writeReleaseEntryToChangelog, writeUnreleasedItemToChangelog} = require('./update')
+const {
+  emitUpdatedDepsEntry: makeUpdatedDepsEntry,
```
amitzur, line 4: why export the function in a different name than how you want it to be exported? :)

amitzur (commented):
Review is WIP (I actually went over only 6/20 files).
But in the meantime:
- The list in the description looks great.
- I don't think it's worth investing in preserving existing behavior around the changelog. I don't know if this is something you made extra effort to do, but I do think it makes sense to do the extra effort (if it's small enough) to clean up the code from existing behavior.

packages/sdk-release-kit/src/changelog/index.js
tourdedave: Consider it art :-)

I was trying on different function names to see what felt more readable in the cli. I was planning on changing it. I'll do it now.

tourdedave:
> * I don't think it's worth investing in preserving existing behavior around the changelog. I don't know if this is something you made extra effort to do, but I do think it makes sense to do the extra effort (if it's small enough) to clean up the code from existing behavior.

I didn't work hard to keep existing behavior. I just made the new behavior opt-in w/ args because it wasn't clear to me how we wanted to handle internal packages. I've been assuming that the pending-changes.yaml file is only for SDKs. Is that true?

amitzur:
> I've been assuming that the pending-changes.yaml file is only for SDKs. Is that true?
That is correct. Following our last adventures I'm happy to drop changelogs for internal packages, and maintain only a git history page for them.

So let's move this forward. What I understand is the following:
1. Developers release internal packages without maintaining a changelog for them. Instead, they need to add entries to `pending-changes.yaml` with text from an SDK perspective about their change.
2. When getting to release the SDK, the following will do the job:
  a) `bongo deps --includeCommitHistoryInChangelog` will add the internal git commits to the **Unreleased** section in the SDK's changelog.
  b) `bongo preversion --verifyPendingChanges` will verify there is something to paste in the features/bugfixes subsection of **Unreleased**.
  c) `bongo version --withPendingChanges` will paste the changes and cut a new section with the version number.

If so, that sounds great and let's push it forward. I'm thinking of 2 issues:
1) Currently, the git commits section has a lot of duplications. Running `bongo deps --includeCommitHistoryInChangelog` on Selenium today brings 200+ lines of commits, where the internal ones appear for every sub package. Do you think that's desired?
2) What if `preversion` fails? Will the changelog now contain garbage? Is it even an issue given that it runs only in CI?

Next steps:
- I'll continue with the code review
- We should agree on the points above
- Implement the remaining unchecked boxes (except non-JS SDK's which is not first priority)
- Merge, share knowledge with the team, and start using (aim for Tuesday).

tourdedave:
> That is correct. Following our last adventures I'm happy to drop changelogs for internal packages, and maintain only a git history page for them.

Okay, great! But what do you mean by `git history page`? Are you referring to the list of commits that get listed in an SDK changelog entry or something else?

In terms of your understanding, 1 & 2 are correct. One risk we haven't discussed/addressed is the potential for an SDK to be released with a changelog entry for a feature/bug fix that it doesn't contain (e.g., the internal package that lead to the pending changes entry for an SDK wasn't released before the SDK). Thoughts on how we want to address this?

Regarding the 2 issues you noted:

1. Who is the commits section for? SDK devs or someone else? From my perspective, it's SDK devs, and the question I was trying to answer was "of the code in our repo, what changes are included in a given version of an SDK?". Right now this isn't something that can be easily found in a consolidated way, even looking through the git history. All you see are automated commit messages of a dependency being updated. So then you need to go to the package that was updated and look through its changes, and guess what? You find yet another automated commit message for a dependency being updated, etc, etc. Listing all commits, including internal ones, was by design. I also built this functionality into `bongo`, maybe it's better suited there than in the changelog entry? Thoughts?

Regardless, there shouldn't be duplicates. If there are, then it's a bug. I'll take a look.

2. If `preversion` fails it should be OK because it's in a branch in CI that won't get merged.

amitzur:
> But what do you mean by `git history page`?

Something like this: https://github.com/applitools/eyes.sdk.javascript1/blob/master/packages/eyes-storybook/History.md

Regarding the duplications etc:
I think you were right in the way you presented it. The "duplications" are indeed part of the correct output.
Since the audience for this is SDK devs, let's drop this from the SDK changelog altogether, and maintain it as a utility script for developers.

> One risk we haven't discussed/addressed is the potential for an SDK to be released with a changelog entry for a feature/bug fix that it doesn't contain (e.g., the internal package that lead to the pending changes entry for an SDK wasn't released before the SDK). Thoughts on how we want to address this?

Good point. I think it should be solved by publishing the entire subtree (all internal pacakges) after merging a PR. That way we won't find ourselves in this situation at all.
I once wrote a verification that "no unreleased commits" are present for a pacakge. But it's not working great so it's no longer blocking release. We can talk about it independently from this task and maybe you can find a way to incorporate it better.

Let's move forward:
- Implement the remaining unchecked boxes (except non-JS SDK's which is not first priority)
- Merge, share knowledge with the team, and start using (aim for today).

PR #798, opened 2022-02-25
Add better support for Eyes.abort

tourdedave:
To provide feature parity with the Java SDK, so they can move to the universal-server.


amitzur (commented):
review for myself and @tourdedave

PR #804, opened 2022-03-04
Fix/screenshoter stitched screenshotting

tourdedave:
Fixes failing web and android tests.


amitzur:
@tourdedave I see what you did here. There are 2 commits in this story:
1. 6399103ab51f5b036ec7f8040583ff65a216d5b8 which was intended to fix DOM location when we found about incorrect values in page coverage. This was released from Kyrylo's machine as version 3.3.10
2. 53eaab499c037cb8c5daddaae89092ec06651731 this looks like an attempt to solve the problem better (and maybe address coverage test failures?). But it wasn't released.

So you reverted the changes introduced in commit (1) and modified the changed in commit (2) so that we don't use the main context scrolling element for moving to (0,0) before taking the screenshot.
Plus, add error handling if the image turns out 0x0 after cropping.

And the outstanding issues right now are:
1) Failing to start iOS tests in screenshoter
2) Failing coverage test: pageCoverage data is correct (which is the one that commit (1) was fixing).
3) Failing coverage test: should send dom and location when check region by selector fully with custom scroll root

We'll continue the investigation...

PR #806, opened 2022-03-11
Support data URI in Iframe

SophieDePaula:
Support data URI in Iframes.
[Trello card that first introduced this requirement](https://trello.com/c/XfWUEGRS/1225-canva-pocbroken-iframe-image?filter=due:notdue)


amitzur:
@SophieDePaula what should be done in this PR?

SophieDePaula:
@amitzur we should be able to support the use case in the Trello card I linked in the description.
So far I just created a test to reproduce the problem.

SophieDePaula:
@chaim-aharonson-applitools overall looks good to me, I will say that I am not as familiar with the custom attribute we need to send to the grid such as `data-applitools-src`.
Also, re `elementNode.setAttribute('srcdoc', newSrc)` , if I understand correcly it can either be `src` or `srcdoc`

chaim-aharonson-applitools:
> @chaim-aharonson-applitools overall looks good to me, I will say that I am not as familiar with the custom attribute we need to send to the grid such as `data-applitools-src`. Also, re `elementNode.setAttribute('srcdoc', newSrc)` , if I understand correcly it can either be `src` or `srcdoc`

That is correct, I'll be happy to discuss about it, this functioality was Intentionally ignored/deprecated, my flows worked well but will be happy to explain and hear your feedback

amitzur:
Done in PR #855

PR #809, opened 2022-03-17
Support batchId, batchName and batchSequenceName

SophieDePaula:
Support batchId, batchName and batchSequenceName from the config file, eyesOpen and environment variables.
This is to fix [this issue](https://trello.com/c/H5xJSTsD/1271-cypress-batchid-configuration-not-being-honored-in-new-cypress-version).
I initially thought that we should only support sending an object with the above properties.


packages/eyes-cypress/src/browser/eyesOpenMapping.js
```diff
@@ -1,6 +1,11 @@
 function eyesOpenMapValues({args, appliConfFile, testName, shouldUseBrowserHooks}) {
   let browsersInfo = args.browser || appliConfFile.browser;
   let accessibilitySettings = args.accessibilityValidation || appliConfFile.accessibilityValidation;
+  let batch = {
+    id: args.batchId || appliConfFile.batch.id,
```
amitzur, line 5: what about `args.batch.id` ?

packages/eyes-cypress/src/browser/eyesOpenMapping.js
```diff
@@ -1,6 +1,11 @@
...
+    name: args.batchName || appliConfFile.batch.name,
```
amitzur, line 6: what about `args.batch.name`?

packages/eyes-cypress/src/browser/eyesOpenMapping.js
```diff
@@ -1,6 +1,11 @@
...
+    sequenceName: args.batchSequenceName || appliConfFile.batch.sequenceName,
```
amitzur, line 7: what about `args.batch.sequenceName`?

packages/eyes-cypress/src/browser/eyesOpenMapping.js
```diff
@@ -1,6 +1,11 @@
 function eyesOpenMapValues({args, appliConfFile, testName, shouldUseBrowserHooks}) {
   let browsersInfo = args.browser || appliConfFile.browser;
   let accessibilitySettings = args.accessibilityValidation || appliConfFile.accessibilityValidation;
+  let batch = {
```
amitzur, line 4: why not `const`?

packages/eyes-cypress/test/unit/browser/eyesOpenMapping.test.js
```diff
@@ -19,6 +19,9 @@ describe('eyes open mapping', () => {
       accessibilityValidation: {level: 'AAA', guidelinesVersion: 'WCAG_2_0'},
       matchLevel: 'Layout',
       enablePatterns: true,
+      batchId: '1234',
```
amitzur, line 22: I would do this in a separate test

packages/eyes-cypress/src/plugin/config.js
```diff
@@ -15,8 +15,20 @@ function makeConfig() {
     ],
   });
 
-  if (!config.batch || !config.batch.id) {
-    config.batch = {id: uuid.v4(), ...config.batch};
+  if ((!config.batch || !config.batch.id) && !config.batchId) {
+    config.batch = {
+      id: uuid.v4(),
+      ...config.batch,
+      name: config.batchName,
```
amitzur, line 22: you are overriding `name` with `undefined` in case there is `config.batch.name` but not `config.batchName`

packages/eyes-cypress/src/plugin/config.js
```diff
@@ -15,8 +15,20 @@ function makeConfig() {
...
+      sequenceName: config.batchSequenceName,
+    };
+  } else {
+    config.batch = {
```
amitzur, line 26: no reason for this `else` block. The only purpose of the code here is to create a single batch with an ID in case the user didn't specify the batch ID. If they **did** specify a batch ID, then this would be handled in `eyesOpenMapping.js`.
Alternatively, handle everything related to the batch here, and do nothing about it in `eyesOpenMapping.js`. But don't spread the logic across multiple places.

amitzur (changes requested):
- Missing handling of `cy.eyesOpen({batch: {id,name,sequenceName}})` (plus a test for it)
- Missing a test in `eyesOpenMapping.test.js` for having `batchId`, `batchName`, `batchSequenceName` in the applitools.config.js file.
- Missing tests for the logic in `config.js`

pending-changes.yaml
```diff
@@ -14,7 +14,7 @@ applitools-for-selenium-ide:
   feature:
     - 
   bug-fix:
-    - 
+    - Support batchId, batchName and batchSequenceName from the config file, eyesOpen and environment variables
```
amitzur, line 17: @benbabayoff8 :point_up:

benbabayoff8: @amitzur Awesome.

SophieDePaula:
Thanks @amitzur ready for review again.

packages/eyes-cypress/src/browser/eyesOpenMapping.js
```diff
@@ -1,6 +1,11 @@
 function eyesOpenMapValues({args, appliConfFile, testName, shouldUseBrowserHooks}) {
   let browsersInfo = args.browser || appliConfFile.browser;
   let accessibilitySettings = args.accessibilityValidation || appliConfFile.accessibilityValidation;
+  const batch = {
+    id: args.batchId || args.batch?.id || appliConfFile.batchId || appliConfFile.batch?.id,
```
amitzur, line 5: optional chaining might not be supported in the user's browser (it's supported from [Chrome 80](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#browser_compatibility), and Cypress supports Chrome from [version 64](https://docs.cypress.io/guides/guides/launching-browsers#Browser-versions-supported))

Our eslint catches this, since we have `ecmaVersion:2018` in `.eslintrc.json`. Also this code should go through prettier.

packages/eyes-cypress/test/unit/plugin/config.test.js
```diff
@@ -48,4 +48,26 @@ describe('config', () => {
       fs.unlinkSync(filePath);
     }
   });
+
+  it('should create random batch id when batch id is not defined in config file', () => {
+    const filePath = path.join(__dirname, '../../../applitools.config.js');
+    fs.writeFileSync(filePath, 'module.exports = {};');
+    const {config} = makeConfig();
+    expect(config.batch.id).not.undefined;
```
amitzur, line 56: I didn't know you could write it this way..
Another way is `expect(config.batch.id).not.to.be.undefined;`
(just for correct english, but actually the shorter way seems clearer to me)

packages/eyes-cypress/test/unit/plugin/config.test.js
```diff
@@ -48,4 +48,26 @@ describe('config', () => {
...
+  });
+
+  it('should not overwrite bach id from config file when passed in an object', () => {
+    const filePath = path.join(__dirname, '../../../applitools.config.js');
+    fs.writeFileSync(filePath, "module.exports = {batch: {id: '1234'}};");
+    const {config} = makeConfig();
+    expect(config.batch.id).to.be.equal('1234');
```
amitzur, line 63: no need for `.be`:
`expect(config.batch.id).to.equal('1234');`

packages/eyes-cypress/test/unit/plugin/config.test.js
```diff
@@ -48,4 +48,26 @@ describe('config', () => {
...
   });
+
+  it('should create random batch id when batch id is not defined in config file', () => {
+    const filePath = path.join(__dirname, '../../../applitools.config.js');
+    fs.writeFileSync(filePath, 'module.exports = {};');
+    const {config} = makeConfig();
+    expect(config.batch.id).not.undefined;
+  });
+
+  it('should not overwrite bach id from config file when passed in an object', () => {
```
amitzur, line 59: bach --> batch

packages/eyes-cypress/test/unit/browser/eyesOpenMapping.test.js
```diff
@@ -36,6 +36,7 @@ describe('eyes open mapping', () => {
         matchLevel: 'Layout',
         useDom: true,
       },
+      batch: {id: undefined, name: undefined, sequenceName: undefined},
```
amitzur, line 39: hmmm, is this how you would like `eyesOpenMapping` to behave? I'm not even sure this leads to correct behavior when sent to the server.
Since `config.js` will populate `batch.id`, it will never have only undefined values. But I think it's still not a good behavior of `eyesOpenMapping`.

packages/eyes-cypress/test/unit/browser/eyesOpenMapping.test.js
```diff
@@ -154,4 +161,132 @@ describe('eyes open mapping', () => {
...
+    };
+
+    const coreConfig = eyesOpenMapValues({
+      args,
+      appliConfFile: {},
+      testName,
+      shouldUseBrowserHooks,
+      defaultBrowser,
+    });
+    expect(coreConfig).to.be.deep.equal(expected);
```
amitzur, line 195: This can be written like this: `expect(coreConfig).to.eql(expected);` (eql is like .deep.equal).

amitzur (changes requested):

packages/eyes-cypress/test/unit/plugin/config.test.js
SophieDePaula: So I'm sticking with the shorter way

packages/eyes-cypress/test/unit/browser/eyesOpenMapping.test.js
SophieDePaula: That's right, the batchId will always have a value. In regards to `name` and `sequenceName`, I now delete the property from the object in case it's undefined.
I updated the test accordingly.

amitzur (approved):

PR #810, opened 2022-03-18
eyes-storybook: preserve URL query params

tourdedave:
[Trello](https://trello.com/c/FghP1h9c)


SophieDePaula (approved):
💯

amitzur (changes requested):
The changes in `processResults.test.json` contain only changes of color as far as I could see.
The reason for these changes is that you need to run with STDOUT_LEVEL=0 when running storybook tests. And the reason for that is that the expected output is based on the colors of the shell in the CI. So we need to downgrade our machines' shell color ability by setting this env var.

amitzur:
@tourdedave following a comment from @SophieDePaula , the value of the env variable might be `STDOUT_LEVEL=1`. I may have been wrong. Try it out and see what value doesn't make you need to change the fixture.

amitzur:
@tourdedave fire at will

PR #811, opened 2022-03-20
remove waitBeforeCapture tests

chaim-aharonson-applitools:


amitzur:
@chaim-aharonson-applitools the test `waitBeforeCapture works from applitools.config file` is not covered by the generic coverage test, therefore I think it should either:
1. not be deleted, OR
2. covered by a unit test in `eyesOpenMapping.test.js`

PR #812, opened 2022-03-20
Add new fixes for getAllTestResults to changelog

amitzur:


benbabayoff8:
@amitzur Looks great.

PR #813, opened 2022-03-20
remove waitBeforeCapture tests (moved to generic coverage tests)

chaim-aharonson-applitools:


packages/eyes-cypress/test/fixtures/testApp/cypress/integration-run/waitBeforeCaptureConfigFile.js
```diff
@@ -6,7 +6,7 @@ describe('works with waitBeforeCapture', () => {
       appName: 'some app',
       testName: 'test waitBeforeCapture, config file',
       browser: {width: 1200, height: 800},
-      layoutBreakPoints: true,
+      layoutBreakpoints: true,
```
amitzur, line 9: @SophieDePaula how did this test pass then, if this typo was there all along? Maybe the web page is not good enough in checking that the page was resized?

pending-changes.yaml
```diff
@@ -52,7 +52,7 @@ Apollo SDK:
     - 
 "@applitools/eyes-cypress":
   feature:
-    - 
+    - remove waitBeforeCapture tests (moved to generic coverage tests)
```
amitzur, line 55: @benbabayoff8 FYI

amitzur (approved):

pending-changes.yaml
benbabayoff8: @amitzur I'm not sure it should be exposed in the changelog for the end user.

chaim-aharonson-applitools: @benbabayoff8 this is what actually done, something needs to be written, any suggestion?

packages/eyes-cypress/test/fixtures/testApp/cypress/integration-run/waitBeforeCaptureConfigFile.js
SophieDePaula: @amitzur @chaim-aharonson-applitools , I did have a bug in the test.
The reason it worked for me, is that I changed the viewport prior to the test and then in `eyesOpen` the viewport was changed again to the value I passed to `browser` in `eyesOpen`.
I think the test is valid still, as when I run it with the proper `layoutBreakpoints`, without changing the viewport and without `waitBeforeCapture` it still fails.
I think in most cases the user will not change the viewport priop to `eyesOpen`.
Sorry about the confusion @chaim-aharonson-applitools.

pending-changes.yaml
amitzur: @chaim-aharonson-applitools and @SophieDePaula - this was not approved by @benbabayoff8 .
The new workflow as we defined it means that any changelog entry we add should be approved by Product.
I realize there is confusion around the tooling - if we don't add anything then the GH hook fails. But this is not a reason to add unwanted entries to the changelog. We will improve the tooling - but when adding entries to the `pending-changes.yaml` file, please think what should the user read in the SDK's changelog.

PR #814, opened 2022-03-22
Switch chromedriver

chaim-aharonson-applitools:
used Docker v4.6
https://docs.docker.com/desktop/mac/release-notes/


chaim-aharonson-applitools:
@amitzur, there is no reason to add anything to changelog, no reason but the merge will fail unless something was changed/added to yaml.
there are too many conflicts with this file, each and every commit has conflicts before able to push it, don't have a solution for that :-(, just mentioning

PR #815, opened 2022-03-22
fix shadowRoot new logic for all SDKs

SophieDePaula:
When we need to create an `Element` with `shadowRoot`, there is no need to manipulate the `shadowRoot` `element` we are getting which is of format `"shadow-6066-11e4-a52e-4f735466cecf": "909d2534-7e1b-4ef5-831e-d5ed066cf633"`.
Therefore, the simplest and minimum change that I found , is to simply set the `shadowRoot`to the `target` without any manipulations.
From what I see in different `spec-driver`s that does have `transformShadowDom`, we try to get to the structure I mentioned above.
I ran `CheckRegionBySelectorWithinShadowDomWithVg.spec.js` on `eyes-selenium`, `eyes-webdriverio-4/5`, `eyes-puppeteer`, `eyes-playwright` and `eyes-nightwatch` and all passed.


amitzur:
@SophieDePaula where did this surface?

SophieDePaula:
@amitzur , @tourdedave mentioned that `CheckRegionBySelectorWithinShadowDomWithVg` is failing when running against `wdio` so I looked into it and found the problem.
Essentially, this issue did not surface in the first place as I was checking the test against `eyes-selenium` where we don't call `transformElement` as this menthod doesn't exsist in `spec-driver-selenium`. 
`tranformElement` does exist in `wdio`, but when we try to call it on `shadowRoot` it throws an error. Then I noticed that we have `transformShadowRoot` in `wdio` and other SDKs, so I looked into it but saw that this method is not exposed. So I looked for the simplest way to resolve this without changing the `spec-driver` interface and I noticed that we already get the `shadowRoot` in a correct format (I check `selenium, wdio 4 and 5, puppeteer, playwright and nightwatch`, I noticed that we do not run this test in `testCafe` so I did not check it).
After I found all of the above, I conculded that all we need to do when dealing with `shadowRoot` is to just pass it as is to the `target` in `Element`.

tourdedave:
@SophieDePaula In wdio-5 that test doesn't throw now, but I get diffs from Eyes

https://eyes.applitools.com/app/batches/00000251754084089078/00000251754084088875?accountId=xIpd7EWjhU6cjJzDGrMcUw~~

tourdedave:
> @tourdedave , the [test](https://eyes.applitools.com/app/test-results/00000251754077187360/00000251754077187126/steps/1?accountId=xIpd7EWjhU6cjJzDGrMcUw~~&mode=step-editor) pass for me.
> Are you sure you are linked to everything? I added to `link.sh` to also link `wdio` for me.

Yeah, pretty sure. I'll try again.

tourdedave:
Relinked and all good now. Thanks!

PR #818, opened 2022-03-23
make sure to check batchId property in global hooks when closing batch

SophieDePaula:
Consider `batchId` property (and not just `batch.id`) when closing the batch in global hooks.
This issue was raised [here](https://trello.com/c/H5xJSTsD/1271-cypress-batchid-configuration-not-being-honored-in-new-cypress-version)

While working on this issue I found that we don't support closing the batch form global hooks when the user sends the `batchId` from `eyesOpen`, so this should be addressed as well.

### DIFF
--- a/pending-changes.yaml
+++ b/pending-changes.yaml
@@ -95,7 +95,8 @@ Apollo SDK:
   feature:
     - 
   bug-fix:
-    - 
+    - consider batchId from config file and environments variable when closing batch
+    - consider batchId from eyesOpen when closing batch
 "@applitools/eyes-storybook":
   feature:
     - 
--- a/packages/eyes-cypress/src/browser/commands.js
+++ b/packages/eyes-cypress/src/browser/commands.js
@@ -97,6 +97,7 @@ Cypress.Commands.add('eyesOpen', function(args = {}) {
   Cypress.log({name: 'Eyes: open'});
   Cypress.config('eyesOpenArgs', args);
   const {title: testName} = this.currentTest || this.test || Cypress.currentTest;
+  const batchIdEyesOpen = args.batchId || (args.batch && args.batch.id);
 
   if (Cypress.config('eyesIsDisabled') && args.isDisabled === false) {
     throw new Error(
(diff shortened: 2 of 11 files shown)


pending-changes.yaml
```diff
@@ -95,7 +95,8 @@ Apollo SDK:
   feature:
     - 
   bug-fix:
-    - 
+    - consider batchId from config file and environments variable when closing batch
+    - consider batchId from eyesOpen when closing batch
```
SophieDePaula, line 99: @benbabayoff8 FYI

benbabayoff8: @SophieDePaula Great.

amitzur:
@SophieDePaula I don't think this was ever supported. I don't know how much I like this overhead. Let's discuss. If it was never supported then this is not a regression, and I wouldn't want to rush into fixing it.
cc @benbabayoff8

SophieDePaula:
@amitzur this PR attempts to solve two issues:

1) When we use `batchId` from either the config file or from an environment variable, an error is thrown because I only send `appliConfFile.batch.id` rather than also send `appliConfFile.batchId` when closing the batch in the global hooks, and therefore the batch is not closed.
2) When we send `batchId` from `eyesOpen` and we use `globalHooks` the batch is not closed as I never addressed this scenario.

When running both scenarios using `eyes-cypress` version `3.23.9` I see that the batch is closed.
Re the first point, the fix is straight forward:
`batchIds: [config.appliConfFile.batchId || config.appliConfFile.batch.id]` instead of `batchIds: [config.appliConfFile.batch.id]`
I understand that the second scenario is more complex.

jrnail23:
@SophieDePaula thanks for addressing this.  I was constantly getting this error when simply following the CircleCI setup documentation (setting `APPLITOOLS_BATCH_ID` to the GIT SHA):
```
TypeError: Cannot read property 'id' of undefined
    at Object.after:run [as handler] (/home/circleci/project/node_modules/@applitools/eyes-cypress/src/plugin/hooks.js:29:51)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
```
@amitzur, I don't know if this was ever actually supported, but your [documentation](https://applitools.com/docs/topics/integrations/github-integration-ci-setup.html#) is telling us to use it this way, which is currently broken.

SophieDePaula:
@jrnail23 this was resolved in a different PR, you can upgrade `eyes-cypress` to version `3.25.3`

SophieDePaula:
@benbabayoff8 when sending batch info (such as batch name, id and so on) from `eyesOpen` the batch will not be closed.
Do we want to add logic to resolve this?

PR #819, opened 2022-03-24
Fix shared mutable appliConfFile in cypress

amitzur:
The bug is that `Cypress.config('appliConfFile')` is what is called **shared mutable state**. It is shared between multiple tests, and therefore **should never** be mutated.

This PR avoids mutating this object, and instead works with a copy of it.


amitzur:
@SophieDePaula I'd appreciate your help with tests for this and shipping it

SophieDePaula:
@amitzur I added a `unit` test for this, I think it's sufficient for this use case. 
Your fix seems straightforward.

Lmk, if anything else is needed.

SophieDePaula (approved):

pending-changes.yaml
```diff
@@ -102,7 +102,8 @@ Apollo SDK:
   feature:
     - 
   bug-fix:
-    - 
+    - |
+      Fix config file parameters (e.g. `browser`) not being passed
```
SophieDePaula, line 106: @benbabayoff8 ^

PR #826, opened 2022-03-25
Add ability to bypass changelog PR check through a magic keyword in commit messages

tourdedave:
There are times when we open a PR and there is no need to update `pending-changes.yaml` (e.g., updating an internal package and there is no SDK/customer facing impact that needs to be communicated to our users). BUT there is a check that always runs to see that we did update this file as part of a PR. So the tooling makes us think that we need to update this file regardless, which can be confusing.

This PR is a means to improve the tooling to address this issue through a magic keyword in a commit message. Once the magic keyword is present in the commit history in the PR, the check will be bypassed. It will still run, but it will short-circuit and pass.

You just need to provide the magic keyword `NO_CHANGELOG` in a commit message in the PR branch. The easiest way to do this is through an empty commit: `git commit --allow-empty -m 'NO_CHANGELOG'`

What do you think?


SophieDePaula (approved):
Great Idea!

amitzur:
@tourdedave looks great. I'd recommend everyone to add an alias in the shell like `alias nochangelog="git commit --allow-empty -m 'NO_CHANGELOG'"` so that we can align on terminology - "did you run `nochangelog`?"

amitzur (approved):

PR #828, opened 2022-03-28
adjust deleteTest parameters eyes-api

SophieDePaula:
`deleteTest` should get `settings` and `logger`, but we don't send these values properly from `eyes-api`.
I found this while I was working on the generic coverage tests for `getAllTestResults`

### DIFF
--- a/packages/eyes-api/src/Runners.ts
+++ b/packages/eyes-api/src/Runners.ts
@@ -7,6 +7,7 @@ import {RunnerOptions, RunnerOptionsFluent} from './input/RunnerOptions'
 import {TestResultsData} from './output/TestResults'
 import {TestResultsSummaryData} from './output/TestResultsSummary'
 import {Eyes} from './Eyes'
+const {makeLogger} = require('@applitools/logger')
 
 type EyesRunnerSpec<TDriver = unknown, TElement = unknown, TSelector = unknown> = types.Core<
   TDriver,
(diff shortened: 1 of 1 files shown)


tourdedave:
Is there a generic coverage test that exercises this code path?

SophieDePaula:
@tourdedave I started implementing coverage test for this and found this issue, so no coverage test yet.

packages/eyes-api/src/Runners.ts
```diff
@@ -48,10 +49,13 @@ export abstract class EyesRunner {
...
-        serverUrl: eyes.configuration.serverUrl,
-        apiKey: eyes.configuration.apiKey,
-        proxy: eyes.configuration.proxy,
+        settings: {
+          ...options,
+          serverUrl: eyes.configuration.serverUrl,
+          apiKey: eyes.configuration.apiKey,
+          proxy: eyes.configuration.proxy,
+        },
+        logger: makeLogger({level: eyes.getConfiguration().getShowLogs() ? 'info' : 'silent', label: 'eyes'}),
```
kyryloonufriiev, line 58: We should not create a new logger here. It actually should use the same logger as we use for the other core functionality. Require a deeper refactoring.

SophieDePaula:
As this will require deeper refactoring, I'm closing this PR and will reopen it at a later time.

PR #835, opened 2022-03-30
Fix extractText for x, y regions without hints

tourdedave:
Test coverage added in https://github.com/applitools/sdk.coverage.tests/pull/142


amitzur (changes requested):
I think if the user passes `''` then it means we don't want to get the hint from the element. So `===undefined` is a good check. Did you have another use case in mind?
Is there a generic coverage test?

tourdedave:
> I think if the user passes `''` then it means we don't want to get the hint from the element. So `===undefined` is a good check. Did you have another use case in mind?

I didn't have a case in mind. Just thought this was cleaner, and didn't see a test covering '', so thought it might be an OK change. I'll set it back.

> Is there a generic coverage test?

It's linked in the description.

tourdedave:
Ben approved this offline.

PR #836, opened 2022-04-01
Allow platformName with lowercase

SophieDePaula:
While investigating the issue described [here](https://trello.com/c/JZujW51R/1268-eyes-universal-python-ios-status-bar-on-custom-grid-internet-brands), I notice that since we don't allow mobile `platformName` to be lowercase , `screenshoter` gets into an incorrect flow and therefor, we get an incorrect screenshot.

In this case, in `take-viewport-screenshot` in `screenshoter` we check if `driver.iOS`, since the customer passed `ios` with lowercase ,  the returned value is `false` and we fall to `makeTakeDefaultScreenshot` intead of calling `makeTakeMarkedScreenshot` and therefore, the customer gets an iOS screenshot with the url bar.


amitzur:
@SophieDePaula why didn't you mark @benbabayoff8 as reviewer for the pending changes?

amitzur (approved):

tourdedave (changes requested):
LGTM, but tag Ben to review the pending changes prose

benbabayoff8:
@SophieDePaula Where do the user pass "ios"? The platform name isn't something public as far as I remember.

SophieDePaula:
@benbabayoff8 that sounds good I would just mention with lowercase, what do you think?

pending-changes.yaml
```diff
@@ -10,22 +10,22 @@ Apollo SDK:
   feature:
     - 
   bug-fix:
-    - accept ios and android platformNames with lowcases 
+    - accept ios and android lowercase as driver platformName capability when using custom grid 
 "@applitools/eyes-selenium":
   feature:
```
SophieDePaula, line 15: @benbabayoff8 please approve, thanks.

benbabayoff8: Awesome.

PR #837, opened 2022-04-01
add test to make sure browserInfo stays intact when running more than one test

SophieDePaula:
This is in continuation to PR #819 , I've added an `e2e` test.


packages/eyes-cypress/test/e2e/cypress-run-make-sure-appliConfFile-stays-intact.test.js
```diff
@@ -0,0 +1,74 @@
...
+      throw ex;
+    }
+  });
+
+  after(async () => {
+    fs.rmdirSync(targetTestAppPath, {recursive: true});
+  });
+
+  it('appliConfFile, browserInfo stays intact', async () => {
+    const config = {
```
amitzur, line 55: I would just create the config file you want instead of mangling with an existing one.

amitzur (approved):
Small suggestions inline, looks good :+1:

PR #838, opened 2022-04-04
add 'waitBeforeCapture' as an option in 'Eyes.Open' and 'Eyes.Check' …

chaim-aharonson-applitools:
…(for Typescript)


SophieDePaula:
@chaim-aharonson-applitools looks good, just make sure to tag Ben for the pending changes

benbabayoff8:
@chaim-aharonson-applitools Is it only for Cypress? Do other SDKs have this type?

Plakhota:
The file is part of https://www.npmjs.com/package/@applitools/visual-grid-client, which is used by other JS SDKs, including Universal, according to the Dependents tab in NPM.
@benbabayoff8

PR #848, opened 2022-04-06
Fix normalizeRegion logic in iOS native app

SophieDePaula:
This PR attempts to resolve the issue that is raised [here](https://trello.com/c/Xt6GC9At/1294-eyes-selenium-native-app-unable-to-perform-region-check-on-ipad)

When capturing a region in iOS, we make sure that the region in question is within a “safe area” on the screen, so that the element that we are trying to capture is within the display area and does not overlap / go under a top or button bar that could have special behaviors (this functionality was added in PR #724 )

 **The way safe area is calculated**:

You can find the logic in [driver.ts in init method](https://github.com/applitools/eyes.sdk.javascript1/blob/master/packages/driver/src/driver.ts#L186).
Basically the safe area is the display size minus the dimensions of XCUIElementTypeNavigationBar on the top and XCUIElementTypeTabBar at bottom if they exist in the app.

**Where do we use the safe area**:

In [driver.ts in normalizeRegion](https://github.com/applitools/eyes.sdk.javascript1/blob/master/packages/driver/src/driver.ts#L397), we check if the region we need to capture intersects with the safeArea we calculated, and we limit the area we capture according to it.

In some cases (like the case that was raised in the trello card above), the user would like to capture XCUIElementTypeNavigationBar / XCUIElementTypeTabBar which are outside of the safe area. In this PR I added logic to not check if that region intersects with the safe area as it will not. We will simply capture that region as is instead.

### WIP:

- [x] Make sure all tests in `screenshoter` package pass
- [x] Write a test to cover this scenario


amitzur (approved):
The explanation in the description is excellent.
I think the changes make sense. In terms of API I would do:
```
// driver.ts
normalizeRegion(region, shouldIgnoreSafeRegion?: boolean)

// element.ts
getRegion(shouldIgnoreSafeRegion?: boolean)

// getTarget.js
await element.getRegion(navBar || tabBar),
```

I intentionally didn't change to object destructuring (`normalizeRegion({region, shouldIgnoreSafeRegion})`) just becuase I think that would make it a much larger changeset which I'd like to avoid at this point (a lot of places use the `element.getRegion` method). But if you think it's important you can go for it.

Regarding tests: `yarn test:e2e:ios` passes on this branch. But indeed tests need to be added.

pending-changes.yaml
```diff
@@ -10,7 +10,7 @@ Apollo SDK:
   feature:
     - 
   bug-fix:
-    - 
+    - Native iOS, allow capturing NavigationBar and TabBar regions
```
SophieDePaula, line 13: @benbabayoff8 ^

benbabayoff8: Looks good to me

SophieDePaula:
[Coverage test for this PR](https://github.com/applitools/sdk.coverage.tests/pull/144)

PR #855, opened 2022-04-11
Support data urls in iframes

amitzur:
[Trello](https://trello.com/c/XfWUEGRS)

The issue was a tricky one, but the solution is only 3 lines of code.

When there is a data url iframe in a page, e.g. `<iframe src="data:text/html,"></iframe>`, then this iframe is considered cross origin. This is because, well, it's really cross origin (its [origin](https://developer.mozilla.org/en-US/docs/Glossary/Origin) is different than the host document), and we detect this by seeing that `frameEl.contentDocument` is null ([here](https://github.com/applitools/dom-scripts/blob/ff5f7ce837d23976a28f939deceb9e6890c53978/packages/dom-shared/src/browser/isAccessibleFrame.js#L5)).

So far there's no problem. Because we handle cross origin iframes. But the issue happens when we switch to this frame and take its dom snapshot. The URL we get for the frame is `data:text/html,...` which is a correct value "on paper". **But** this URL would be used in the resource mapping for UFG, as the key for the frame's contents resource. And that's bad.

*(This is the mechanism for frames in the UFG - we upload the frame's contents (dom and resource mapping) **as a resource**, and then include that resource in the general resource mapping of the page. What URL do we use? the frame's `src` usually. Unless it's special cases, **like data urls**. It doesn't really matter what the URL is. As long as the resource is mapped.)*

The UFG doesn't like data urls as keys in resource maps, and righteously so. See this (`vg-cli view-rendering UAujt6tHnEKUivQXIz7G6A~~/8e123795-3f58-4c1d-a644-71e5d682fa1c`):
Screenshot from 2022-04-11 05-32-37

What we would like to have is this (`vg-cli view-rendering UAujt6tHnEKUivQXIz7G6A~~/2b8b54db-36fe-4631-ba1d-5fde1e96eba9`):
Screenshot from 2022-04-11 05-33-47

There are 2 new tests here:
1. unit test for takeDomSnapshot with a mock of a page with data url iframe.
2. e2e test that shows this solution works (see this result - baseline is with the fix, and checkpoint is without the fix: https://eyes.applitools.com/app/test-results/00000251752657141257/00000251752657141101/steps/1?accountId=UAujt6tHnEKUivQXIz7G6A~~&mode=step-editor)


chaim-aharonson-applitools:
@amitzur cool !!!
The original ticket mentioned 'HTML' (not 'IMG') as the iframe src:  
`<iframe src="data:text/html,<html><body></body></html>"></iframe>`
Additionally, it appears as **encoded** in ticket, something like:
`<iframe src="data:text/html;charset=utf-8,%3Chtml%3E%3Cbody%3E%3Cimg%20style='width:170px;height:250px;'%20src='${location.origin}/smurfs.jpg'%3E%3C/body%3E%3C/html%3E"></iframe>`
**Glad to say that I tested both with your code and it works** :-)
So, _with your permission_, I added another test for data url as encoded html and moved the test html files to 'fixtures/frames' directory

amitzur:
@chaim-aharonson-applitools I don't think that makes a difference, but it's always good to validate as much a possible the actual issue we intend to solve. So great!
I'll ship this later today once @SophieDePaula and @tourdedave take a look

pending-changes.yaml
```diff
@@ -21,6 +21,7 @@ Apollo SDK:
       `extractText` now supports regions that don't use hints while using `x`/`y` coordinates
     - accept ios and android lowercase as driver platformName capability when using custom grid
     - Native iOS, allow capturing NavigationBar and TabBar regions
+    - Support data urls in iframes
```
tourdedave, line 24: There should be an entry for eyes-universal too, no?

packages/eyes-sdk-core/test/e2e/check.e2e.spec.js
```diff
@@ -1,20 +1,34 @@
 const setupTests = require('./utils/core-e2e-utils')
 
-// this is an example
-describe.skip('check e2e', () => {
-  const {getDriver, getSDK} = setupTests({before, beforeEach, afterEach})
+describe('check e2e', () => {
```
tourdedave, line 3: I've actually been relying on this file as a starting point for authoring new tests or debugging issues reported in the universal-server. Can we keep it or add this test as a new file with a more descriptive name?

tourdedave (changes requested):
Nice!

A couple of small tweaks requested.

SophieDePaula (approved):
:100:

pending-changes.yaml
chaim-aharonson-applitools: @tourdedave 
Sure, it's me :-)

amitzur: yes, thanks

packages/eyes-sdk-core/test/e2e/check.e2e.spec.js
amitzur: :+1:

tourdedave (approved):

amitzur:
@tourdedave I made the requested changes. Merging.

pending-changes.yaml
amitzur: And also:
1. storybook
2. cypress
3. wdio5-service

I'm adding that post-merge directly to master.
The original ticket was for storybook

pending-changes.yaml
```diff
@@ -29,6 +30,7 @@ Apollo SDK:
...
```
amitzur, line 33: @benbabayoff8 I merged before getting your approval. Please see this entry

PR #856, opened 2022-04-11
Fix/screenshoter off by one error

tourdedave:
There's an off-by-one error that can happen when taking a stitched screenshot using CSS stitching for a full check region if the target element has an `x` coordinate that ends in `.5` pixels 🤯 

That's because the `region` (e.g., the region to be captured, which is passed in as an argument to `takeStitchedScreenshot`) and `targetRegion` (e.g., the calculation done in `takeStitchedScreenshot` to determine the intersection between the region and where the page was scrolled to) can drift slightly due to a rounding error, but they shouldn't because they are similar and rely on each other. The `targetRegion` is used to determine the `cropRegion` and the `region` is used to determine the `partRegions`.

When there is a disparity between them it can cause incorrect `partRegions` being calculated and extraneous 1px borders showing up in the final screenshot. Here are some details from the specific issue where this surfaced:

```
// initialOffset
{ x: 128.5, y: 391.796875 }

// postMoveOffset { x: 129, y: 392 }

// region passed into takeStitchedScreenshot -- ultimately derived from window.getComputedStyle
{ x: 128.5, y: 391.796875, width: 942.984375, height: 3294.34375 }

// targetRegion -- ultimately derived from a combination of the postMoveOffset and the region
{ x: 129, y: 392, width: 942.484375, height: 800 }

// partRegions
[
  { x: 1071, y: 392, width: 1, height: 750 },
  { x: 129, y: 1142, width: 942, height: 740 },
  { x: 1071, y: 1142, width: 1, height: 740 },
  { x: 129, y: 1882, width: 942, height: 740 },
  { x: 1071, y: 1882, width: 1, height: 740 },
  { x: 129, y: 2622, width: 942, height: 740 },
  { x: 1071, y: 2622, width: 1, height: 740 },
  { x: 129, y: 3362, width: 942, height: 324 },
  { x: 1071, y: 3362, width: 1, height: 324 }
]
```

Screen Shot 2022-04-11 at 21 33 11

Thankfully we already track the difference that gets rounded off as part of scrolling (e.g., `expectedRemainingOffset`), so we can use it to compensate the region if there's a difference between the `region` `x`/`width` and `targetRegion` `x`/`width`. Here are the partRegions with the compensation applied.

```
[
  { x: 128, y: 1142, width: 942, height: 740 },
  { x: 128, y: 1882, width: 942, height: 740 },
  { x: 128, y: 2622, width: 942, height: 740 },
  { x: 128, y: 3362, width: 942, height: 324 }
]
```

And a successful screenshot.

Screen Shot 2022-04-11 at 21 33 37

I've confirmed that all e2e screenshoter and generic coverage tests pass with this fix applied/linked. There are no `pending-changes.yaml` entries since they are already handled in the parent branch.


packages/screenshoter/test/e2e/web/element-fully.spec.js
```diff
@@ -36,4 +36,18 @@ describe('screenshoter web', () => {
       logger,
     })
   })
+
+  it('take full element screenshot with "css" scrolling of an element with a position coordinate ending in .5px', async () => {
+    await driver.visit('data:text/html,<div%20style="position:absolute;left:50.5;top:12.80;width:942.98;%20height:3284.36;">blah</div>')
```
amitzur, line 41: ooooh, nice. I'm pretty sure you can decode the `%20` there. you're already using `<` etc.

packages/screenshoter/src/take-stitched-screenshot.js
```diff
@@ -65,6 +65,11 @@ async function takeStitchedScreenshot({
   if (region) region = utils.geometry.intersect(region, contentRegion)
   else region = contentRegion
 
+  if (targetRegion.x !== region.x && targetRegion.width !== region.width) {
+    logger.verbose('Compensating for region width differences...')
```
amitzur, line 69: let's add the numbers to the log (`targetRegion.x=${targetRegion.x} region.x=${region.x} targetRegion.width=${targetRegion.width} region.width=${region.width} expectedRemainingOffset.x=${expectedRemainingOffset.x}` or something like that). WDYT?

amitzur (approved):
Very nice! So trivial

PR #857, opened 2022-04-12
Native viewport size

chaim-aharonson-applitools:
add geometry.ceil for region


packages/driver/src/driver.ts
```diff
@@ -209,7 +209,9 @@ export class Driver<TDriver, TContext, TElement, TSelector> {
     if (this.isMobile) {
       this._driverInfo.orientation ??= await this.getOrientation().catch(() => undefined)
     }
-
+    // TODO set a multiple options for closed list of universal config variables and loop over it
+    const envUniversalConfig = process.env.UNIVERSAL_VIEWPORT_SIZE_CALCULATION
```
amitzur, line 213: let's call this `useCeilForViewportSize` and make it be passed in config for `manager.openEyes` and then propagated to driver

packages/driver/src/driver.ts
```diff
@@ -452,7 +454,14 @@ export class Driver<TDriver, TContext, TElement, TSelector> {
           }
         }
       }
-      size = utils.geometry.round(size)
+      //let calculateViewportMethod = 'round' <== currently being ignored but should be default as'round' and else taken from env
+      if (this._driverInfo.universalWrapperConfig?.viewportSizeCalculation) {
+        // TODO - make it work as dynamic function name
+        //if (viewportSizeCalculation && typeof utils.geometry[viewportSizeCalculation] === 'function') {
```
amitzur, line 460: Not recommended, this is harder to track and find in the code. It's just being "smart" :)
I'd prefer it to be
```
if (this.useCeilForViewprotSize) {
...
} else {
...
}
```

packages/eyes-sdk-core/lib/utils/takeVHSes.js
```diff
@@ -14,7 +14,19 @@ async function takeVHSes({driver, browsers, apiKey, waitBeforeCapture, logger})
     if (!apiKeyInput) {
       throwError('UFG_Apikey element could not be found')
     }
-    await apiKeyInput.type(apiKey)
+    const inputJson = {
+      apiKey,
+      url: serverUrl || '',
```
amitzur, line 19: why not `serverUrl`? No biggy, I just think it's better if we use the same terminology with the NMG team

packages/eyes-sdk-core/lib/utils/takeVHSes.js
```diff
@@ -14,7 +14,19 @@ async function takeVHSes({driver, browsers, apiKey, waitBeforeCapture, logger})
...
+    }
+    const proxyObject = proxy && proxy.toProxyObject()
+    if (proxyObject) {
```
amitzur, line 22: what about username and password? and `isHttpOnly`?
I spoke with Roy about them supporting it

packages/types/src/data.ts
```diff
@@ -27,6 +27,11 @@ export type AutProxy = {
   AUTProxyMode?: 'Allow' | 'Block'
 }
 
+export type UniversalWrapperConfig = {
+  wrapperType?: string
```
amitzur, line 31: what is this?

pending-changes.yaml
```diff
@@ -8,9 +8,9 @@ Apollo SDK:
     - Remove kitchen to accommodate space for PlayStation
 "@applitools/eyes-universal":
   feature:
-    - 
+    - Internal 'takeVHSes' extend the sent data
```
amitzur, line 11: Support passing `serverUrl` and `proxy` to VHS.

pending-changes.yaml
```diff
@@ -8,9 +8,9 @@ Apollo SDK:
...
   bug-fix:
-    - 
+    - Native viewport size calculation changed from round to ceil
```
amitzur, line 13: Use `Math.ceil` instead of `Math.round` for viewportSize calculation when passed `useCeilForViewportSize:true`

packages/utils/src/geometry.ts
```diff
@@ -40,6 +40,22 @@ export function round(target: Region | RectangleSize | Location): typeof target
   return result
 }
 
+export function ceil(region: Region): Region
+export function ceil(region: RectangleSize): RectangleSize
+export function ceil(region: Location): Location
+export function ceil(target: Region | RectangleSize | Location): typeof target {
+  const result = {...target} as any
+  if (types.has(target, ['x', 'y'])) {
+    result.x = Math.round(target.x)
```
amitzur, line 49: Oh, this is very confusing. I realize it might make sense, but from first look it's unexpected.
Not sure what to do about it.
Oh well...
I would add a comment here saying:
> intentionally using Math.round and not Math.ceil here, because the point is that for width and height it makes sense to use ceil, but not for x and y

amitzur (changes requested):

packages/eyes-sdk-core/lib/utils/takeVHSes.js
```diff
@@ -11,15 +11,30 @@ async function takeVHSes({driver, browsers, apiKey, waitBeforeCapture, logger})
...
+        }
+      }
+      await apiKeyInput.type(JSON.stringify(inputJson))
+      const ready = await context.element({type: 'accessibility id', selector: 'UFG_ApikeyReady'})
+      if (!ready) {
+        throwError('UFG_ApikeyReady element could not be found')
+      }
+      await ready.click()
+    } else {
+      logger.log('apiKeyInput was skipped')
```
amitzur, line 36: Use the term `UFG_Apikey` instead of `apiKeyInput`, to be consistent with the other places in this file which throw exception.

chaim-aharonson-applitools: done

packages/utils/src/geometry.ts
chaim-aharonson-applitools: done

packages/driver/src/driver.ts
```diff
@@ -28,11 +29,13 @@ export class Driver<TDriver, TContext, TElement, TSelector> {
     spec: types.SpecDriver<TDriver, TContext, TElement, TSelector>
     driver: Driver<TDriver, TContext, TElement, TSelector> | TDriver
     logger?: any
+    customConfig?: any
   }) {
     if (options.driver instanceof Driver) return options.driver
 
     this._spec = options.spec
 
+    this._customConfig = options.customConfig || {}
```
amitzur, line 38: you can do this with a default value in the parameter list

packages/eyes-api/src/input/Configuration.ts
```diff
@@ -71,6 +71,8 @@ export type OpenConfiguration = {
   saveDiffs?: boolean
   /** @undocumented */
   dontCloseBatches?: boolean
+  /** @undocumented */
```
amitzur, line 74: no need for changes in eyes-api at all

packages/eyes-sdk-core/lib/utils/takeVHSes.js
```diff
@@ -10,16 +10,25 @@ async function takeVHSes({driver, browsers, apiKey, waitBeforeCapture, logger})
...
   if (driver.isAndroid) {
-    const apiKeyInput = await context.element({type: 'accessibility id', selector: 'UFG_Apikey'})
-    if (!apiKeyInput) {
-      throwError('UFG_Apikey element could not be found')
-    }
-    await apiKeyInput.type(apiKey)
-    const ready = await context.element({type: 'accessibility id', selector: 'UFG_ApikeyReady'})
-    if (!ready) {
-      throwError('UFG_ApikeyReady element could not be found')
+    const UFG_Apikey = await context.element({type: 'accessibility id', selector: 'UFG_Apikey'})
```
amitzur, line 13: I didn't mean the variable name in my previous comment. I meant the message to the log. No biggy.

packages/eyes-sdk-core/test/e2e/universal-overrides.e2e.spec.js
```diff
@@ -0,0 +1,29 @@
+const setupTests = require('./utils/core-e2e-utils')
+
+describe.skip('universal overrides', () => {
+  process.env.UNIVERSAL_VIEWPORT_SIZE_CALCULATION = true
```
amitzur, line 4: you can remove this test

amitzur (approved):
See inline comments. But generally it's approved.

amitzur (approved):

PR #858, opened 2022-04-12
fix capturing window and region screenshots in android native landscape

SophieDePaula:
This PR attempts to solve the issues that were raised in [this Trello card](https://trello.com/c/Xsv4v0UK/1303-android-native-landscape-unable-to-check)

I found two main problems that caused the error we see in the Trello card above:

1) We crop out the button navigation bar, however, in landscape mode, we need to crop it from the left.
2) We use `Appium` API to get the dimensions of the system bars, however, I see that this API is buggy and in landscape mode, it returns the wrong value, therefore, I check if we are in landscape mode, then I flip to portrait to get the correct dimensions and then flip back to landscape.

As a result of this change, I also had to make adjustment to `normalizeRegion` in `driver` to adjust `x` offest in case we are in landscape mode.

### TODO
- [x] check if all test pass for `spec-driver-selenium`
- [x] check if all test pass for `driver`
- [x] check if all test pass for `screenshoter`
- [ ] write tests


pending-changes.yaml
```diff
@@ -20,15 +20,18 @@ Apollo SDK:
     - |
       `extractText` now supports regions that don't use hints while using `x`/`y` coordinates
     - accept ios and android lowercase as driver platformName capability when using custom grid
-    - Native iOS, allow capturing NavigationBar and TabBar regions
+    - native iOS, allow capturing NavigationBar and TabBar regions
+    - native android, in case we test a device in ladscape mode, make sure to account for the navigation bar on the left and not at the button of the image.
+    - native android, when fetching the dimensions of system bars in landscape mode, webdriver returns the wrong deminetion, to workaround it, we now flip to portrait to capture these demiensions and then flip back to landscape.
```
SophieDePaula, line 25: @benbabayoff8 let me know if you think the comment about the dimensions is unnecessary for the user. I wasn't sure.

packages/spec-driver-selenium/src/spec-driver.ts
```diff
@@ -242,16 +243,36 @@ export async function waitUntilDisplayed(driver: Driver, selector: Selector, tim
...
+  if (orientation.toLocaleLowerCase() === 'landscape') {
+    await setOrientation(driver, 'LANDSCAPE')
+  }
   return {
     statusBarHeight: statusBar.visible ? statusBar.height : 0,
     navigationBarHeight: navigationBar.visible ? navigationBar.height : 0,
   }
 }
+
+async function setOrientation(driver: Driver, orientation: string) {
```
amitzur, line 268: type of `orientation` parameter should be `'PORTRAIT' | 'LANDSCAPE'` and there should be a default value of `PORTRAIT` IMO.

packages/spec-driver-selenium/src/spec-driver.ts
```diff
@@ -242,16 +243,36 @@ export async function waitUntilDisplayed(driver: Driver, selector: Selector, tim
...
+  if (orientation.toLowerCase() === 'landscape') {
+    await setOrientation(driver, 'PORTRAIT')
+  }
   const getSystemBarsCommand = new Command('getSystemBars')
   const {statusBar, navigationBar} =
     process.env.APPLITOOLS_SELENIUM_MAJOR_VERSION === '3'
       ? await (driver as any).schedule(getSystemBarsCommand)
       : await driver.execute(getSystemBarsCommand)
+
+  if (orientation.toLocaleLowerCase() === 'landscape') {
```
amitzur, line 259: `toLowerCase` is just fine :) we're not handling special chars here

pending-changes.yaml
```diff
@@ -10,7 +10,7 @@ Apollo SDK:
   feature:
     - 
   bug-fix:
-    - 
+    - native android, in case we test a device in ladscape mode, make sure to account for the navigation bar on the left and not at the button of the image.
```
amitzur, line 13: My suggestion + a few typo fixes:
When running a native app on Android, in case we test a device in landscape mode, make sure to account for the navigation bar on the left and not at the bottom of the image. Also account for an Appium bug when calculating system bars height.

pending-changes.yaml
amitzur: see my suggestion above

pending-changes.yaml
```diff
@@ -20,15 +20,18 @@ Apollo SDK:
...
+    - native android, when fetching the dimensions of system bars in landscape mode, webdriver returns the wrong deminetion, to workaround it, we now flip to portrait to capture these demiensions and then flip back to landscape.
 "@applitools/eyes-webdriverio":
   feature:
     - Support UFG for native mobile
   bug-fix:
     - |
       `extractText` now supports regions that don't use hints while using `x`/`y` coordinates
     - accept ios and android lowercase as driver platformName capability when using custom grid
-    - Native iOS, allow capturing NavigationBar and TabBar regions
+    - native iOS, allow capturing NavigationBar and TabBar regions
```
amitzur, line 33: Guideline is to use proper capitalization and grammar as much as possible.

amitzur (changes requested):
I think this is very good.
A bit concerned about the orientation flip, but I have no other suggestion.
Just a few minor changes requested

benbabayoff8:
@amitzur @SophieDePaula I wouldn't describe the "how" part (that we flip and flip back).
Only what is the scenario and that it is improved/fixed.

pending-changes.yaml
SophieDePaula: @amitzur I am not entirely sure what I need to change here.
I changed `Native` to `native` as I see all the other comments start with lowercase.

SophieDePaula: I updated this entry, hopefully, this was your intention.

packages/spec-driver-selenium/src/spec-driver.ts
SophieDePaula: I used `ScreenOrientation` type under `types`, I didn't set any default value to this type as it's probably used in other places.

PR #871, opened 2022-04-14
fix beforeCapture hook mapping

SophieDePaula:
[Trello card](https://trello.com/c/vcJjdJxY/1300-cypress-beforecapturescreenshot-hook-not-working)
The mapping for `beforeCaptureScreenshot` hook was incorrect.


pending-changes.yaml
```diff
@@ -111,6 +111,7 @@ Apollo SDK:
       `extractText` now supports regions that don't use hints while using `x`/`y` coordinates
     - add 'waitBeforeCapture' as an available type for 'Eyes.Open' and 'Eyes.Check' (for Typescript)
     - Support data urls in iframes
+    - fix beforeCapture hook not being honored
```
SophieDePaula, line 114: @benbabayoff8 fyi

amitzur: 1. beforeCaptureScreenshot is the correct term.
2. honored --> respected

amitzur (approved):

amitzur:
@SophieDePaula actually, I don't think these should be specifix tests. There should be a generic test for this

pending-changes.yaml
```diff
@@ -107,7 +107,7 @@ Apollo SDK:
   feature:
     - 
   bug-fix:
-    - 
+    - fix beforeCaptureScreenshot hook not being respected
```
SophieDePaula, line 110: @benbabayoff8 fyi

PR #876, opened 2022-04-19
Fix Appium settings/caps in universal

tourdedave:
This PR does the following:
- implicitly set the Appium setting for `allowInvisibleElements` on Android native (so users won't have to specify it as a capability)
- fix broken capabilities in iOS when running on a newer version of Appium

I still need to finish porting the iOS ufg native test over. I have it working locally but hitting a timeout error when running on Sauce.


packages/eyes-universal/src/spec-driver/webdriver.ts
```diff
@@ -219,7 +229,9 @@ export async function getCookies(driver: Driver, context?: boolean): Promise<Coo
   })
 }
 export async function getCapabilities(browser: Driver): Promise<Record<string, any>> {
-  return browser.getSession?.() ?? browser.capabilities
+  const caps = await browser.getSession?.()
+  if (caps && utils.types.isObject(caps)) return caps
+  return browser.capabilities
```
SophieDePaula, line 234: @tourdedave I am not very familiar with getting the capabilities from the browser.
will `browser.capabilities` not work for all scenarios?

SophieDePaula (commented):
@tourdedave the logic here makes sense to me, I only had one questions below

packages/eyes-universal/src/spec-driver/webdriver.ts
tourdedave: As far as I know, yes, and the way it was written before assumes `browser.capabilities` is a safe fallback.

PR #878, opened 2022-04-25
change exception property in testResultsSummary to be of type TestResultsException

SophieDePaula:
[Trello](https://trello.com/c/WV7ixoDy/1325-universal-sdk-runtime-errors-in-closemanager-results-returned-as-empty-dicts)

The property `exception` in testResultsSummary is of type `Error` and therefore the content of this object is lost when sending it in the response to the universal clients.

I created a new type `TestResultsException` that contains only `stack` and `message` properties as `strings` to resolve that.

Also, I added a fix on the `universal` side to map `reason` when we need to throw an error that issue was raised in [this card](https://trello.com/c/UuZrhplh/1326-universal-sdk-eyesmanagerclosemanager-command-does-not-priovide-exception-reason-type-hint)


packages/eyes-universal/src/socket.ts
```diff
@@ -133,7 +133,7 @@ export function makeSocket(ws: WebSocket): Socket {
         const result = await fn(payload)
         emit({name, key}, {result})
       } catch (error) {
-        emit({name, key}, {error: {message: error.message, stack: error.stack}})
+        emit({name, key}, {error: {message: error.message, stack: error.stack, reason: error.reason}})
       }
```
SophieDePaula, line 137: @kyryloonufriiev I see that in `eyes-selenium-universal` we need to map `reason` property value on the [receiving `socket` as well](https://github.com/applitools/eyes.sdk.javascript1/blob/master/packages/eyes-selenium-universal/src/socket.ts#L95).
Do you think that this is something we need to add to our [docs](https://github.com/applitools/eyes.sdk.javascript1/blob/master/packages/eyes-universal/README.md#eyesmanagerclosemanager)?

pending-changes.yaml
```diff
@@ -12,7 +12,8 @@ Apollo SDK:
   feature:
     - 
   bug-fix:
-    - 
+    - |
+     exception in TestResultsSummary now should be accepted properly by universal clients, and we now also map `reason` when emitting an error for Eyes exceptions.
```
SophieDePaula, line 16: @benbabayoff8 fyi

SophieDePaula:
adjusted [this](https://github.com/applitools/sdk.coverage.tests/pull/150) generic test to make sure we get the relevant exception in `TestResultsSummary`

tourdedave (changes requested):
Looks great! Just a small tweak to the test in `eyes-selenium-universal`

packages/eyes-selenium-universal/test/play/play.js
```diff
@@ -0,0 +1,41 @@
+// adopted styleSheets on chrome
+const path = require('path')
+const spec = require(path.resolve('./dist/spec-driver'))
+const setupEyes = require('@applitools/test-utils/src/setup-eyes')
+const {VisualGridRunner} = require('@applitools/eyes-api')
+
+describe('play', () => {
```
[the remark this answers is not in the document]
SophieDePaula: good idea!

packages/eyes-sdk-core/lib/sdk/EyesClassic.js
```diff
@@ -191,7 +191,7 @@ class EyesClassic extends EyesCore {
       .then(results => {
         if (isErrorCaught) {
           if (results.info && results.info.testResult) return {testResults: results.info.testResult}
-          else return {exception: results}
+          else return {exception: results ? {stack: results.stack, message: results.message} : null}
```
amitzur, line 194: In the previous line we verify `results.info`. So you can be sure that in this line `results` is defined.

packages/eyes-api/src/output/TestResultContainer.ts
```diff
@@ -21,10 +21,10 @@ export class TestResultContainerData implements Required<types.TestResultContain
     return new TestResultsData(this.testResults, this._deleteTest)
   }
 
-  get exception(): Error {
+  get exception(): types.TestResultsException {
```
amitzur, line 24: Isn't this PR an internal detail? Why should it propagate to the user? This is essentially a breaking change. In `eyes-api` we should still hold an Error

packages/types/src/data.ts
```diff
@@ -14,6 +14,8 @@ export type AccessibilityStatus = 'Passed' | 'Failed'
 
 export type TestResultsStatus = 'Passed' | 'Unresolved' | 'Failed'
 
+export type TestResultsException = {stack?: string; message?: string}
```
amitzur, line 17: Should these fields really be optional?

packages/eyes-api/src/output/TestResultContainer.ts
SophieDePaula: @amitzur I can undo the type changes from here are `data.ts` in `types`, but isn't it a misrepresentation of the type we use in `TestResultContainer`?

amitzur: @SophieDePaula Here's what I mean:
Currently the user API is like this - `summary.getAllResults()[0].exception` is an Error instance (if there was an exception).
This means our users know they can do the following:
```
const summary = await runner.getAllTestResults()
const err = summary.getAllResults()[0].exception
if (err instanceof Error) {
  // handle error
}
```
I'm not saying this is the encouraged way to write user code, but **it works**.

What you do in this PR is change the Error to be a plain object. That's a breaking change. So either you knowingly do this and then ask for approval from Product, or you avoid doing it.

Once you realize you don't want to break the user API, you have an anchor - TestResultContainer in eyes-api should have an `Error` instance and nothing else.
The requirement is that in universal the exception field should be serializable. Then this means that either (a) core should not return an Error instance, since it would pass as an empty object in the websocket, or (b) core should still return an Error instance, but the universal-server should serialize it.
I would go with (a) which is what you chose as well. But then you must create an Error instance out of this object on the eyes-api side.
Another way to do it is have a property of `originalError` on the plain object. It would probably not serialize for the universal, but it could serve as an API between core and eyes-api, so that eyes-api would take the `originalError` into the `exception` field of `TestResultContainer` (instead of creating a new one).

SophieDePaula: @amitzur universal users are also getting the same `summary` object so I won't be able to keep the type of `exception` in `TestResultContainer` and fix this problem. Therefore I actually think option `b` is the only option we can go with here.
Regarding serialization, it seems that the only place to do it is in `socket.ts`. Let me know what you think.

packages/eyes-sdk-core/lib/sdk/EyesVisualGrid.js
```diff
@@ -262,7 +262,7 @@ class EyesVisualGrid extends EyesCore {
         }
         return results.map((result, i) => {
           const testResults = result instanceof Error ? result.info && result.info.testResult : result.toJSON()
-          const exception = testResults ? null : result
+          const exception = testResults ? null : {stack: result.stack, message: result.message}
```
kyryloonufriiev, line 266: I don't think this is right because later we are trying to throw this exception. And this is just wrong to throw something that is not an instance of Error.

packages/types/src/data.ts
kyryloonufriiev: Errors name doesn't make sense, there is nothing about tests or results. It is basically just a normal js error type - https://github.com/microsoft/TypeScript/blob/main/lib/lib.es5.d.ts#L972-L976.

kyryloonufriiev (changes requested):
I don't agree with the approach that was chosen here. This will brake cores behavior just to solve serialization problems in a different package. I think the right approach would be to wrap an error with a class with the `.toJSON()` method. Since all of the errors we expect to have are already custom errors it means that we can just add a method to the `EyesError`/`TestFailedError` implementation.

SophieDePaula:
@kyryloonufriiev I created a new type `SerializedError` which has `toJSON` method and extends `Error` and reverted back the changes I did.
Lmk what you think.
@amitzur fyi

packages/eyes-sdk-core/lib/sdk/EyesClassic.js
```diff
@@ -192,7 +193,7 @@ class EyesClassic extends EyesCore {
       .then(results => {
         if (isErrorCaught) {
           if (results.info && results.info.testResult) return {testResults: results.info.testResult}
-          else return {exception: results}
+          else return {exception: new SerializedError(results)}
```
kyryloonufriiev, line 196: hmmm. Tell me if I'm wrong, we will find ourselves at this line of code only if the error is not an ordinary error object, but the one extended from EyesError. If this is true, isn't it easier to add `toJSON` method to the already created error?

kyryloonufriiev (changes requested):

packages/eyes-sdk-core/lib/sdk/EyesClassic.js
SophieDePaula: @kyryloonufriiev from what I understand this could be any error that is thrown from `close`.

packages/types/src/data.ts
```diff
@@ -323,3 +323,8 @@ export type TestResultSummary = {
   missing: number
   matches: number
 }
+
+export interface SerializedError extends Error {
+  message: string
+  stack: string
+}
```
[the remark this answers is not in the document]
SophieDePaula: I removed this type.

packages/eyes-api/src/errors/SerializedError.ts
```diff
@@ -0,0 +1,12 @@
+import * as utils from '@applitools/utils'
+
+export class SerializedError extends Error {
```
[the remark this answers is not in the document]
SophieDePaula: Yes, that's correct, I removed this form `eyes-api` and kept it only in `core`.

packages/eyes-sdk-core/lib/sdk/EyesClassic.js
kyryloonufriiev: Ah, ok I see. We have a completely different logic in `EyesVisualGrid`. Then I think you are right errors should be wrapped, but only those which are not `EyesError`'s. I also believe that the better place to do it would be `closeManager`.

SophieDePaula: @kyryloonufriiev `exception` is `readonly` so I had to do it in `EyesVisualGrid` and `EyesClassic`

kyryloonufriiev: I think you didn't understand what I meant. In closeManager you will have access to all the same data. The fact that something is readonly in types is irrelevant, because you can do whatever you want at this point.

packages/eyes-sdk-core/lib/new/close-manager.js
```diff
@@ -18,28 +19,31 @@ function makeCloseManager({runner}) {
...
+          else throw new DiffsFoundError(container.testResults)
+        } else if (container.testResults.status === TestResultsStatuses.Failed) {
+          throw new TestFailedError(container.testResults)
         }
       }
 
-      if (exception) summary.exceptions += 1
+      if (container.exception) {
+        summary.exceptions += 1
+        container.exception = new SerializedError(container.exception)
```
SophieDePaula, line 36: @kyryloonufriiev I took your advice and now I cast `exception` here. Lmk what you think.

packages/eyes-sdk-core/lib/errors/EyesError.js
```diff
@@ -15,6 +16,9 @@ class EyesError extends Error {
       this.stack = error.stack
     }
   }
+  toJSON() {
+    return utils.general.toJSON({message: this.message, stack: this.stack, reason: this.reason, info: this.info})
```
kyryloonufriiev, line 20: Why do we need `utils.general.toJSON`?

packages/eyes-sdk-core/lib/new/close-manager.js
```diff
@@ -18,28 +19,31 @@ function makeCloseManager({runner}) {
...
-        if (exception) throw exception
-
-        if (testResults.status === TestResultsStatuses.Unresolved) {
-          if (testResults.isNew) throw new NewTestError(testResults)
-          else throw new DiffsFoundError(testResults)
-        } else if (testResults.status === TestResultsStatuses.Failed) {
-          throw new TestFailedError(testResults)
-        }
+        if (container.exception) throw container.exception
       }
```
kyryloonufriiev, line 33: Why not?

```
if (throwErr && container.exception) throw exception
```

kyryloonufriiev (approved):
Looks good. You can merge it. Also this is another card you probably fixed with this PR - https://trello.com/c/wvosVkhj

PR #880, opened 2022-04-26
Account for an Appium bug when calculating system bars height

SophieDePaula:
Trello cards: [wdio](https://trello.com/c/rvlrdtce/1256-saucelabs-android-emulator-landscape-cannot-extract-empty-region) and [universal](https://trello.com/c/wbcpTjbm/1329-universal-sdk-appium-android-landscape-mode-tests-produce-incorrectly-cropped-screenshots)

I moved the logic that I initially implemented in `spec-driver-selenium` to `driver`. Still, I had to add `setOrientation` method to `spec-driver` interface and implement this method in different spec-drivers.

Background story: when we capture an image of a native mobile android device, we need to account for the system bars. We use the webdriver do to that, but there is a bug in `Appium` that when we get the dimensions of the system bars in landscape, we get the wrong coordinates.
To overcome that , I change the orientation to portrait, retrieve the dimensions of the system bars and flip again to landscape.

There are generic tests to cover these scenarios.


pending-changes.yaml
```diff
@@ -27,6 +27,7 @@ Apollo SDK:
     - accept ios and android lowercase as driver platformName capability when using custom grid
     - Fixed check region fully in classic execution when using CSS stitching
     - Support data urls in iframes
+    - Account for an Appium bug when calculating system bars height
```
SophieDePaula, line 30: @benbabayoff8 fyi

chaim-aharonson-applitools: Maybe resizing the viewport and resizing back to original dimensions will do the same effect? (something like a 'built-in' `layoutBreakPoints` functionality for these cases)
It feels wrong to get the system bars dimensions in a different orientation, maybe they're getting resized and the dimensions are actually wrong?   
Do we need to somehow set a reminder to revisit this in a few weeks/months/years in order to confirm Appium fixed this bug and we can remove this hack? 
And... how simple it is to remove this functionality from the code? maybe it should be more isolated?

SophieDePaula:
@chaim-aharonson-applitools , @kyryloonufriiev thanks for taking a look.
[I opened an issue with `appium`](https://github.com/appium/appium/issues/16867), [here](https://app.saucelabs.com/tests/4e6c0a024efb4cc8a097c5eb86698232#7) you can see the recording of the test in SauceLabs if that's helpful (with me flipping the device).

It seems that the `navigationBar` `height` and `width` are switched when on `landscape` mode, but that doesn't happen with `statusBar`.

Instead of flipping the device, I can simply take the `width` instead of the `height` in landscape mode for `navigationBar` until I get a reply on this issue. 
What do you think?

pending-changes.yaml
SophieDePaula: @chaim-aharonson-applitools I am not sure what you mean by resizing, I am not chaining the viewport size

SophieDePaula:
@kyryloonufriiev , @chaim-aharonson-applitools instead of flipping the device, I now use the width as the value that needs to be cropped out when in landscape mode.
It seems that android switches the height and width in landscape mode for the navigation bar (and not status bar).
I ran my tests on `eyes-selenium`, `eyes-selenium-universal` and `eyes-webdriverio` and all seems to work with this change.
Also, appium have another bug in Android 12 appium version 1.x, that the height of the systemBars is returned as 1 px. I did not implement any workaround for that.
Please lmk what you think about this solution.
@amitzur fyi

SophieDePaula:
@kyryloonufriiev yes, I'll change the name of the method.
The status bar always returns the same height and width in both `landscape` and `portrait` modes.

PR #882, opened 2022-04-27
Support `failTestcafeOnDiff` as environment variable

chaim-aharonson-applitools:
Support `failTestcafeOnDiff` as environment variable `APPLITOOLS_FAIL_TESTCAFE_ON_DIFF`


packages/eyes-testcafe/src/legacy.ts
```diff
@@ -201,6 +201,10 @@ export type TestCafeConfiguration = {
   tapDirPath?: string
 }
 
+function getFailTestcafeOnDiff(_testcafeConfig?: TestCafeConfiguration): boolean {
```
tourdedave, line 204: It would be better to follow the existing paradigm in `transformConfig`. There's already a case where we update the config from an environment variable. And, it is unit test/unit testable.

tourdedave (changes requested):

packages/eyes-testcafe/src/legacy.ts
tourdedave: example: https://github.com/applitools/eyes.sdk.javascript1/blob/master/packages/eyes-testcafe/src/legacy.ts#L215
test: https://github.com/applitools/eyes.sdk.javascript1/blob/master/packages/eyes-testcafe/test/unit/legacy.spec.js#L132

Although, I don't think the unit test is checking the env var case. But it would be easy enough to do so

tourdedave (approved):
👍 

Assuming all the tests pass, ship it shabtay!

PR #887, opened 2022-04-28
Redo safe area fix

kyryloonufriiev:
This PR should solve the design and functional problems of PR https://github.com/applitools/eyes.sdk.javascript1/pull/848.

__Problems of the previous solution:__
 - Safe area abstraction is exposed to the outer world, while it is meant to be an internal thing.
 - In `screenshoter` there are checks for specific selectors that should determine if the region should ignore a safe area or not, and we don't really have information about selector types.
 - Selectors that were used in `screenshoter` are not cover all of the ways to selector those bars.
 - Elements inside the bars should also be handled.
 
__Solution:__
For me the biggest problem of the previous solution was the exposition of the safe area, I believe that this kind of decision should be internal. It is important to understand why the safe area exists in the first place. The reason is that some views on ios could rewrite the safe area for itself to occupy all of the screen space, but elements such as `TopBar` and `NavBar` on the os level will overwrite whatever view is specifying. So it is very important to normalize the element region in relation to a safe area (the test with demonstration is easy to find in the `screenshoter`). But if the whole element is already out of a safe area it doesn't mean it does not exist or is not visible, probably it is just a part of `TopBar`/`NavBar`, so we don't need to intersect it with the safe area. It is something I didn't predict when I introduced this feature.


SophieDePaula:
@kyryloonufriiev I chose to check if the user tries to capture the nav bar or tab bar specifically and if so I skip the check for the intersection with the `safe area`.
I didn't know if there are some cases where we should fail as the region that we need to capture is problematic.
If this is not the case, then this solution is fine with me.

PR #888, opened 2022-04-29
Ignore self-signed certificates errors

kyryloonufriiev:
Trello: https://trello.com/c/LQiV3Fcn

Added an assignment for the `NODE_TLS_REJECT_UNAUTHORIZED` environment variable as suggested in a ticket. My only concern is a warning message produced by the node js
```
(node:39481) Warning: Setting the NODE_TLS_REJECT_UNAUTHORIZED environment variable to '0' makes TLS connections and HTTPS requests insecure by disabling certificate verification.
(Use `node --trace-warnings ...` to show where the warning was created)
```


pending-changes.yaml
```diff
@@ -12,13 +12,14 @@ Apollo SDK:
...
   bug-fix:
-    - 
+    - ignore error produced by nodejs when sending requests to https server with self-signed certificates
 "@applitools/eyes-selenium":
   feature:
     - 
   bug-fix:
     - Fixed check region fully in classic execution when using CSS stitching
 "@applitools/eyes-webdriverio":
+    - ignore error produced by nodejs when sending requests to https server with self-signed certificates
```
amitzur, line 22: @benbabayoff8 FYI ([Trello](https://trello.com/c/LQiV3Fcn/1314-js-multiple-customers-ssl-issue-w-universal-sdk)]

amitzur (commented):
@kyryloonufriiev is there a way to disable this warning?

kyryloonufriiev:
@amitzur I believe this is only possible when you run node with a specific flag (`--no-warnings`)

kyryloonufriiev:
@tourdedave exactly what I was imagining. 👍

packages/eyes-sdk-core/test/unit/server/ServerConnector.spec.js
```diff
@@ -830,4 +828,29 @@ render height & width are required when deviceEmulationInfo is not provided, req
     )
     assert.ok(renderErr.message.includes(`Error: combination of url, dom, resources is invalid`))
   })
+
+  it('works with self signed certificates', async () => {
```
amitzur, line 832: Does this test fail without Axios' custom `httpsAgent` in `ServerConnector`?

packages/visual-grid-client/src/sdk/resources/createFetchOptions.js
```diff
@@ -22,6 +23,12 @@ function createFetchOptions(resource, {referer, userAgent, proxy, autProxy, cook
     fetchOptions.headers['Cookie'] = createResourceCookieHeader(resource.url, cookies)
   }
 
+  if (resource && /^https/.test(resource.url)) {
+    fetchOptions.agent = new https.Agent({
```
amitzur, line 27: There's a conflict here with the agent we create when `proxy && proxy.getIsHttpOnly()`, if I understand correctly

amitzur: Also, if there is a proxy, then you probably want to set `rejectUnauthorized: false` in any case (maybe if the proxy itself is `https`?) because it's usually the proxy's certificate that is self signed, not the resource's certificate.

In other words - why not always set it?

amitzur (commented):
Looks very good! See my inline comments. From my perspective this can be shipped, either with my comments addressed or not.

tourdedave:
> Looks very good! See my inline comments. From my perspective this can be shipped, either with my comments addressed or not.

Okay. I'll try and ship it tonight.

PR #889, opened 2022-04-29
execution-grid-client

tourdedave:
Here are some open items that I'm noting for reference (in no particular order):

- Exposing this as a binary

For the initial launch, are we expecting the user to have a node run-time and call it good enough by using `bin`? (@amitzur)

- e2e testing

We'll want to include some simple happy/sad path e2e tests.

And, also some that exert a certain amount of load and potential boundary conditions like large DOM-snapshot payloads (stretch).

- Logging strategy

What do we want to log to troubleshooting customer issues? Writing a file to disk in the same tmp dir as the universal server? Potentially logging performance metrics for commands/routes we know to be potential hot spots (e.g., execute script)?

- Adding a README

(and noting in it that this is a complement to the visual grid -- read: not meant for classic SDK execution at the moment)

- Spawning the tunnel binary, communicating with the tunnel server with each driver setup/teardown, and tracking tunnel/driver state as needed

- Waiting for CDP support in the EG


packages/execution-grid-client/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "scripts": {
+    "test": "mocha --no-timeouts test/selenium-proxy.spec.js",
+    "lint": "eslint . --ext .js"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "axios": "^0.27.2",
+    "eslint": "^8.14.0",
```
amitzur, line 15: can be moved to `devDependencies`

packages/execution-grid-client/src/selenium-proxy.js
```diff
@@ -0,0 +1,106 @@
+const express = require('express')
+const morgan = require('morgan')
+const {createProxyMiddleware, responseInterceptor} = require('http-proxy-middleware')
+const axios = require('axios')
+const Queue = require('./queue')
+
+function createServer({host, port, forwarding_url, withQueue, withRetry} = {}) {
+  let queue
+  async function q(req, res, next) {
+    if (req.path === '/wd/hub/session' && req.method === 'POST') {
```
amitzur, line 10: path might not have `/wd/hub`. The spec just says `POST /session` https://www.w3.org/TR/webdriver/#endpoints

packages/execution-grid-client/test/selenium-proxy.spec.js
```diff
@@ -0,0 +1,44 @@
...
+const {Builder} = require('selenium-webdriver')
+const chrome = require('selenium-webdriver/chrome')
+const chromedriver = require('chromedriver')
+const assert = require('assert')
+
+describe('selenium-proxy', () => {
+  let egServer, egClient
+  before(async () => {
+    await chromedriver.start(['--port=4446', '--url-base=wd/hub'], true)
+    egServer = await createSeleniumProxy({
```
amitzur, line 11: I think this is extra level, since the chromedriver itself is enough to represent the server.
The queue thing is just for testing purposes, right?

packages/execution-grid-client/test/selenium-proxy.spec.js
```diff
@@ -0,0 +1,44 @@
...
+      host: 'localhost',
+      port: 4445,
+      forwarding_url: 'http://localhost:4446/wd/hub',
```
amitzur, line 14: why not `forwardingUrl` as the JS gods intended?

amitzur (commented):
This shows the idea well. I think it's in the right direction.
Not sure why the queue is needed. Left a few inline comments.

packages/execution-grid-client/package.json
tourdedave: Whoops.

packages/execution-grid-client/src/selenium-proxy.js
tourdedave: Yeah, I just mimicked what we see from other grid providers. We can drop the `/wd/hub`

tourdedave: > Not sure why the queue is needed. Left a few inline comments.

This was just for testing. Trying to simulate what will be happening in the EG server, so we can adapt it as needed for integration testing.

packages/execution-grid-client/test/selenium-proxy.spec.js
tourdedave: Yep

packages/execution-grid-client/test/selenium-proxy.spec.js
tourdedave: Didn't even realize I did that 😂

packages/execution-grid-client/test/proxy.spec.js
```diff
@@ -0,0 +1,113 @@
...
+
+  afterEach(async () => {
+    nock.cleanAll()
+    await proxy.close()
+  })
+
+  it('proxies webdriver requests', async () => {
+    proxy = await createProxyServer()
+
+    nock('https://exec-wus.applitools.com')
```
tourdedave, line 16: Does nock implicitly handle assertions? Or is there something explicit we need to add?

packages/execution-grid-client/src/proxy.js
```diff
@@ -0,0 +1,84 @@
...
+      const responseBody = await parseBody(proxyResponse, {encoding: 'utf-8'})
+      response.body = JSON.parse(responseBody)
+      if (
+        response.body.value.error === 'session not created' &&
+        ['CONCURRENCY_LIMIT_REACHED', 'NO_AVAILABLE_DRIVER_POD'].includes(response.body.value.data.appliErrorCode)
+      ) {
+        await utils.general.sleep(RETRY_BACKOFF[Math.min(request.retry, RETRY_BACKOFF.length - 1)])
+        request.retry += 1
+        request.removeAllListeners()
+        proxy.web(request, response, {buffer: request.body && streamify(JSON.stringify(request.body))})
```
tourdedave, line 38: 💯 the secret sauce for making the retry. Nice

tourdedave (commented):
A no-fluff implementation with a more dialed in testing solution 👌

Well done!

Here are some open items that I'm noting for reference (in no particular order):

- Exposing this as a binary

For the initial launch, are we expecting the user to have a node run-time and call it good enough?

- e2e testing

We'll want to include some simple happy/sad path e2e tests.

And, also some that exert a certain amount of load and potential boundary conditions like large DOM-snapshot payloads (stretch).

- Logging strategy

Writing to a file to disk in the same tmp dir as the universal server?

Potentially logging performance metrics for commands/routes we know to be potential hot spots (e.g., execute script)

- Adding a README

(and noting in it that this is a complement to the visual grid -- read: not meant for classic SDK execution at the moment)

- Spawning the tunnel binary, communicating with the tunnel server, and tracking tunnel/driver state

packages/execution-grid-client/test/proxy.spec.js
kyryloonufriiev: I don't think it contains any kind of assertions by default. What assertions do you have in mind?

packages/execution-grid-client/src/proxy.js
kyryloonufriiev: Yep, this is something I had realized as the last step. Once you read the body from the request stream it doesn't have it anymore.

packages/execution-grid-client/test/proxy.spec.js
tourdedave: Primarily - how do we know when our code isn't working? 😄 

I saw this in the nock docs: https://github.com/nock/nock#expectations

It seemed to imply that if the scope specified wasn't called that it would throw. If that's true, then it's probably enough.

kyryloonufriiev: Hmmm. The tests are designed in a way that if our code will not work selenium will throw an error during starting a session.

tourdedave: True, that's probably enough.

PR #890, opened 2022-04-29
Make sure android platformName always starts with uppercase

SophieDePaula:
[Trello card](https://trello.com/c/jl8jTFUr/1334-universal-android-w3c-os-is-lowercase)


SophieDePaula:
@amitzur this is ready for review.

pending-changes.yaml
```diff
@@ -13,6 +13,7 @@ Apollo SDK:
     - 
   bug-fix:
     - Account for an Appium bug when calculating system bars height
+    - Make sure android platformName starts with uppercase if `config.useUppercaseForPlatformName` is set to true
```
SophieDePaula, line 16: @benbabayoff8 this is an internal change, but tagging you here just in case.

benbabayoff8: @SophieDePaula It will affect customers' baselines. So I'm not sure this is an internal change. is it the Universal server's changelog?

SophieDePaula: @benbabayoff8 yes

PR #891, opened 2022-05-01
support `cy.eyesGetAllTestResults` in TS

chaim-aharonson-applitools:
@benbabayoff8 please review


benbabayoff8:
@chaim-aharonson-applitools We used the term "TypeScript" in previous releases, didn't we? We should be consistent.

packages/eyes-cypress/eyes-index.d.ts
```diff
@@ -30,5 +30,16 @@ declare namespace Cypress {
...
+      /**
+       * Returns an object with the applitools test results from a given test / test file. This should be called after close.
+       * @example
+       * after(() => {
+       *  cy.eyesGetAllTestResults().then(summary => {
+       *    console.log(summary)
+       *  })
+       * })
+       */
+      eyesGetAllTestResults(): any
```
amitzur, line 43: Actually this returns a promise to a `TestResultsSummary`. See here:
https://github.com/applitools/eyes.sdk.javascript1/blob/master/packages/types/src/data.ts#L314

amitzur (changes requested):

packages/eyes-cypress/eyes-index.d.ts
chaim-aharonson-applitools: done

amitzur (changes requested):
1. I think it's better to get the type from `eyes-api`. The type from `types` is used in eyes-api code (to define the parameter to the constructor) but it's not exactly like the TestResultsSummary class in eyes-api. For example, there are not properties `passed`, `unresolved`, etc. on the return value.

2. The function returns a sort of `Promise`. It's a Cypress object. Can you check the type of that Cypress promise-like object and put it here. It would be applicable also to the other methods.

chaim-aharonson-applitools:
@amitzur all requested changes are done, also added TS basic test, please review

amitzur (commented):
1. The folder under test/e2e/ts should not be a project with readme, pacakge.json etc. It's not discoverable, it's just hiding there. If there's anything you want to document about the workflow, then put it somewhere else (either we should have a CONTRIBUTING.md, or discourse, or a google doc, or a gist, reference from the sdk knowledge share spreadsheet, etc. You get the idea). the `test/e2e/ts` folder should just be a container that holds what it needs to test cypress with typescript.

2. The folder structure was confusing. There was double `integration`, `support`, and `plugins` folders, both inside the `cypress` folder and outside it.

3. Cypress can run TypeScript OOTB without a pre compilation step. It's all written here: https://docs.cypress.io/guides/tooling/typescript-support So `@cypress/webpack-preprocessor` is not needed, nor is `@bahmutov/add-typescript-to-cypress` (that package was last published 3 years ago, when TS was not supported OOTB in Cypress. You can see in the link that Cypress added this support for TS OOTB from version 4.4.0). btw, Gleb Bahmutov is not just a random guy - he used to be the VP engineering at Cypress).

4. We do want to verify that compilation passes, so we configure `noEmit` in `tsconfig.json`.

5. The `plugins` and `support` files should reference the **source code**, to actually test it.

6. The test itself is very complicated and also not accurate:
  - `cy.eyesGetAllTestResults` should not be inside `afterEach` but inside `after`. It's misleading to put it in `afterEach`.
  - Don't use `assert` in Cypress tests. It's running in the browser and Cypress provides its own assertion builtin library.

7. The line `import { TestResultsSummary } from '@applitools/eyes-api';` inside the test is very wrong. Think of the test as something our user will do. And you run it automatically to verify that if our user writes the same thing as this file, then he will be successful. Our user will not import from eyes-api. And it's also not needed for the test. To be honest, I tried to make the test understand the type of `summary` and could not. I tried it also on a clean Cypress project and I think they have an issue there. But that doesn't mean we should instruct our users to import eyes-api. We could export this type from `eyes-cypress` if we want. But we don't at this point just because we're swamped.

8. The whole thing was not incorporated into the test pipeline. It was only a preparation and was done to run it manually. Notice the change in package.json scripts:
  - `test:ts:compile` - tests compilation (does NOT emit JS code)
  - `test:ts:run` - tests run (does NOT depend on compilation)
  - `test:ts` - runs both of them
  - `test` - runs `test:ts` as part of the pipeline.

pending-changes.yaml
```diff
@@ -91,7 +91,7 @@ Apollo SDK:
   feature:
     - 
   bug-fix:
-    - 
+    - Support `cy.eyesGetAllTestResults` in TypeScript
```
amitzur, line 94: @benbabayoff8 FYI

PR #892, opened 2022-05-02
check if browser include setOrientation method for wdio5

SophieDePaula:
Seems that `setOrientation` does not exist in `wdio5`


kyryloonufriiev:
What about wdio 6?

SophieDePaula:
@kyryloonufriiev I don't see that other methods like `getSystemBars` and `getOrientation` were added to `v6.d.ts`. Did you handle it differently in `wido6`?

kyryloonufriiev (changes requested):
Ah sorry. For some reason, I thought that wdio 6 also has some methods missed in the types.

Doesn't `setOrientation` accept an argument?

SophieDePaula:
@kyryloonufriiev thanks for catching that, I fixed it.
Lmk, if there is anything else and if not I'll merge and release.

SophieDePaula:
@kyryloonufriiev I changed it to just `string`, it does work with lowercase as well.

kyryloonufriiev:
It works in terms of ts, but actually did you ever run this command with lowercase values? Because I see that you are passing lowercase values when I believe it should be uppercase.

SophieDePaula:
@kyryloonufriiev yes, before I changed the logic to use the `width` in `landscape` mode the device flipping worked.

kyryloonufriiev:
Ah. Then fine, however, it might not work for some older appium versions, but not important.

kyryloonufriiev (approved):

PR #893, opened 2022-05-05
[eg-client] Add tunnel support

kyryloonufriiev:
- [x] Implement logic to create and delete tunnels
- [ ] Spawn tunnel manager server


packages/execution-grid-client/package.json
```diff
@@ -42,7 +42,7 @@
     "@applitools/logger": "1.1.1",
     "@applitools/utils": "1.2.14",
     "http-proxy": "1.18.1",
-    "node-fetch": "3.2.4",
+    "node-fetch": "2",
```
tourdedave, line 45: Why `2`? Because there's not an equivalent `@types/nod-fetch` version?

tourdedave (commented):
Looks good. I'm assuming we don't have an e2e test for the tunnel because of the fact that we need to spawn it manually. Is that right?

packages/execution-grid-client/package.json
kyryloonufriiev: Ah. Since `node-fetch@3` it is esm-only module. And we have to support commonjs so it is recommended to use the version before 3. But I will fix it to latest minor of 2 as well as sync `@types/node-fetch` version.

PR #894, opened 2022-05-09
[driver] Fix an issue with viewport size on native devices

kyryloonufriiev:
[Trello](https://trello.com/c/jf2pM29b)
The reason for the issue was not sanitized status bar height. If for some reason it is absent (we usually treat it as 0 height) it was affecting viewport height since we are getting a difference between display height and status bar height, and when status bar height is `undefined` we are getting `NaN` as a result.

I'm also improving logging a little bit as part of this PR.


amitzur (commented):
A brief review:
1. I liked very much the new logs in various places. Very informative.
2. `app-no-navbar.spe.js` - there is a typo which might cause the file not to run (spe.js --> spec.js)
3. Where did you get `viewportTestApp.apk` from?

I actually looked at everything except the important part in `driver.ts` where the calculations were changed :)

packages/driver/src/driver.ts
```diff
@@ -157,31 +159,43 @@ export class Driver<TDriver, TContext, TElement, TSelector> {
...
-      }
       if (this.isAndroid) {
-        this._driverInfo.statusBarHeight /= this.pixelRatio
-        this._driverInfo.navigationBarHeight /= this.pixelRatio
+        // bar sizes could be extracted only on android
+        const barsSize = await this._spec.getBarsSize?.(this.target).catch(() => undefined as never)
+        if (barsSize) {
+          this._logger.log('Driver bars size', barsSize)
+
+          // navigation bar height is confused with the with in landscape orientation on android
```
SophieDePaula, line 172: @kyryloonufriiev typo, `with the width` rather than with.
I would also use `height is replaced with the width` , I think it makes more sense.

packages/driver/src/driver.ts
```diff
@@ -157,31 +159,43 @@ export class Driver<TDriver, TContext, TElement, TSelector> {
...
+          this._driverInfo.navigationBarHeight &&= this._driverInfo.navigationBarHeight / this.pixelRatio
+        }
       }
 
       // calculate viewport size
       if (!this._driverInfo.viewportSize) {
         this._driverInfo.viewportSize = {
           width: displaySize.width,
-          height: displaySize.height - this._driverInfo.statusBarHeight,
+          height: displaySize.height - this.statusBarHeight,
```
SophieDePaula, line 198: @kyryloonufriiev we don't need to consider the `navigationBarHeight`?

SophieDePaula:
@kyryloonufriiev looks good to me, just added one comment and one question

packages/driver/src/driver.ts
kyryloonufriiev: Display size is returned already without navbar height.

packages/driver/src/driver.ts
kyryloonufriiev: Fixed

kyryloonufriiev:
@amitzur 
1. Nice to read!
2. Fixed
3. `viewportTestApp.apk` this is an app from the reproducer in the [Trello](https://trello.com/c/VdEyKCn2/1344-universal-sdk-error-when-checking-andoroid-app-without-navbar) card I've attached

PR #895, opened 2022-05-09
Adjust android landscape native

SophieDePaula:


kyryloonufriiev:
I would love you to point me on the test which proves it is useful at all, I think we should do something like this only when we 100% sure this is related to the version of the platform.

SophieDePaula:
@kyryloonufriiev, I found something that feels more reliable than this current solution. I see that `getSystemBars` return also the coordinates of the bar. So if `x === 0` we know that the `navigationBar` is on the left side and if it's greater than `0` it's on the right. I can return this value from `getSystemBars`.
What do you think?

amitzur:
@SophieDePaula what trello is this addressing?

SophieDePaula:
@amitzur this is not addressing any Trello card. This is in regards to the discussion re `platformVersion`.
I rely on `platformVersion` to understand if the `navigationBar` is on the right or the left side of the display. But I found that I get the coordinates of the bar and I can determine the side by that which is a better solution.
I still need to implement it, but was waiting to hear what Kytylo thinks.

SophieDePaula:
closing this one and will open a new PR to implement the logic I mentioned above.

PR #902, opened 2022-05-12
Cypress -support element as target region

SophieDePaula:
[Trello](https://trello.com/c/hX9Ys8bh/30-support-passing-cypress-element-as-target-region)

### DOTO:

- [x] README for target
- [x] README for other regions (ignore, floating, accessibility)
- [x] Universal server cannot start with the latest version


packages/eyes-cypress/src/browser/eyesCheckMapping.js
```diff
@@ -21,17 +21,21 @@ function toCheckWindowConfiguration(config = {}) {
...
+    contentRegions: covertRegions(config.content),
     accessibilityRegions: convertAccessabilityRegions(config.accessibility),
   };
 
   if (config.target === 'region') {
     if (!Array.isArray(config.selector)) {
-      if (!config.hasOwnProperty('selector')) {
+      if (config.element) {
+        regionSettings = {
+          region: {selector: config.element.selector},
```
kyryloonufriiev, line 36: isn't there a case when an element doesn't have a selector?

SophieDePaula:
@amitzur , @kyryloonufriiev this is ready for review

pending-changes.yaml
```diff
@@ -100,7 +100,7 @@ Apollo SDK:
     - Support data urls in iframes
 "@applitools/eyes-cypress":
   feature:
-    - 
+    - add support to send elements as Target, ignore, floating, accessability and match regions.
```
amitzur, line 103: My proposal:

`cy.eyesCheckWindow` now supports passing DOM elements and jQuery objects (which are the result value of `cy.get` and similar Cypress DOM queries) for target and regions (ignore, floating, accessibility, and match regions).

Happy to hear feedback @SophieDePaula @benbabayoff8

packages/eyes-cypress/src/browser/eyesCheckMapping.js
```diff
@@ -1,104 +1,173 @@
...
+  }
+
+  function refElements(regions) {
+    if (!regions) return regions;
+    if (!(regions instanceof Array)) regions = [regions];
+    const elements = [];
+    regions.map(region => {
+      if (isHTMLElement(region)) {
+        elements.push(refer.ref(region));
+      } else if (region.constructor.name === 'jQuery') {
```
kyryloonufriiev, line 153: What if I did `jQuery([1,2,3,4])`? This also will have constructor name `jQuery`.

kyryloonufriiev (changes requested):
I'm glad you decide to go with a ref representation of elements instead of selectors. I think the next step should be to decide with which one of the two (native and jQ) supported element types you want to work with. I think it should be native elements, so in the very beginning, you will convert jQ objects to normal dom and continue wor from there. I suggest this [jQuery#toArray()](https://api.jquery.com/toArray/#toArray).

amitzur:
@SophieDePaula I'm going to push a few changes to reflect my thoughts, and to avoid some ping-pong in the review. I'll indicate what I've done

packages/eyes-cypress/src/browser/eyesCheckMapping.js
```diff
@@ -161,13 +160,13 @@ function makeEyesCheckMapping(refer) {
...
-      return false;
-    }
-  }
+  // #endregion
+}
+
+function isHTMLElement(element) {
+  // Avoiding instanceof here since the element might come from an iframe, and `instanceof HTMLElement` would fail.
+  // This check looks naive, but if anyone passes something like {nodeType: 1} as a region, then I'm fine with them crashing :)
+  return element.nodeType && element.nodeType === Node.ELEMENT_NODE;
```
kyryloonufriiev, line 168: Ah, nice observation.

amitzur (changes requested):
1. This is the first time I notice the custom coverage tests code style. Check out my changes in `ShouldSendAccessibilityRegionsByDomElementWithVg.spec.js`. We can change the existing tests in another PR, but I'd like the new tests to be written like that for simplicity. No reason to copy the code style of emitted code. They are over complicated because it's generated code.
Specifically:
  - not using `assert` since there's no need, and also **this is code in the browser**. Cypress already provides assertions, and there is no builtin `assert` module in the browser.
  - accessing properties not as string accessors.
  - add display name to test
  - map jQuery using `.toArray` and not specifying each element manually.

3. Why are tests for ignore regions separate from the tests for accessibility and floating? You are also not checking the output of the regions in those tests.

4. If I understand correctly, you noticed that we were missing support for accessibility regions by coordinates and fixed it. So please add a test for that.

5. I reorganized `eyesCheckMapping`.
  - The default of `{}` now sits at the top in the custom command. That makes it easier to reason about. 
  - No need for the inner function `toCheckWindowConfiguration`
  - Change logic to detect HTMLElement and jQuery
  - Change `.map` to `for/of` loop. Don't use `map` if you are not mapping.


Lastly, I don't know if you noticed, but `cy.get` returns elements inside iframe. That was surprising to me. But it works! See the accessibility test that I changed - the first region is actually an element inside iframe

packages/eyes-cypress/src/browser/commands.js
```diff
@@ -5,14 +5,14 @@ const Refer = require('./refer');
 const Socket = require('./socket');
 const {socketCommands} = require('./socketCommands');
 const {eyesOpenMapValues} = require('./eyesOpenMapping');
-const {eyesCheckMapValues} = require('./eyesCheckMapping');
+const {makeEyesCheckMapping} = require('./eyesCheckMapping');
```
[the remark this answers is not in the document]
SophieDePaula: @amitzur I was doing that to not sent the refer for every `check` call, but only once at the beginning of the file.

pending-changes.yaml
SophieDePaula: @amitzur I think it's great, I would only leave out `(which are the result value of cy.get and similar Cypress DOM queries)` as I feel it's a bit too long, and also we have examples in the `README`

benbabayoff8: Agree with Sophie. I'd recommend: "cy.eyesCheckWindow now supports passing DOM elements and jQuery object for target and regions". We will explain further in the ReadMe

packages/eyes-cypress/README.md
```diff
@@ -310,6 +310,17 @@ cy.eyesCheckWindow({ tag: 'Login screen', target: 'your target' })
   });
   ```
 
+Pass an element as the target
```
amitzur, line 313: @SophieDePaula this is misplaced. I will make the necessary changes

amitzur, line 313: @SophieDePaula this is misplaced. I will make the necessary changes

packages/eyes-cypress/test/fixtures/testApp/cypress/integration-run/region.js
```diff
@@ -41,6 +41,28 @@ describe('eyes-cypress', () => {
     });
   });
 
+  it('region by JQuery element', () => {
```
amitzur, line 44: So you kept these tests even though you wrote new ones?

amitzur, line 44: So you kept these tests even though you wrote new ones?

amitzur: @SophieDePaula :point_up:

amitzur: @SophieDePaula :point_up:

PR #914, opened 2022-05-19
[browser-extension] Update manifest to v3

kyryloonufriiev:
I was using this [guide](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/#updating-manifest-dot-json) provided by Google to update `manifest.json`.
The problem here is that migration to manifest v3 also means changes in extension api. Fortunately, we use only one of the APIs that were changed, the required changes are described [here](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/#executing-arbitrary-strings). Since we don't use chrome api directly but instead through the browser extension polyfill, I'm not sure if we need to perform any changes there, and also how it will affect our firefox support.


tourdedave (changes requested):
Based on [this issue in webextension-polyfill](https://github.com/mozilla/webextension-polyfill/issues/329#issuecomment-884938069) it looks like we will need to update our call for `executeScript` to match what's documented. Also - how do we test the new manifest version? It's not clear to me from the docs if this is something that's already shipped that we can play with.

pending-changes.yaml
```diff
@@ -113,7 +113,7 @@ Apollo SDK:
     - Allow running with self-signed certificates
 eyes-browser-extension:
   feature:
-    - 
+    - Migrate to manifest v3
```
amitzur, line 116: @benbabayoff8

kyryloonufriiev:
@tourdedave I actually see that the new `scripting.executeScript` thing is the part of web extension configuration, so I believe we have to start using it in the code.
About tests, I believe it simply doesn't work right. And our coverage test will demonstrate it. If not I will check how we can test it.

kyryloonufriiev:
@tourdedave @amitzur This is actually a big change since v3 forbids us from dynamic code generation (`new Function`). I was able to avoid it everywhere except `executeScript` command. I believe this is also possible because we need it to unref args and ref results. Another problem might be with dom-snapshot and dom-capture scripts which we intentionally use as a string, but we always can start using them as functions.

kyryloonufriiev:
There is another challenge. Since v3 extensions are changed the background script, to run in a web worker instead of a page, `axios` stopped working. It happens because workers don't have the `XMLHttpRequest` object. I don't know how it could be fixed without making changes in the core. The preferable change would be to refactor `ServerConnector` to use `fetch` instead of `axios`.

zamboney:
@kyryloonufriiev the `core` build will be failed until we will merge [this pr](https://github.com/applitools/dom-scripts/pull/112)

.github/actions/tests/action.yml
```diff
@@ -68,7 +68,7 @@ runs:
           yarn --cwd ../scripts install
           yarn --cwd ../scripts link
           yarn link @applitools/scripts
-          npx link "${{inputs.links}}" --build --install
```
kyryloonufriiev, line 71: isn't that killing the logic of linking it in the line above?

zamboney: No, this is done the same thing just it's set the `@applitools/scripts` as the root `package`, and just like any other `npx` command, it will first try to resolve the package in `./node_modules` and because `@applitools/scripts` was `yarn link` before it will resolve the linked `@applitools/scripts` package and would not install it globally

js/packages/driver/src/driver.ts
```diff
@@ -136,6 +136,9 @@ export class Driver<TDriver, TContext, TElement, TSelector> {
   get isEdgeLegacy(): boolean {
     return /edge/i.test(this.browserName) && Number(this.browserVersion) <= 44
   }
+  get canExecuteStringScript(): boolean {
+    return this._driverInfo.features.canExecuteStringScript ?? true
+  }
```
kyryloonufriiev, line 141: No, need for a specific getter for that. There is already a getter for `fetuares`

js/packages/core/src/classic/utils/take-dom-capture.ts
```diff
@@ -1,6 +1,13 @@
 import {type Driver} from '@applitools/driver'
 import {type Logger} from '@applitools/logger'
-import {getCaptureDomPoll, getPollResult, getCaptureDomPollForIE, getPollResultForIE} from '@applitools/dom-capture'
+import {
+  getCaptureDomPoll,
+  getPollResult,
+  getCaptureDomPollForIE,
+  getPollResultForIE,
+  pollResultCjs,
+  captureDomPollCjs,
```
kyryloonufriiev, line 9: I am thinking that cjs versions of the script should be dynamically imported only if needed.

js/packages/core/src/ufg/utils/take-dom-snapshot.ts
```diff
@@ -37,13 +38,15 @@ export async function takeDomSnapshot<TContext extends Context<unknown, unknown,
   }
   const scripts = {
     main: {
-      script: `return (${
-        isLegacyBrowser ? await getProcessPagePollForIE() : await getProcessPagePoll()
-      }).apply(null, arguments);`,
+      script: canExecuteStringScript
+        ? `return (${isLegacyBrowser ? await getProcessPagePollForIE() : await getProcessPagePoll()}).apply(null, arguments);`
+        : await import('@applitools/dom-snapshot').then(({processPagePoll}) => processPagePoll),
```
kyryloonufriiev, line 43: I love the fact that you don't use `require`! But once built it will be `require`, right?

js/packages/core/src/ufg/utils/take-dom-snapshot.ts
```diff
@@ -23,6 +23,7 @@ export async function takeDomSnapshot<TContext extends Context<unknown, unknown,
 }): Promise<DomSnapshot> {
   const driver = context.driver
   const isLegacyBrowser = driver.isIE || driver.isEdgeLegacy
+  const {canExecuteStringScript} = driver.features
```
kyryloonufriiev, line 26: This will be always false because you never give it a default value of `true`. You should design it in a way falsy is an acceptable default (e.g. `canExecuteOnlyFunctionScripts` or any other naming with a similar meaning), or simply give this flag a default value of `true`.

zamboney: no problem i will fix it

js/packages/core/src/ufg/utils/take-dom-snapshot.ts
zamboney: it will keep the `await import` as the `moduleResolution` set to `node16` [ref](https://www.typescriptlang.org/tsconfig#moduleResolution)
if you think it should be changed to `require` i will change it.

kyryloonufriiev: I actually said that I love using `import` instead of `require`.  But it should build it to `require` because we support commonjs. And `moduleResolution` and `module` are two different things. So since we use `module: "commonjs"` i would expect it to be built as `require` and not `import`.

PR #922, opened 2022-05-25
[cypress] Improve type definitions

kyryloonufriiev:
[Trello](https://trello.com/c/uaqa0K5N)


pending-changes.yaml
```diff
@@ -109,7 +109,7 @@ Apollo SDK:
     - Fixed bug when navigation bar was presented in screenshot on Android 12
 "@applitools/eyes-cypress":
   feature:
-    - 
+    - Improve type definitions
```
amitzur, line 112: :+1: approved

packages/eyes-cypress/index.d.ts
```diff
@@ -0,0 +1,89 @@
...
+import type * as api from '@applitools/eyes-api'
+
+type MaybeArray<T> = T | T[]
+
+type LegacyRegion = {left: number; top: number; width: number; height: number}
+type Selector = {selector: string; type?: 'css' | 'xpath', nodeType?: 'element' | 'shadow-root'} | 'string'
+type Element = HTMLElement | JQuery<HTMLElement>
+
+interface CypressCheckSettings extends types.CheckSettings<Element, Selector> {
+  tag?: CypressCheckSettings['name']
```
amitzur, line 13: I didn't know it was possible to do something like this

packages/eyes-cypress/index.d.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  target?: 'window' | 'region'
+  selector?: Selector
+  element?: Element
+
+  ignore?: MaybeArray<CypressCheckSettings['ignoreRegions'][number] | LegacyRegion>
```
amitzur, line 19: what is the `[number]` used for?

amitzur (commented):
Structure looks good. Try to run `yarn test:ts:compile`. It fails for me. It's something I added recently. Here's what I get:

```
➜  eyes-cypress git:(cypress-types) ✗ yarn test:ts:compile      
yarn run v1.22.17
$ tsc --project test/e2e/ts/cypress
test/e2e/ts/cypress/integration/typescript.spec.ts:12:7 - error TS2345: Argument of type '{ appName: string; testName: string; }' is not assignable to parameter of type 'CypressEyesConfig'.
  Object literal may only specify known properties, and 'appName' does not exist in type 'CypressEyesConfig'.

12       appName: 'Cypress typescript App',
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

test/e2e/ts/cypress/integration/typescript.spec.ts:15:5 - error TS2769: No overload matches this call.
  Overload 1 of 2, '(tag?: string): null', gave the following error.
    Argument of type '{ target: string; fully: boolean; waitBeforeCapture: number; }' is not assignable to parameter of type 'string'.
  Overload 2 of 2, '(settings?: CypressCheckSettings): null', gave the following error.
    Argument of type '{ target: "window"; fully: boolean; waitBeforeCapture: number; }' is not assignable to parameter of type 'CypressCheckSettings'.
      Object literal may only specify known properties, and 'fully' does not exist in type 'CypressCheckSettings'.

 15     cy.eyesCheckWindow({
        ~~~~~~~~~~~~~~~~~~~~
 16       target: 'window',
    ~~~~~~~~~~~~~~~~~~~~~~~
... 
 18       waitBeforeCapture: 1000,
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 19     })
    ~~~~~~



Found 2 errors in the same file, starting at: test/e2e/ts/cypress/integration/typescript.spec.ts:12

error Command failed with exit code 2.
```

I didn't run `yarn test:ts:run` to see what it outputs.

packages/eyes-cypress/index.d.ts
kyryloonufriiev: This is needed here because the type of `CypressCheckSettings['ignoreRegions']` is an array. And to get the type of the element we can use `[number]`, or even just `[0]`.

kyryloonufriiev: This is needed here because the type of `CypressCheckSettings['ignoreRegions']` is an array. And to get the type of the element we can use `[number]`, or even just `[0]`.

kyryloonufriiev:
@amitzur I've fixed this failed test, problem was due to tsconfig didn't have `moduleResolution` property set to `node` as it intended.

packages/eyes-cypress/index.d.ts
amitzur: Ahh got it

packages/eyes-cypress/index.d.ts
```diff
@@ -0,0 +1,89 @@
+/// <reference types="cypress" />
+
+import type * as types from '@applitools/types'
+import type * as api from '@applitools/eyes-api'
+
+type MaybeArray<T> = T | T[]
+
+type LegacyRegion = {left: number; top: number; width: number; height: number}
```
SophieDePaula, line 8: @kyryloonufriiev Shouldn't this be called `CoordinatesRegion`?

packages/eyes-cypress/index.d.ts
```diff
@@ -0,0 +1,89 @@
...
+  content?: MaybeArray<CypressCheckSettings['contentRegions'][number] | LegacyRegion>
+  strict?: MaybeArray<CypressCheckSettings['strictRegions'][number] | LegacyRegion>
+  floating?: MaybeArray<CypressCheckSettings['floatingRegions'][number] | (({element: Element} | Selector | LegacyRegion) & {maxUpOffset?: number; maxDownOffset?: number; maxLeftOffset?: number; maxRightOffset?: number})>
+  accessibility?: MaybeArray<CypressCheckSettings['accessibilityRegions'][number] | (({element: Element} | Selector | LegacyRegion) & {accessibilityType?: types.AccessibilityRegionType})>
+
+  scriptHooks?: CypressCheckSettings['hooks']
+}
+
+interface CypressEyesConfig extends types.EyesConfig<Element, Selector> {
+  browser?: MaybeArray<CypressEyesConfig['browsersInfo'][number] | {deviceName: string; screenOrientation?: types.ScreenOrientation; name?: string}>
```
SophieDePaula, line 30: @kyryloonufriiev we can also have

`batch: {name, id, sequenceName, notifyOnComletion, properties}`

kyryloonufriiev: Your comment is valid. And this type allows you to do it because it is derivative from `EyesConfig` which has all of the basics. And in this file I'm specifying just cypress specifics

packages/eyes-cypress/index.d.ts
kyryloonufriiev: No. First of all, we already use this naming in `eys-api`. The `Legacy` prefix here to differentiate from a "new" region type which has `x`& `y` instead of `top` & `left`. We support the "old" way just for backward compatibility, but the core's internals is already `x` & `y` only.

kyryloonufriiev:
@amitzur Short answer - yes. But the `"moduleResolution": "node"` is required if you want to import anything. I believe most of the users already have it.

kyryloonufriiev:
@SophieDePaula I've fixed this test. But e2e tests stucks on the global hooks test.

packages/eyes-cypress/index.d.ts
amitzur: If we want to remove it some day we can use `Omit`

kyryloonufriiev:
@SophieDePaula The problem with the new `getConfig` was only because `require` caches files, so it wasn't possible to get the same file with different configurations.

PR #923, opened 2022-05-25
Fix coded region calculation bug in CSS stitching

tourdedave:
[Trello](https://trello.com/c/tHqghsNu)

When performing a check window in classic execution with CSS stitching, if the page is scrolled prior to performing the check command (e.g., if the test interacts with an element and the driver implicitly scrolls the element into view), the coded region value that we send to Eyes ends up being incorrect.

This PR fixes that. Generic tests for CSS and Scroll stitching were added [here](https://github.com/applitools/sdk.coverage.tests/pull/157).


kyryloonufriiev:
To be honest I don't think that introduction of the flag that skips some behavior is an acceptable solution. If you say the problem is in the `contextLocationInViewport` value, then we should fix that. 
I think an actual problem here is that in this case when we transform regions we should be scrolled at a position where the screenshot was taken, which means to (0,0) in full-page mode.

tourdedave:
> I think an actual problem here is that in this case when we transform regions we should be scrolled at a position where the screenshot was taken, which means to (0,0) in full-page mode.

Agreed. I wasn't satisfied with the solution myself but wasn't sure about a better way. I don't think the issue is in `contextLocationInViewport`. It's more of what you said - the page being in the incorrect position when we transform. Are you thinking the fix would be in screenshoter? If so, do you think the fix would be to scroll to 0,0 after taking a full page screenshot (when scrollingMode is CSS)? Or just not restoring the scroller state after?

tourdedave:
When previously stabilizing screenshoter tests, we removed scrolling to `0,0` before taking a full page stitched screenshot and added compensation to the screenshotRegion. I believe this change (the removal of scrolling to `0,0`) created this issue with CSS stitching & coded regions but we missed it because it was a case not covered by any tests (until now).

Restoring scrolling to `0,0` fixes the issue and doesn't introduce any side-effects (confirmed with screenshoter e2e tests and coverage tests in eyes-selenium when linked). The compensation used in calculating the screenshot region is still needed though for accurate page coverage calculations.

pending-changes.yaml
```diff
@@ -14,6 +14,7 @@ Apollo SDK:
     - Increased max payload size from 100Mb to 254Mb
   bug-fix:
     - Fixed `CheckSetting`'s `fully` being overridden by `Configuration`'s `forceFullPageScreenshot`
+    - Fixed incorrect calculation of coded regions in classic mode when using CSS stitching
```
tourdedave, line 17: @benbabayoff8 ☝️

tourdedave:
Failures 1 & 3 were caused by my sdk.coverage.tests branch being outdated. They pass now. For failure 2, I made it so we don't scroll to `0,0` on element fully. All coverage tests pass now. @kyryloonufriiev can you please review?

packages/screenshoter/package.json
```diff
@@ -63,7 +63,7 @@
     "@applitools/logger": "1.1.5",
     "@applitools/snippets": "2.2.3",
     "@applitools/utils": "1.3.0",
-    "jpeg-js": "^0.4.3",
+    "jpeg-js": "0.4.3",
```
tourdedave, line 66: I made this fixed so I could run `yarn deps` (it errors otherwise)

pending-changes.yaml
tourdedave: @amitzur Per Ben's request while he's OOO - can you review this changelog entry?

packages/screenshoter/src/take-screenshot.js
```diff
@@ -63,7 +63,9 @@ async function takeScreenshot({
     if (!window && !driver.isNative) {
       await scrollIntoViewport({context: target.context, region: target.region, scroller: target.scroller, logger})
     }
-
+    if (fully && !region && target.scroller) {
+      await target.scroller.moveTo({x: 0, y: 0})
+    }
```
kyryloonufriiev, line 68: @tourdedave 
```suggestion
    if (fully && !target.region && target.scroller) {
      await target.scroller.moveTo({x: 0, y: 0})
    }
```

I also this that if you scrolled to 0,0 here we don't need to scroll to 0,0 again in `takeStitchedScreenshot`.

packages/eyes-sdk-core/test/e2e/repro.e2e.spec.js
```diff
@@ -13,27 +13,24 @@ describe.skip('repro', () => {
       driver,
       config: {
         appName: 'check e2e',
-        testName: 'check e2e test',
+        testName: 'repro 1277',
```
kyryloonufriiev, line 16: what is the point of this file? Is that something I don't know about?

kyryloonufriiev (approved):
Looks good. I've requested one change. Can you please validate if it makes sense? If it doesn't, then feel free to merge?

packages/eyes-sdk-core/test/e2e/repro.e2e.spec.js
tourdedave: It was just something used to initially reproduce the issue. It's a file that lives in core/test/e2e but is always skipped

packages/screenshoter/src/take-screenshot.js
tourdedave: I'm not sure if it doesn't make sense. It's hard for me to reason about. It would require a lot of re-testing to verify.

PR #927, opened 2022-06-01
Use ceil instead of round when getting size in image.js

SophieDePaula:
[Trello card](https://trello.com/c/rvlrdtce/1256-saucelabs-android-emulator-landscape-cannot-extract-empty-region)

In some cases the `divide` method that is called in `take-stitched-screenshot` receives different widths of `region` and `image.size` it is different in one pixel. 
We get the region dimensions from `getElementContentSize` snippet and the `image.size` by scaling down the image.
Since we are dealing with fractions here we round the numbers we get along the way and it seems that the browser rounding and our rounding is different.
Therefore, it was decided, that instead of using `round` method, we will now use `ceil`. 
This resolved the issue that was presented in the trello card above (I added a test to reproduce the issue) and also ran all `screenshoter` tests. 
All the tests passed except one, which I had to update the image as now it is bigger in 1 pixel.


packages/screenshoter/test/e2e/android-web/landscape-fully-scroll.spec.js
```diff
@@ -0,0 +1,35 @@
...
+describe('web fully landscape', () => {
+  let driver, destroyDriver
+
+  before(async () => {
+    ;[driver, destroyDriver] = await makeDriver({
+      type: 'android',
+      app: 'chrome',
+      logger,
+      sauceDeviceName: 'Google Pixel 4 GoogleAPI Emulator',
+      sauceAndroidVersion: '12.0',
```
kyryloonufriiev, line 12: I think betted just do `deivceName` and `platformVersion` so we can use it not only for sauce and android environmets.

packages/screenshoter/test/e2e/e2e.js
```diff
@@ -32,6 +32,7 @@ exports.sleep = utils.general.sleep
 
 exports.test = async function test({type, tag, driver, ...options} = {}) {
   if (options.withStatusBar) tag += '-statusbar'
+  options['debug'] = true
```
kyryloonufriiev, line 35: Do you need it here?

kyryloonufriiev (changes requested):
Alos fixture should be in `android-web` directory and not just `android`.

pending-changes.yaml
```diff
@@ -13,21 +13,25 @@ Apollo SDK:
     - Dorp support for Node.js versions <=12
   bug-fix:
     - Fixed incorrect calculation of coded regions in classic mode when using CSS stitching
+    - Change the way we round image size when dealing with fractions in mobile devices
```
SophieDePaula, line 16: @amitzur / @kyryloonufriiev , Ben is OOO can you please approve this?

pending-changes.yaml
```diff
@@ -13,25 +13,25 @@ Apollo SDK:
     - Dorp support for Node.js versions <=12
   bug-fix:
     - Fixed incorrect calculation of coded regions in classic mode when using CSS stitching
-    - Change the way we round image size when dealing with fractions in mobile devices
+    - Change the way we round image size when dealing with fractions
```
SophieDePaula, line 16: @amitzur can you please approve this one?

amitzur: `Fix rounding error of image size when scaling introduces fractions`
Is that accurate? I think you should emphasize what is fixed. Bluntly speaking, nobody cares that you changed the way you do something :)

amitzur (changes requested):

kyryloonufriiev (approved):

PR #929, opened 2022-06-03
[universal] Added alpine binary

kyryloonufriiev:
[Trello](https://trello.com/c/xCmJE2dH)
This is a small PR, but I'm wondering how to test it. I would like to run it the same way we are running tests for other platforms, which means in the matrix of the GitHub actions. The problem is that GitHub actions don't allow you to specify `runs-on: alpine`, but maybe something like that would help - https://github.com/marketplace/actions/run-on-architecture.
Really want it to be as seamless as possible. 

- [x] Build for alpine
- [x] Test alpine binary


tourdedave:
Yeah, running in a Docker alpine image would be a good way to go. What's the check to perform? Make sure the binary runs and returns the correct thing (e.g., stdout string, exit code, or some a heartbeat ping)?

PR #932, opened 2022-06-03
[Cypress] Throw an exception for CORS iframe and add the ability to send driver to check

SophieDePaula:
[Trello card](https://trello.com/c/xxTyiKjc/1312-wlw-cypress-universal-bad-render)

This PR should solve an implementation problem in cypress `spec-driver` related to how we handle the context.
Up until this PR, every time I had to use the `document` in cypress `spec-driver` (which is for almost every action `spec-driver` performs) I first validate that it exist and if not, I set it back to the window that contains the page that is tested.
The reason for that, is that the `driver` in cypress `spec-driver` is actually the document of the page and if the user navigates to a different page or call `open` before `visit` I lose the `document`.
This create a problem when dealing with a page that contains CORS Iframe as Cypress do not support CORS iframes. When we try to switch to such a frame, we get null context and then we restore the context to the main window which results in an incorrect screenshot.

To resolve this, now if we cannot get the context of a frame, I throw an exception.
In addition, to establish a more robust implementation of `spec-driver`, I now send the `driver` in check instead of refreshing the context for every call to a `spec-driver` method.


packages/eyes-sdk-core/lib/sdk/EyesCore.js
```diff
@@ -26,13 +26,13 @@ class EyesCore extends EyesBase {
...
-  async check(checkSettings = {}) {
-    return this._check(checkSettings)
+  async check(checkSettings = {}, driver) {
+    return this._check(checkSettings, driver)
   }
 
   async checkAndClose(checkSettings, throwEx) {
     this._logger.log(`checkAndClose(checkSettings) - begin`)
-    return this._check(checkSettings, true, throwEx)
+    return this._check(checkSettings, undefined, true, throwEx)
```
SophieDePaula, line 35: I decided to send here `undefined` for the driver as I am not sure who calls this method.

pending-changes.yaml
```diff
@@ -52,7 +52,7 @@ Apollo SDK:
   feature:
     - 
   bug-fix:
-    - 
+    - Fix incorrect test results when dealing with a page that contains CORS iframes
```
SophieDePaula, line 55: @amitzur fyi

packages/eyes-sdk-core/lib/sdk/EyesClassic.js
```diff
@@ -76,8 +76,15 @@ class EyesClassic extends EyesCore {
...
-    await this._driver.init()
+  async _check(checkSettings = {}, driver, closeAfterMatch = false, throwEx = true) {
+    if (driver) {
+      const useCeilForViewportSize = this._configuration.getUseCeilForViewportSize()
+      const keepPlatformNameAsIs = this._configuration.getKeepPlatformNameAsIs()
+      const customConfig = {useCeilForViewportSize, keepPlatformNameAsIs}
+      this._driver = await new Driver({spec: this.spec, driver, logger: this._logger, customConfig}).init()
+    } else {
+      await this._driver.init()
+    }
```
kyryloonufriiev, line 87:
```suggestion
    if (driver) {
      const useCeilForViewportSize = this._configuration.getUseCeilForViewportSize()
      const keepPlatformNameAsIs = this._configuration.getKeepPlatformNameAsIs()
      const customConfig = {useCeilForViewportSize, keepPlatformNameAsIs}
      this._driver = new Driver({spec: this.spec, driver, logger: this._logger, customConfig})
    }
    await this._driver.init()
    
```

packages/eyes-cypress/src/browser/spec-driver.ts
```diff
@@ -28,13 +27,20 @@ export function mainContext(): Context {
 }
 
 export function parentContext(context: Context): Context {
-  if (!context) return; // because Cypress doesn't support cross origin iframe, then childContext might return null, and then the input to parentContext might be null
+  // because Cypress doesn't support cross origin iframe, then childContext might return null, and then the input to parentContext might be null
+  if (!context) {
+    throw 'got null context from parentContext'
```
kyryloonufriiev, line 32: We should always throw `new Error(...)`. I also would recommend a more general message such as `'Context is not accessible'`, since we don't need to use implementation details (names of the functions) in error messages.

pending-changes.yaml
SophieDePaula: @benbabayoff8 please approve

PR #962, opened 2022-06-06
Cypress add support for secured ws

SophieDePaula:
[Trello card](https://trello.com/c/yKNkzel0/1281-cypress-paigeai-operation-is-insecure-error-on-firefox-when-opening-websocket)

This PR add support for secured ws for both the local proxy server and sends `key` and `cert` to the universal server.


pending-changes.yaml
```diff
@@ -52,7 +53,7 @@ Apollo SDK:
   feature:
     - 
   bug-fix:
-    - 
+    - add support for secure websocket
```
SophieDePaula, line 56: @benbabayoff8 fyi

benbabayoff8: @SophieDePaula I'd vote to move it to the feature section. What do you think?

kyryloonufriiev (approved):
👍🏼

PR #964, opened 2022-06-07
make sure to call waitBeforeCapture when unable to setViewportSize

SophieDePaula:
[Trello card](https://trello.com/c/Fw8o2343/1399-nasdaq-waitbeforecapture-feature-not-working-as-expected)

When the user uses `layoutBreakpoints` with environments that includes width less than `500px` and heaful browser, we cannot set the viewport size. As a result, we don't call `waitBeforeCapture`.


packages/eyes-sdk-core/lib/utils/takeDomSnapshots.js
```diff
@@ -73,13 +72,17 @@ async function takeDomSnapshots({
...
       } else {
         const failedBrowsers = browsersInfo.map(({name}) => `(${name})`).join(', ')
         const message = chalk.yellow(
           `The following configurations [${failedBrowsers}] have a viewport-width of ${requiredWidth} pixels, while your local browser has a limit of ${actualViewportSize.width} pixels, so the SDK couldn't resize it to the desired size. As a fallback, the resources that will be used for these checkpoints have been captured on the browser's limit (${actualViewportSize.width} pixels). To resolve this, you may use a headless browser as it can be resized to any size.`,
         )
         logger.console.log(message)
+        logger.log(message)
       }
+    } finally {
+      if (waitBeforeCapture) await waitBeforeCapture()
```
kyryloonufriiev, line 85: Makes sense!

kyryloonufriiev (approved):

pending-changes.yaml
```diff
@@ -30,6 +32,8 @@ Apollo SDK:
     - Fixed broken links to enums implementation in the README.md
     - |
       Fixed `forceFullPageScreenshot` option behavior
+    - |
+      Make sure to call `waitBeforeCapture` when the viewport size is less than 500 px
```
SophieDePaula, line 36: @benbabayoff8 fyi

benbabayoff8: @SophieDePaula I'd continue with the same conventions above - "fixed calling 'waitBeforeCapture' when the viewport size is less than 500px

kyryloonufriiev: @benbabayoff8 I think it should be something like `when failed to set required viewport size`. Because the same issue could happen as well when you trying to set too big in headful mode

benbabayoff8: @kyryloonufriiev Agreed.

PR #965, opened 2022-06-08
Improve the logic to detect navBar location in native android

SophieDePaula:
In continuation to the discussion [here](https://github.com/applitools/eyes.sdk.javascript1/pull/895) I now rely on the navigation bar X coordinate to determine the side of the navigation bar in native Android.


pending-changes.yaml
```diff
@@ -12,7 +12,7 @@ Apollo SDK:
   feature:
     - 
   bug-fix:
-    - 
+    - Improve the logic that detects the side of native Android navigation bar
```
SophieDePaula, line 15: @benbabayoff8 fyi

packages/types/src/driver.ts
```diff
@@ -103,9 +104,12 @@ export interface SpecDriver<TDriver, TContext, TElement, TSelector> {
...
   setOrientation?(driver: TDriver, orientation: ScreenOrientation): Promise<void>
-  getBarsSize?(
-    driver: TDriver,
-  ): Promise<{statusBarHeight: number; navigationBarHeight: number; navigationBarWidth: number}>
+  getBarsSize?(driver: TDriver): Promise<{
+    statusBarHeight: number
+    navigationBarHeight: number
+    navigationBarWidth: number
+    navBarXcoordinate: number
+  }>
```
kyryloonufriiev, line 112: No need to invent your own format. It would be better to just use the object as-is from appium.

packages/screenshoter/src/take-viewport-screenshot.js
```diff
@@ -167,8 +167,8 @@ function makeTakeNativeScreenshot({driver, stabilization = {}, debug, logger}) {
         image.crop({
           top: driver.statusBarHeight,
           bottom: driver.orientation === 'landscape' ? 0 : driver.navigationBarHeight,
-          left: driver.platformVersion >= 8 ? driver.navigationBarHeight : 0,
-          right: driver.platformVersion < 8 ? driver.navigationBarHeight : 0,
+          left: driver._driverInfo.navBarXcoordinate === 0 ? driver.navigationBarHeight : 0,
+          right: driver._driverInfo.navBarXcoordinate > 0 ? driver.navigationBarHeight : 0,
```
kyryloonufriiev, line 171: I think we should have a better abstraction here. Maybe something like `viewportLocation` that is calculated on the driver's side.

kyryloonufriiev (changes requested):

pending-changes.yaml
benbabayoff8: @SophieDePaula Looks good to me :)

packages/types/src/driver.ts
SophieDePaula: @kyryloonufriiev looking at the `spec-driver` implementation I'm inclined against it because we have some logic there, for example:

`navigationBarWidth: navigationBar.visible ? navigationBar.width : 0`

which we would need to implement everywhere we use one of these values.

kyryloonufriiev:
I've pushed my commit. This way we can abstract out bar sizes almost everywhere, by just using `viewportRegion` which has coordinates that depend on bars placement.

kyryloonufriiev:
@SophieDePaula I had added a test in `screeenshoter` to test how the new algorithm detects nav bar position. So here are some points I want to share.
The previous implementation which bound navbar placement with the platform version was flown. There is indeed a different navbar placement in landscape mode in older versions of **appium** which we have to use with old versions of android. However, this is just because older appium default for landscape orientation is 90 degrees while the new appium default is 270 degrees. So the difference in landscape mode is old good landscape-right vs. landscape-left, which could happen on any device old or new. So I created two tests to test in both landscape orientations, and then I find out that the navbar placement (that we get with the `getSystemBars` method) doesn't really work that well, so I had to find a better way to detect the orientation of the screen.
Long story short, works really well right now. We don't need to check bar sizes and orientation in a bunch of places anymore, only during initialization where we perform all basic calculations.

kyryloonufriiev (approved):

SophieDePaula:
@kyryloonufriiev got it, thanks for the letting me know.

PR #969, opened 2022-06-13
Padded regions

chaim-aharonson-applitools:
### [WIP] 
Allow adding padding to regions.
relevant to: `ignoreRegions`, `layoutRegions`, `strictRegions`, `contentRegions`
### API (for JS)
`ignoreRegions: [{ selector: '#ignore_me', padding: 15 }]` // this will apply the value of padding to all sides of element
`ignoreRegions: [{ selector: '#ignore_me', padding: {left: 10, top: 10} }]` // this will add padding only to relevant sides
### Validations
- Values are numbers
- Values are not empty or zero, in case a 'side' (aka: left, right) has no value or '0' the 'side' will be ignored
- In case all values are empty or zero the whole padding will be totally ignored
 
https://trello.com/c/PYKqJLbg/42-padded-regions


kyryloonufriiev (changes requested):
I believe you have miss understood this feature. First of all, there is no need to change driver or snippet packages. They don't need to know about the padding.
The point where you modify a region should be definitely in core, I think somewhere in the CheckSettingsUtils file.
Also, I see that you properly implemented a change in eyes-api, but for some reason described the api as `{selector, padding}`. This would mean that you interfere with selector type and this is not right at all. Because this feature could be used also with elements or even {x, y, width, height} objects.

chaim-aharonson-applitools:
> I believe you have miss understood this feature. First of all, there is no need to change driver or snippet packages. They don't need to know about the padding. The point where you modify a region should be definitely in core, I think somewhere in the CheckSettingsUtils file. Also, I see that you properly implemented a change in eyes-api, but for some reason described the api as `{selector, padding}`. This would mean that you interfere with selector type and this is not right at all. Because this feature could be used also with elements or even {x, y, width, height} objects.

thanks, I fixed according to the CR
` ignoreRegions: [{ region: '#section_2', padding: 15 }],` is supported as well if that is what you meant.

chaim-aharonson-applitools:
tests [WIP]:

### eyes-sdk-core > unit
`toCheckWindowConfiguration handles regions padding` verify that given region padding will result the same padding values, if padding 'side' is not set: ensure it receives 0, also ensure that case padding is Object and padding is Number will have the same values at the end
### eyes-sdk-core > it
`check region with padding` perform two 'eyes.check', first without padding, second with padding, ensure that the **first** result element `Rect` dimensions are equal to **second** result element `Rect` dimensions once adding the 'padding' values, in short: first width is 900, padding is 10 -> expect the second width to be 900 + 10 + 10
### GCT (generic coverage tests) [WIP]
I have a ready structure, need some modifications in order to use it 

I guess there are many tests can be added...

packages/eyes-sdk-core/lib/sdk/CheckSettingsUtils.js
```diff
@@ -231,7 +242,24 @@ async function toScreenshotCheckSettings({checkSettings, context, screenshot}) {
...
+  const assignToAll = utils.types.isNumber(padding) && padding > 0
+  return ['top', 'right', 'bottom', 'left'].reduce((mem, prop) => {
+    mem[prop] = assignToAll
+      ? padding
+      : padding[prop] && utils.types.isNumber(padding[prop]) && padding[prop] > 0
+      ? padding[prop]
+      : 0
+    return mem
+  }, {})
+}
```
kyryloonufriiev, line 255: I don't think you need it here. You should pass whatever you will get from the user to the function an then process it there.

packages/eyes-sdk-core/lib/sdk/CheckSettingsUtils.js
```diff
@@ -211,7 +217,12 @@ async function toScreenshotCheckSettings({checkSettings, context, screenshot}) {
         if (elements.length > 0) {
           const contextLocationInViewport = await elements[0].context.getLocationInViewport()
           for (const element of elements) {
-            const region = utils.geometry.offset(await element.getRegion(), contextLocationInViewport)
+            const elementRegion = appendPaddingToRegion(
+              await element.getRegion(),
+              getSerializedPadding(reference.padding),
+            )
+            console.log('>>> elementRegion', elementRegion, reference.padding)
```
kyryloonufriiev, line 224: You should use `logger` and note `console`

packages/visual-grid-client/src/sdk/calculateSelectorsToFindRegionsFor.js
```diff
@@ -93,6 +94,10 @@ function regionWithUserInput({regionObject, userRegion, regionName}) {
       maxRightOffset: userRegion.maxRightOffset,
     })
   }
+  regionObject.left -= padding.left
+  regionObject.width += padding.left + padding.right
+  regionObject.top -= padding.top
+  regionObject.height += padding.top + padding.bottom
```
kyryloonufriiev, line 100: the utils I had recommended would be also helpful here

packages/types/src/setting.ts
```diff
@@ -32,6 +33,11 @@ type AccessibilityRegion<TRegion> = {
   type?: AccessibilityRegionType
 }
 
+type PaddedRegion<TRegion> = {
+  region: TRegion
+  padding?: number | PaddingRect
+}
```
kyryloonufriiev, line 39: Better place it in the `data.ts` file. And call it `OffsetRect`, to not attach this type to padding only. at the end of the day, it is just an offset from each side. We already use the same type for image cropping you can find it in `data.ts`, would be nice if you integrate those types (use `OffsetRect` to define `ImageCropRect`).

kyryloonufriiev (changes requested):

packages/driver/src/spec-utils.ts
```diff
@@ -71,7 +74,7 @@ export function splitSelector<TSelector>(
     }
   }
 
-  return {contextSelectors, elementSelector}
+  return {contextSelectors, elementSelector, padding}
```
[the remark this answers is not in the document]
chaim-aharonson-applitools: totally forgot...

packages/eyes-sdk-core/lib/sdk/CheckSettingsUtils.js
chaim-aharonson-applitools: removed

packages/types/src/setting.ts
chaim-aharonson-applitools: done

packages/eyes-sdk-core/lib/sdk/CheckSettingsUtils.js
chaim-aharonson-applitools: done

chaim-aharonson-applitools:
@benbabayoff8 please review the change-log

benbabayoff8:
@chaim-aharonson-applitools Looks good to me

chaim-aharonson-applitools:
also add support for GCT --> test.config.visualgrid as Boolean

packages/utils/src/geometry.ts
```diff
@@ -247,3 +247,21 @@ export function divide(region: Region, size: RectangleSize, padding: {top?: numb
   }
   return subRegions
 }
+export function withPadding(region: Region, padding: any): Region {
```
kyryloonufriiev, line 250: Don't use `any` if it is possible
```suggestion
export function withPadding(region: Region, padding: number | {top?: number; right?: number; bottom?: number; left?: number}): Region {
```

kyryloonufriiev: When you will use a padding object without knowing if all of the fields are specified you can use `?? 0` to replace it with 0 if it wasn't specified

packages/types/src/data.ts
```diff
@@ -57,13 +57,20 @@ export type TextRegion = Region & {text: string}
...
+  top?: number
+  right?: number
+  bottom?: number
+  left?: number
 }
 
+export type PaddedRegion<TRegion> = {
+  region: TRegion
+  padding?: number | OffsetRect
+}
```
kyryloonufriiev, line 70: No need to put it here. This is related to settings only same as FloatedRegion and AccessibilityRegion

packages/sdk-coverage-tests/src/generate/emit.js
```diff
@@ -26,6 +26,7 @@ function emitTest(test, {makeSpecEmitter, makeFile}) {
     throw new Error(`Missing implementation for test ${test.name}`)
   }
   test.config.baselineName = test.config.baselineName || test.key
+  test.config.visualgrid = test.vg || false
```
kyryloonufriiev, line 29: Why?

kyryloonufriiev (changes requested):

packages/utils/src/geometry.ts
```diff
@@ -247,3 +247,21 @@ export function divide(region: Region, size: RectangleSize, padding: {top?: numb
...
+}
+function getSerializedPadding(padding: any) {
+  const assignToAll = types.isNumber(padding) && padding > 0
+  return ['top', 'right', 'bottom', 'left'].reduce((mem, prop) => {
+    mem.set(
+      prop,
+      assignToAll ? padding : padding[prop] && types.isNumber(padding[prop]) && padding[prop] > 0 ? padding[prop] : 0,
+    )
+    return mem
+  }, new Map<string, number>())
```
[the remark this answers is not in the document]
chaim-aharonson-applitools: your suggestion will not work: 1. in case there is no padding at all 2. in case the padding object has partial 'sides' (like: `{left: 10}`) the result I wanted to achieve here is a unified object for all cases, once padding is not set it'll return object with all sides as 0, on missing 'side' it will still return it as a 0.
Validation is a habit I got used to... always suspect users are humans... I limited the values to positive since we can't predict the results and failures negative might cause.
Map is due to the IDE errors I received once it was simple `{}`

packages/types/src/data.ts
chaim-aharonson-applitools: that's where it was, I thought you asked me to move it out... now it's back

packages/sdk-coverage-tests/src/generate/emit.js
chaim-aharonson-applitools: in one of the first GCT it implemented there was a diff in one of the measurements comparing  UFG vs. Classic, back then I added `isVg` to config of test itself abusing the fact that any parameter in config object will automatically passed to generated test, back then I had a condition like: `...left: config.isVg ? 120 : 100...` it was a hack and in the end Amit spit that test (with two variations) to two separated tests. 
Now, for the GCT I added for padding, it appears like UFG and Classic have the same issue, a single pixel in `height` value. I added this parameter to the test and IMO it is useful.
Additionally, your comment caused to me to rethink that this pixel diff might be a cause of rounding the numbers, I guess it happens before adding the padding values in one case and on the other case it is done after adding the padding values... will re-check that now

packages/types/src/data.ts
kyryloonufriiev: A did ask you to move `OffsetRect`. but no problem

packages/utils/src/geometry.ts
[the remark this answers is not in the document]
kyryloonufriiev: Ok, then you can do

```
if (types.isNumber(padding)) {
  padding = {left: padding, right: padding, top: padding, bottom: padding}
} else {
  padding = {left: paddinng.left ?? 0, right: paddinng.right ?? 0, top: paddinng.top ?? 0, bottom: paddinng.bottom ?? 0,}
}
```

I understand your point about the validation. It is a classic dilemma in development. But this is not a user-facing function this is for us (developers) and we have typescript to check if we pass something that makes sense. All user inputs that could propagate here should be validated much earlier so we don't have the validation in every function.

> Map is due to the IDE errors I received once it was simple {}

You should never change an implementation due to types. Only if it is for good, but this is not the case.

packages/sdk-coverage-tests/src/generate/emit.js
chaim-aharonson-applitools: I just saw that 1 pixel diff in `height` (UFG vs. Classic) is a know issue.
here:
https://github.com/applitools/sdk.coverage.tests/blob/universal-sdk/coverage-tests.js#L864
https://github.com/applitools/sdk.coverage.tests/blob/universal-sdk/coverage-tests.js#L882
first is the **Classic** test, look at the expected: `...height: 184`
second is the **UFG** test, same element but expected height is **`185`**
so, might not be perfect but not a regression I caused :-)
BTW, my fix will solve the duplication above, copy config record from second test and add it to second + wrapping the 'expected height' with new condition. bingo :-)
### I know there is at least one more deprecated duplication, I will refactor once done with this branch and it GCT

kyryloonufriiev (approved):
👏🏼 🥳 Feel free to merge, but please do not release it for now.

packages/visual-grid-client/src/sdk/calculateSelectorsToFindRegionsFor.js
```diff
@@ -55,7 +67,8 @@ function calculateSelectorsToFindRegionsFor({
               regionObject.left = Math.max(0, regionObject.left - imageLocation.x)
               regionObject.top = Math.max(0, regionObject.top - imageLocation.y)
             }
-            regions.push(regionWithUserInput({regionObject, userRegion, regionName}))
+            const padding = userRegion.padding
+            regions.push(regionWithUserInput({regionObject, userRegion, regionName, padding}))
```
amitzur, line 71: @chaim-aharonson-applitools why do you need to add another parameter, `padding`, here? It is already available in the `userRegion` parameter.

packages/visual-grid-client/src/sdk/calculateSelectorsToFindRegionsFor.js
```diff
@@ -1,10 +1,22 @@
...
+  const res = utils.geometry.withPadding(tempRegionObject, padding)
+  return {
+    width: res.width,
+    height: res.height,
+    left: res.x,
+    top: res.y,
+  }
+}
 
 function calculateSelectorsToFindRegionsFor({
```
amitzur, line 21: @chaim-aharonson-applitools I would expect additional unit tests to be added to `calculateSelectorsToFindRegionsFor.test.js`

packages/utils/src/geometry.ts
```diff
@@ -247,3 +247,23 @@ export function divide(region: Region, size: RectangleSize, padding: {top?: numb
...
 }
+export function withPadding(
+  region: Region,
+  padding: number | {top?: number; right?: number; bottom?: number; left?: number},
+): Region {
+  if (types.isNumber(padding)) {
+    padding = {left: padding, right: padding, top: padding, bottom: padding}
+  } else {
+    padding = {
+      left: padding?.left ?? 0,
```
amitzur, line 258: @chaim-aharonson-applitools how could `padding` be undefined at this point?

amitzur: And if so, then how come it's not defined as an optional parameter?

chaim-aharonson-applitools: this is an optional parameter, I will remind you that we spoke about it and you suggested to always expect to have full padding object at the end of the flow, I thought that in case its not need why should it be set to anything, eventually you were right (what else :-) ) it was easier to always have padding object even in case it is with zero values

packages/visual-grid-client/src/sdk/calculateSelectorsToFindRegionsFor.js
chaim-aharonson-applitools: will add

packages/visual-grid-client/src/sdk/calculateSelectorsToFindRegionsFor.js
chaim-aharonson-applitools: you are correct, fixing that

PR #971, opened 2022-06-14
Support boolean for LazyLoadOptions

tourdedave:
In eyes-api we have a sensible default value of `true` (see [here](https://github.com/applitools/eyes.sdk.javascript1/blob/fix/lazyloadTypes/packages/eyes-api/src/input/CheckSettings.ts#L610)). In order for this to compile, we will need update the type to optionally accept booleans. This is currently blocking us from releasing eyes-api ([link](https://github.com/applitools/eyes.sdk.javascript1/runs/6883799569?check_suite_focus=true#step:4:115)). I'm opening a PR for this because I know this item was a point of contention when finalizing the PR for lazy loading.


tourdedave:
@kyryloonufriiev please review as soon as you can

kyryloonufriiev (approved):
Ah. I always wanted it!

PR #972, opened 2022-06-15
Add regionId to coded regions

tourdedave:
[Trello](https://trello.com/c/pv51sCYA)


tourdedave:
@kyryloonufriiev A few things to note.

I did:

- verify the behavior through integration tests using the mock driver.

- add e2e tests for this behavior in core

I did not:

- get a chance to verify the e2e tests since I don't have access to the test2 environment. So I'm waiting for this functionality to be deployed to the public cloud.

- get a chance to check on serialization of the selector (or deriving it from an element) in different drivers. So the behavior is still localized to core (it's [here](https://github.com/applitools/eyes.sdk.javascript1/blob/b9b5e2933739ed3dda6407187f29df7dc8af6e98/packages/eyes-sdk-core/lib/sdk/CheckSettingsUtils.js#L6-L8)).

- add the ability for us to specify a regionId manually

packages/eyes-sdk-core/lib/sdk/CheckSettingsUtils.js
```diff
@@ -3,6 +3,10 @@ const utils = require('@applitools/utils')
 const snippets = require('@applitools/snippets')
 const ImageMatchSettings = require('../config/ImageMatchSettings')
 
+function getRegionIdFromElement(element) {
+  return element && element.target && element.target.selector
+}
```
kyryloonufriiev, line 8: You don't need to use `target`, just `element.selector`.

packages/eyes-sdk-core/lib/geometry/Region.js
```diff
@@ -162,6 +162,7 @@ class Region {
       this._top = top
       this._width = width
       this._height = height
+      this._regionId = regionId
```
kyryloonufriiev, line 165: I really don't want to add more stuff to this class since it should be removed soon. But it seems reasonable to do it if we don't want deep dive into the refactoring.

kyryloonufriiev (changes requested):

packages/eyes-sdk-core/lib/geometry/Region.js
tourdedave: Did you see that I also added it to [`AccessibilityMatchSettings`](https://github.com/applitools/eyes.sdk.javascript1/blob/a6094696b838e20d841ea142fe06f1e27e53a2d1/packages/eyes-sdk-core/lib/config/AccessibilityMatchSettings.js#L31) and [`FloatingMatchSettings`](https://github.com/applitools/eyes.sdk.javascript1/blob/a6094696b838e20d841ea142fe06f1e27e53a2d1/packages/eyes-sdk-core/lib/config/FloatingMatchSettings.js#L33)?

packages/eyes-sdk-core/lib/sdk/CheckSettingsUtils.js
tourdedave: The integration tests disagree. With just `element.selector` they all fail.

tourdedave: It returns `{selector: 'selector-value'}` instead of just the selector string value.

kyryloonufriiev: It returns you, a `CommonSelector` type. You can check it here 
https://github.com/applitools/eyes.sdk.javascript1/blob/7cda72a9313973d14817e8be62722c80aa52d7e5/packages/types/src/driver.ts#L55-L58

tourdedave: `element.selector` isn't always populated though, is it? It looks like when running against the vg that `element.selector` could be undefined but `element.target` (and subsequently `element.target.selector`) is populated.

tourdedave: Something like d758d50e24925cb6db04c08381e7c64c773054da works. If the goal is to initially serialize Selenium and ship quickly, then wouldn't this be enough for an initial release? Then we could update the spec drivers in a follow-up version?

kyryloonufriiev: I don't understand how `element.target.selector` could return you something. `element.target` is selenium WebElement, but selenium doesn't store selector in element instance.

kyryloonufriiev: Ah. Then of course. We add the selector to the mock element. But the bottom line is you should not use `.target` at all.

tourdedave: @kyryloonufriiev For some reason, the new spec function is not showing up for me in driver/element.ts when testing. I'm pretty sure I have everything linked/built correctly - but I couldn't find the shell script we used to use so I did it by hand. Is there something else you run to do linking? Or did I miss a step in wiring this up?

packages/spec-driver-selenium/src/spec-driver.ts
```diff
@@ -63,6 +63,16 @@ export function transformSelector(selector: Selector | CommonSelector): Selector
   }
   return selector
 }
+export function untransformSelector(selector: Selector): CommonSelector {
+  if (utils.types.isString(selector)) {
+    return {type: 'css', selector: selector}
+  } else if (byHash.includes(Object.keys(selector)[0])) {
+    const s = Object.entries(selector)[0]
+    return {type: s[0], selector: s[1]}
+  } else if (utils.types.has(selector, 'using') && utils.types.has(selector, 'value')) {
```
kyryloonufriiev, line 72:
```suggestion
  } else if (utils.types.has(selector, ['using', 'value'])) {
```

packages/types/src/driver.ts
```diff
@@ -71,6 +71,7 @@ export interface SpecDriver<TDriver, TContext, TElement, TSelector> {
   transformDriver?(driver: any): TDriver
   transformElement?(element: any): TElement
   transformSelector?(selector: Selector<TSelector>): TSelector
+  untransformSelector?(selector: any): any
```
kyryloonufriiev, line 74:
```suggestion
  untransformSelector?(selector: TSelector | Selector<TSelector>): Selector
```

packages/spec-driver-selenium/src/spec-driver.ts
```diff
@@ -63,6 +63,16 @@ export function transformSelector(selector: Selector | CommonSelector): Selector
   }
   return selector
 }
+export function untransformSelector(selector: Selector): CommonSelector {
+  if (utils.types.isString(selector)) {
+    return {type: 'css', selector: selector}
+  } else if (byHash.includes(Object.keys(selector)[0])) {
+    const s = Object.entries(selector)[0]
+    return {type: s[0], selector: s[1]}
```
kyryloonufriiev, line 71: The problem with that is the fact that `By.linkText('...')` and `{linkText: '...'}` will generate different common selectors.
I would better do something like this
```suggestion
    const [[method, selector]] = Object.entries(selector)
    const by = Selenium.By[method](selector)
    return {type: by.using, selector: by.value}
```

packages/eyes-sdk-core/lib/sdk/CheckSettingsUtils.js
```diff
@@ -3,6 +3,13 @@ const utils = require('@applitools/utils')
 const snippets = require('@applitools/snippets')
 const ImageMatchSettings = require('../config/ImageMatchSettings')
 
+function stringifySelector(element) {
+  if (element) {
+    //console.log('HERE', element)
+    return element.commonSelector && element.commonSelector.selector
```
kyryloonufriiev, line 9: So you decide that we want just to specify the selector string, and don't need to say anything about the type of the selector? And what about an index of the element under the same selector?

packages/eyes-sdk-core/lib/sdk/CheckSettingsUtils.js
```diff
@@ -219,17 +230,17 @@ async function toScreenshotCheckSettings({checkSettings, context, screenshot}) {
...
+            const scaledRegion = utils.geometry.scale(
+              {
+                x: Math.max(0, region.x - screenshot.region.x),
+                y: Math.max(0, region.y - screenshot.region.y),
+                width: region.width,
+                height: region.height,
+              },
+              context.driver.viewportScale,
             )
+            scaledRegion.regionId = selector ? selector : stringifySelector(element)
```
kyryloonufriiev, line 242: I see you saved `selector` from the region in the beginning and then use it to understand if you have to stringify the selector from an element. I would better always use a selector from an element.
```suggestion
            scaledRegion.regionId = element.selector && stringifySelector(element)
```

packages/spec-driver-selenium/src/spec-driver.ts
```diff
@@ -63,6 +63,16 @@ export function transformSelector(selector: Selector | CommonSelector): Selector
   }
   return selector
 }
+export function untransformSelector(selector: Selector): CommonSelector {
```
kyryloonufriiev, line 66: I think you also need to add a passthrough logic for selectors that are already common.
```suggestion
export function untransformSelector(selector: Selector | CommonSelector): CommonSelector {
```

kyryloonufriiev (changes requested):
LGTM, my comments are mostly in the types and **un**transformation area. Only one is really important about the index of the elements, so please address it first.

packages/eyes-sdk-core/lib/sdk/CheckSettingsUtils.js
tourdedave: Agreed, this was a hold-out from the initial implementation. Although I prefer a tell-don't-ask style of code. I would assume that stringifySelector (or whatever it calls) would work if element.selector exists otherwise return undefined. I'll commit a fix.

packages/spec-driver-selenium/src/spec-driver.ts
tourdedave: Okay

packages/eyes-sdk-core/lib/sdk/CheckSettingsUtils.js
tourdedave: I don't understand your first question. Can you elaborate?

As for the indexing, it was discussed that this wouldn't be part of this initial release. So I didn't design for it. There were some open questions that we didn't address, IIRC.

tourdedave:
@kyryloonufriiev A few things:

1.  In spec-driver-selenium, the test for `untransformSelector` passes. But when trying to build the package I get the following errors and I'm not sure how to resolve them:

```
src/spec-driver.ts:71:16 - error TS7053: Element implicitly has an 'any' type because expression of type 'string' can't be used to index type 'typeof By'.
  No index signature with a parameter of type 'string' was found on type 'typeof By'.

71     const by = Selenium.By[how](what)
                  ~~~~~~~~~~~~~~~~

src/spec-driver.ts:76:3 - error TS2322: Type '(Function & { __applitoolsBrand?: never; }) | (RelativeBy & { __applitoolsBrand?: never; }) | ({ className: string; } & { __applitoolsBrand?: never; }) | ({ css: string; } & { ...; }) | ... 7 more ... | { ...; }' is not assignable to type 'CommonSelector'.
  Type 'Function & { __applitoolsBrand?: never; }' is not assignable to type 'CommonSelector'.

76   return selector
     ~~~~~~~~~~~~~~~


Found 2 errors in the same file, starting at: src/spec-driver.ts:71
```

2. You didn't respond to [my earlier comment](https://github.com/applitools/eyes.sdk.javascript1/pull/972#discussion_r899403824) regarding issues with testing the behavior across packages. It's still unresolved for me.

3. There's still the matter of getting the integration test for coded regions updated (so the vg tests get the same region coordinates for elements, and to get the `check region with padding` tests working).

I'm off on Monday (US holiday) so I'm assuming you will take this the rest of the way. Otherwise, I can pick it back up on Tuesday.

packages/eyes-sdk-core/lib/sdk/CheckSettingsUtils.js
kyryloonufriiev: Actually, it is my bad. What I recommended is not doing what I was imagining. 
it should be like this 
```
element.selector ? stringifySelector(element) : undefined
```

packages/eyes-sdk-core/lib/sdk/CheckSettingsUtils.js
kyryloonufriiev: I don't usually work with everything linking. I'm trying to link only packages that are actually needed for my work, and reset everything after.
So to make it shows up in the `driver` you just have to lint `types` to the `driver` package (types should be built of course).

packages/eyes-sdk-core/lib/sdk/CheckSettingsUtils.js
kyryloonufriiev: We have two pieces of information about a selector: type and selector string. I was asking if we really want to exclude type from the stringified value. I guess I would like to have it, it will reduce ambiguity in some edge cases.

pending-changes.yaml
```diff
@@ -10,20 +10,21 @@ Apollo SDK:
       An example of how to inline `markdown code highlights`
 "@applitools/eyes-universal":
   feature:
-    - 
+    - Add support for dynamic coded regions
```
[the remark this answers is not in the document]
benbabayoff8: @amitzur looks good to me

packages/eyes-sdk-core/test/e2e/codedRegions.e2e.spec.js
```diff
@@ -1,5 +1,6 @@
 const setupTests = require('./utils/core-e2e-utils')
 const {getTestInfo} = require('@applitools/test-utils')
+const {By, locateWith} = require('../../node_modules/@applitools/spec-driver-selenium/node_modules/selenium-webdriver') // because RelativeBy
```
kyryloonufriiev, line 3: Why? This will not work in gh actions

amitzur: Because there are 2 instances of `selenium-webdriver` (core and spec-driver-selenium), and they use `instanceof` :(

https://github.com/SeleniumHQ/selenium/blob/6d2a7829f9b00f76fb5baf04d9c31135d81ac1ce/javascript/node/selenium-webdriver/lib/webdriver.js#L989

Do you have a suggestion?
Why wouldn't this work in GH actions? There should be this file tree also there.

amitzur:
@tourdedave @kyryloonufriiev oh, another thing I didn't do, and was actually not communicated to anyone outside of the team, is to include the locator strategy in the region ID. Something like `xpath=/html/body`, or `xpath:/html/body`, or `linkPartialText:This is a link.`

I'm wondering if to communicate this to product.

packages/eyes-sdk-core/test/e2e/codedRegions.e2e.spec.js
amitzur: @kyryloonufriiev WDYT?

amitzur:
@tourdedave @kyryloonufriiev In my last commit I added the expected output. The tests will fail, of course, but that's the spec.

packages/eyes-sdk-core/test/e2e/codedRegions.e2e.spec.js
tourdedave: For context, this is the error you get when running locally and just specifying `const {By, locateWith} = require('selenium-webdriver')`

```
  1) coded regions
       region id works - classic:
     TypeError: Invalid locator
      at Object.check [as checkedLocator] (/Users/me/Documents/_dev/applitools/eyes.sdk.javascript1/packages/spec-driver-selenium/node_modules/selenium-webdriver/lib/by.js:421:9)
      at Object.findElements (/Users/me/Documents/_dev/applitools/eyes.sdk.javascript1/packages/spec-driver-selenium/node_modules/selenium-webdriver/lib/webdriver.js:1050:20)
      at Object.findElements (/Users/me/Documents/_dev/applitools/eyes.sdk.javascript1/packages/spec-driver-selenium/dist/spec-driver.js:166:17)
      at Context.elements (/Users/me/Documents/_dev/applitools/eyes.sdk.javascript1/packages/driver/dist/context.js:275:47)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
      at async referencesToRegions (lib/sdk/CheckSettingsUtils.js:220:26)
      at async Object.toScreenshotCheckSettings (lib/sdk/CheckSettingsUtils.js:198:20)
      at async Object.afterScreenshot (lib/sdk/EyesClassic.js:173:33)
      at async takeScreenshot (node_modules/@applitools/screenshoter/src/take-screenshot.js:86:7)
      at async Object.getScreenshot (lib/sdk/EyesClassic.js:168:24)
      at async Object._getAppOutputWithScreenshot (lib/sdk/EyesBase.js:1734:24)
      at async MatchWindowTask.match (lib/MatchWindowTask.js:52:54)
      at async Object.checkWindowBase (lib/sdk/EyesBase.js:1130:25)
      at async Object._check (lib/sdk/EyesClassic.js:107:20)
      at async Object.check (lib/new/check.js:20:20)
      at async Context.<anonymous> (test/e2e/codedRegions.e2e.spec.js:57:7)
```

tourdedave: We're definitely linked. That's the only way to test it locally.

kyryloonufriiev: Ah. Of course, because you added the spec-driver method...

kyryloonufriiev: I think the way to go here would be passing the selenium module in `spec.build(...)` instead of requiring it in there.

packages/eyes-sdk-core/test/e2e/codedRegions.e2e.spec.js
```diff
@@ -0,0 +1,145 @@
...
+      const regions = [
+        '.region.one:nth-child(1)',
+        {selector: '.region.one:nth-child(2)'},
+        {selector: '//div[@class="region one"][3]', type: 'xpath'},
+        By.css('.region.one:nth-child(4)'),
+        By.xpath('//div[@class="region one"][5]'),
+        By.js('return document.querySelector(".two:nth-child(1)")'),
+        // Commenting this locator out for the moment because (oddly) it is not making its way into the resulting output
+        // TODO: fix/re-enable
+        // By.linkText('two'),
```
kyryloonufriiev, line 51: This didn't work because there was no such test on the page. However, it doesn't work even when I changed it to text that actually presents on the page, probably because it is looking for links only - `<a>`.

packages/eyes-sdk-core/test/e2e/codedRegions.e2e.spec.js
```diff
@@ -0,0 +1,145 @@
...
+        el,
+        locateWith(By.css('.two')).toRightOf(el),
+        '.region.three:nth-child(3n)', // 4 regions are targeted by this selector
```
kyryloonufriiev, line 54: also what about shadow dom selectors shouldn't we be able to stringify then too?
I mean 
```
{selector: By.css('.shadow-host'), shadow: By.css('.shadow-child')}
```
should generate something like 
```
css:.shadow-host >>> css:.shadow-child
```

tourdedave:
> Remaining work:
> 
> 1. Support the index and sorting functionality also for UFG.
> 2. Provide core API for passing `regionId` from the outside (it should take precedence over the selector. It should be provided similarly to how `padding` is provided). Makes sense to generalize the type of the RegionReference wrapper there for more than just padding.
> 3. Stabilize integration tests.

I added support for item 1, and addressed most of item 3. Still need to address item 2. There's also adding Shadow DOM selectors (if we want to include it for this iteration).

re: item 3. The padding test still needs some attention, and I was unable to get the region coordinates in ufg execution to be consistent with classic mode (so the assertions are conditionalized by execution mode - not my favorite, but workable for now).

packages/eyes-sdk-core/test/e2e/codedRegions.e2e.spec.js
amitzur: Of course, it's **linkText** :) I put only one link in the page...

packages/eyes-sdk-core/test/e2e/codedRegions.e2e.spec.js
amitzur: Let's keep it as RFE for now? What would happen if this is passed without supporting it? only the host will show up?
cc @benbabayoff8

kyryloonufriiev: Yes.

packages/eyes-api/src/input/CheckSettings.ts
```diff
@@ -31,9 +31,10 @@ type AccessibilityRegionReference<TElement, TSelector> = {
   type?: AccessibilityRegionType
 }
 
-type PaddedRegionReference<TElement, TSelector> = {
+type CodedRegionReference<TElement, TSelector> = {
   region: RegionReference<TElement, TSelector>
   padding?: number | types.OffsetRect
+  regionId?: string
```
kyryloonufriiev, line 37: Shouldn't it be also in floating and accessibility regions?

kyryloonufriiev (changes requested):
LGTM. But we also should add `regionId` to core types in `types` package.

packages/eyes-api/src/input/CheckSettings.ts
tourdedave: Yes. But that begs the question, is padding supported in floating and accessibility regions?

kyryloonufriiev: I asked this question. And the answer was explicit - no.

tourdedave: Okay

tourdedave:
@kyryloonufriiev Okay, I updated with your feedback. All good?

kyryloonufriiev (approved):
Just one suggestion from me. This is not required, I just think it would be nice to formally bind those types together.

packages/eyes-api/src/input/CheckSettings.ts
```diff
@@ -24,16 +24,19 @@ type FloatingRegionReference<TElement, TSelector> = {
...
   maxLeftOffset?: number
   maxRightOffset?: number
+  regionId?: string
 }
 
 type AccessibilityRegionReference<TElement, TSelector> = {
   region: RegionReference<TElement, TSelector>
   type?: AccessibilityRegionType
+  regionId?: string
 }
```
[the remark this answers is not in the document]
tourdedave: I like it for its simplicity, but worry that it's not as readable/not as easy to understand at a glance

tourdedave:
@kyryloonufriiev don't we also need to add `untransformSelector` to `eyes-universal/src/spec-driver/webdriver.ts`?

tourdedave:
> Makes sense. If we want it to work for something else than js selenium sdk

I re-read the initial customer requirement. It's for the Java SDK. So I'll add it.

tourdedave:
> > Makes sense. If we want it to work for something else than js selenium sdk
> 
> I re-read the initial customer requirement. It's for the Java SDK. So I'll add it.

And we'll want to convert the e2e core test to a generic coverage test as well, now that I think about it.

kyryloonufriiev:
I have fixed some problems with implementation. It is ready to merge

PR #979, opened 2022-06-21
Support cypress 10

SophieDePaula:
This PR now runs all the generic tests with `Cypress` `10`.

TODO:
- [x] Add support for `setup` script for `Cypress` `10`  
- [x] Update the manual setup instructions in `README`
- [x] `globalHooks` is now configured in `cypress.config.json` under `setupNodeEvents`. We should make sure to call the users hooks. For example [see this](https://docs.cypress.io/api/plugins/after-run-api#Syntax)
    **note**: Since in `Cypress 10` the global hooks are defined in `setupNodeEvents` and so is the method that execute our internal setup logic, I added a flag to make sure we call the setup logic in `cypress.config.js` only once.
- [x] Add tests for new setup functionality


pending-changes.yaml
```diff
@@ -118,7 +118,7 @@ Apollo SDK:
     - Prevented navbar from appearing on Android 12 screenshots
 "@applitools/eyes-cypress":
   feature:
-    - 
+    - Add support for cypress 10
```
SophieDePaula, line 121: @benbabayoff8

packages/eyes-cypress/README.md
```diff
@@ -67,7 +67,58 @@ Add this file to your project with either:
...
+```
+
+**In setupNodeEvents:**
+
+```js
+if(!eyesSetup) {
+    eyesSetup = true
+     require('@applitools/eyes-cypress')(module)
+     return module.exports(on, config) 
+  }
```
SophieDePaula, line 90: I was thinking that we can add a tutorial example for cypress 10 and link the `cypress.config.js` file here so the user will have a complete example of how the file should look like.

chaim-aharonson-applitools: 'Cypress 10 using UFG' example based on the existing 'Cypress UFG' tutorial is attached.
[Cypress_10_tutorial.zip](https://github.com/applitools/eyes.sdk.javascript1/files/8986683/Cypress_10_tutorial.zip)
### Issues I ran into while adding this test:
1. `yarn generate:tests` many tests compilations returned error, I know it was the same before adding support to Cypress 10 but it should be an opportunity to fix, I saw commented out functionality and I guess it was for a reason... so adding the irrelevant tests to ignore should solve this (?) also, generated tests created as '.spec.js' and not in their new format '.cy.js' these tests are passing and .ext is configurable but still, it might be better to align with Cypress setup instructions.
2. Could not `yarn cypress:new` and `yarn cypress:run:new` got conflicts saying `cypress.json` and `cypress.config.js` are at the same directory, I changed the paths in package.json scripts  
3. `yarn cypress:run` (not new) - fails with `Could not find a Cypress configuration file.` (although the presented path is correct) also, `yarn cypress` open IDE of Cypress 10 although configured to use old version
4. `packages/eyes-cypress/src/plugin/pluginExport.js` and `packages/eyes-cypress/bin/eyes-setup.js` appear as erroneous files, as far as I remember: 'type' in 'package.json' should be added in order to be able to use this syntax, I wondered how it worked in your end with no errors (@SophieDePaula) due to that I did not commit any change or added 'type' to ensure this is not a local issue over my machine
5. after yarn-linking my local 'eyes-cypress' on the new tutorial folder and running `npx eyes-setup` I received this error: `sh: /Users/....../tutorial-cypress-2/node_modules/.bin/eyes-setup: Permission denied`, which made me wonder whether the new example I created was really working with Cypress 10 and the right version of eyes-cypress ?

SophieDePaula: @chaim-aharonson-applitools 
1) I created generic tests in `custom` folder for the failed generated tests. As for the naming, it's not a big deal as I can set the path of the folder that contains the tests in `cypress.config.js`
2) I fixed `cypress:new` and `cypress:run:new`, but now `yarn cypress` is hanging for some reason when using cypress version `6.5.0` (does not happen if I try cypress 8). This only happens in this particular test as some of the `e2e` tests run against cypress `6.5.0` without errors.
4) I updated the logic there to be compatible with the lint rules

I need more time to address your other points.

SophieDePaula: @chaim-aharonson-applitools here is another update
I was able to link my local version of `eyes-cypress` to your tutorial and did not get `access denied` when running the setup script. The script worked as expected for me.
Also, I upgraded the local version of cypress to 8 when running `yarn cypress`. For some reason this particular test hangs with 6.5.0, however I verified that our e2e tests that are running with version `6.5.0` work as expected.
Do you have any other comments?

chaim-aharonson-applitools: thanks for the update, I will retry it and update in case there are issues

SophieDePaula:
@kyryloonufriiev this PR is ready for review. 
Fyi the Cypress github action here is failing over `Unable to open X display`. I saw this error when I tried to use the `test` workflow a while ago, so I don't believe this is anything to do with the changes I made in this PR. Also, the `it` tests pass locally.

js/packages/eyes-cypress/.eslintrc.json
```diff
@@ -7,7 +7,16 @@
...
     "*.d.ts",
-    "*.ts"
+    "*.ts",
+    "test/coverage/generic/cypress.config.js",
+    "/test/unit/setup/fixtures/pluginCypress10.js",
+    "/test/unit/setup/fixtures/cypressConfig-legacy-file/cypress.config.js",
+    "test/play/support.js",
+    "test/unit/setup/fixtures/supportFile-handle-commands.js",
+    "test/it/setup/fixtures/cypress.config.js",
+    "test/play/cypress"
```
kyryloonufriiev, line 17: Why do you need to use both `.eslintignore` and this? Isn't it two ways to do the same thing?

kyryloonufriiev (approved):

js/packages/eyes-cypress/src/setup/handlePlugin.js
```diff
@@ -20,4 +23,25 @@ function handlePlugin(cwd) {
...
+function handlerPluginCypress10(cwd) {
+  const configContent = fs.readFileSync(path.resolve(cwd, 'cypress.config.js'), 'utf-8');
+  const legacyPluginsFilePath = getLegacyPuginFilePath(cwd, configContent);
+  let content;
+  let pathToContent;
+
+  if (legacyPluginsFilePath) {
+    content = readFileSync(legacyPluginsFilePath, 'utf-8');
+    pathToContent = legacyPluginsFilePath;
+  } else {
```
kyryloonufriiev, line 35: I didn't get why would you care about the legacy file in cypress 10

js/packages/eyes-cypress/.eslintrc.json
SophieDePaula: No, `eslintrc` is for the eslint rules, even if a file is not tracked by git , I can still get a failure for `lint` rules when I try to push a commit.

js/packages/eyes-cypress/src/setup/handlePlugin.js
kyryloonufriiev: But you don't care how they specify their `setupNodeEvents` with function in-place, or by exporting it from a different file. At the end you just need to handle this function in the `cypress.config.js`

js/packages/eyes-cypress/.eslintrc.json
kyryloonufriiev: I wasn't talking about `.gitignore`, but about `.eslintignore` you made changes in this file

js/packages/eyes-cypress/src/setup/handlePlugin.js
SophieDePaula: But if they use a plugin file then they might have their global hooks there

kyryloonufriiev: If it is Cypress 10 the only way to configure global hooks would be through the function in config file. Am I right?
If so it is doesn't matter where they declare their function, because it will always end up in config file, and this is where you will patch it with your code. They could declare function in plugins file or blablabla file, it doesn't matter, but what matters is where this function will be used.

SophieDePaula: yeah, you are right, fixed

js/packages/eyes-cypress/.eslintrc.json
SophieDePaula: fixed

PR #987, opened 2022-06-22
Update emit.js

chaim-aharonson-applitools:
following CR, since there currently no consumers for `config.visualgrid` its the best time to fix it  (GCTs which support/consume the parameter in PR)


packages/sdk-coverage-tests/src/generate/emit.js
```diff
@@ -26,7 +26,7 @@ function emitTest(test, {makeSpecEmitter, makeFile}) {
     throw new Error(`Missing implementation for test ${test.name}`)
   }
   test.config.baselineName = test.config.baselineName || test.key
-  test.config.visualgrid = test.vg || false
+  test.vg && (test.config.vg = true)
```
kyryloonufriiev, line 29: On a second thought about that, I think this is not right to add custom stuff in the config object.

kyryloonufriiev (changes requested):
```
test.test.call(utils, {...sdk, config: test.config, env: test.env, meta: test.meta})
```
I think you just need to pass everything from the test object to the test function.

packages/sdk-coverage-tests/src/generate/emit.js
```diff
@@ -45,7 +44,7 @@ function emitTest(test, {makeSpecEmitter, makeFile}) {
   const sdk = makeSpecEmitter(emitter, test)
   test.output = output
   if (test.page) sdk.driver.visit(test.page)
-  test.test.call(utils, {...sdk, config: test.config, env: test.env, meta: test.meta})
+  test.test.call(utils, {...test, ...sdk})
```
kyryloonufriiev, line 47: @chaim-aharonson-applitools What do you think? this is what I propose.

chaim-aharonson-applitools: great, it works perfectly

PR #989, opened 2022-06-27
Import Python SDK

ikseek:
It's the code snapshot from from eyes.sdk.python repo, rev #145895c65771952d512d416c54a0e3094d7655fa
All code is placed in python/ directory as https://trello.com/c/0CPMi4sz suggests. It still has it's own coverate-tests suite, up to be replaced.
The only files I had to place outside python directory are:
* Github Workflows configuration is placed in _py-tests.yaml_ and _py-publish.yaml_ files. Currently tests are being run automatically on pull requests. Both tests and publishing runs can be dispatched manually.
* _.pre-commit-config.yaml_ in the root directory and _pre-commit.haml_ Github Workflow. It's a configuration for nicely managed set of automatic code gardening tools (formatters, sorters, convention checkers, static analyzers) that we actively use.

Besides the files we will also need to **configure secrets** and **Github Pages publishing** configurations.
I didn't switch to bongo tool for changelog management, I guess it should be the next step, under the supervision of someone's who knows how it works.


kyryloonufriiev:
@ikseek I implement a change we were talking about, regarding `APPLITOOLS_API_KEY` vs. `APPLITOOLS_API_KEY_SDK`.  I think if everything will work for you this PR could be merged, and then we can continue work on integrating our pipelines.

PR #993, opened 2022-06-28
Fix/egc crash

tourdedave:
re: [Trello](https://trello.com/c/egP42c2n)


kyryloonufriiev (approved):

kyryloonufriiev:
@tourdedave Do you have any idea why the test fails?

tourdedave:
@benbabayoff8 we don't have a changelog for the egc. Should we add one? Or should we just put feature/bug-fix entries for the egc as part of the universal-server?

kyryloonufriiev:
@tourdedave I'm merging it since the issue is fixed, we shouldn't wait for the chrome fix

PR #994, opened 2022-06-29
Single aborted test fails all tests

chaim-aharonson-applitools:


kyryloonufriiev:
@chaim-aharonson-applitools I see tests are failing after your changes:) Please fix those tests, and I don't know if we need to add new tests for this change.

chaim-aharonson-applitools:
> @chaim-aharonson-applitools I see tests are failing after your changes:) Please fix those tests, and I don't know if we need to add new tests for this change.

- This PR is a 'draft' now
- Fixed, now tests are passing
-By saying 'add new tests for this change', do you mean GCT or package tests? asking since GCT is already in a PR and I think it is an important change, considering the upcoming change that Or's team about to deliver and related to same card.

benbabayoff8:
@chaim-aharonson-applitools Maybe something like:" Fixed bug where a failure in a single UFG environment fails all other environments in the **same** configuration"

What do you think?

tourdedave:
> considering the upcoming change that Or's team about to deliver and related to same card.

Can you link to the relevant Trello ticket(s) in your PR description?

chaim-aharonson-applitools:
> Can you link to the relevant Trello ticket(s) in your PR description?

https://trello.com/c/zHV8qGQN/44-single-aborted-test-fails-all-tests-for-the-same-eyesopen-method

chaim-aharonson-applitools:
Added support for wrong environment configuration will not fail all test, this is corresponding with server implementation.
- '/job-info' API became '/v2/job-info'
- upon the server API returns an error for one (or more) tests, the test becomes an erroneous test and due to that does not continue the 'normal' flow and does not trigger '/render' or '/status' APIs, at this point it is certain that this test should not perform `abort` since it's session never started and never will. to avoid calling 'abort' we trigger the 'testController' with `setShouldSkipAbort` and performing a check whether 'abort' should be triggered (default) or not. Please note: calling 'abort' does not harm the functionality, this is just an unneeded functionality which result irrelevant errors in console 
- VGC-e2e test uses 2 environments, one of them is wrong, it expects the result will be an Array of a single 'TestResults' element and a single 'Error' element

kyryloonufriiev (approved):

PR #1003, opened 2022-07-05
Cypress remove element from config

SophieDePaula:
[Trello card](https://trello.com/c/5rpqNnRN/1448-island-cypress-universal-ufg-checkregion-region-serialization-issue-when-checking-react-web-app?filter=due:notdue)

When we get an element as a region, we `ref` it and we return an `object` from `eyesCheckMapping.js` that includes the different parts of the `config` we converted. But we also need to remove the original values we got from the user.
In this case we did not remove the original element, which caused a problem later in `socket` when we tried to serialize the `config` object.


pending-changes.yaml
```diff
@@ -69,7 +69,7 @@ Apollo SDK:
   feature:
     - 
   bug-fix:
-    - 
+    - fix region as element serialization issue
```
SophieDePaula, line 72: @benbabayoff8

benbabayoff8: @SophieDePaula I'm not sure I would write it for this specific issue. Maybe something like: "Improved serialization for Target elements". What do you think?

SophieDePaula: @benbabayoff8 I didn't improve anything, I just removed the original value (which is the element) form the config as we don't need it.

benbabayoff8: Ok. So something like "fixed use cases where fail to serialize elements"
I would prefer something more generic.

SophieDePaula: Got it , not prob

kyryloonufriiev (approved):
💯

PR #1004, opened 2022-07-05
Feat/upgrade SIDE Eyes deps

tourdedave:
re: [Trello](https://trello.com/c/ltyy4CmX)


tourdedave:
> Looks like a pretty minor update. What about adding a GitHub action for publishing? We can do at least what we did in the `browser-ext` package, but we can also go the full way and also integrate publishing to the chrome and mozilla stores using actions from github marketplace.

It would be nice to have this be automated, but given the priority of the task I can publish it manually for now and we can sort out improved publishing later. We'll also need to make sure source maps don't get included so the final size of the zip file is below the 4MB limit for the Mozilla store.

There's also the need for us to update the extension to support v3 of manifests. But I'm worried about Selenium IDE - I'm not sure if the Selenium project is planning to put in the work to update the extension to support v3. Especially since the plan was to ditch the browser extension in favor of an Electron app.

PR #1010, opened 2022-07-07
[WIP] Universal e2e tests

amitzur:
This is instead of PR #873 which was on top of the old structure of the project.

There is one big challenge we still need to overcome - have a server process per client process, and have the server (child) process exit when the client (parent) process exits.

Another challenge: `UniversalClient` needs to know what the driver URL is. Right now we set it explicitly in its constructor. But this is dirty. Worst case we'll control it with an environment variable, but there has to be a way for us to know the driver URL from the driver instance.

TODO:
- [ ] server process per client process
- [ ] detect `driverUrl` from driver
- [ ] ability to run e2e test with selenium
- [ ] ability to run e2e test with playwright
- [ ] unit tests for `UniversalClient`
- [ ] emit generic coverage tests for selenium
- [ ] emit generic coverage tests for selenium


kyryloonufriiev:
> detect driverUrl from driver
@amitzur I don't think it is possible. They have it somewhere in closure and not in the instance of the driver, at least it was like this last time I checked it.

amitzur:
@kyryloonufriiev I created a new package for universal client.
The current status and open issues are:

Run the following scenario:
1. Build and link `spec-driver-selenium` to `eyes-universal` and `universal-client`.
2. Build `universal-client`
3. Link `universal-client` to `eyes-universal`
4. Run `test/e2e/selenium.spec.ts` test in `eyes-universal`
5. The test runs fine but the process hangs

Status:
1. **server process per client process**: I use the technique of ref/unref socket during requests. But it still doesn't work and the process hangs.
2. **detect driverUrl from driver**: I added `__serverUrl` on the driver when returned from `spec.build`.
3. **ability to run e2e test with selenium**: Almost. Besides the process hanging, I also had trouble with the `exports` field in package.json of `universal-client`. I wanted to consume it like `import {Eyes} from '@applitools/universal-client/webdriver` but it doesn't work.
4. **unit tests for UniversalClient**: I started something, happy to hear feedback about the direction.

kyryloonufriiev:
@amitzur I have one question about the new universal-client package. Why did you decide to put the framework-specific stuff there?
I think it should be a framework-agnostic package with a really small setup to work with a specific framework. All you need to setup a universal client is basically things that we pass to `Core.makeSDK`, and of course utility functions from spec-driver.

kyryloonufriiev:
@amitzur Ah. So do you create this package just for testing? I am thinking about something that could be used also in production.
I agree that universal-client shouldn’t depend on universal-server, since it is a separate unit on its own.

amitzur:
@kyryloonufriiev yes, I had only testing in mind. I didn't think about it for production, since only Cypress is a candidate for it, and it's such a different creature that I'm not sure we want this coupling.

js/packages/universal-client/tsconfig.json
```diff
@@ -16,5 +16,5 @@
   "ts-node": {
     "transpileOnly": true
   },
-  "exclude" : ["dist", "types", "test"],
```
kyryloonufriiev: I would not create another directory on the root level for something that we already have. Better put stress tests under the `test` directory.

js/packages/universal-client/src/shared/spawn-server.ts
```diff
@@ -1,41 +1,68 @@
...
+  }
+  let server: ChildProcess
+  switch (cliType) {
+    case 'npx':
+      server = spawn(
+        'npx',
+        ['-p', '@applitools/eyes-universal', 'eyes-universal'].concat(eyesUniversalArgs),
+        spawnOptions,
+      )
+      break
```
kyryloonufriiev, line 29: When would we need it? I think we should always run with a specific universal-server executable.

js/packages/universal-client/src/shared/spawn-server.ts
```diff
@@ -1,41 +1,68 @@
-import type {ChildProcess} from 'child_process'
-import {fork} from 'child_process'
+import type {ChildProcess, SpawnOptionsWithStdioTuple, StdioPipe, StdioNull} from 'child_process'
+import {spawn} from 'child_process'
 import {Socket} from './socket'
 import type {Logger} from '@applitools/logger'
+import {Readable} from 'stream'
+type ReadableWithRef = Readable & {unref: () => void}
+type CliType = 'npm_local' | 'npx'
+export {CliType}
```
kyryloonufriiev, line 8: You can `export` type in the same line you declare it.

js/packages/universal-client/src/shared/spawn-server.ts
zamboney: Using it with `npm_local` means, we depend on the file structure. I think that if we have the possibility to run it without depending on the file structure is a good approach (just in case)

js/packages/universal-client/src/shared/spawn-server.ts
zamboney: np

js/packages/universal-client/tsconfig.json
zamboney: i will move it

js/packages/universal-client/src/shared/spawn-server.ts
kyryloonufriiev: It is not running it "without dependency to the file structure". It is installing it from the npm, this is something we don't need at all.

zamboney:
@kyryloonufriiev can help with [that](https://github.com/applitools/eyes.sdk.javascript1/runs/7877931831?check_suite_focus=true#step:5:155) failure?

kyryloonufriiev:
@zamboney you need to add `@applitools/scripts` to dev dependencies

zamboney:
I'm closing this pr and opening a new one with the same code, just not in a new package `universal-client` but inside the `eyes-universal`. This is for simplification of this process. We want, for now, the `universal-client` just to be used as an `e2e` test, but in the future, we will make it a separate package that would be consumed by the outside world

PR #1011, opened 2022-07-08
Universal server idle timeout

SophieDePaula:
[Trello card](https://trello.com/c/1KtpPXqD/1442-infogrid-cypress-universal-server-idle-timeout)


js/packages/eyes-universal/src/universal-server.ts
```diff
@@ -24,6 +24,7 @@ const LOG_DIRNAME = process.env.APPLITOOLS_LOG_DIR ?? path.resolve(os.tmpdir(),
 export type ServerOptions = HandlerOptions & {
   debug?: boolean
   idleTimeout?: number
+  detached?: boolean
```
kyryloonufriiev, line 27: It shouldn't be here because you never use it here. This is clearly an option of `makeServerProcess`.

js/packages/eyes-universal/src/universal-server-process.ts
```diff
@@ -5,7 +5,7 @@ import path from 'path'
 export function makeServerProcess(options: ServerOptions): Promise<{port: number; close: () => void}> {
```
kyryloonufriiev, line 5:
```suggestion
export function makeServerProcess(options: ServerOptions & {detached?: boolean}): Promise<{port: number; close: () => void}> {
```

js/packages/eyes-universal/src/universal-server.ts
```diff
@@ -49,7 +50,11 @@ export async function makeServer({
   console.log(`Logs saved in: ${LOG_DIRNAME}`)
   baseLogger.log('Server is started')
 
-  let idle = setTimeout(() => server.close(), idleTimeout)
+  let idle = setTimeout(() => {
+    if (idleTimeout) {
+      server.close()
+    }
```
kyryloonufriiev, line 56: You don't have to even set the timeout if it is 0

kyryloonufriiev (changes requested):

js/packages/eyes-universal/src/universal-server.ts
SophieDePaula: @kyryloonufriiev I thought this is cleaner than to ask everywhere if timeout > 0 call `clearTimeout`.

kyryloonufriiev: You will have two if statements the same as you have now, but you will not call `setTimeout` when you don't care about it.

js/packages/eyes-universal/src/universal-server.ts
```diff
@@ -49,11 +49,10 @@ export async function makeServer({
...
 
-  let idle = setTimeout(() => {
-    if (idleTimeout) {
-      server.close()
-    }
-  }, idleTimeout)
+  let idle: any
+  if (idleTimeout) {
+    idle = setTimeout(() => server.close(), idleTimeout)
+  }
```
SophieDePaula, line 55: @kyryloonufriiev fyi

kyryloonufriiev (approved):

pending-changes.yaml
```diff
@@ -73,7 +73,7 @@ Apollo SDK:
     - Fixed various issues during taking screenshots in landscape orientation on some native devices
 "@applitools/eyes-cypress":
   feature:
-    - 
+    - Close the universal server when cypress process is finished
```
SophieDePaula, line 76: @benbabayoff8

benbabayoff8: @SophieDePaula The user shouldn't care that we have a "Universal Server". 
I would consider it as a bug fix and write something like: "Internal architecture fixes".
@amitzur @kyryloonufriiev What do you think?

PR #1012, opened 2022-07-11
add core e2e test for DOM slot

chaim-aharonson-applitools:


chaim-aharonson-applitools:
converted to draft, postponed until feature in 'dom-snapshot' will be published

kyryloonufriiev:
Why do we need it? Isn't it just a functionality of dom-snapshot?

chaim-aharonson-applitools:
the relevant feature in 'dom-snapshot' was published so now tests may be added and core be published to expose the feature

chaim-aharonson-applitools:
@kyryloonufriiev agree, I will close this PR and branch, will move the change-log to a new Branch in order to publish 'core' with the new 'dom-snapshot'

kyryloonufriiev:
@chaim-aharonson-applitools Yep. I don't think that we should add a core test for every dom-snapshot feature.

PR #1014, opened 2022-07-12
[Cypress] Support padded element

SophieDePaula:
[ Trello card](https://trello.com/c/PYKqJLbg/42-padded-regions)

TODO:
- [x] check if it would be possible to use `selector` instead of `region` when passing a `selector`
- [x] update README


pending-changes.yaml
```diff
@@ -72,7 +72,7 @@ Apollo SDK:
     - Fixed various issues during taking screenshots in landscape orientation on some native devices
 "@applitools/eyes-cypress":
   feature:
-    - 
+    - Add support for padded coded regions
```
SophieDePaula, line 75: @benbabayoff8

kyryloonufriiev (approved):

PR #1015, opened 2022-07-13
make sure we ref Node elements intsead of reffing each attribute

SophieDePaula:
[Trello card](https://trello.com/c/9N29mxlp/1470-lexisnexis-maximum-call-stack-size-exceeded-cypress)
In `reffer` we rely on the `constructor name` to check if we need to `ref` each `attribute` or the entire `object`.
In this edge case, we get several different `elements` that have an empty `constructor name` and therefore, when we try to ref each value of that element we get `Maximum call stack size exceeded`

TODO:
- [ ] add a test


pending-changes.yaml
```diff
@@ -74,7 +74,8 @@ Apollo SDK:
   feature:
     - 
   bug-fix:
-    - 
+    - |
+    Fix `Maximum call stack size exceeded` error when `Node Elements` does not have a `constructor` `name`
```
SophieDePaula, line 78: @benbabayoff8

pending-changes.yaml
```diff
@@ -75,7 +75,7 @@ Apollo SDK:
     - 
   bug-fix:
     - |
-    Fix `Maximum call stack size exceeded` error when `Node Elements` does not have a `constructor` `name` 
+    Fix `Maximum call stack size exceeded` error in `eyesCheck`
```
SophieDePaula, line 78: @benbabayoff8

amitzur:
@SophieDePaula what are those elements for which the constructor name is empty?

js/packages/eyes-cypress/src/browser/commands.js
```diff
@@ -8,9 +8,7 @@ const {eyesOpenMapValues} = require('./eyesOpenMapping');
 const {eyesCheckMapValues} = require('./eyesCheckMapping');
 const {TestResultsSummary} = require('@applitools/eyes-api');
 const refer = new Refer(value => {
-  if (!value || !value.constructor || !value.constructor.name) return false;
-  const name = value.constructor.name;
-  return name === 'HTMLDocument' || name === 'Window' || value.ownerDocument;
+  if (value && value.nodeType) return true;
```
kyryloonufriiev, line 11: Previously your implementation was returning `true` for `window`, and now it is `false`.

SophieDePaula: Thanks, nice catch

SophieDePaula:
> @SophieDePaula what are those elements for which the constructor name is empty?

@amitzur the `html` element doesn't have a `constructor name`, I suppose they overwrote it somehow.

amitzur:
@SophieDePaula In this case I would provide as much information as possible about the behavior of this page.
And please note that `src/browser/eyesCheckMapping.js` already has a `isHTMLElement` function.

js/packages/eyes-cypress/src/browser/commands.js
```diff
@@ -8,9 +8,14 @@ const {eyesOpenMapValues} = require('./eyesOpenMapping');
...
-  const name = value.constructor.name;
-  return name === 'HTMLDocument' || name === 'Window' || value.ownerDocument;
+  if (
+    value &&
+    (value.nodeType === Node.ELEMENT_NODE ||
+      value.nodeType === Node.DOCUMENT_NODE ||
+      (value.constructor && value.constructor.name === 'Window'))
+  ) {
+    return true;
+  }
```
kyryloonufriiev, line 18: Maybe something like this? Without an `if` statement

```suggestion
  return (
    value &&
    (value.nodeType === Node.ELEMENT_NODE ||
      value.nodeType === Node.DOCUMENT_NODE ||
      (value.constructor && value.constructor.name === 'Window'))
  )
```

SophieDePaula:
> @SophieDePaula In this case I would provide as much information as possible about the behavior of this page. And please note that `src/browser/eyesCheckMapping.js` already has a `isHTMLElement` function.

@amitzur when inspecting the `html` element I see that the constructor name is empty. It might be some polyfill that cause this behavior, I am not really sure this is the user intention but that's the end result.
I looked at the `isHTMLElement` and had to add a check for `DOCUMENT_NODE`

SophieDePaula:
@kyryloonufriiev I added a test, lmk if you have any comments.

js/packages/eyes-cypress/test/it/browser/fixtures/cypress/integration/refer.spec.js
```diff
@@ -0,0 +1,22 @@
...
+  it('ref element', () => {
+    cy.visit('https://www.applitools.com/helloworld');
+    cy.get('body > div > div:nth-child(1)').then(el => {
+      {
+        const obj = {
+          nodeType: 9,
+          entry1: 'some value',
+          entry2: 'some other value',
+        };
+        const res = refer.ref(obj);
```
amitzur, line 15: @SophieDePaula @kyryloonufriiev why are you testing the refer with Cypress? It feels needless complication to me.

SophieDePaula: @amitzur I needed to test it on the browser side due to usage of `Node.nodeType` that was not available otherwise.

PR #1018, opened 2022-07-15
js/universal-server: fix driver connection with self-signed cert

tourdedave:
[Trello](https://trello.com/c/FP30FPU8)


js/packages/eyes-universal/src/spec-driver/webdriver.ts
```diff
@@ -110,6 +112,14 @@ export function transformDriver(driver: Driver | StaticDriver): Driver {
       return originalCallback(request, {...options, rejectUnauthorized: false}, callback)
     }
     options.agent = {http: agent, https: agent}
+  } else {
+    const httpAgent = new http.Agent({
+      //rejectUnauthorized: false,
+    })
```
kyryloonufriiev, line 118: I think there is no need to create and overide HTTP agent. it is just the same as a default one.

js/packages/eyes-universal/test/it/webdriver-with-self-signed-cert.spec.ts
```diff
@@ -0,0 +1,36 @@
+import assert from 'assert'
+import {testServer} from '@applitools/test-server'
+import * as spec from '../../src/spec-driver/webdriver'
+
+describe('webdriver with self-signed cert', () => {
```
kyryloonufriiev, line 5: I would prefer to "rebrand" `webdriver-with-proxy` file and put these tests together in the same since they are testing very similar things.

kyryloonufriiev (approved):
Ah, what a shame that I didn't think about it initially... Looks good to me 👍🏼

js/packages/eyes-universal/src/spec-driver/webdriver.ts
tourdedave: Without it you get a build error

```
src/spec-driver/webdriver.ts:122:5 - error TS2741: Property 'http' is missing in type '{ https: https.Agent; }' but required in type '{ http: Agent; https: Agent; }'.

122     options.agent = {https: httpsAgent}
        ~~~~~~~~~~~~~

  node_modules/@wdio/types/build/Options.d.ts:169:9
    169         http: http.Agent;
                ~~~~
    'http' is declared here.
```

kyryloonufriiev: @tourdedave Then you should use [global agent](https://nodejs.org/api/http.html#httpglobalagent) instead of a custom one

PR #1023, opened 2022-07-19
[Cypress] - fix padded coded regions tests and add regionId support

SophieDePaula:


js/packages/driver/src/element.ts
```diff
@@ -55,7 +55,9 @@ export class Element<TDriver, TContext, TElement, TSelector> {
       throw new TypeError('Element constructor called with argument of unknown type!')
     }
 
-    if (this._selector && this._spec.untransformSelector) {
+    if (specUtils.isSimpleCommonSelector(this._selector)) {
+      this._commonSelector = this._selector.selector
```
SophieDePaula, line 59: @kyryloonufriiev I have a problem here with the type, can you take a look, I am not sure how to fix this.

kyryloonufriiev: You should not do `this._selector.selector`, just `this._selector.selector`, I understand that you are addressing the fact that for some reason all of the selectors in cypress look like `{selector
: {selector: 'blabla'}}`. But you should not have it in the first place.

SophieDePaula: @kyryloonufriiev did you mean that is should just be `this._commonSelector = this._selector`?
That's fine, but still, there is a problem with the `type`

SophieDePaula:
@kyryloonufriiev this is ready for review.

pending-changes.yaml
```diff
@@ -77,6 +77,7 @@ Apollo SDK:
 "@applitools/eyes-cypress":
   feature:
     - Add support for padded coded regions
+    - Add support for regionId
```
SophieDePaula, line 80: @benbabayoff8

benbabayoff8: In other SDKs we wrote:"Add support for dynamic coded regions", so let's be consistent :)

js/packages/eyes-cypress/src/browser/socketCommands.js
```diff
@@ -54,6 +54,11 @@ function socketCommands(socket, refer) {
     return await spec.getCookies();
   });
 
+  socket.command('Driver.untransformSelector', async selector => {
+    const result = await spec.untransformSelector(selector);
+    return result;
+  });
```
kyryloonufriiev, line 60: Do we need it?

kyryloonufriiev (approved):
Looks good to me. I  just think you forget to remove something. please see my comment.

pending-changes.yaml
SophieDePaula: Yes sir!

js/packages/eyes-cypress/src/browser/socketCommands.js
SophieDePaula: No, I forgot to remove it. Thanks

PR #1024, opened 2022-07-19
support adding 'userTestId' to eyes.open config

chaim-aharonson-applitools:


chaim-aharonson-applitools:
change-log already approved by Ben

kyryloonufriiev:
@chaim-aharonson-applitools What about classic eyes? We should have it there as well.

chaim-aharonson-applitools:
> @chaim-aharonson-applitools What about classic eyes? We should have it there as well.

Ben said we don't have to support it there

kyryloonufriiev:
I think we should we have to be consistent. Please add this feature to classic as well.

chaim-aharonson-applitools:
> I think we should we have to be consistent. Please add this feature to classic as well.

done. also fixed the 'types' package.json

js/packages/types/package.json
```diff
@@ -41,13 +41,16 @@
...
+    "test:sanity": "yarn build",
     "deps": "bongo deps",
     "lint": "eslint \"**/*.ts\"",
     "preversion": "bongo preversion && yarn build",
     "version": "bongo version",
     "postversion": "bongo postversion --skip-release-notification"
   },
-  "dependencies": {},
+  "dependencies": {
+    "@applitools/scripts": "1.1.0"
```
kyryloonufriiev, line 52: Please put it in devDeps

kyryloonufriiev (changes requested):
I don't see two important tests, userTestId being added in the TestResults object. And `userTestId` being added in TestResultsContainer object.

js/packages/eyes-sdk-core/lib/sdk/EyesClassic.js
```diff
@@ -65,6 +65,10 @@ class EyesClassic extends EyesCore {
       this._configuration.setViewportSize(utils.geometry.round(utils.geometry.scale(vs, this._driver.viewportScale)))
     }
 
+    if (!this._configuration.getUserTestId()) {
```
skhalymon, line 68: If a single Eyes instance would be used for multiple `eyes.open/close` on the first step, we set `usertTestId` and on the second call of `eyes.open` it will use previous  `usertTestId` value.

js/packages/types/package.json
chaim-aharonson-applitools: done

kyryloonufriiev (approved):
@chaim-aharonson-applitools Looks good to me:) Sorry for the confusion around this feature.

PR #1040, opened 2022-07-21
Fix/take screenshot puppeteer

tourdedave:
[Trello](https://trello.com/c/SiYAh0Bh)


pending-changes.yaml
```diff
@@ -61,6 +61,7 @@ Apollo SDK:
   bug-fix:
     - Fixed bug where a failure in a single UFG environment fails all other environments in the same configuration
     - Better support in DOM slot element
+    - Fixed issue that prevented screenshots from being taken
```
tourdedave, line 64: @benbabayoff8 ☝️

benbabayoff8: @tourdedave Looks good

tourdedave:
I tested this fix against the coverage tests for eyes-puppeteer and they passed

https://github.com/applitools/eyes.sdk.javascript1/runs/7449133781?check_suite_focus=true

kyryloonufriiev:
@tourdedave I find a way to create a test that will work for both browsers. I think it could be merged

PR #1041, opened 2022-07-21
Py rfe add user test id

skhalymon:
It's adding identifier `user_test_id` to `Configuration`. It's more internal feature which allow tracking which test is related to what TestResults in the `get_all_test_results` step.

By default, in USDK, the `user_test_id` is set with UUID if no value provided. As it happens inside USDK config, python don't have access to it. We've decided to do not generate another UUID inside open step and leave it as is for now. 

Generally speaking, we have a quorum with @ikseek that `userTestId` shouldn't be generated automatically on JS side as well. It basically has no sense for user, as they will need wrappers to keep `userTestId` values anyway. While keeping this value with `null` will show that no value was settled.

USDK implementation is here https://github.com/applitools/eyes.sdk.javascript1/pull/1024


ikseek:
Lgtm

ikseek (approved):

kyryloonufriiev:
@skhalymon I don't understand the problem with auto-generation. If you never pass this value, then you probably don't care and can ignore it. We have our use for this value, this feature is not only about your use case. It is just nice that you wanted something that we were talking about for a while.

skhalymon:
@kyryloonufriiev this what I wanted to understand is it required by your side or not. If so, have no objections. Thanks!

ikseek:
@kyryloonufriiev the problem with auto generation is that it's result is saved to configuration. Previously eyes only did read configuration, never used it to store any state except test and app names which are explicitly provided as a part of open call. In python SDK configuration might be shared among multiple eyes instances or reused with reused eyes instance. So eyes open might see testId set by prior open call by another eyes instance or the same eyes instance opened twice. This way multiple tests will get the same auto-generated test id.
If you need this id for other purposes, can you please also explain them? Will user break anything if he passes same userTestId in multiple eyes?

kyryloonufriiev:
@ikseek @skhalymon I have a feeling we are not talking about the same thing. First of all, this feature is not user-facing, so the user should not pass this value ever. Then I still don't understand how auto-generation on our side could possibly affect your configuration object. Or you don't like that USDK "made" you store something in the configuration object, then it is simply not true. Since you always can implement it differently, basically it is your implementation detail you can do whatever you want.

ikseek:
@kyryloonufriiev yes, there seems to be misunderstanding. What we asked for is a feature that we needed ourself and that might be used by other customers who write their testing frameworks around eyes. An opaque piece of data that can travel from eyes open back to result. What we put in, that we get out, if we put None, we get back Non. In our case where we are going to utilize it, we will never put None there, but we couldn't understand why you needed to return random ID there. And now there are additional details that you use it for something else internally, so we would like to know what limitations that your use might put on this field?

kyryloonufriiev:
So the main difference is that we didn't design it for users and you was meant to create a public API from it.
We use it for logging, there no limitation from our side, you can pass what ever you want there.
I think what you are saying is makes sense. If we would consider this feature as user-facing we would not generate it on our side. But i don't think Ben or anyone from product is aware of the fact that we wanna have another user-facing API. Do you know that someone actually needs it, I mean from a real customers.

PR #1043, opened 2022-07-22
[py-robot] Rfe 2898 post poscessing of robot test resulsts

skhalymon:
This PR introduces propagation of Eyes test results to RobotFramework `report.html` and `log.html` results. 

RobotFramework don't have direct way to access report from the place where we collect all tests results. But it is possible to edit report after with `prerebotmodifier`. So, the decision was to make an Intermediate store of test data. This data then is feeding through `prerebotmodifier` and then appears inside `report.html` and `log.html`.

`prerebotmodifier` usually used as argument of CLI, but thanks to @ikseek, we've found another way to turn it on. So, no need to add extra CLI arguments, but it is possible to use it in this way as well. By default, the feature is turned on through patching, but it could be turned off if required.

To turn off feature, need to specify `propagate_eyes_test_results` option to **false** inside `applitools.yaml` (it was updated to version 1.2).

Also, it's possible to use it with `robot` CLI by adding `--prerebotmodifier EyesLibrary.PropagateEyesTestResults` option.

The result looks like:
image
Report page

image
Log page


python/eyes_robotframework/src/EyesLibrary/keywords/session.py
```diff
@@ -17,13 +17,8 @@ class RunnerKeywords(LibraryComponent):
     @keyword("Eyes Get All Test Results")
     def get_all_tests_results(self):
         # type: () -> TestResultsSummary
-        try:
-            test_results = self.eyes_runner.get_all_test_results()
-            self.log_to_console(test_results)
+        with self.eyes_runner_get_all_test_results() as test_results:
```
skhalymon, line 20: Add `test_results_manager` here

python/eyes_robotframework/src/EyesLibrary/test_results_manager.py
```diff
@@ -0,0 +1,201 @@
...
+    )
+
+
+class EyesToRobotTestResultsManager(object):
+    """Collect all test results from the eyes library and map them to the robot tests"""
+
+    def __init__(self, configure):
+        # type: (Optional[Configuration]) -> None
+        self.suites = []  # type: list[RobotTestsSuite]
+        self.current_suite = None  # type: Optional[RobotTestsSuite]
```
skhalymon, line 149: We could remove `self.current_suite` and `self.current_test`

python/eyes_robotframework/src/EyesLibrary/test_results_manager.py
```diff
@@ -0,0 +1,201 @@
...
+        return "EyesToRobotTestResultsManager:<{}>".format(
+            list([s.suite_name for s in self.suites])
+        )
+
+    def register_robot_suite_started(self, data, result):
+        # type: (TestSuite,TestSuite) -> None
+        self.current_suite = RobotTestsSuite(suite_name=data.name)
+        self.suites.append(self.current_suite)
+
+    def register_robot_suite_ended(self, data, result):
```
skhalymon, line 167: Move logic to `register_robot_suite_started`

python/eyes_robotframework/src/EyesLibrary/keywords/session.py
skhalymon: No needed

ikseek (approved):

PR #1047, opened 2022-07-25
Fix/eyes puppeteer GCT failures

tourdedave:
I had difficulty running the `test packages` GH action w/ linking between eyes-puppeteer and spec-driver-puppeteer. So I tested this with a dirty workaround (e.g., in spec-driver-puppeteer I performed a `yarn build && yarn pack`, moved the binary into eyes-puppeteer and did a `yarn add ./binary`, removed the `deps` script from eyes-puppeteer to avoid an error during the job's setup, and committed the changes). A passing result can be seen [here](https://github.com/applitools/eyes.sdk.javascript1/runs/7500652654?check_suite_focus=true).

Changing `takeScreenshot` from `page.screenshot` to the public CDP API resolves the GCT failures (and also resolves the original issue, which was a broken API call because the private CDP end-point on `page` is no longer available in newer versions of puppeteer).


kyryloonufriiev:
@tourdedave I also submit a PR to fix the same thing. https://github.com/applitools/eyes.sdk.javascript1/pull/1046. I think my solution is a little better because it uses pptr api and not cdp

PR #1048, opened 2022-07-25
[universal] Remove settings update

kyryloonufriiev:


tourdedave (changes requested):
@kyryloonufriiev This came up during release testing for the NMG. It was a necessary fix to overcome cases where the SDK would fail to locate the necessary elements to perform VHS extraction. I was asked to make it so the server handled it implicitly (rather than have the customer specify the capability to set it). To remove it, we would require testing from @Yarden-Ingber's team to verify we don't introduce any regressions.

So the action item here, is to have a broader discussion.

js/packages/eyes-universal/package.json
```diff
@@ -50,13 +50,14 @@
...
     "build:zip": "zip -j ./bin/eyes-universal.zip ./bin/eyes-universal-alpine ./bin/eyes-universal-linux ./bin/eyes-universal-macos ./bin/eyes-universal-win.exe",
     "build:tgz": "tar -czf ./bin/eyes-universal.tar.gz -C ./bin ./eyes-universal-alpine ./eyes-universal-linux ./eyes-universal-macos ./eyes-universal-win.exe",
-    "docker:setup": "node ../scripts/scripts/generate-docker-compose-config.js && docker-compose up -d",
-    "docker:teardown": "docker-compose down",
     "test": "yarn test:unit && yarn test:it && yarn test:e2e",
     "test:sanity": "yarn test:unit && yarn test:it",
     "test:unit": "mocha ./test/unit/*.spec.ts --exit --no-timeouts -r ts-node/register",
     "test:it": "mocha ./test/it/*.spec.ts --no-timeouts -r ts-node/register",
     "test:e2e": "mocha ./test/e2e/*.spec.ts --exit --no-timeouts -r ts-node/register",
+    "setup": "yarn docker:setup",
```
tourdedave, line 58: IIRC, having `setup` like this causes issues in the build/test pipeline

kyryloonufriiev: Without it doesn't work at all. I had added the mocha hook to the relevant test to wait for docker to start.

tourdedave:
We'll also need to document (somewhere, for the field at least) what the capability is for the user to set this since it's an obscure appium specific capability. It's `'settings[allowInvisibleElements]': true`

kyryloonufriiev:
@tourdedave Customer also can update settings in runtime as we do it - https://appium.io/docs/en/advanced-concepts/settings/

tourdedave:
One other thought I had was to present a helpful error to users when they reach a situation where they need to change this setting (e.g., when running with NMG on Android and the SDK is unable to find the necessary element). The error message should contain details about how to set the setting (or a link to some sort of documentation about it). What do you think?

kyryloonufriiev:
The thing is that we don't know if they don't have our library at all, or if they have it but we are just not able to find it due to this issue. 
Of course, we could include both, I think it would be great and very helpful.

tourdedave:
Per [Trello](https://trello.com/c/9L9JC09L/1495-biogen-appium-universal-allowinvisibleelements-breaks-xpath#comment-62fa2de20e0f0067571d5f53), merging/releasing.

PR #1052, opened 2022-07-27
add robotframework sdk to the list of supported sdks

Artem-applitools:


kyryloonufriiev (approved):
LGTM

pending-changes.yaml
```diff
@@ -1,3 +1,8 @@
+"@applitools/sdk-coverage-tests":
+  feature:
+    - Support reporting for the robotframework sdk
+  bug-fix:
+    -
```
kyryloonufriiev, line 5: @Artem-applitools Please remove it. There is no need for a changelog for something else than SDK. This is an internal tooling package so we don't maintain changelog for them

kyryloonufriiev (changes requested):

pending-changes.yaml
Artem-applitools: @kyryloonufriiev Removed

Artem-applitools:
@kyryloonufriiev Pls merge the PR when you will be publishing new package version

PR #1055, opened 2022-07-28
[js-testcafe] allow class destructuring with backwards compatibility

tourdedave:
[Trello](https://trello.com/c/HMuSVzo3)

Tested in GH actions ([link](https://github.com/applitools/eyes.sdk.javascript1/runs/7562248592?check_suite_focus=true)) and locally by building/packing and installing it against both [the tutorial](https://github.com/applitools/tutorial-testcafe) (JS) and the example provided by the field in the Trello ticket (TS).


kyryloonufriiev:
`index.ts` generates `exports.default = api_1.Eyes`
`index-legacy.ts` generates `module.exports = index_1.Eyes;`

So with `index.js` you will have to do `const Eyes = require('@applitools/eyes-testcafe').default`

tourdedave:
> 

Ah, I see. 

This will work: `import Eyes from '@applitools/eyes-testcafe'`
But this will not `const Eyes = require('@applitools/eyes-testcafe')`

So it is a breaking change. Okay.

tourdedave:
Closing this PR since we'll address this as part of a larger effort to unify the API (that and a workaround was found to expose the class needed for the support ticket).

PR #1057, opened 2022-08-02
add hover tests

zamboney:
Add test to cover `hover` for `web`

## on the way add more stuff:
- Add ignored files to `.gitignore`
- Add explain on where the images save once tests fails


.gitignore
```diff
@@ -1,4 +1,6 @@
 .vscode
 .idea
 .history
-.DS_Store
\ No newline at end of file
+.DS_Store
+node_modules
```
kyryloonufriiev, line 5: I'm not sure I understand why we need it here.

js/packages/screenshoter/test/e2e/e2e.js
```diff
@@ -42,7 +42,11 @@ exports.test = async function test({type, tag, driver, ...options} = {}) {
     const actual = await screenshot.image.toObject()
 
     const expected = await makeImage(`./test/fixtures/${type}/${tag}.png`).toObject()
-    assert.strictEqual(pixelmatch(actual.data, expected.data, null, expected.width, expected.height), 0)
+    assert.strictEqual(
+      pixelmatch(actual.data, expected.data, null, expected.width, expected.height),
```
kyryloonufriiev, line 46: I love the message. But sometimes it is important to know the reason why it failed. There might be different image sizes or pixel differences. Can you include this information?

js/packages/screenshoter/test/e2e/web/hover-element.spec.js
```diff
@@ -0,0 +1,27 @@
...
+  before(async () => {
+    ;[driver, destroyDriver] = await makeDriver({type: 'chrome', logger})
+    await driver.visit('https://applitools.github.io/demo/TestPages/StickyHeaderWithRegions')
+  })
+
+  after(async () => {
+    await destroyDriver()
+  })
+
+  it('take element in frame screenshot with hover element', async () => {
```
kyryloonufriiev, line 15: I would rather call it
```suggestion
  it('take hovered element screenshot', async () => {
```

js/packages/screenshoter/test/e2e/web/hover-element.spec.js
```diff
@@ -0,0 +1,27 @@
...
+    const element = await driver.element('#input')
+    await driver.target.$(element.target).moveTo()
+
+    await test({
+      type: 'web',
+      tag: 'hover-element',
```
kyryloonufriiev, line 21: The convention here is 
```
`${what's on the screenshot (page | frame | element)}-${fully or not}-${what's special about it}`
```

```suggestion
      tag: 'element-hovered',
```

kyryloonufriiev: The same for file name

kyryloonufriiev (changes requested):

.gitignore
zamboney: `.DB_Store` i will remove but `node_modules` should not be added to the repo

js/packages/screenshoter/test/e2e/e2e.js
zamboney: np

.gitignore
zamboney: the `.DB_Store` was [before](https://github.com/applitools/eyes.sdk.javascript1/blob/f07a743710fda3590eb6d0014e41855586a07b8e/.gitignore#L4) this pr

kyryloonufriiev: I was talking about `node_modules`. It wasn't there and it wasn't a problem, so I don't see why we should add it.

zamboney: I don't understand why the `node_moudles` are ignore right now but this will make sure it will ignore in the future.

And, unless you want to keep the `node_moudles`, it doesn't do anything.

kyryloonufriiev: We need to understand how it will affect us first, before doing it. Especially in cases like this, when there are no real problem solved. I actually did the check why we don't have node_modules included in the repo, and why we didn't have node_modules ignored here. To clarify it for your self you have to look at /js/.gitignore file.

tourdedave: Put another way - this repo is a monorepo of multiple programming languages (right now just Python and JS, but soon to also include Java, C#, and Ruby). The .gitignore in the `js` directory already handles `node_modules`. The top-level .gitignore for the repo does not need this since all languages do not need this.

zamboney: ok, i will remove it

kyryloonufriiev:
Looks good to merge. Make sure you use the "Squash and merge" button

kyryloonufriiev (approved):

PR #1059, opened 2022-08-03
Add `typesVersions` and `exports` to the `package.json`

zamboney:
Now to `import` the new `SDK` (a.k.a not `index-legacy`) with all of the types inside it, all needed to do is `import` from  `@applitools/eyes-testcafe/next`

For example, `BatchClose`:

```
import { BatchClose } from "@applitools/eyes-testcafe/next";
```

> import to note: in order for `packge.json` `exports` to work need to add in `compilerOptions` on `tsconfig.json` `    "moduleResolution": "Node16",` [link](https://stackoverflow.com/questions/72193784/why-doesnt-the-exports-field-of-npm-work-in-typescript)


kyryloonufriiev:
@zamboney It comes to me that if the customer uses node < 16 they will not be able to run `import {BatchClose} from '@applitools/eyes-testcafe/next'`, even if types are recognized. Because node < 16 will not respect the `exports` configuration, so it will think that `./next` is an actual file or dir in the package. 
In order to help customers with node < 16, we should help typescript understand where the types live when they will import from `@applitools/eyes-testcafe/dist/index`. This could be done with `typesVersions` i believe.

kyryloonufriiev:
@zamboney in order to solve the `Verify changelog` check, you should add a changelog entry in `pending-changes.yml` file at the root of the repo. Obviously, you should add it to `eyes-testcafe` only, I think you will get it.

zamboney:
> @zamboney It comes to me that if the customer uses node < 16 they will not be able to run `import {BatchClose} from '@applitools/eyes-testcafe/next'`, even if types are recognized. Because node < 16 will not respect the `exports` configuration, so it will think that `./next` is an actual file or dir in the package. In order to help customers with node < 16, we should help typescript understand where the types live when they will import from `@applitools/eyes-testcafe/dist/index`. This could be done with `typesVersions` i believe.

yes of course that's why I added both `typesVersions` and `exports`.

kyryloonufriiev:
You added `typesVersions` for `next`, and node < 16 doesn't know what `@applitools/eyes-testcafe/next` is, it thinks that there should be a file/dir with name `next`. But there are not. This is why customers who use node < 16 should use our directory structure instead of this fancy `next` sub-module.

kyryloonufriiev (changes requested):

zamboney:
@kyryloonufriiev I tested it on `node@14`, and it works. I understand what you are saying about the file structure and how it's worked for customers before this change, but i don't understand why they should use it after this change if it works fine with `typesVersions` (as long as they use `ts@<3.1` [link](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#version-selection-with-typesversions)). of course they can keep work with the file/dir as before.

kyryloonufriiev:
@zamboney I have no doubt that TS works, but I don't think that JS will actually work. Because node < 16 will throw when you try to import/require `eyes-testcafe/next`.

zamboney:
@kyryloonufriiev good point, but what is the difference from what it was before? It will just work for `node@16` and above

But for sure this solution does not cover `node@14`

If you want to cover `node@14` I can add a `next` folder that `expose` `dist/index.js`.

This will make the whole solution as a setup for the upcoming deprecation of `node@14` in April 23

Does it sound good?

kyryloonufriiev:
I don't want a directory next in the file system. This is just a work around for the user. If they have node 16 the work around will be nicer, if not then you should deal with a file system.

zamboney:
So you can approve this PR 🧑‍🎤?

kyryloonufriiev:
I cannot approve. We need to support `@applitools/eyes-testcafe/dist` for customers with node < 16

zamboney:
But isn't the `dist` folder already created by `tsc`?

zamboney:
@kyryloonufriiev I think i figure it out

check the video below

[asciicast](https://asciinema.org/a/tsLlN3RMCb28hXE6mM34DrByM)

It looks like everything is working as expected!

I will add the `dist` export after you approve that this is what you had in mind.

js/packages/eyes-testcafe/package.json
```diff
@@ -30,6 +30,27 @@
...
+      "default": "./dist/index-legacy.js"
+    },
+    "./dist": {
+      "types": "./types/index.d.ts",
+      "default": "./dist/index.js"
+    },
+    "./next": {
+      "types": "./types/index.d.ts",
+      "default": "./dist/index.js"
+    }
```
kyryloonufriiev, line 52: What about all of this but in a `typesVesions`? because it still doesn't solve the problem for node < 16?

js/packages/eyes-testcafe/package.json
```diff
@@ -30,6 +30,27 @@
...
+  },
+  "exports" : {
+    ".": {
+      "types": "./types/index-legacy.d.ts",
+      "default": "./dist/index-legacy.js"
+    },
+    "./dist": {
+      "types": "./types/index.d.ts",
+      "default": "./dist/index.js"
+    },
```
kyryloonufriiev, line 48: Also, I would allow importing any file from `dist`.
```suggestion
    "./dist/*": {
        "types": "./types/*.d.ts",
        "default": "./dist/*.js"
    },
```

kyryloonufriiev (changes requested):

js/packages/eyes-testcafe/package.json
zamboney: isn't `typeVersions` for `typesctipt` only use?

kyryloonufriiev: No, how it node resolves modules and how typescript resolves types are two completely different topics. You can try it out, `import {Eyes} from '@applitools/eyes-testcafe/dist'` will not resolve types in node < 16, without tsconfig's `"moduleResolution": "node16"`.

zamboney: Ok good to know! i will add it to the `typeVersions`

kyryloonufriiev (approved):

PR #1060, opened 2022-08-04
`eyes-storybook`: Improved `start-storybook` command path resolution algorithm

giladgd:
# Changes
Improved `start-storybook` command path resolution algorithm, using an implementation of npm's resolution algorithm.
Fixes a problem where `eyes-storybook` crashed on load when npm workspaces is used.

Given the following file tree:
```txt
root/
  |-- package.json
  |-- node_modules/
    |-- .bin/
      |-- start-storybook
  |-- subProject/
    |-- package.json
    |-- node_modules/
      |-- .bin/
```

When running `eyes-storybook` inside of `root/subProject`:
* The current implementation will check if `root/subProject/node_modules/.bin/start-storybook` exists and will then fail if it doesn't
* The new implementation will search the file tree for a `./node_modules/.bin/start-storybook` file like npm's own implementation, and in this example it will find and use `root/node_modules/.bin/start-storybook`


kyryloonufriiev:
@giladgd What about using `npx` instead?

giladgd:
> @giladgd What about using `npx` instead?

Running `npx --no start-storybook` gets stuck in case `start-storybook` is not found in the file tree (from my own testings), and we cannot predict which node module is going to supply the `start-storybook` command, as it can come from `@storybook/react` and can also come from `@storybook/vue` for example.
I originally wanted to use npm's implementation of this, but it's not easy to do, so I implemented it myself.

amitzur:
@giladgd and I spoke yesterday about this. I'd like us to have rigorous tests for `startStorybookServer` function. It seems to be not the most testable code right now. So I think we should make it testable, run tests with a **mock** `start-storybook` bin file, and run this test on both Linux and Window in the CI.
Since this isn't trivial or immediate, and I wouldn't like @giladgd to spend a lot of time on it, plus he has a workaround, then we can put this on hold for the moment.
I think it would be a good task for @zamboney .

@kyryloonufriiev FYI, happy to hear your thoughts.

giladgd:
@kyryloonufriiev Can we merge this PR soon?

NoamGaash:
@kyryloonufriiev 
any updates?

amitzur:
@NoamGaash
This PR requires adding tests to an area in the code that is not currently very testable. I think it's a 2 day effort. I believe It's worth the effort, but it needs to be prioritized, and since our frontend already has a workaround, I suggest we find the proper timing to do this. We can keep this PR open until we feel it's the right time to address it (a customer complains about the same area in the code, OR it becomes more painful for the frontend team).

NoamGaash:
Cool
I just wanted to make sure it wasn't already fixed :+1:

NoamGaash:
@kyrylodolynskyi 
why did we delete this branch?
Its needed.

kyrylodolynskyi:
It was unintentional, i just cleaned up old branches.

giladgd:
@kyrylodolynskyi Can we then merge it soon? We're atill waiting for this

amitzur:
@NoamGaash it was deleted unintentionally as part of a script, sorry

amitzur:
@giladgd we'll provide it shortly. AFAIK you're not blocked, you just want to remove the workaround. Is that correct?

js/packages/eyes-storybook/src/initPage.js
```diff
@@ -38,7 +38,9 @@ function makeInitPage({iframeUrl, config, browser, logger, getTransitiongIntoIE,
         logger.log(
           `Puppeteer page closed [page ${pageId}] while still in page pool, creating a new one instead`,
         );
-        await pagePool.removeAndAddPage(pageId);
+        if (!pagePool.getIsBrowserClose()) {
```
SophieDePaula, line 41: On the latest major version 19 of Puppeteer, when we try to create a new page after closing the browser, we get an exception that the target was closed

js/packages/eyes-storybook/test/e2e/happy-config/applitools.config.js
```diff
@@ -13,5 +13,5 @@ module.exports = {
     }
   },
   ignoreRegions: [{selector: '.global-ignore-this'}],
-  // puppeteerOptions: {headless: false, devtools: true},
+  puppeteerOptions: {headless: 'new'},
```
SophieDePaula, line 16: I needed to add this to all the tests config so that we won't get the warning in the snapshot about the new chrome headless.

giladgd:
> @giladgd we'll provide it shortly. AFAIK you're not blocked, you just want to remove the workaround. Is that correct?

That's correct.
I want to fix this here since this issue can also happen to customers, and there's no way for a customer to know how to work around this issue as there was no indicative log that explains this issue

js/packages/eyes-storybook/test/e2e/eyes-storybook.e2e.test.js
```diff
@@ -100,7 +103,7 @@ describe('eyes-storybook', () => {
   it('fails with proper message when failing to get stories because of navigation timeout', async () => {
     const promise = presult(
       utils.process.sh(
-        `node ./bin/eyes-storybook --read-stories-timeout=10 -u http://localhost:9001`,
+        `node ./bin/eyes-storybook --read-stories-timeout=5 -u http://localhost:9001`,
```
SophieDePaula, line 106: from what I see in the test runs, this test is flaky and randomly we will get `CONNECTION_REFUSED` as the page loads in less than 10 mili seconds and will throw this error.

SophieDePaula:
@amitzur this is ready for review

amitzur:
@SophieDePaula @kyrylodolynskyi I see changes in `pagePool` related to the upgrade of puppeteer. Can we please avoid dealing with 2 problems, and revert the puppeteer upgrade?

kyrylodolynskyi:
@amitzur Yes, I suppose we can.

js/packages/eyes-storybook/test/e2e/snapshots/eyes-storybook.e2e.test.json
```diff
@@ -106,8 +106,16 @@
     ""
   ],
   "navigation timeout stderr": [
+    "\u001b[1m\u001b[43m\u001b[30m",
```
amitzur, line 109: @kyrylodolynskyi but you added `process.env.PUPPETEER_DISABLE_HEADLESS_WARNING = true`. Why is this still showing up in the snapshot?

PR #1061, opened 2022-08-04
Pull makeImage into its own package (from screenshoter)

tourdedave:
- [x] pull it into its own package
- [x] convert it to TS (e.g., no build errors and tests passing)
- [x] specify concrete types (not just `any`)
- [x] update screenshoter to use the new package
- [x] publish new package and update screenshoter package.json to use it (once code review is complete)


js/packages/image/src/index.ts
```diff
@@ -664,4 +693,4 @@ function _scaleImageIncrementally(src, dst) {
   return dst
 }
 
-module.exports = makeImage
+export = makeImage
```
kyryloonufriiev, line 696: It would be better to use named export.

js/packages/image/src/index.ts
```diff
@@ -1,13 +1,42 @@
...
+
+type TransformData = {
+  rotate: number,
+  scale: number,
+  crop?: any,
+  modifiers: any[],
+  added?: any,
+}
+
+type Image = {
```
kyryloonufriiev, line 16: I would specify it as an `interface`, not `type`. There is no big difference, but I usually use `type` to specify the shape of a data object, and `interfaces` to specify a functional object as we have here.

kyryloonufriiev (changes requested):
Looks good to me. One small thing that is maybe more about esthetic is I wouldn't put the whole implementation in the index file, but rather create an `image.ts` file with the implementation of the image logic. And the same for test `index.d.ts` is not informative enough as for me.

js/packages/image/src/image.ts
```diff
@@ -1,13 +1,42 @@
...
+
+type TransformData = {
+  rotate: number
+  scale: number
+  crop?: any
+  modifiers: any[]
+  added?: any
+}
+
+interface Image {
```
kyryloonufriiev, line 16: This should be exported, so we can use the time in other packages
```suggestion
export interface Image {
```

kyryloonufriiev (changes requested):
Looks good to me, the only missing point is a dedicated github action to publish this pacakge.

tourdedave:
> Looks good to me, there are a few missing points:
> 
> * Dedicated github action to publish `image` package.
> * Add `@applitools/scripts` to the `screenshoter` as a dev dependency.
> * Add `test:sanity` command to the `image`, it could run all of the tests, since they are fast.

`@applitools/scripts` is already a dev dep in `screenshoter`. The other items are done.

js/packages/image/package.json
```diff
@@ -0,0 +1,72 @@
...
+    "directory": "js/packages/image"
+  },
+  "license": "SEE LICENSE IN LICENSE",
+  "author": {
+    "name": "Applitools Team",
+    "email": "team@applitools.com"
+  },
+  "aliases": [
+    "screenshoter"
+  ],
```
kyryloonufriiev, line 21: Aha. I see what confused me. We use the first alias as the name of the job, so it says that `screenshoter` doesn't have `scripts` dependency when the `image` is the one that doesn't have it. 
```suggestion
  "aliases": [
    "image"
  ],
```

kyryloonufriiev (changes requested):
If after merging my proposed change all tests are passing in the PR, feel free to merge.

kyryloonufriiev (approved):
Great, let's merge and release it!

PR #1062, opened 2022-08-04
when log file is created by another process, don't throw an error

SophieDePaula:
[Trello card](https://trello.com/c/m9SMe586/1516-universal-sdk-does-not-respond-to-coremakemanager-command)


js/packages/logger/src/handler-rolling-file.ts
```diff
@@ -65,6 +65,15 @@ function ensureDirectoryExistence(filename: string) {
...
-    fs.mkdirSync(dirname)
+    try {
+      fs.mkdirSync(dirname)
+    } catch (ex) {
+      if (ex.message.includes('file already exists')) {
+        // another process already created this file
+        // do nothing
+      } else {
+        throw ex
+      }
```
kyryloonufriiev, line 76: I believe an empty `if` block is a bad practice 
```suggestion
      // if a file already exists ignore the error
      if (!ex.message.includes('file already exists')) {
        throw ex
      }
```

kyryloonufriiev (approved):

amitzur:
@SophieDePaula @kyryloonufriiev this seems pretty fragile. Is there no explicit API for this?
Also, the comment is confusing IMO since the *directory* already exists, not the log file.

SophieDePaula:
@amitzur there is an `api` and we are using it `if (!fs.existsSync(dirname))`, the problem is that after we pass this check, another process already created the dir so we get the exception.
Re the changelog, now I see that what I wrote is indeed misleading, can I simple update this entry in `universal` changelog in `master`?

PR #1065, opened 2022-08-08
NML integration

tourdedave:
[Trello](https://trello.com/c/VG7yc6Vy)


js/packages/nml-client/package.json
```diff
@@ -0,0 +1,78 @@
...
+    "test:sanity": "yarn test",
+    "test:watch": "yarn test --watch --extension ts",
+    "deps": "bongo deps",
+    "preversion": "bongo preversion && yarn build",
+    "version": "bongo version",
+    "postversion": "bongo postversion --skip-release-notification"
+  },
+  "dependencies": {
+    "@applitools/req": "1.0.1",
+    "uuid": "^8.3.2"
```
kyryloonufriiev, line 49: You can use our `@applitools/utils` package, as we do everywhere else

js/packages/test-utils/src/parse-env.js
```diff
@@ -67,6 +67,25 @@ const DEVICES = {
...
+        args: [],
+        env: {
+          DYLD_INSERT_LIBRARIES:
+            '@executable_path/Frameworks/UFG_lib.xcframework/ios-arm64_x86_64-simulator/UFG_lib.framework/UFG_lib',
+          NML_API_KEY: process.env.APPLITOOLS_API_KEY_SDK,
+        },
+      },
+      ...SAUCE_CREDENTIALS,
+    },
+  },
```
kyryloonufriiev, line 88: I think the better way to go here would be to use regular iPhone 12 env. You can add `processArguments` in the options of the `parseEnv` function; this way, you can add it to any environment.

js/packages/nml-client/src/broker.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+  return response
+}
+
+export async function pollForMessageResult(url: string): Promise<any> {
+  url = /-response$/.test(url) ? url : url + '-response'
+  return req(url, {
+    retry: [{statuses: [404]}],
+  })
+}
```
kyryloonufriiev, line 22: You can encapsulate polling logic using `hooks` property and `makeReq` constructor function. This is definitely not something required, but as a see, it will create a nicer function that will handle end-to-end communication through the broker.

js/packages/nml-client/package.json
```diff
@@ -0,0 +1,78 @@
...
+    "./package.json": "./package.json"
+  },
+  "scripts": {
+    "lint": "eslint '**/*.ts'",
+    "build": "tsc",
+    "test": "mocha --no-timeouts -r ts-node/register 'test/**/*.spec.ts'",
+    "test:sanity": "yarn test",
+    "test:watch": "yarn test --watch --extension ts",
+    "deps": "bongo deps",
+    "preversion": "bongo preversion && yarn build",
```
kyryloonufriiev, line 43: No need to build in here. We are building it separately from versioning during release.

kyryloonufriiev (changes requested):

js/packages/nml-client/package.json
tourdedave: Use it for what?

js/packages/test-utils/src/parse-env.js
tourdedave: Oh, I see that there is an option in parseEnv that handles injection for NMG. I'll do something with that.

https://github.com/applitools/eyes.sdk.javascript1/blob/feat/nml-integration/js/packages/test-utils/src/parse-env.js#L660-L668

js/packages/nml-client/package.json
tourdedave: `uuid` isn’t part of `@applitools/utils` - should I add support for it there and then use it through `@applitools/utils`?

Or did you mean something else?

tourdedave: `utils.general.guid()` in `@applitools/utils`

tourdedave:
core tests fail because bongo doesn't support `file:../..` paths for packages in its fixed dependency check

nml-client tests fail because of a change in test-utils that needs to be published (works when linked)

kyryloonufriiev (approved):
Looks good to me. But the test in nml-client is failing right here in PR. Is it fine? Feel free to merge it if you will be able to publish

PR #1069, opened 2022-08-13
move toCheckSettings logic to screenshoter and add tests

SophieDePaula:
TODO:
- [x]  Run tests for all modified packages and make necessary adjustments that don't pass
- [x] Add tests around `regionId` and `padding` in `getBaseCheckSettings`

All tests pass except [this assertion](https://github.com/applitools/eyes.sdk.javascript1/blob/master/js/packages/eyes-sdk-core/test/it/codedRegions.spec.js#L194). The reason it is no longer passing is that the `padding` logic was separated from the coordinate extraction part and for some reason we relay on the fact here that the reference to this element changed after we call `check`.
I am not entirely sure why we perform this assertion.


kyryloonufriiev:
@SophieDePaula I think you can merge your PR directly in master. This way we can already start taking advantage of this.

SophieDePaula:
> @SophieDePaula I think you can merge your PR directly in master. This way we can already start taking advantage of this.

@kyryloonufriiev I branched out of Dave's NML branch, so I also have his changes here

js/packages/eyes-sdk-core/lib/sdk/CheckSettingsUtils.js
```diff
@@ -196,60 +196,49 @@ function toMatchSettings({checkSettings = {}, configuration}) {
...
-                y: Math.max(0, region.y - screenshot.region.y),
-                width: region.width,
-                height: region.height,
-              },
-              context.driver.viewportScale,
-            )
-            referenceRegions.push(scaledRegion)
-          }
+function toBaseCheckSettings(settings) {
+  const regionTypes = ['ignore', 'layout', 'strict', 'content', 'floating', 'accessibility']
```
SophieDePaula, line 200: @kyryloonufriiev notice that I had to make a few changes to this function.

SophieDePaula:
@kyryloonufriiev this PR is ready for review.

js/packages/utils/src/general.ts
```diff
@@ -83,3 +83,12 @@ export function pluralize(object: [] | number, config?: [manyCase: string, singl
...
 }
+
+export function stringifySelector(element: any): string {
+  if (element && element.commonSelector) {
+    if (typeof element.commonSelector === 'string') {
+      return element.commonSelector
+    }
+    return element.commonSelector.selector
+  }
+}
```
kyryloonufriiev, line 94: this function shouldn't be in the utils package containing general utilities, and this is an applitools specific one. You could feel it when you put `any` in argument type. Because you don't have access to `types`.

js/packages/screenshoter/src/take-screenshot.js
```diff
@@ -110,6 +115,34 @@ async function takeScreenshot({
     // restore focus on original active context
     await activeContext.focus()
   }
+  async function extractCoodinatesForCodedRegions(regionsToCalculate, screenshot) {
+    const codedRegionsByCoordinates = []
+    for (const codedRegion of regionsToCalculate) {
+      const {region, ...options} = codedRegion.region ? codedRegion : {region: codedRegion}
```
kyryloonufriiev, line 121: In this line, I see that you exposed the check settings sub-structure to `screenshoter`. The idea was to avoid it, but passing just elements/selectors and not the whole object of coded region with unnecessary metadata.

js/packages/screenshoter/test/e2e/e2e.js
```diff
@@ -56,6 +56,27 @@ exports.test = async function test({type, tag, driver, ...options} = {}) {
...
+    const result = []
+    for (const r of region) {
+      result.push({
+        region: {
+          x: Math.round(r.region.x),
+          y: Math.round(r.region.y),
+          width: Math.round(r.region.width),
+          height: Math.round(r.region.height),
+        },
+        regionId: r.regionId,
```
kyryloonufriiev, line 71: Same here. You shouldn't test coded region structure, since this is not the feature of screenshoter.

kyryloonufriiev (changes requested):

js/packages/utils/src/general.ts
SophieDePaula: @kyryloonufriiev we use this function in both `core` and `screenshoter`, where would be the appropriate place to put it?

js/packages/screenshoter/src/take-screenshot.js
SophieDePaula: @kyryloonufriiev this info about the region such as `regionId` and `padding` is needed in `screenshoter` as there we also add the padding to the coordinates we find and calculate the `regionId`.

SophieDePaula: did you want the logic around `padding` and `regionId` to be separated from extracting the coordinates? Today it's calculated together:

```
        const elementRegion = utils.geometry.withPadding(await element.getRegion(), options.padding)
        const region = utils.geometry.offset(elementRegion, contextLocationInViewport)
        const scaledRegion = utils.geometry.scale(
          {
            x: Math.max(0, region.x - screenshot.region.x),
            y: Math.max(0, region.y - screenshot.region.y),
            width: region.width,
            height: region.height,
          },
          context.driver.viewportScale,
        )
```

js/packages/utils/src/general.ts
kyryloonufriiev: This shouldn't be used in `screenshoter` in the first place

js/packages/screenshoter/src/take-screenshot.js
kyryloonufriiev: @SophieDePaula Exactly, I want this logic to be spreaded. I understand that it might look like an unnecessary complication, but I believe that this is a unification, because 1 - this is something that already happening when we are working with ufg, and 2 this is what going to happen when we will work with nml.

SophieDePaula:
@kyryloonufriiev please take another look at this PR and also the note I added to the description.
Thanks!

js/packages/eyes-sdk-core/lib/sdk/CheckSettingsUtils.js
```diff
@@ -196,60 +196,74 @@ function toMatchSettings({checkSettings = {}, configuration}) {
...
+    regionTypes.flatMap(async regionType => {
+      /* filter:
+        in case we were not able to find a region coordinates we set undefined as a place holder to not lose order,
+        however, we don't want to keep this entry in the returned checkSettings */
+
+      /* flat
+        in case a selector is a match for more than one elmenet in the page */
+
+      transformedSettings[`${regionType}Regions`] = transformedSettings[`${regionType}Regions`].filter(Boolean).flat()
+    })
```
kyryloonufriiev, line 229: 1. Don't use `map` / `flatMap` if you don't care about the outcome.
2. How it could be undefined if you explicitly set there at least `[]`

js/packages/eyes-sdk-core/lib/sdk/CheckSettingsUtils.js
```diff
@@ -196,60 +196,74 @@ function toMatchSettings({checkSettings = {}, configuration}) {
...
+  })
+
+  function getBaseCheckSettings(regionsByCoordinates) {
+    const transformedSettings = {...settings}
+    for (const regionType of regionTypes) {
+      const enrichedRegions = []
+      transformedSettings[`${regionType}Regions`] = transformedSettings[`${regionType}Regions`]
+        ? transformedSettings[`${regionType}Regions`]
+        : []
+      for (const region of transformedSettings[`${regionType}Regions`]) {
```
kyryloonufriiev, line 215: if the point of this is just to have something to use in `for`, then better do this 
```suggestion
      for (const region of transformedSettings[`${regionType}Regions`] || []) {
```

js/packages/screenshoter/src/take-screenshot.js
```diff
@@ -84,11 +86,13 @@ async function takeScreenshot({
...
     screenshot.image.scale(driver.viewportScale)
 
+    const codedRegionsByCoordinates = await extractCoodinatesForCodedRegions(regionsToCalculate, screenshot)
+
     if (hooks && hooks.afterScreenshot) {
       await hooks.afterScreenshot({driver, scroller: target.scroller, screenshot})
     }
 
-    return screenshot
+    return {screenshot, codedRegionsByCoordinates}
```
kyryloonufriiev, line 95: I wouldn't call it "coded regions", because this way you merge eyes terminology and low-level screenshoter terminology. As I see it it is just "region" / "caltulatedRegions". Coded regions have their own features and types, and this is not what you return here.

js/packages/screenshoter/src/take-screenshot.js
```diff
@@ -110,6 +114,35 @@ async function takeScreenshot({
...
+                width: region.width,
+                height: region.height,
+              },
+              context.driver.viewportScale,
+            )
+            scaledRegions.push({region: scaledRegion})
+          }
+          codedRegionsByCoordinates.push({regions: scaledRegions, commonSelector: elements[0].commonSelector})
+        }
+      }
```
kyryloonufriiev, line 142: I think this deserves its own file, and separate integration test

kyryloonufriiev (changes requested):

js/packages/eyes-sdk-core/lib/sdk/CheckSettingsUtils.js
SophieDePaula: @kyryloonufriiev , what do you mean if I don't care about the outcome? I do care, I want all the regions in each type to not have nested arrays.
I actually had a bug here, but now if I can't find an element in `take-screenshot`, I return an array with `{regions:[]}`, and I set `undefined` in `enrichRegion` in `getBaseCheckSettings` as a place holder.

js/packages/screenshoter/src/take-screenshot.js
SophieDePaula: I separated this logic to a new file. I already have `e2e` tests for this logic, so I don't think we need integration tests as well.

js/packages/eyes-sdk-core/lib/sdk/CheckSettingsUtils.js
kyryloonufriiev: Ok, my first question why do you use `flatMap` here? Why not just `map`?
Both of them do not make sense here, because those methods construct some new value out of the array, but you don't use the value it returns, so you shouldn't use this method in the first place. Maybe `forEach` is what you need.

But even if you get this, there is a bigger question why do you need to do the second iteration over all of the region is you could do this during the first itteration.

kyryloonufriiev (changes requested):

PR #1078, opened 2022-08-21
universal e2e tests client

zamboney:
In order to test the communication between the `eyes-universal` and other platform (like `Python` or `Ruby` for example) this `client` will test the communication between `eyes-universal` (a.k.a. server) and a `client` (a.k.a. `nodejs` client).

This change is here to replace [this close](https://github.com/applitools/eyes.sdk.javascript1/pull/1010) pr that try to make this `client` as a standalone package.

Also adding a `stress tests` that use this `client` test.


js/packages/spec-driver-selenium/src/spec-driver.ts
```diff
@@ -3,7 +3,7 @@ import type {Size, Region, Cookie, DriverInfo, WaitOptions, ScreenOrientation} f
 import * as Selenium from 'selenium-webdriver'
 import * as utils from '@applitools/utils'
 
-export type Driver = Selenium.WebDriver & {__applitoolsBrand?: never}
+export type Driver = Selenium.WebDriver & {__serverUrl?: string} & {__applitoolsBrand?: never}
```
kyryloonufriiev, line 6: This should not be here in the type. Because it is user-facing thing

kyryloonufriiev (changes requested):

js/packages/spec-driver-selenium/src/spec-driver.ts
zamboney: ok so where it's should be? without it `tsc` break

kyryloonufriiev: You can skip declaring it because this is already a hack. So no need to pretend it is a real thing.
BTW if you really want to declare it you can declare it in the return value of the build function.

js/packages/spec-driver-selenium/src/spec-driver.ts
```diff
@@ -379,7 +379,10 @@ export async function build({selenium, ...env}: any): Promise<[Driver, () => Pro
       noProxy: proxy.bypass,
     })
   }
-  const driver = await builder.build()
+  const driver: Driver = await builder.build()
+  // this is for the nodejs client
+  // we want to expose the `serverUrl` without decalre it as part of the type
+  Object.defineProperty(driver, '__serverUrl', {get: () => url})
```
zamboney, line 385: @kyryloonufriiev
return the `serverUrl` value in the `build` function, meaning passing this value through the `nodejs` client. I think it's overhead and it makes the code less readable (IMO).
I think using `Object.defineProperty` to bind to the `driver` the `serverUrl` is a hacky solution but not as hacky as changing the `Driver` type.
WTYT.

kyryloonufriiev: I was thinking about something else. instead of return `[Driver, () => void]` - return `[Driver & {__serverUrl: string}, () => void]`. This way you don't change the user-facing `Driver` type and have all of the declarations you need during testing.

zamboney: Np will do on Sunday 😎

js/packages/spec-driver-selenium/src/spec-driver.ts
zamboney: @kyryloonufriiev done please take a look

js/packages/spec-driver-selenium/src/spec-driver.ts
zamboney: @kyryloonufriiev done please take a look, 10x

js/packages/spec-driver-selenium/src/spec-driver.ts
```diff
@@ -379,7 +381,9 @@ export async function build({selenium, ...env}: any): Promise<[Driver, () => Pro
       noProxy: proxy.bypass,
     })
   }
-  const driver = await builder.build()
+  const driver: Driver = await builder.build()
+  ;(driver as DriverWithServerUrl).__serverUrl = url
+  Object.defineProperty(driver, '__serverUrl', {get: () => url})
```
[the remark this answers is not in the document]
zamboney: my bad i will remove it

js/packages/spec-driver-selenium/src/spec-driver.ts
```diff
@@ -332,7 +333,8 @@ const browserOptionsNames: Record<string, string> = {
   chrome: 'goog:chromeOptions',
   firefox: 'moz:firefoxOptions',
 }
-export async function build({selenium, ...env}: any): Promise<[Driver, () => Promise<void>]> {
+
+export async function build({selenium, ...env}: any): Promise<[DriverWithServerUrl | Driver, () => Promise<void>]> {
```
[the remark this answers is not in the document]
zamboney: will do

js/packages/eyes-universal/src/socket.ts
```diff
@@ -16,12 +16,14 @@ export interface Socket {
...
-export function makeSocket(ws: WebSocket, {logger}: {logger?: Logger} = {}): Socket {
+export function makeSocket(ws?: WebSocket, {logger}: {logger?: Logger} = {}): Socket {
   let socket: WebSocket = null
   const listeners = new Map<string, Set<(...args: any[]) => any>>()
   const queue = new Set<() => any>()
 
-  attach(ws)
+  if (ws) {
+    attach(ws)
+  }
```
kyryloonufriiev, line 26: If you do this, that makes sense, you would probably want to remove  `if (!ws) return` from the `attach` function.

js/packages/eyes-universal/test/stress/run.sh
```diff
@@ -0,0 +1,26 @@
...
+
+done <<< "$(seq $1)"
+
+
+# wait for all pids
+for pid in ${pids[*]}; do
+    wait $pid
+done
+
+echo "stress test done"
```
kyryloonufriiev, line 26: It is hard for me to understand what exactly it is doing. Why did you choose to do it with bash and not js/ts?

kyryloonufriiev (approved):
Looks good to me!

PR #1079, opened 2022-08-22
Store the chrome log files in the running dir

zamboney:
Adding an example of how to save the chrome log file using chrome flags


kyryloonufriiev:
Wasn't it about storing our logs, those that we write in the console of the extension?

zamboney:
Yes, this pr is an example for the client to write the logs from the extension to the disk

PR #1081, opened 2022-08-23
Allow viewport size to be greater than the screen size in `UFG`.

zamboney:
Before this change, the test fails if the viewport is set to be greater than the window size both in `classic` and `UFG`.

This fix allows the viewport to be greater than the screen in `UFG` (but less than the `UFG` limitation).


js/packages/eyes-sdk-core/lib/sdk/EyesVisualGrid.js
```diff
@@ -117,9 +117,12 @@ class EyesVisualGrid extends EyesCore {
         testConcurrency: this._runner.testConcurrency,
       })
 
-    if (this._configuration.getViewportSize()) {
-      const vs = this._configuration.getViewportSize()
-      await this.setViewportSize(vs)
+    const vsFromConfiguration = this._configuration.getViewportSize()
+    if (vsFromConfiguration) {
+      const error = await this.setViewportSize(vsFromConfiguration, false)
```
amitzur, line 122: I recommend to just `try/catch` instead of changing the API of `setViewportSize`.

pending-changes.yaml
```diff
@@ -10,7 +10,7 @@ Apollo SDK:
       An example of how to inline `markdown code highlights`
 "@applitools/eyes-universal":
   feature:
-    - 
+    - Allow viewport size greater than the screen size in `UFG`.
```
amitzur, line 13:
```suggestion
    - Don't fail `eyes.open` when there is a failure to set viewport size in `UFG`.
```

amitzur (changes requested):
The core e2e test is :100: 
I think the implementation should be more minimal. I don't see a reason to do any change to `setViewportSize`.

js/packages/eyes-sdk-core/lib/sdk/EyesVisualGrid.js
amitzur: Not sure I understand. I mean doing something like this:

```
if (this._configuration.getViewportSize()) {
  const vs = this._configuration.getViewportSize()
  try {
    await this.setViewportSize(vs)
  } catch(err) {
    this._logger.warn(`Failed to "setViewportSize" to ${vsFromConfiguration}, but still continue with the test`)
  }
}
```

Do you see anything wrong with it?

zamboney: Without my change [this error](https://github.com/applitools/eyes.sdk.javascript1/blob/17d2b6d3aac364636b92493669bd1fc3486130ac/js/packages/eyes-sdk-core/lib/sdk/EyesCore.js#L245) will thrown no matter if it's catch the `error` or not

amitzur: IMO it's more important to keep the API as clean as possible. It makes things easier to reason about. In this case, I would change that place from `logger.error` to `logger.log`. Because you will call `logger.warn` here anyway.

js/packages/eyes-sdk-core/test/e2e/viewportSizeFallback.e2e.spec.js
```diff
@@ -0,0 +1,37 @@
...
+      driver,
+      config: {
+        appName: 'core e2e',
+        testName: 'viewport size greater then screen size',
+        browsersInfo: [
+          {
+            width: UFG_MAX_WIDTH_HEIGHT_SIZE - 1,
+            height: UFG_MAX_WIDTH_HEIGHT_SIZE - 1,
+            name: 'chrome',
+            headless: false,
```
amitzur, line 28: This does nothing, it can be removed.

js/packages/eyes-sdk-core/test/e2e/viewportSizeFallback.e2e.spec.js
```diff
@@ -0,0 +1,37 @@
...
+          },
+        ],
+      },
+    })
+    await driver.get('https://example.org')
+    await eyes.check()
+    await eyes.close({throwErr: true})
+  })
```
amitzur, line 36: I would add the following at the end:

```
const actualViewportSize = await driver.executeScript(getViewportSize)
expect(actualViewportSize.width).to.be.lessThan(UFG_MAX_WIDTH_HEIGHT_SIZE - 1)
expect(actualViewportSize.height).to.be.lessThan(UFG_MAX_WIDTH_HEIGHT_SIZE - 1)
```
It's important to verify that the conditions you assume happen in the test, actually happened

amitzur: This requires adding at the top:
```
const {expect} = require('chai')
const {getViewportSize} = require('@applitools/snippets')
```

amitzur (changes requested):
There's another comment about the test. I was running this in docker and wasn't successful in getting the `setViewportSize` to fail, even though it is NOT headless. Only on my local machine's chrome I was able to verify it.
So a verification of the actual viewport size is very important in the test.

js/packages/eyes-sdk-core/test/e2e/viewportSizeFallback.e2e.spec.js
zamboney: About the `headless` you're right, but about the other properties, I'm not sure. You can see in the [eyes test](https://eyes.applitools.com/app/test-results/00000251740978189603/00000251740978189447/steps/1?accountId=UAujt6tHnEKUivQXIz7G6A~~&mode=step-editor) that the width and height are set as expected...
image

zamboney:
@amitzur, I change all your request on the running code
On the test side, I understand what you saying about the way I tested it, and I moved the tests to `it` tests, and now it's testing the code as expected (you can remove my `try catch` [code](https://github.com/applitools/eyes.sdk.javascript1/blob/34987ecaa28a7d759b5212dcf5bdce430c8c5e71/js/packages/eyes-sdk-core/lib/sdk/EyesVisualGrid.js#L123-L129) and the test will fail.

js/packages/eyes-sdk-core/test/e2e/viewportSizeFallback.e2e.spec.js
zamboney: @amitzur  i remove this test and create a new one in `it`

js/packages/eyes-sdk-core/test/e2e/viewportSizeFallback.e2e.spec.js
zamboney: i see what i miss and remove this test

amitzur: I only referred to the `headless: false`

js/packages/eyes-sdk-core/lib/sdk/EyesVisualGrid.js
```diff
@@ -117,9 +118,15 @@ class EyesVisualGrid extends EyesCore {
...
 
-    if (this._configuration.getViewportSize()) {
-      const vs = this._configuration.getViewportSize()
-      await this.setViewportSize(vs)
+    const vsFromConfiguration = this._configuration.getViewportSize()
+    if (vsFromConfiguration) {
+      try {
+        await this.setViewportSize(vsFromConfiguration)
+      } catch (error) {
+        if (error instanceof EyesError)
```
amitzur, line 126: I don't think this is needed. The contract is that `setViewportSize` throws if it fails. No need to look into **what** it throws.

amitzur (changes requested):
In addition to the inline comment about the `error instanceof EyesError`, I have a comment about the test.
I liked that it was an e2e test. Kyrylo is working on core redesign and I would like this test to stay the same even after his work.
I don't like the integration tests that check `EyesVisualGrid` because that class is not how I would like core to be designed in the first place. This is why we are doing the final redesign now :)
The only comment I have about that e2e test is that the test should also verify the assumption that the viewport size was NOT set to what was asked.

zamboney:
The `e2e` test I did don't really test anything ,you can remove the code I change and everything stay the same.

I think it's because the `selenium` `docker ` image using a VM like env that not really have a screen size, but I'm not sure.

I can keep the `e2e` test just for a `UFG` smoke test but I didn't figure out a way to test my changes using it

js/packages/eyes-sdk-core/lib/sdk/EyesVisualGrid.js
zamboney: This is for safety measure, what happens if there is a `property undefined` error? If I didn't `catch` the type of the `error` it would continue

amitzur: Fair enough

amitzur:
@zamboney the e2e does test what it needs, it's the docker that's "misbehaving".

I would keep the e2e with the assertion that I suggested, in addition to your good integration test.
I would skip the e2e test and then release this. After that we should insist on understanding how to make the test behave well in docker.

zamboney:
@amitzur i added the `e2e` test as a `smoke test`. the changes you suggested do not work. the value of the `width` and `height` remained the same.

kyryloonufriiev:
@zamboney did you try this - https://github.com/SeleniumHQ/docker-selenium#setting-screen-resolution?

zamboney:
As @amitzur suggested, I added the `e2e` test that tests this change but `skip` it. I will continue investigating this issue

zamboney:
> @zamboney did you try this - https://github.com/SeleniumHQ/docker-selenium#setting-screen-resolution?
@kyryloonufriiev  its not working 😢 i will investigate more

pending-changes.yaml
```diff
@@ -10,7 +10,7 @@ Apollo SDK:
       An example of how to inline `markdown code highlights`
 "@applitools/eyes-universal":
   feature:
-    - 
+    - Don't fail `eyes.open` when there is a failure to set viewport size in `UFG`.
```
kyryloonufriiev, line 13: This is relevant for all of the sdks except storybook, not only universal.

kyryloonufriiev (commented):
@zamboney Let's fix the changelog and merge it today.

pending-changes.yaml
zamboney: done

PR #1082, opened 2022-08-24
Add description to universal-server releases

tourdedave:
The GH action we use for creating the universal server release in GitHub is [softprops/action-gh-release](https://github.com/softprops/action-gh-release). It has options to populate the description of a release either by providing a text blurb, a link to a file, or by letting GitHub create the release notes ([link](https://github.com/softprops/action-gh-release#-customizing)).

- text blurb: This is the simplest/quickest way to add a helpful description (and what is done in this PR). It's not the ideal end-state, but an incremental step which can help people outside of our team more easily find info on their own.

- file: we can't use the changelog file as it is today since it contains entries for all releases. But we could easily add the ability for `bongo` to maintain a new file (e.g., `LATEST`) which contains just the changelog entries for the latest release.

- let GitHub create the release notes: this option is not suitable since it creates a list of changes from the commit history for the entire monorepo


kyryloonufriiev:
I think the ideal solution would be to add some command to the bongo, to extract the latest change log text, so we can call it after release and get the string, which we can use in gh release and slack notification. I was thinking about doing it for a long time.
If you think that it is hard to do, then we can easily go with `text blurp` option for now.

tourdedave:
@kyryloonufriiev Getting the latest text blurb from bongo isn't difficult. The functionality already exists, it's just not exposed to the cli. How do you run a command and save/use the result in the GH action?

kyryloonufriiev:
@tourdedave I was thinking about adding this information as an output value for the `publish` action (not workflow). We already have some output values there, like `tag`  - https://github.com/applitools/eyes.sdk.javascript1/blob/master/.github/actions/publish/action.yml#L79
you can read about actions output here - https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#outputs-for-composite-actions

.github/actions/publish/action.yml
```diff
@@ -78,6 +81,10 @@ runs:
           echo "::notice::Publish tag - $TAG"
           echo "::set-output name=tag::$TAG"
         fi
+        CHANGELOG="$(yarn run --silent bongo log --latest-changelog)"
+        if [ ! -z "$CHANGELOG" ]
+        then
+          echo "::set-output name=changelog::$CHANGELOG"
```
kyryloonufriiev, line 87:
```suggestion
        CHANGELOG="$(yarn run --silent bongo log --latest-changelog)"
        if [ ! -z "$CHANGELOG" ]
        then
          echo "::set-output name=changelog::$CHANGELOG"
        fi
```

pending-changes.yaml
```diff
@@ -15,6 +15,7 @@ Apollo SDK:
     - 
 "@applitools/eyes-selenium":
   feature:
+    - Dynamically include changelog entries in a release description and publish notifications to Slack
```
kyryloonufriiev, line 18: why in selenium?

tourdedave: That was a mistake. Fixed.

PR #1086, opened 2022-08-28
Disable `HelperAndroid` for `NMG`

zamboney:
Disable the `NMG` by `HelperAndroid` by disabling it through the `UFG`

this seem like an urgent change so tests TBD


js/packages/driver/src/driver.ts
```diff
@@ -260,9 +260,16 @@ export class Driver<TDriver, TContext, TElement, TSelector> {
...
+      switch (true) {
+        case this.isIOS: {
+          this._helper = await HelperIOS.make({spec: this._spec, driver: this, logger: this._logger})
+          break
+        }
+        case this.isAndroid && this._customConfig?.disableHelper !== true: {
+          this._helper = await HelperAndroid.make({spec: this._spec, driver: this, logger: this._logger})
+          break
+        }
+      }
```
kyryloonufriiev, line 272: No reason to avoid it on android.
```suggestion
     if (!this._customConfig?.disableHelper) {
       this._helper = this.isIOS
          ? await HelperIOS.make({spec: this._spec, driver: this, logger: this._logger})
          : await HelperAndroid.make({spec: this._spec, driver: this, logger: this._logger})
     }
```

zamboney: @kyryloonufriiev np about that it just means that the `HelperIOS` would be `null` and, as far as I understand, the issue is in `HelperAndroid`. If you think this is fine then I will change it

kyryloonufriiev: The issue is not about android only. It is about everything.

PR #1088, opened 2022-08-30
Added support for lazy loading views in android

zamboney:
We create this communication to communicate with `AndoirdHelper` lazyloading `async` operations.

## Running Step By Step

1. The `driver` write the command to the `AndroidHelper.input` (just like before)
2. if the `command` is `async`, it writes to `AndroidHelper.input` `'WAIT'` immediately. 
3. the `driver` waits until the `AndroidHelper.input` `'WAIT'` change
   1. if it is taking more than 5 min, it continues.
 4. clears the `AndroidHelper.input` (just like before)
 5. return the response from `Android` to the `driver` `command`.

[](https://mermaid.live/edit#pako:eNpdkcFuwjAMhl_F8hkkLlyqCYlSQFx2GRKHloPXmBFBkspNNU2l776k6aRqOTm_P_-2kx5rpxgz_BJq7nDOKwvhbMuLaM_g7wy1M4asevuUjXegbdP5KyyXG8hLYVL_GRBuG2dbhps4A0EVp9U1-eZj4a6_bE_nIUm7KL28dPyCovwm7WG9Wpl2qijGimMfE9HeOBnHsvGyNtpONsfR5kbPNvjkcy1ZH-bdJmxfenqkJdNaidlH5lDWTyaZJaN4eo8ILtCwGNIqPFwflQoDZ7jCLISK5FFhZYfAdY0iz3ulvRPMxr4LpM67jx9bYxZn-4MKTeETzEQNvzF_hC0)

## API
we created a new `command` `offset_async` that requests (just like the `offset` `command`) the `elementId` and returns the height of the element plus if the request is `lazy load`. The response is separated by `;`.

## Using `LazyLoadOptions` check
by using [`LazyLoadOptions.waitingTime`](https://github.com/applitools/eyes.sdk.javascript1/blob/02adb3544f58215842ab6a74f802f26c5b0c31b4/js/packages/types/src/data.ts#L327) the `driver` pass to the `AndroidHelper` how much time do we need to wait between each scroll


kyryloonufriiev:
Please give this PR a more self-explanatory name. When we squash and merge it the make of the PR will be the only thing we will see, and it will be hard to guess what is "Deloy before extract value". I would recommend, "Add delay before extraction value from android helpre library"

js/packages/driver/src/helper-android.ts
```diff
@@ -61,20 +81,55 @@ export class HelperAndroid<TDriver, TContext, TElement, TSelector> {
...
+    element: Element<TDriver, TContext, TElement, TSelector>,
+    waitTimeout = 0,
+  ): Promise<types.Region> {
+    const elementId = await this._getElementId(element)
+    if (!elementId) return null
+
+    let contentHeightString = ''
+
+    const offsetCommand: 'offset_async' | 'offset' =
+      semverGte(this._version, '1.8.0') && waitTimeout > 0 ? 'offset_async' : 'offset'
```
kyryloonufriiev, line 118: Instead of parsing the version every time this function is called, I would suggest doing it once when you extracted the version string.

js/packages/driver/src/helper-android.ts
```diff
@@ -61,20 +81,55 @@ export class HelperAndroid<TDriver, TContext, TElement, TSelector> {
...
     } else {
-      const elementId = await this._getElementId(element)
-      if (!elementId) return null
-      contentHeightString = await this._command(`offset;${elementId};0;0;0`)
+      contentHeightString = await this._command(`${offsetCommand};${elementId};0;0;0;0`)
     }
+    // TODO remove it before merging
+    // contentHeightString = await this._command(`offset;${elementId};0;0;0;${waitTimeout}`)
+
+    if (contentHeightString === null) return null
```
kyryloonufriiev, line 132: Any falsy value should lead to returning `null`. But the real condition to return null was already written down in the code - `Number.isNaN(contentHeight)`

js/packages/eyes-sdk-core/lib/config/Configuration.js
```diff
@@ -31,6 +31,9 @@ const DEFAULT_VALUES = {
   concurrentSessions: 1,
   isThrowExceptionOn: false,
   waitBeforeCapture: 100, //ms
+
+  // mobile
+  waitForMoblieAsync: 500, //ms
```
kyryloonufriiev, line 36: You probably forget to remove this

js/packages/driver/src/element.ts
```diff
@@ -203,8 +203,9 @@ export class Element<TDriver, TContext, TElement, TSelector> {
     return region
   }
 
-  async getContentSize(): Promise<types.Size> {
+  async getContentSize(options: {contentRegionWaitTimeout?: number} = {}): Promise<types.Size> {
```
kyryloonufriiev, line 206: `contentRegionWaitTimeout` - isn't a good naming, no way I will understand what it actually means. I'm all pro does not invent a new name for things that already exist. So you will pass here `LazyLoad.waitTimeout`. I would vote for the option to accept an object with `lazyLoad` property because at the end helper lib should support not just `waitTimeout` but all other `lazyLoad` properties as well.

js/packages/driver/src/helper-android.ts
```diff
@@ -61,20 +81,55 @@ export class HelperAndroid<TDriver, TContext, TElement, TSelector> {
...
+        this._logger.warn(
+          `waiting for text change from WAIT command ${command} for ${
+            maxTimeout - start
+          }ms, clear the text and continue`,
+        )
+        break
+      }
+      await utils.general.sleep(timeout)
+      text = await this._input.getText()
+    }
```
kyryloonufriiev, line 102: I'm personally not a big fan of `break`. Why not incorporate this condition into a `while`? And add a log after the loop if the `text` is still `WAIT`. Another thing I think it is better to calculate the end time from the beginning instead of summing up start and timeout on each itteration.

kyryloonufriiev (changes requested):

js/packages/driver/src/helper-android.ts
```diff
@@ -38,13 +57,16 @@ export class HelperAndroid<TDriver, TContext, TElement, TSelector> {
...
+    version: string
   }) {
     this._spec = options.spec
     this._input = options.input
     this._action = options.action
     this._legacy = options.legacy
     this._logger = options.logger
+    this._version = options.version || '0.0.0'
     this.name = this._legacy ? 'android-legacy' : 'android'
+    this._supportAsync = semverGte(this._version, '1.8.0')
```
kyryloonufriiev, line 69: Better pass `supportAsync` in the constructor, this will allow you to avoid passing `version`

js/packages/driver/src/helper-android.ts
```diff
@@ -61,20 +83,54 @@ export class HelperAndroid<TDriver, TContext, TElement, TSelector> {
       await this._action.type('1').catch(() => null)
       text = await this._input.getText()
     }
+
+    // wait until the JAVA complete the async operation
+    const timeout = 500,
+      start = Date.now(),
+      maxTimeout = 5 * 60 * 1000 + start
```
kyryloonufriiev, line 90: This is not, in fact, a `maxTimeout`, but something like `finishAt`, because this is not a duration, but a date. Also `start` variable is needed only once, so not really make sense to create it.

js/packages/driver/src/helper-android.ts
```diff
@@ -61,20 +83,54 @@ export class HelperAndroid<TDriver, TContext, TElement, TSelector> {
...
+
+    let now = Date.now()
+    while (text === 'WAIT' && maxTimeout > now) {
+      now = Date.now()
```
kyryloonufriiev, line 94: Why not?
```suggestion
    while (text === 'WAIT' && maxTimeout > Date.now()) {
```

js/packages/driver/src/helper-android.ts
```diff
@@ -61,20 +83,54 @@ export class HelperAndroid<TDriver, TContext, TElement, TSelector> {
...
+  async getContentRegion(
+    element: Element<TDriver, TContext, TElement, TSelector>,
+    options?: {lazyLoad?: types.LazyLoadOptions},
+  ): Promise<types.Region> {
+    const elementId = await this._getElementId(element)
+    if (!elementId) return null
+
+    let contentHeightString = ''
+
+    const waitingTime = options?.lazyLoad?.waitingTime ?? 0
```
kyryloonufriiev, line 117: I think it is important to keep the variable as close as possible to the place it belongs. Otherwise, it will create constant discomfort while reading such core, especially when someone else put their logic in between.

js/packages/driver/src/helper-android.ts
kyryloonufriiev: there is no need to continue if the result is `''` or `'blablabla'` or anything else that could not be cast to a number. This is why I'm suggesting casting it to the number as soon as you get the string, so you can check if it is not `NaN` in order to continue.

kyryloonufriiev (changes requested):

js/packages/driver/src/helper-android.ts
```diff
@@ -11,23 +13,39 @@ export class HelperAndroid<TDriver, TContext, TElement, TSelector> {
...
+      type: 'xpath',
+      selector: '//*[@content-desc="EyesAppiumHelper_Version"]',
+    })
     if (!input) {
       legacy = true
       input = await driver.element({type: 'xpath', selector: '//*[@content-desc="EyesAppiumHelper"]'})
     }
+    if (versionElement) {
+      version = await versionElement.getText()
+    }
```
kyryloonufriiev, line 28: I believe this logic has to be much more compact, as well as not spreaded
```suggestion
    let input = await driver.element({type: 'xpath', selector: '//*[@content-desc="EyesAppiumHelperEDT"]'})
    if (!input) {
      legacy = true
      input = await driver.element({type: 'xpath', selector: '//*[@content-desc="EyesAppiumHelper"]'})
    }
    const versionElement = await driver.element({
      type: 'xpath',
      selector: '//*[@content-desc="EyesAppiumHelper_Version"]',
    })
    const version = await versionElement?.getText() ?? '0.0.0'
```

PR #1091, opened 2022-09-06
Try to remove  `__applitoolsBrand`

zamboney:
It seems to break the typing in `.ts` file that use it

### DIFF
--- a/js/packages/eyes-cypress/src/browser/spec-driver.ts
+++ b/js/packages/eyes-cypress/src/browser/spec-driver.ts
@@ -1,7 +1,7 @@
 type EyesSelector = {selector: string; type?: string}
-export type Selector = (string | EyesSelector) & {__applitoolsBrand?: never};
-export type Context = Document & {__applitoolsBrand?: never};
-export type Element = HTMLElement & {__applitoolsBrand?: never};
+export type Selector = (string | EyesSelector)
+export type Context = Document
+export type Element = HTMLElement
 
 export function executeScript(context: Context, script: string, arg: any): any {     
 
--- a/js/packages/eyes-nightwatch/src/spec-driver.ts
+++ b/js/packages/eyes-nightwatch/src/spec-driver.ts
@@ -2,13 +2,13 @@ import type {Size, Cookie, DriverInfo} from '@applitools/types'
 import type * as Nightwatch from 'nightwatch'
 import * as utils from '@applitools/utils'
 
-export type Driver = Nightwatch.NightwatchAPI & {__applitoolsBrand?: never}
+export type Driver = Nightwatch.NightwatchAPI
 export type Element = (
   | {ELEMENT: string}
   | {'element-6066-11e4-a52e-4f735466cecf': string}
   | Nightwatch.NightwatchTypedCallbackResult<{ELEMENT: string} | {'element-6066-11e4-a52e-4f735466cecf': string}>
-) & {__applitoolsBrand?: never}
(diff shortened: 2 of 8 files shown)


giladgd (approved):
LGTM

kyryloonufriiev:
@zamboney What is the point? It was there for a reason!

zamboney:
@giladgd can you please add an example for the problem please

giladgd:
@kyryloonufriiev You can see in the example code below that our SDK rejects the `ElementHandle` type from being passed to `Target.window().ignoreRegion( ... )`:
image

But you can see that the TS types are supposed to support such a case:
image
image

In fact, the above code actually does work, but the TS types are incompatible because of the `& {__applitoolsBrand?: never}` type used when referencing Puppeteer's `ElementHandle` type.

Another example of this problem exists in `eyes.open( ... )`, in this case, you must pass a `puppeteer.Page` element to `eyes.open( ... )`, although the TS types prohibit it:
image

This basically means that the SDK cannot be properly used in TypeScript without manually doing something like this:
image
Which we shouldn't deliberately ask users to do.

Usually, something like `& {__applitoolsBrand?: never}`  type is used when you want to expose some function that returns an object with such a type, which is a type the user cannot construct by themselves, while the underlying JS object could theoretically be constructed by the user but we want to prohibit that.

For example:
image

In our case, I couldn't find a good reason why we use `& {__applitoolsBrand?: never}`.
It was added as part of a big refactor of the codebase, without giving an explicit reason why that change was made, and I couldn't find any comments or documentation with an explanation of why this type is in place.
Furthermore, I tried finding a way to construct such a type from the SDK but couldn't find anything.

Can you please explain why we have `& {__applitoolsBrand?: never}` in the first place?
I think we can find a solution that serves the original purpose of this while still allowing users to easily use our SDKs in TS without any type casting needed.

kyryloonufriiev:
@giladgd Thank you for the explanation. I don't want to start by explaining why I've added this brand instead I want to get to the bottom of why you have those ts errors. It doesn't seem to me like the brand is the problem, because the error explicitly says what exactly is different in those two types in both cases and this is not the brand.
Also if you have a closer look you will see that such a brand cannot cause any problem, because this is not the kind of brand you referring to in your vision of "how it is usually used". What I've added here is actually prohibiting the property `__applitoolsBrand` from appearing in the type (and all of this _optionanly_, double safety).
The bottom line - I'm not convinced that the brand causes any problems. There might be some mess with your dependencies, so typescript gets types from a different pptr version than you actually use. 
I don't have pptr + ts demo right now, but here is our playwright tutorial, according to what you wrote should have the same problem, but it is not.
image

giladgd:
@kyryloonufriiev Thanks for the explanation. You're right, I missed the optionality of `__applitoolsBrand` detail, so this shouldn't cause problems either way.
Thanks for pointing me to check the Puppeteer module version, there indeed was a mess with the installed module versions in our project, which I now fixed and the problem seems to be resolved now.

amitzur:
@giladgd what was the problem with the module versions? Is it related to npm workspaces?

giladgd:
@amitzur For some reason, the structure of the packages saved in `package-lock.json` made an older version of Puppeteer needed by an unrelated package be shared with `@applitools/eyes-puppeteer` instead of the newer version of Puppeteer that was installed in our project.
I solved it by deleting `package-lock.json` and constructing it again from scratch.
I suspect the root cause for this is related to `--legacy-peer-deps` we used to use to construct `package-lock.json` in the first place

PR #1092, opened 2022-09-07
js/core: add nml-client support for ios

tourdedave:


tourdedave:
@kyryloonufriiev I had trouble getting a local build going to verify my e2e test in core. Any pointers on what to do?

PR #1093, opened 2022-09-07
Nightwatch support v2

SophieDePaula:
We should support Nightwatch v1 and v2
[Breaking changes](https://github.com/nightwatchjs/nightwatch/wiki/Migrating-to-Nightwatch-2.0)
[Trello card](https://trello.com/c/LMzJuAm9/71-support-for-nightwatch2)

What I noticed so far that is effecting us that is not mentioned in the breaking changes above is the returned value of `driver.execute`.
In V1, it used to be `this` and not it's `Awaitable<this, T>`. Therefore, we should not wrap `call` returned value with a `Promise`.

TO DO:

- [x] Fix element handling
- [x] Fix `call` method in `spec-driver`
- [x] Run all tests against V1 and V2
- [x] `startWebDriver` is not supported in V2, I need to see how it effects us


js/packages/driver/src/element.ts
```diff
@@ -674,7 +674,10 @@ export class Element<TDriver, TContext, TElement, TSelector> {
       const result = await operation()
       // Some frameworks could handle stale element reference error by itself or doesn't throw an error
       if (this._spec.isStaleElementError(result, this.selector as TSelector)) {
-        await this.refresh()
+        const refreshed = await this.refresh()
+        if (!refreshed) {
+          throw result
```
SophieDePaula, line 679: I think it makes sense to throw here the `stale element exception` if we are unable to refresh the context, the same as we do in `catch`. Otherwise, the error we are getting is less helpful.

kyryloonufriiev: I believe we could do something like this.

js/packages/eyes-nightwatch/test/e2e/cucumber/step_definitions/steps.js
```diff
@@ -1,18 +1,57 @@
...
+})
 
 Given(/^I open Applitools`s demo page$/, async () => {
-  console.log(client.transport)
-  await client.url('https://applitools.github.io/demo/')
+  if (process.env.APPLITOOLS_NIGHTWATCH_MAJOR_VERSION === '1') {
+    await client.url('https://applitools.github.io/demo/')
+  } else {
+    await browser.url('https://applitools.github.io/demo/')
+  }
```
kyryloonufriiev, line 40: wouldn't it be simpler to create a global variable where you will save the browser or client?

SophieDePaula:
@kyryloonufriiev this PR is ready for review. The only thing that I am missing is that when running Nightwatch v1 from the `github` `action`, the major version env var is still missing. What did I miss?

pending-changes.yaml
```diff
@@ -46,6 +46,7 @@ Apollo SDK:
 "@applitools/eyes-nightwatch":
   feature:
     - Don't fail `eyes.open` when there is a failure to set viewport size in `UFG`.
+    - Support Nightwatch major version 2
```
SophieDePaula, line 49: @benbabayoff8

benbabayoff8: Looks good to me. Thanks

.github/workflows/publish-nightwatch.yml
```diff
@@ -69,16 +69,17 @@ jobs:
...
-            echo "::set-output name=environments::[{\"node-version\":16}]"
+            echo "::set-output name=environments::[{\"framework-version\":2,\"node-version\":16}]"
           fi
   test:
     needs: [verify]
     strategy:
       max-parallel: 3
       fail-fast: false
       matrix:
+        framework-version: [1, 2]
```
kyryloonufriiev, line 82: You don't need to do this, because you completely defined it in `include` property.

.github/workflows/publish-nightwatch.yml
```diff
@@ -69,16 +69,17 @@ jobs:
...
         include: ${{fromJSON(needs.verify.outputs.environments)}}
     name: 'Test (node: ${{matrix.node-version}})'
     runs-on: ubuntu-latest
```
kyryloonufriiev, line 85: Here is what you've missed, to have env var for v1
```suggestion
    name: 'Test (node: ${{matrix.node-version}})'
    runs-on: ubuntu-latest
    env:
      APPLITOOLS_NIGHTWATCH_MAJOR_VERSION: ${{matrix.framework-version}}
```

js/packages/eyes-nightwatch/package.json
```diff
@@ -43,18 +43,19 @@
...
     "test": "yarn test:it && yarn test:e2e && yarn test:coverage",
     "test:sanity": "yarn test:it",
     "test:it": "mocha ./test/it/spec-driver.spec.js --exit --no-timeouts -r @applitools/test-utils/mocha-hooks/docker",
     "test:e2e": "yarn test:e2e:nightwatch && yarn test:e2e:cucumber",
     "test:e2e:cucumber": "cucumber-js test/e2e/cucumber/features/*.feature --require test/e2e/cucumber/cucumber.config.js --require test/e2e/cucumber/step_definitions",
     "test:e2e:nightwatch": "nightwatch --config test/e2e/nightwatch.conf.js --eyes-config applitools.config.js test/e2e/*.spec.js",
-    "test:coverage": "yarn generate:tests && APPLITOOLS_BATCH_NAME='JS Coverage Tests: eyes-nightwatch' APPLITOOLS_BATCH_ID=$(uuidgen) XUNIT_FILE=coverage-test-report.xml mocha --config .mocharc.cvg.js",
-    "report": "coverage-tests report https://raw.githubusercontent.com/applitools/sdk.coverage.tests/master/js/config.js --name 'eyes-nightwatch'",
+    "test:coverage": "yarn generate:tests && APPLITOOLS_BATCH_NAME='JS Coverage Tests: eyes-nightwatch' APPLITOOLS_BATCH_ID=$(uuidgen) XUNIT_FILE=coverage-test-report.xml mocha --config .mocharc.cvg.js -c test/nightwatch.conf.js",
```
kyryloonufriiev, line 53: What is `-c test/nightwatch.conf.js` ?

js/packages/eyes-nightwatch/src/spec-driver.ts
```diff
@@ -10,19 +10,21 @@ export type Element = (
 ) & {__applitoolsBrand?: never}
 export type Selector = {locateStrategy: Nightwatch.LocateStrategy; selector: string} & {__applitoolsBrand?: never}
 
-type ShadowRoot = {'shadow-6066-11e4-a52e-4f735466cecf': string}
+type ShadowRoot = {'shadow-6066-11e4-a52e-4f735466cecf': string} | {id_: string}
```
kyryloonufriiev, line 13: Hmmm. I'm curious why it is needed here.

js/packages/eyes-nightwatch/src/spec-driver.ts
```diff
@@ -169,7 +188,18 @@ export async function getCookies(driver: Driver, context?: boolean): Promise<Coo
 }
 export async function getCapabilities(driver: Driver): Promise<Record<string, any>> {
   try {
-    return await call(driver, 'session')
+    const session = await call(driver, 'session')
+    if (process.env.APPLITOOLS_NIGHTWATCH_MAJOR_VERSION === '1') {
+      return session
+    } else {
+      const iterator = session.caps_.map_[Symbol.iterator]()
```
kyryloonufriiev, line 195: I don't like the fact that we are using internal properties here. Isn't there a way to do it officially?

js/packages/eyes-nightwatch/src/spec-driver.ts
```diff
@@ -264,10 +294,25 @@ export async function build(env: any): Promise<[Driver, () => Promise<void>]> {
...
+    const client = Nightwatch.client(settings)
+    client.isES6AsyncTestcase = true
+    await client.createSession()
+    return [client.api, () => client.session.close()]
+  } else {
+    // for CheckRegionBySelectorInFrameFullyOnFirefoxLegacy
+    settings.desiredCapabilities.browserName =
+      settings.desiredCapabilities.browserName === 'Firefox'
+        ? settings.desiredCapabilities.browserName.toLowerCase()
+        : settings.desiredCapabilities.browserName
```
kyryloonufriiev, line 308: Why not lowercase it all the time?

js/packages/eyes-nightwatch/test/it/spec-driver.spec.js
```diff
@@ -361,7 +386,7 @@ describe('spec driver', async () => {
       value: 'world',
       domain: input && input.context ? '.applitools.github.io' : 'google.com',
       path: '/',
-      expiry: 4025208067,
+      expiry: IS_NIGHTWATCH_1 ? 4025208067 : 1698334725,
```
kyryloonufriiev, line 389: Look how it is done in selenium/wdio. This way you will not need to hardcode the value.

js/packages/eyes-nightwatch/src/spec-driver.ts
SophieDePaula: In v2, I do not longer see `shadow-6066-11e4-a52e-4f735466cecf` as the key to the id of the element, but I see `id_`.
Screen Shot 2022-09-22 at 1 29 10 PM

js/packages/eyes-nightwatch/package.json
SophieDePaula: I removed it, thanks

js/packages/eyes-nightwatch/src/spec-driver.ts
SophieDePaula: I noticed that it is needed only in FF legacy, I know that some of the other browser names are already lowered case, but for example, `MicrosofyEdge` is not and there is no need to lower case it.

js/packages/eyes-nightwatch/test/it/spec-driver.spec.js
SophieDePaula: I see it is hardcoded in selenium and wdio as well

js/packages/eyes-nightwatch/src/spec-driver.ts
SophieDePaula: I found a way to get the caps with a getter, however, I didn't find a way to get the map.
I found [this example](https://github.com/nightwatchjs/nightwatch/blob/da1bdc9c3d5297468828a88db8f6c6bbb66aa305/lib/transport/selenium-webdriver/session.js#L25) from `Nightwatch` repo, and it seems that they call a getter for each property separately.
` Object.getOwnPropertyNames` didn't give me anything other than `map_`

.github/workflows/publish-nightwatch.yml
kyryloonufriiev: yep

js/packages/eyes-nightwatch/src/spec-driver.ts
kyryloonufriiev: I don't think we should use undocumented private values that hard. You would better find the type definition for ShadowRoot type

js/packages/eyes-nightwatch/test/it/spec-driver.spec.js
kyryloonufriiev: https://github.com/applitools/eyes.sdk.javascript1/blob/6844fe6db97b224d052165ccf149f5c41942ea37/js/packages/spec-driver-webdriverio/test/e2e/spec-driver.spec.ts#L563

js/packages/eyes-nightwatch/src/spec-driver.ts
kyryloonufriiev: Doc says it should be a simple object... I really don't want to use private props if we don't have to. I believe there should be a better way.

SophieDePaula: done

js/packages/eyes-nightwatch/src/spec-driver.ts
SophieDePaula: I use a getter instead of this private property now

js/packages/eyes-nightwatch/test/it/spec-driver.spec.js
SophieDePaula: Got it, thanks

js/packages/eyes-nightwatch/src/spec-driver.ts
```diff
@@ -10,19 +10,21 @@ export type Element = (
...
-type ShadowRoot = {'shadow-6066-11e4-a52e-4f735466cecf': string}
+type ShadowRoot = {'shadow-6066-11e4-a52e-4f735466cecf': string} | {getId: () => string}
 type CommonSelector<TSelector = never> = string | {selector: TSelector | string; type?: string}
 
 // #region HELPERS
 
 const LEGACY_ELEMENT_ID = 'ELEMENT'
 const SHADOW_ROOT_ID = 'shadow-6066-11e4-a52e-4f735466cecf'
 const ELEMENT_ID = 'element-6066-11e4-a52e-4f735466cecf'
+const SHADOW_ROOT_V2_ID = 'getId'
```
kyryloonufriiev, line 21: I don't think there is a need for it. The reason it was existed because keys are not really readable. `getId` is fine naming

js/packages/eyes-nightwatch/src/spec-driver.ts
```diff
@@ -38,13 +40,20 @@ function call<
 >(driver: Driver, command: TCommand, ...args: any[]): Promise<TResult> {
   return new Promise<TResult>((resolve, reject) => {
     const promise = (driver[command] as any)(...args, (result: Nightwatch.NightwatchCallbackResult<TResult>) => {
+      if (!result) reject('Got empty result')
```
kyryloonufriiev, line 43: You should not reject primitives such as string, if you want you has to reject `Error` object

js/packages/eyes-nightwatch/src/spec-driver.ts
```diff
@@ -157,10 +166,20 @@ export async function setWindowSize(driver: Driver, size: Size): Promise<void> {
...
   } catch {
-    await call(driver, 'setWindowPosition' as 'windowPosition', 0, 0)
-    await call(driver, 'setWindowSize' as 'windowSize', size.width, size.height)
+    if (process.env.APPLITOOLS_NIGHTWATCH_MAJOR_VERSION === '1') {
+      await call(driver, 'setWindowPosition' as 'windowPosition', 0, 0)
+      await call(driver, 'setWindowSize' as 'windowSize', size.width, size.height)
+    } else {
+      await driver.setWindowPosition(0, 0)
+      await driver.setWindowSize(size.width, size.height)
+    }
```
kyryloonufriiev, line 182: what is the difference? in call it will do the same thing as you are doing in `else`

SophieDePaula: @kyryloonufriiev I wrote a comment in line 172, for `setWindowRect`. I did the same in the `catch` to be consistent as this might also result in the same issue.

PR #1097, opened 2022-09-08
Storybook add tiemout and retry for take dom snapshots

SophieDePaula:
[Trello card](https://trello.com/c/QN2p1v0O/1401-cvent-storybook-execution-hanging)
To solve the issue above, I set a timeout on `takeDomSnapshots` and if we fail, we retry in `renderStories`


pending-changes.yaml
```diff
@@ -61,7 +61,7 @@ Apollo SDK:
   feature:
     - 
   bug-fix:
-    - 
+    - Add a timeout when taking DOM per story and add a retry
```
SophieDePaula, line 64: @benbabayoff8

benbabayoff8: @SophieDePaula I would write it only as retry, the timeout is implicit, isn't it?

SophieDePaula: @benbabayoff8 not sure what you mean

benbabayoff8: So I'm fine with the current changelog. Thanks

kyryloonufriiev:
`takeDomSnapshot` has timeout already - https://github.com/applitools/eyes.sdk.javascript1/blob/master/js/packages/eyes-sdk-core/lib/utils/takeDomSnapshot.js#L15

SophieDePaula:
> `takeDomSnapshot` has timeout already - https://github.com/applitools/eyes.sdk.javascript1/blob/master/js/packages/eyes-sdk-core/lib/utils/takeDomSnapshot.js#L15

@kyryloonufriiev from the investigation Amit and I did, we saw that even though we have a timeout where you mentioned, the promise for `takeDomSnapshots` in `getStoryData` is never rejected and therefore we hang.
We saw two missing stories that the last log line from them was `unning takeDomSnapshot(s) for story` and we saw two page crash `Puppeteer error for page 10: Error: Page crashed!`.

kyryloonufriiev:
@SophieDePaula Got it. Thank you!

pending-changes.yaml
amitzur: @SophieDePaula I think @benbabayoff8 meant that from the user's perspective, the timeout is not interesting since it's an implementation detail, and the important part is about the retry. So I think something like this:

> Prevent Eyes-Storybook from hanging when there is a failure to take DOM.

amitzur:
@kyryloonufriiev @SophieDePaula yes, it was very strange. This is exactly where we were looking, but Puppeteer seems to behave in a weird way, the poll script is not resolving or rejecting, as Sophie said.
Without deeply understanding Puppeteer's behavior here, we went with a solution that will first solve the issue for the customer, and then we can investigate offline.

kyryloonufriiev (approved):

PR #1100, opened 2022-09-13
Add the scroll offset to region

zamboney:
Add to the region offset the scroll offset.

The calculation of the region base of the [`getBoundingClientRect`](https://github.com/applitools/eyes.sdk.javascript1/blob/07d0fb36b5ce7ecb88c3d4854871686d082b664c/js/packages/snippets/src/browser/getElementRect.js#L13), the `getBoundingClientRect` calculate the offset ["relative to the viewport"](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect).

This is a problem if the test scroll the page before before `eyes.check` because it's try to move to region while SER is already scrolled

This fix the issue by add to the `x` and `y` coordinates of the `region` the
scroll `x` and `y` coordinates.

## Testing

TBD;


amitzur (commented):
This looks good to me. 2 comments:
1. The case of `fully` region should also be covered. You can see in line 44 that it's the same logic as the `else` when `isScrollable` is `null`.
2. The test IMO should be just like other screenshoter tests, where before taking the screenshot you would scroll the page. I would cover both default *and* custom `scrollingElement`.

js/packages/screenshoter/src/get-target.js
```diff
@@ -37,27 +37,34 @@ async function getTarget({window, context, region, fully, scrollingMode, logger}
...
       } else {
-        const scrollingElement = await context.getScrollingElement()
-        return {
-          context: elementContext,
-          region: await element.getRegion(),
-          scroller: makeScroller({element: scrollingElement, scrollingMode, logger}),
+        scrollingElement = await context.getScrollingElement()
+        returnRegion = await element.getRegion()
+      }
+      scroller = makeScroller({element: scrollingElement, scrollingMode, logger})
```
kyryloonufriiev, line 53: No reason for the `scroller` to be `let`

js/packages/screenshoter/src/get-target.js
```diff
@@ -37,27 +37,34 @@ async function getTarget({window, context, region, fully, scrollingMode, logger}
...
+      const isScrollerRoot = await scroller.element.isRoot()
+      if (!isScrollerRoot && returnRegion) {
+        const scrollerOffset = await scroller.getScrollOffset()
+        returnRegion = {
+          ...returnRegion,
+          x: returnRegion.x + scrollerOffset.x,
+          y: returnRegion.y + scrollerOffset.y,
         }
```
kyryloonufriiev, line 61: We already have utility functions for this kind of geometry manipulation.
```suggestion
        returnRegion = utils.geometry.offset(returnRegion, scrollerOffset)
```

js/packages/screenshoter/src/get-target.js
```diff
@@ -37,27 +37,34 @@ async function getTarget({window, context, region, fully, scrollingMode, logger}
...
-        return {
-          context: elementContext,
-          region: await element.getRegion(),
-          scroller: makeScroller({element: scrollingElement, scrollingMode, logger}),
+        scrollingElement = await context.getScrollingElement()
+        returnRegion = await element.getRegion()
+      }
+      scroller = makeScroller({element: scrollingElement, scrollingMode, logger})
+      const isScrollerRoot = await scroller.element.isRoot()
+      if (!isScrollerRoot && returnRegion) {
```
kyryloonufriiev, line 55: no need to call `isRoot` from the `scroller`'s element, when you already have the `scrollingElement` variable.
I would instead move the `scroller` creation back where it was, in the return statement, since there is no need to work with the `scroller` object here.

kyryloonufriiev (changes requested):

zamboney:
@kyryloonufriiev  i fix your comment. 
Also, can you help me decide what to add `pending-changes.yaml`? i was thinking:

- [ ] `"@applitools/eyes-selenium"`
- [ ] `"@applitools/eyes-webdriverio"`
- [ ] `"@applitools/eyes-playwright"`
- [ ] `"@applitools/eyes-protractor"`
- [ ] `"@applitools/eyes-puppeteer"`
- [ ] `"@applitools/eyes-testcafe"`
- [ ] `"@applitools/eyes-nightwatch"`
- [ ] `"@applitools/eyes-cypress"`

WDYT? (fill free to check which package you want me to add the changes)

js/packages/screenshoter/test/e2e/web/region-fully.spec.js
```diff
@@ -36,4 +36,17 @@ describe('screenshoter web', () => {
       logger,
     })
   })
+
+  it('take full region screenshot with "css" scrolling', async () => {
```
kyryloonufriiev, line 40: I don't like that this test is called exactly the same as the one above. I recommend
```suggestion
  it('take full region screenshot after manual scroll with "css" scrolling', async () => {
```

js/packages/screenshoter/test/e2e/web/region.spec.js
```diff
@@ -34,4 +34,25 @@ describe('screenshoter web', () => {
...
+  it('take region by selection with force offset', async () => {
+    await driver.execute(`document.querySelector('html').scrollTop = 200`)
+    await test({
+      type: 'web',
+      tag: 'region-select',
+      region: '#centered',
+      driver,
+      logger,
+    })
+  })
```
kyryloonufriiev, line 57: This is not an appropriate place for this test. In the naming scheme `region` is always `{x, y, width, height}`. You should put it in the `element.spec.js` file. Also, the need for the first test is questionable, because we already have exactly the same test.

js/packages/screenshoter/test/e2e/web/region-inside-scrollable-element.spec.js
```diff
@@ -0,0 +1,62 @@
...
+
+  it('take region by selection inside scrollabe element', async () => {
+    await test({
+      type: 'web',
+      tag: 'region-inside-scrollable-element',
+      region: '#lipsum',
+      driver,
+      logger,
+    })
+  })
```
kyryloonufriiev, line 25: I'm not sure what it is testing. This is just an element screenshot.

js/packages/screenshoter/test/e2e/web/region-inside-scrollable-element.spec.js
```diff
@@ -0,0 +1,62 @@
...
+  it.skip('take region by selection inside scrollabe element with fully', async () => {
+    await test({
+      fully: true,
+      type: 'web',
+      tag: 'region-inside-scrollable-element',
+      region: '#lipsum',
+      driver,
+      logger,
+    })
+  })
```
kyryloonufriiev, line 37: You see the "bug in the stitching" because you didn't specify `scrollingElement`, we are not able to detect it automatically at the moment.
I recognize the need for such a test because at the moment we don't test a full screenshot of the element, which is not scrollable itself. (we actually test it but when an element is inside a frame). I think such a test should go in `element-fully.spec.js` file, or in its own file. Also, the case with `css` stitching should be tested.

kyryloonufriiev (changes requested):
I left some comments regarding the tests you've added. And here is some comments in general:

- I believe it is better to mark tests with `after manual scroll` then `with force offset`. I would never gues what is the force offset if I saw the name of this test failed in the report.
- `region by selection` should actually be `region by selector`. But I would just call it `element` as we call it in all other tests in this package.

kyryloonufriiev:
> @kyryloonufriiev i fix your comment. Also, can you help me decide what to add `pending-changes.yaml`? i was thinking:
> 
> * [ ]  `"@applitools/eyes-selenium"`
> * [ ]  `"@applitools/eyes-webdriverio"`
> * [ ]  `"@applitools/eyes-playwright"`
> * [ ]  `"@applitools/eyes-protractor"`
> * [ ]  `"@applitools/eyes-puppeteer"`
> * [ ]  `"@applitools/eyes-testcafe"`
> * [ ]  `"@applitools/eyes-nightwatch"`
> * [ ]  `"@applitools/eyes-cypress"`
> 
> WDYT? (fill free to check which package you want me to add the changes)

It should simply go in all of the SDKs that have classic support. Means all of them except cypress, storybook, and testcafe. Pay attention that there are some SDKs that are not on your list.

pending-changes.yaml
```diff
@@ -18,40 +18,40 @@ Apollo SDK:
     - Don't fail `eyes.open` when there is a failure to set viewport size in `UFG`.
     - Added support for lazy loading views in android native apps
   bug-fix:
-    - 
+    - If the `SRE` is not the `root` and his offset to the region.
```
kyryloonufriiev, line 21: I'm not sure what it means... I can guess since I know what you've done.
I think we don't need to highlight technical details of the fix/feature, but prescribe what problem we solved product-wise. 
In this case, I would write
`Fixed incorrect calculation of the target element position`

@benbabayoff8 What you would say about the general guideline?

benbabayoff8: @kyryloonufriiev I agree, we shouldn't include technical details or implementation stuff.
The updated suggestion looks good to me.

js/packages/screenshoter/test/e2e/web/region-inside-scrollable-element.spec.js
zamboney: it's testing it with the `region`

js/packages/screenshoter/test/e2e/web/region-inside-scrollable-element.spec.js
zamboney: i specify the `scrollingElement` [here](https://github.com/applitools/eyes.sdk.javascript1/blob/msft_scrolltop_issue/js/packages/screenshoter/test/e2e/web/region-inside-scrollable-element.spec.js).
i will add a `css` case and move this test to `element-fully.spec.js`

js/packages/screenshoter/test/e2e/web/region.spec.js
zamboney: your right i will remove this test

js/packages/screenshoter/test/e2e/web/region-inside-scrollable-element.spec.js
kyryloonufriiev: Ah. I didn't see this part of the test. Sorry.

js/packages/screenshoter/test/e2e/web/element.spec.js
```diff
@@ -34,4 +34,15 @@ describe('screenshoter web', () => {
...
+  it('take element by selection with manual scroll', async () => {
+    await driver.execute(`document.querySelector('html').scrollTop = 200`)
+    await test({
+      type: 'web',
+      tag: 'element-select',
+      region: '#centered',
+      driver,
+      logger,
+    })
+  })
```
kyryloonufriiev, line 47: Can you preserve a naming scheme of the tests above? It should be `take element screenshot after manual scroll with "scroll" scrolling`. I don't see a reason to use a different fixture here, since this could be tested as well with the `#overflowing-div-image` element.

js/packages/screenshoter/test/e2e/web/region-fully.spec.js
```diff
@@ -36,4 +36,17 @@ describe('screenshoter web', () => {
...
+    await test({
+      type: 'web',
+      tag: 'region-fully',
+      region: {x: 30, y: 500, height: 700, width: 200},
+      fully: true,
+      scrollingMode: 'css',
+      driver,
+      logger,
+    })
+  })
```
kyryloonufriiev, line 51: should be also tested with "scroll" scrolling

js/packages/screenshoter/test/e2e/web/region-inside-scrollable-element.spec.js
```diff
@@ -0,0 +1,50 @@
...
+  it('take region by element inside scrollable element with force offset', async () => {
+    await driver.execute(`document.querySelector('.wrapper').scrollTop = 200`)
+    await test({
+      type: 'web',
+      tag: 'region-inside-scrollable-element',
+      region: '#lipsum',
+      driver,
+      logger,
+    })
+  })
```
kyryloonufriiev, line 36: We should test things like this with both "css" and "scroll" scrolling modes.

kyryloonufriiev (approved):
I left some comments, that are more related to the way you call things, or coverage improvements. 
I ask you to take into consideration the tests we already have as well as the terminology and structure we use there. You came up with a great test case - "what if we are going to take screenshots after the page was already scrolled". I love this, but we should test it thoroughly in all places this could take effect.
The bottom line, you can merge it if you want. But I would also ask you to run all of the screenshoter tests before you will merge your PR, just to be sure that everything will work fine.

zamboney:
@kyryloonufriiev i run all the `test:e2e:web` test locally and they all passed

kyryloonufriiev:
@zamboney I mean all of them. Native as well.

zamboney:
> @zamboney I mean all of them. Native as well.

i will create new PR for that

PR #1101, opened 2022-09-13
Assign `lazyLoad.waitingTime` to `waitBeforeScreenshots`

zamboney:
The idea is that once the `lazyloadWaitingTime` and the `waitBeforeScreenshots` isn't sync it causes a stitching inconsistency.

This PR will make sure that `lazyloadWaitingTime` equals to `waitBeforeScreenshots`, only if the user didn't pre-set `waitBeforeScreenshots`.


kyryloonufriiev:
I didn't deep dive into the implementation. But here is what I wanna say, even putting aside that the idea isn't valid in 100% of cases, I don't want to implement features like that in the old core when we are about to have a new one.
I also would like to know more about the motivation from the description. At the moment I'm assuming that you decided on your own to change the thing you find confusing during the implementation of the native lazy loading feature.

zamboney:
I updated the description.

About the idea behind it, it was requested by @danielputerman and approve by @benbabayoff8 

If you think that the new core redesign will fix it then I will close this PR

PR #1102, opened 2022-09-13
NML proxy support

tourdedave:
TODO:

- [x] add e2e coverage in core
- [x] ~~see if nock can support a proxy configuration so it can be tested through an integration test in the nml-client instead of an e2e test~~
- [x] add pending changes entry for the universal server


js/packages/eyes-sdk-core/test/e2e/ufg-native-nml.spec.js
```diff
@@ -49,7 +50,34 @@ describe('UFG native NML', () => {
       const manager = await sdk.makeManager({type: 'vg', concurrency: 5})
       const eyes = await manager.openEyes({driver, config})
       await eyes.check()
+      await eyes.check()
```
kyryloonufriiev, line 53: What is the point of the second check?

kyryloonufriiev (approved):
LGTM! Let's merge

js/packages/eyes-sdk-core/test/e2e/ufg-native-nml.spec.js
tourdedave: It was a performance item we checked as part of the initial NMG integration.

PR #1105, opened 2022-09-19
WebView support

tourdedave:
[Trello](https://trello.com/c/uEWiEF5s)
[Generic Coverage Tests](https://github.com/applitools/sdk.coverage.tests/pull/167)

Use cases:

1. Customer switches to web view in their test before calling Eyes.check (and they do not specify `Target.webview` in their check settings)
2. Customer does not switch to web view in their test and calls Eyes.check with `Target.webview` in their check settings


pending-changes.yaml
```diff
@@ -25,6 +25,7 @@ Apollo SDK:
     - Using `lazyLoad.waitingTime` as a delay between stitches by default
     - Added `Sony Xperia 10 II` emulation device
     - Added `iPhone 14`  and `iPhone 14 Pro Max` ios devices
+    - Added support for webview switching in classic execution
```
tourdedave, line 28: @benbabayoff8 ☝️

tourdedave: Ben approved this offline

js/packages/screenshoter/src/take-screenshot.js
```diff
@@ -23,11 +23,18 @@ async function takeScreenshot({
...
+  webview,
 }) {
   debug =
     debug ||
     (process.env.APPLITOOLS_DEBUG_SCREENSHOTS_DIR ? {path: process.env.APPLITOOLS_DEBUG_SCREENSHOTS_DIR} : debug)
   logger = logger ? logger.extend({label: 'screenshoter'}) : makeLogger({label: 'screenshoter'})
+
+  if (driver.isMobile) {
+    if (webview) await driver.switchWorld(typeof webview === 'string' ? {id: webview} : null)
+    else await driver.refresh()
```
kyryloonufriiev, line 35: I'm not sure that the driver should ever be reinitialized in screenshoter. Since one of the requirements of the screenshoter package is that driver is already initialized and the state of the driver is correct.

kyryloonufriiev (commented):
Looks really great, I like the simplicity. There is one important comment regarding the spec-driver, let's discuss it.
I also think we should handle this case - if the user switched to the webview, and asks for a viewport screenshot it should get the whole viewport, not just the webview. This is how it works with frames. I believe we should be consistent here.

js/packages/screenshoter/src/take-screenshot.js
tourdedave: Okay. But isn't there a case where this could happen with webviews? Like when `driver.init()` has been called, then a user switches to a webview, and then they perform a check? This is the case that presented when testing e2e in screenshoter. `refresh` only reinitializes in this case ([link](https://github.com/applitools/eyes.sdk.javascript1/blob/js/feat/webview-support/js/packages/driver/src/driver.ts#L324-L333)).

tourdedave:
> if the user switched to the webview, and asks for a viewport screenshot it should get the whole viewport, not just the webview. This is how it works with frames.

@kyryloonufriiev Just to make sure I understand - you mean if the user specifies something like `Target.window().webview()`? Can you point me to an example e2e test for frames that does something similar for reference?

js/packages/screenshoter/src/take-screenshot.js
kyryloonufriiev: `driver.init()` will be performed in check, so `screenshoter` will always get a driver with an actual state. And this is not a coincidence this is the requirement of the `screenshoter` package.

tourdedave: Okay. This requirement was not clear to me. Is it written down somewhere? Seems like a helpful thing to put in a readme.

tourdedave:
> @tourdedave I had this case in mind.
> 
> ```js
> await driver.switchContext('webview')
> await eyes.check(Target.window())
> ```
> 
> I think we don't have a test for such case. But you can see this very clear in screenshoter when we calculate a `window` value.

@kyryloonufriiev Are you saying we don't have a test that demonstrates what you're describing for frames (but it's something we support for frames)?

I wanted a test so I could see what the final image looks like in the case of frames (since you said it's similar behavior). It should be a viewport screenshot which contains the contents of the webview, right?

What's the desired outcome of doing what I described (`Target.window().webview()`)?

kyryloonufriiev:
@tourdedave 
- There is nothing special about it - `Target.window().webview()`, same as `Target.webview()`.
- Yes, there is no test for taking a viewport screenshot after manual switch to a frame.

When the user doing `Target.window()` it shouldn't matter where the user is in webview or frame. It is always the topmost context or world for that matter. `Target.window().fully()` will take a full screenshot of the native world. BTW speaking of full screenshots, there is a way to specify the scroll root element for a target context and the main context separately. I think we should give it some thought in the context of webviews.

tourdedave:
@kyryloonufriiev

So, to put into my own words.

For the implicit case (where a user switches into a webview before calling check), we currently support:

```
await driver.switchContext('webview')
await eyes.check()
```

Which gets us just the webview. And we also want to support:

```
await driver.switchContext('webview')
await eyes.check(Target.window())
```

Which would get us the viewport (not just the webview). Is that right?

> there is a way to specify the scroll root element for a target context and the main context separately. I think we should give it some thought in the context of webviews.

Okay, but I committed to shipping webview support by Wednesday. If we're looking to increase the scope further then I think we should save this functionality for a follow-on version. What do you think?

kyryloonufriiev:
Ok. I'm fine with not implementing scroll root thing.

The examples you gave have no difference. `Target.window()` is the same as passing nothing. What I was saying is that if you are going to check WINDOW it should be topmost world, doesn't matter what is the current state of the driver. But if you are going to check region or frame, it should be related to the current state of the driver.
Here is the example
`Target.window()` - is `cd /` 
`Targer.region('bla')` or `Target.frame('bla') is `cd ./bla`

tourdedave:
> The examples you gave have no difference. `Target.window()` is the same as passing nothing. What I was saying is that if you are going to check WINDOW it should be topmost world, doesn't matter what is the current state of the driver. But if you are going to check region or frame, it should be related to the current state of the driver. Here is the example `Target.window()` - is `cd /` `Targer.region('bla')` or `Target.frame('bla') is `cd ./bla`

I thought that the implicit use case we wanted to support was preserving the driver state that the user switched to. Did I misunderstand? That's what is tested [here](https://github.com/applitools/eyes.sdk.javascript1/blob/js/feat/webview-support/js/packages/core/test/e2e/classic/webview.spec.ts#L42-L62) which gets [this result Eyes](https://eyes.applitools.com/app/test-results/00000251738079447364/00000251738079447177/steps/1?accountId=xIpd7EWjhU6cjJzDGrMcUw~~&display=details&fb=eyJkYXRlUmFuZ2UiOnsiZnJvbURhdGUiOiIyMDIyLTA5LTI2VDA0OjAwOjAwLjAwMFoiLCJ0b0RhdGUiOiIyMDIyLTA5LTI2VDA0OjAwOjAwLjAwMFoifSwidGV4dFNlYXJjaCI6IndlYnZpZXcifQ%3D%3D&mode=step-editor&top=00000251738079447364%2816%29).

If there is still confusion, then let's hop on a call tomorrow to discuss.

kyryloonufriiev:
As I said state is important when you are checking "not a window". Because window is special – it always means root.

tourdedave:
I feel like we're talking past each other. I'll sync up with you tomorrow.

kyryloonufriiev (changes requested):

pending-changes.yaml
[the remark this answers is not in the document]
tourdedave: We'll need to add it at some point (not for this release). We should keep track of this in the Trello card. I'll note it there.

PR #1106, opened 2022-09-22
`Lazyload` native, change stitch szie

zamboney:
try to fix the issue of `lazyLoad` native feature by scrolling half way to to the screen between each stitch


js/packages/screenshoter/src/take-stitched-screenshot.js
```diff
@@ -80,7 +80,18 @@ async function takeStitchedScreenshot({
...
+  if (lazyLoad && driver.isNative) {
+    ;[, ...partRegions] = utils.geometry.divide(
+      region,
+      {
+        width: image.size.width,
+        height: image.size.height * 0.5,
+      },
+      overlap,
+    )
+  }
```
kyryloonufriiev, line 94: Isn't it an over complication? I believe it will be much easier to create a variable for the size (second argument), it value will depend on the condition you wrote

js/packages/screenshoter/test/e2e/android/app-fully-lazyload.spec.js
```diff
@@ -35,8 +35,8 @@ describe('screenshoter androidx app', () => {
       fully: true,
       framed: true,
       scrollingMode: 'scroll',
-      wait: 2000,
-      lazyLoad: {waitingTime: 1000},
+      wait: 2500,
+      lazyLoad: {waitingTime: 3000},
```
kyryloonufriiev, line 39: Why those values are different?

kyryloonufriiev (changes requested):

js/packages/screenshoter/test/e2e/android/app-fully-lazyload.spec.js
zamboney: it's made more consistent result locally, i can change them back because in sauce lab it's work before

js/packages/screenshoter/src/take-stitched-screenshot.js
zamboney: but i have to remove the first argument from the array, can you explain in code please?

js/packages/screenshoter/test/e2e/android/app-fully-lazyload.spec.js
zamboney: i don't think so i will try

js/packages/screenshoter/src/take-stitched-screenshot.js
zamboney: @kyryloonufriiev, the `initialRegion` must be from the origin `image` because this i how the rest of the algorithm works.

js/packages/driver/src/helper-android.ts
```diff
@@ -71,6 +71,10 @@ export class HelperAndroid<TDriver, TContext, TElement, TSelector> {
   }
 
   private async _command(command: string): Promise<string> {
+    // clean the input before passing value there could
+    // be a left over from previous async calls
+    await this._input.type('')
```
zamboney, line 76: this is to fix the issue that Yan saw we are passing `command` with previous results

PR #1107, opened 2022-09-26
update types

SophieDePaula:
Update cypress types in order to release cypress after the `core` redesign


js/packages/eyes-cypress/index.d.ts
```diff
@@ -22,26 +22,22 @@ interface CypressCheckSettings extends types.CheckSettings<Element, Selector> {
...
+interface CypressEyesConfig extends types.Batch, types.OpenSettings<'ufg'> {
+  batchId?: CypressEyesConfig['id']
+  batchName?: CypressEyesConfig['name']
+  batchSequence?: CypressEyesConfig['sequenceName']
+  notifyOnCompletion?:boolean
+  envName?: CypressEyesConfig['environmentName'],
+  browser?: MaybeArray<browserInfo>
+}
+interface browserInfo {
+  browser?: types.DesktopBrowserRenderer | types.ChromeEmulationDeviceRenderer | types.IOSDeviceRenderer | types.AndroidDeviceRenderer
```
kyryloonufriiev, line 39: This is not going to do the trick. You declared this 
```
browser: [{browser: {...}}, {browser: {...}}, {browser: {...}}]
```

js/packages/eyes-cypress/index.d.ts
```diff
@@ -22,26 +22,22 @@ interface CypressCheckSettings extends types.CheckSettings<Element, Selector> {
...
-  batchSequence?: CypressEyesConfig['batch']['sequenceName']
-  notifyOnCompletion?: CypressEyesConfig['batch']['notifyOnCompletion']
-
-  envName?: CypressEyesConfig['environmentName']
-
-  ignoreCaret?: CypressEyesConfig['defaultMatchSettings']['ignoreCaret']
-  matchLevel?: CypressEyesConfig['defaultMatchSettings']['matchLevel']
-  accessibilitySettings?: CypressEyesConfig['defaultMatchSettings']['accessibilitySettings']
-  ignoreDisplacements?: CypressEyesConfig['defaultMatchSettings']['ignoreDisplacements']
+interface CypressEyesConfig extends types.Batch, types.OpenSettings<'ufg'> {
```
kyryloonufriiev, line 30: You also lose all of the legacy properties and all of the properties that were in the Config but do not get it to OpenSettings.

js/packages/eyes-cypress/index.d.ts
```diff
@@ -22,26 +22,22 @@ interface CypressCheckSettings extends types.CheckSettings<Element, Selector> {
   strict?: MaybeArray<CypressCheckSettings['strictRegions'][number] | LegacyRegion>
   floating?: MaybeArray<CypressCheckSettings['floatingRegions'][number] | (({element: Element} | Selector | LegacyRegion) & {maxUpOffset?: number; maxDownOffset?: number; maxLeftOffset?: number; maxRightOffset?: number})>
   accessibility?: MaybeArray<CypressCheckSettings['accessibilityRegions'][number] | (({element: Element} | Selector | LegacyRegion) & {accessibilityType?: types.AccessibilityRegionType})>
-
   scriptHooks?: CypressCheckSettings['hooks']
+  ignoreCaret?: boolean
+  ignoreDisplacements?: boolean
```
kyryloonufriiev, line 27: I believe those properties are already in `CheckSettings`.

kyryloonufriiev (changes requested):

js/packages/eyes-cypress/index.d.ts
```diff
@@ -9,24 +9,25 @@ type LegacyRegion = {left: number; top: number; width: number; height: number}
...
   selector?: Selector
   element?: Element
 
-  ignore?: MaybeArray<CypressCheckSettings['ignoreRegions'][number] | LegacyRegion>
-  layout?: MaybeArray<CypressCheckSettings['layoutRegions'][number] | LegacyRegion>
-  content?: MaybeArray<CypressCheckSettings['contentRegions'][number] | LegacyRegion>
-  strict?: MaybeArray<CypressCheckSettings['strictRegions'][number] | LegacyRegion>
-  floating?: MaybeArray<CypressCheckSettings['floatingRegions'][number] | (({element: Element} | Selector | LegacyRegion) & {maxUpOffset?: number; maxDownOffset?: number; maxLeftOffset?: number; maxRightOffset?: number})>
-  accessibility?: MaybeArray<CypressCheckSettings['accessibilityRegions'][number] | (({element: Element} | Selector | LegacyRegion) & {accessibilityType?: types.AccessibilityRegionType})>
+  ignore?: MaybeArray<CypressCheckSettings['ignoreRegions'] | LegacyRegion>
```
kyryloonufriiev, line 20: What made you remove `[number]` here? Without `[number]` you declare that ignore will accept this - `[[{selector: 'bla', type: 'css'}, {region: {x: 0, y: 0, width: 100, height: 100}, regionId: 'blabla}]]`

js/packages/eyes-cypress/index.d.ts
```diff
@@ -9,24 +9,25 @@ type LegacyRegion = {left: number; top: number; width: number; height: number}
...
+  layout?: MaybeArray<CypressCheckSettings['layoutRegions'] | LegacyRegion>
+  content?: MaybeArray<CypressCheckSettings['contentRegions'] | LegacyRegion>
+  strict?: MaybeArray<CypressCheckSettings['strictRegions'] | LegacyRegion>
+  floating?: MaybeArray<CypressCheckSettings['floatingRegions']| (({element: Element} | Selector | LegacyRegion) & {maxUpOffset?: number; maxDownOffset?: number; maxLeftOffset?: number; maxRightOffset?: number})>
+  accessibility?: MaybeArray<CypressCheckSettings['accessibilityRegions'] | (({element: Element} | Selector | LegacyRegion) & {accessibilityType?: types.AccessibilityRegionType})>
 
-  scriptHooks?: CypressCheckSettings['hooks']
+  scriptHooks?: CypressCheckSettings['hook']
```
kyryloonufriiev, line 27: It is called `hooks`

kyryloonufriiev (changes requested):

js/packages/eyes-cypress/index.d.ts
```diff
@@ -9,24 +9,25 @@ type LegacyRegion = {left: number; top: number; width: number; height: number}
 type Selector = {selector: string; type?: 'css' | 'xpath', nodeType?: 'element' | 'shadow-root'} | 'string'
 type Element = HTMLElement | JQuery<HTMLElement>
 
-interface CypressCheckSettings extends types.CheckSettings<Element, Selector> {
+interface CypressCheckSettings extends Omit<api.CheckSettings<Element,Selector>, 
+'ignore' | 'floating' | 'content' | 'layout' | 'strict'> {
```
kyryloonufriiev, line 13: You should extend this interface, not from the `CheckSettings` but from the `CheckSettingsPlain` - https://github.com/applitools/eyes.sdk.javascript1/blob/8a02b0f6b3cad7c13b9e6ee5870cf81d39eadff9/js/packages/eyes-api/src/index.ts#L58-L59. Since `CheckSettings` is a fluent api class for users. And you want a plain object here. I wonder how it even compiles since you cannot extend an interface from a class.

js/packages/eyes-cypress/index.d.ts
```diff
@@ -9,24 +9,25 @@ type LegacyRegion = {left: number; top: number; width: number; height: number}
...
+  strict?: MaybeArray<CypressCheckSettings['strictRegions'] | LegacyRegion>
+  floating?: MaybeArray<CypressCheckSettings['floatingRegions']| (({element: Element} | Selector | LegacyRegion) & {maxUpOffset?: number; maxDownOffset?: number; maxLeftOffset?: number; maxRightOffset?: number})>
+  accessibility?: MaybeArray<CypressCheckSettings['accessibilityRegions'] | (({element: Element} | Selector | LegacyRegion) & {accessibilityType?: types.AccessibilityRegionType})>
 
-  scriptHooks?: CypressCheckSettings['hooks']
+  scriptHooks?: CypressCheckSettings['hook']
 }
 
-interface CypressEyesConfig extends types.EyesConfig<Element, Selector> {
+interface CypressEyesConfig extends api.Configuration<Element, Selector> {
```
kyryloonufriiev, line 30: Same here - `PlainConfiguration`
https://github.com/applitools/eyes.sdk.javascript1/blob/8a02b0f6b3cad7c13b9e6ee5870cf81d39eadff9/js/packages/eyes-api/src/index.ts#L60

js/packages/types/src/core-universal.ts
```diff
@@ -65,7 +65,7 @@ type UniversalEyes<TDriver, TElement, TSelector, TType extends 'classic' | 'ufg'
 >
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-export type Ref<TValue = never> = {'applitools-ref-id': string}
+export type Ref<TValue = never> = {'applitools-ref-id': string; type?: 'element' | 'selector' | 'context'}
```
SophieDePaula, line 68: change this to `string` instead of specific values.

js/packages/core/src/utils/to-base-check-settings.ts
```diff
@@ -45,7 +45,7 @@ export function toBaseCheckSettings<TElement, TSelector>({
         const {selector, regions} = calculatedRegions.shift()
         return regions.map(region => ({
           region,
-          regionId: utils.types.isString(selector) ? selector : selector?.selector,
+          regionId: reference.regionId ?? (utils.types.isString(selector) ? selector : selector?.selector),
           ...options,
```
kyryloonufriiev, line 49: This is shouldn't be needed because `options` will contain `regionId` and it will override the calculated `regionId`. I think you have a reference object formatted poorly in cypress. Can you verify if it matches this format `{region: Selector; regionId: string}`? I am under impression that you have `regionId` property smushed together with `Selector`

js/packages/eyes-cypress/test/fixtures/testApp/cypress/integration-run/appliConfFile.js
```diff
@@ -26,7 +26,7 @@ describe('Hello world', () => {
       let res = '';
       for (const result of summary.getAllResults()) {
         res +=
-          result.getTestResults()._results.name +
+          result.getTestResults()._result.name +
```
kyryloonufriiev, line 29: Why not use a public api here? `.getName()` or just `.name`?

js/packages/eyes-cypress/src/browser/eyesCheckMapping.js
```diff
@@ -36,14 +36,29 @@ function eyesCheckMapValues({args, refer}) {
       if (config.element) {
         if (isHTMLElement(config.element)) {
           regionSettings = {
-            region: refer.ref(config.element),
+            region: Object.assign(refer.ref(config.element), {type: 'element'}),
```
kyryloonufriiev, line 39: I believe this should go into the `refer` implementation. So you don't have to make this decision every time for every ref.

kyryloonufriiev (changes requested):

js/packages/core/src/utils/to-base-check-settings.ts
SophieDePaula: @kyryloonufriiev , In cypress the API is what you describe as "smushed together with selector":
`{
          type: 'css',
          selector: 'body > div:nth-child(3) > div:nth-child(3)',
          regionId: 'my-region-id',
        },
`
This worked before the redesign.
I don't see that I gave this specific example in our docs, so I am not sure if we should worry about breaking the current api.

js/packages/eyes-cypress/test/fixtures/testApp/cypress/integration-run/appliConfFile.js
SophieDePaula: good point

js/packages/eyes-cypress/src/browser/refer.js
```diff
@@ -14,16 +14,21 @@ class Refer {
...
-        value.ownerDocument ||
-        (value.constructor && value.constructor.name === 'Window'))
-    );
+    if (!value) return;
+    if (value.nodeType === Node.ELEMENT_NODE) {
+      return 'element';
+    } else if (
+      value.nodeType === Node.DOCUMENT_NODE ||
+      value.ownerDocument ||
+      (value.constructor && value.constructor.name === 'Window')
```
kyryloonufriiev, line 23: I beliave window is a driver and the document is a context

PR #1127, opened 2022-10-03
Add static webview method to CheckSettingsFluent

tourdedave:
(and drop AndroidVersion latest-2 from enums/data)


js/packages/eyes-api/src/input/CheckSettings.ts
```diff
@@ -89,6 +89,7 @@ export type Target<TElement, TSelector> = {
     scrollRootElement?: ElementReference<TElement, TSelector>,
   ): CheckSettingsFluent<TElement, TSelector>
   shadow(selector: SelectorReference<TSelector>): CheckSettingsFluent<TSelector>
+  webview(option: string | boolean | null): CheckSettingsFluent<TElement, TSelector>
```
kyryloonufriiev, line 92: It actually should be optional
```suggestion
  webview(option?: string | boolean): CheckSettingsFluent<TElement, TSelector>
```

tourdedave: Ah, I see

PR #1129, opened 2022-10-11
Support coded region in region

tourdedave:
[Trello](https://trello.com/c/Bhsdn014)

When using coded regions inside of a check region, the position coordinates that get calculated are relative to the top left of the page, not the top left of the region. This PR fixes that. It also consolidates the logic for converting regions to coordinates (using what is in screenshoter).

We have just 1 generic coverage test which happens to use a coded region inside of a region. It expects a value that is relative to the page. I've updated it in https://github.com/applitools/sdk.coverage.tests/pull/168


pending-changes.yaml
```diff
@@ -12,7 +12,7 @@ Apollo SDK:
   feature:
     - Drop support for Android 9 and Android 10 specific devices
   bug-fix:
-    - 
+    - Fix calculation for determining coded region positions when they are inside of a check region
```
tourdedave, line 15: @benbabayoff8 ☝️

benbabayoff8: looks good.

js/packages/core/src/automation/locate.ts
```diff
@@ -32,7 +32,7 @@ export function makeLocate<TDriver, TContext, TElement, TSelector>({
     }
     // TODO driver custom config
     const driver = await makeDriver({spec, driver: target, logger})
-    const screenshot = await takeScreenshot({driver, settings, logger})
+    const screenshot = await takeScreenshot({driver, settings, logger, regionsToCalculate: []})
```
kyryloonufriiev, line 35: I beliave this meant to be in settings

js/packages/core/src/automation/types.ts
```diff
@@ -62,6 +62,7 @@ export interface ScreenshotSettings<TElement, TSelector> extends BaseCore.ImageS
   waitBetweenStitches?: number
   lazyLoad?: boolean | {scrollLength?: number; waitingTime?: number; maxAmountToScroll?: number}
   webview?: boolean | string
+  regionsToCalculate?: []
```
kyryloonufriiev, line 65: It could not be here. Because this would mean that `eyes.check`, could also accept it

kyryloonufriiev (changes requested):
I've briefly reviewed it, without deep dive into the solution. Does this PR complete?

js/packages/screenshoter/src/extract-coodinates-for-selectors-and-elements.js
```diff
@@ -12,6 +24,12 @@ async function extractCoodinatesForSelectorsAndElements({regionsToCalculate, scr
         for (const element of elements) {
           const elementRegion = await element.getRegion()
           const region = utils.geometry.offset(elementRegion, contextLocationInViewport)
+          if (compensateCoordinatesRelativeToParentRegion) {
+            logger.log('coded region is inside of a region, compensating...')
+            logger.log('coded region before', region)
+            region.x = contextLocationInViewport.x
+            logger.log('coded region after', region)
+          }
```
kyryloonufriiev, line 32: How is it possible that only one coordinate has to be compensated?

js/packages/screenshoter/src/take-screenshot.js
```diff
@@ -92,7 +92,13 @@ async function takeScreenshot({
 
   screenshot.image.scale(driver.viewportScale)
 
-  const calculatedRegions = await extractCoodinatesForSelectorsAndElements({regionsToCalculate, screenshot, context})
+  const calculatedRegions = await extractCoordinatesForSelectorsAndElements({
+    compensateCoordinatesRelativeToParentRegion: !!region && driver.isWeb,
```
kyryloonufriiev, line 96: Isn't compensation also needed for native apps?

js/packages/core-base/src/types.ts
```diff
@@ -172,6 +172,8 @@ export interface ImageSettings<TRegion = Region> {
     scaleRatio?: number
   }
   debugImages?: {path: string; prefix?: string}
+  regionsToCalculate?: any[]
+
```
kyryloonufriiev, line 176: Having it here would also mean that it will go to `CheckSettings`. I meant that this property is not coming from the outside, and is always calculated internally. So it should be declared in an appropriate location. If `screenshoter` had ts this definitely would be the place, but today instead of `screenshoter` ts package we have this `take-screenshot.ts` file in the `core`, that wraps `screenshoter` package.

tourdedave: Yes, I know. This is where I added it. It was originally as a new argument to this function. Your recommendation was to consolidate it into settings. My intention was *not* to expose this property externally, but to address type/build errors I was getting. I'll try again.

js/packages/screenshoter/src/take-screenshot.js
tourdedave: Likely, yes. For the current issue this PR addresses, no. When opening it up to native apps there a few GCT failures that need to be reviewed to determine if the test needs to be updated or if the implementation is missing something.

js/packages/screenshoter/src/extract-coodinates-for-selectors-and-elements.js
tourdedave: It seems likely that `y` would also need to be included in this, but it's not something I tested for. I was focused on the user reported issue.

tourdedave:
> @tourdedave I'm sure, I wasn't sure if the PR is fully completed, but I already saw your comment on the trello. So I left a full review.

I took it out of draft, removed the `WIP` tag, and marked you as a reviewer. When I do that, it's complete (a.k.a. ready for review) -- although I reserve the word "complete" for when something has been reviewed/tested/released. So... 🤷‍♂️

kyryloonufriiev:
@tourdedave Yes, I agree. Sorry about that.

js/packages/screenshoter/src/take-screenshot.js
tourdedave: There were two native mobile GCT tests I think (I can't remember if there were others):

```
test/generic/AppiumIOSCheckRegionWithIgnoreRegion.spec.js
test/generic/AppiumAndroidCheckRegionWithIgnoreRegion.spec.js
```

For the first one, it seems like we already do the right thing. It's handled as part of the scaling logic in `extract-coordinates-for-selectors-and-elements`

```
driver    | 2022-10-24T16:48:57.091Z [INFO ] Extracted region { x: 131, y: 333, width: 113, height: 65 }
screenshoter| 2022-10-24T16:48:57.091Z [INFO ] scaled region { x: 115, y: 35, width: 113, height: 65 }
```

Makes me wonder why this isn't working for web as-is. Taking a closer look.

tourdedave: The main difference I see so far is that a region screenshot on native mobile uses `takeSimpleScreenshot` and web uses `takeStitchedScreenshot`. For the web test I'm using (`core/test/e2e/classic/coded-regions.spec.ts`) it looks like the crop region that's calculated does not contain the correct `x`/`y` coordinates. If that's true, then I'd rather fix that then add compensation. I'll keep digging.

js/packages/screenshoter/src/extract-coodinates-for-selectors-and-elements.js
kyryloonufriiev: I would really prefer to spend more time to have a proper solution for the problem, not just a solution for the card.

tourdedave: We don't need this new compensation. The issue was coming from `calculate-screenshot-region.js`.

js/packages/screenshoter/src/take-screenshot.js
tourdedave: The issue was coming from `calculate-screenshot-region.js`. Fixed in https://github.com/applitools/eyes.sdk.javascript1/pull/1129/commits/e1f2081ab62b978ee845fdeb4e37adec4a6cd879

kyryloonufriiev (approved):
I'm really glad we get to the bottom of the issue.

tourdedave:
I'll wait until https://github.com/applitools/sdk.coverage.tests/pull/168 is approved before merging, since they are intertwined.

PR #1131, opened 2022-10-13
Get `contentSize` without `Helper` in `IOS`

zamboney:
On the older version of the SDK, once there was no `Helper`, there was a fallback for calculating the size of the element ([ref](https://github.com/applitools/eyes.sdk.javascript1/blob/9c6cb79c7eb048a04e64a09c5ec1c7173aaeaa9a/packages/driver/src/element.ts#L169-L179)).
This PR just restores this and added a test for `IOS` to make sure that if there is no `Helper` the `contentSize` can be measured

[Trello ticket](https://trello.com/c/oIjiiQVb/1584-appium-java-universal-ios-cannot-take-full-screen-screenshot)


js/packages/driver/src/element.ts
```diff
@@ -205,6 +188,38 @@ export class Element<TDriver, TContext, TElement, TSelector> {
...
+        })
+        const childElementRegion = await this._spec.getElementRegion(this.driver.target, childElement.target)
+        return {
+          ...elementRegion,
+          height: childElementRegion.y + childElementRegion.height - elementRegion.y,
+        }
+      }
+      default:
+        return this._spec.getElementRegion(this.driver.target, this.target)
+    }
```
kyryloonufriiev, line 221: What is the point of `switch` here if only one case exists? So it should be `if` here.

js/packages/driver/src/element.ts
```diff
@@ -205,6 +188,38 @@ export class Element<TDriver, TContext, TElement, TSelector> {
...
+    } catch (err) {
+      this._logger.warn(`Unable to get the attribute 'contentSize' due to the following error: '${err.message}'`)
+    }
+    switch (await this.getAttribute('type')) {
+      case 'XCUIElementTypeScrollView': {
+        const elementRegion = await this._spec.getElementRegion(this.driver.target, this.target)
+        const [childElement] = await this.driver.elements({
+          type: 'xpath',
+          selector: '//XCUIElementTypeScrollView[1]/*', // We cannot be sure that our element is the first one
+        })
```
kyryloonufriiev, line 212: Can you explain this comment better?

kyryloonufriiev (changes requested):

js/packages/driver/src/element.ts
zamboney: Np will do

js/packages/driver/src/element.ts
zamboney: The layout is arranged like it and this is how you need to get it.

Also this is how it's done in the [past](https://github.com/applitools/eyes.sdk.javascript1/blob/9c6cb79c7eb048a04e64a09c5ec1c7173aaeaa9a/packages/driver/src/element.ts#L173)

kyryloonufriiev: I was asking about the comment in the code. I also don't understand in what exact case it should work. I know that this was implemented like this in Java, but I want to understand what is so special about this case. And couldn't it be solved in a more flexible way?

PR #1138, opened 2022-10-23
Change the `execPath` to support `universal-server` in `cypress@6`

zamboney:
`cypress@6` and lower is bundled with `Electron` version (11.3.0) that [not support `async` shell execution](https://www.electronjs.org/releases/stable?version=12&page=6#:~:text=Added%20a%20new%20asynchronous%20shell.trashItem()%20API%2C%20replacing%20the%20synchronous%20shell.moveItemToTrash().%20%2325114). This fix will change change the `execPath` of the `fork` `eyes-universal` process to the current running version of `node` in the OS so that the `eyes-universal` server can be run on the background.

This isn't an ideal fix but keep in mind that the alternative is that `cypress@6` and below will not work with `eyes-universal`

[Trello card](https://trello.com/c/cIzRoAkV/1523-cypress-691-cant-start-universal-server)


js/packages/eyes-universal/src/universal-server-process.ts
```diff
@@ -1,14 +1,15 @@
...
 export function makeServerProcess(
-  options: ServerOptions & {detached?: boolean},
+  options: ServerOptions,
+  clientForkOptions?: ForkOptions,
 ): Promise<{port: number; close: () => void}> {
   return new Promise((resolve, reject) => {
     const server = fork(path.resolve(__dirname, '../dist/cli.js'), [`--config=${JSON.stringify(options)}`], {
-      detached: options.detached ?? true,
       stdio: [options.shutdownMode === 'stdin' ? 'inherit' : 'ignore', 'ignore', 'ignore', 'ipc'],
+      ...(clientForkOptions || {}),
```
kyryloonufriiev, line 12:
```suggestion
      ...(clientForkOptions ?? {}),
```

js/packages/eyes-cypress/src/plugin/server.js
```diff
@@ -30,13 +32,30 @@ function makeStartServer({logger}) {
...
+    }
+
+    const {port: universalPort, close: closeUniversalServer} = await makeServerProcess(
+      {
+        key: path.resolve(__dirname, '../pem/server.key'),
+        cert: path.resolve(__dirname, '../pem/server.cert'),
+        idleTimeout: 0,
+        shutdownMode: 'stdin',
+      },
+      forkOptions,
```
kyryloonufriiev, line 57: Not a fun double-object argument. `makeServerProcess` should accept only one argument, what you actually want is to add a sub-object - `forkOptions`.

js/packages/eyes-cypress/src/plugin/server.js
```diff
@@ -30,13 +32,30 @@ function makeStartServer({logger}) {
...
-      detached: false,
-      idleTimeout: 0,
-      shutdownMode: 'stdin',
-    });
+    const forkOptions = {
+      detached: true,
+    };
+
+    // taking the version of Cypress from the `execPath`
+    const cypressVersionMatch = process.execPath.match(/Cypress\/(\d+\.\d+\.\d+)/);
```
kyryloonufriiev, line 40: Why we cannot detect the cypress version through a package.json?

kyryloonufriiev (changes requested):

js/packages/eyes-cypress/src/plugin/server.js
zamboney: i thought it more clear to the user that there are custom object for the `makeServerProcess` and a known object for `child_process.fork` if you disagree, then i will put it as part of the custom `makeServerProcess` option. Let me know what you think.

js/packages/eyes-cypress/src/plugin/server.js
kyryloonufriiev: The standard way to detect the version of the framework is `require('cypress/package.json').version`. Wouldn't it work here?

zamboney: yes good point, i will change it

js/packages/eyes-cypress/src/plugin/server.js
kyryloonufriiev: I tend to disagree. If the first argument is an object, this is almost always the only argument. Passing the second object makes no sense to me. There is one argument called options, that could have any parameters (or you cal call it named arguments) you need. I'm not even talking that this will be more aesthetically pleasing since the linter will not make you to create those pointless lines with only one breaket.

js/packages/eyes-cypress/src/plugin/server.js
```diff
@@ -4,8 +4,10 @@ const {makeServerProcess} = require('@applitools/eyes-universal');
 const handleTestResults = require('./handleTestResults');
 const path = require('path');
 const fs = require('fs');
+const semverLt = require('semver/functions/lt');
 const {Server: HttpsServer} = require('https');
 const {Server: WSServer} = require('ws');
+const {sync: whichSync} = require('which');
```
kyryloonufriiev, line 10: BTW, why not use the async version here?

js/packages/eyes-universal/src/universal-server.ts
```diff
@@ -23,6 +24,7 @@ export type ServerOptions = HandlerOptions & {
   shutdownMode?: 'lazy' | 'stdin'
   idleTimeout?: number
   printStdout?: boolean
+  forkOptions?: ForkOptions
```
kyryloonufriiev, line 27: This should not go here. Because this is not part of the ServerOptions. We don't even have a way to use it here.

js/packages/eyes-universal/src/universal-server-process.ts
```diff
@@ -2,13 +2,13 @@ import {type ServerOptions} from './universal-server'
 import {fork} from 'child_process'
 import path from 'path'
 
-export function makeServerProcess(
-  options: ServerOptions & {detached?: boolean},
-): Promise<{port: number; close: () => void}> {
+export function makeServerProcess(options: ServerOptions): Promise<{port: number; close: () => void}> {
```
kyryloonufriiev, line 5: Here is where `forkOptions?: ForkOptions` should go. You can create a type called `ServerInProcessOptions` here, or just simply use inline decl in arguments - `ServerOptions & {forkOptions?: ForkOptions}`

kyryloonufriiev (changes requested):

js/packages/eyes-cypress/src/plugin/server.js
zamboney: i will use the `async` version

PR #1140, opened 2022-10-25
[browser-extension] Coverage tests issues fix 

zamboney:
This PR contains 

- fix issues that found in the coverage tests
- enable coverages tests in `gh action`

### DIFF
--- a/.github/workflows/publish-browser-extension.yml
+++ b/.github/workflows/publish-browser-extension.yml
@@ -67,12 +67,10 @@ jobs:
         run: |
           sudo apt-get install -y xvfb
       - name: Run tests
-        shell: bash
-        working-directory: ./js/packages/eyes-browser-extension
-        run: |
-          yarn install
-          yarn build
-          xvfb-run --auto-servernum -- yarn test
+        uses: ./.github/actions/tests
+        with:
+          working-directory: ./js/packages/eyes-browser-extension
+          test-type: xvfb
   publish:
     needs: [test]
     name: Publish
--- a/.github/workflows/publish-browser-extension.yml
+++ b/.github/workflows/publish-browser-extension.yml
@@ -63,13 +63,16 @@ jobs:
           always-auth: true
       - name: Setup xvfb
         shell: bash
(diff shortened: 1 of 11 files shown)


.github/workflows/publish-browser-extension.yml
```diff
@@ -63,13 +63,16 @@ jobs:
...
       - name: Run tests
-        uses: ./.github/actions/tests
-        with:
-          working-directory: ./js/packages/eyes-browser-extension
+        shell: bash
+        working-directory: ./js/packages/eyes-browser-extension
+        run: |
+          yarn install
+          yarn build
+          xvfb-run --auto-servernum -- yarn test
```
kyryloonufriiev, line 75: We should use this action to run tests - `./.github/actions/tests`

js/packages/snippets/src/browser/hits.js
```diff
@@ -0,0 +1,5 @@
+function hits([element]) {
+  return element.innerText
+}
+module.exports = hits
```
kyryloonufriiev, line 4: hits? What does this name mean?

kyryloonufriiev (changes requested):
We should have a way to test extension with test action, this is for sure.
About the loader for snippets, it might be necessary, but I hope it is not, so I'm investigating it right now.

js/packages/eyes-browser-extension/package.json
```diff
@@ -21,6 +21,7 @@
     "build:dist": "webpack",
     "build:zip": "cd dist; zip -r eyes-browser-extension.zip *; mv eyes-browser-extension.zip ../",
     "generate:tests": "coverage-tests generate ./test/coverage/config.js --name 'eyes-browser-extension'",
+    "test:xvfb": "xvfb-run --auto-servernum -- yarn test",
```
kyryloonufriiev, line 24: nice!

kyryloonufriiev (approved):
It looks like the snippet manipulation is unavoidable. So you can merge this PR.

PR #1142, opened 2022-10-27
[Storybook] add support for StoryStore v7

SophieDePaula:
[Trello card](https://trello.com/c/lATzyeIA/81-storybook-support-storestoryv7)

TODO:
- [x] all integration tests pass
- [x] all unit tests pass
- [x] all e2e tests pass


js/packages/eyes-storybook/src/browser/getStories.js
```diff
@@ -104,9 +104,11 @@ async function getStories({timeout = DEFAULT_TIMEOUT} = {timeout: DEFAULT_TIMEOU
     }
   }
 
-  function getStoriesThroughClientAPI(clientApi) {
-    return clientApi.getStories().map((story, index) => {
+  async function getStoriesThroughClientAPI(clientApi) {
+    const stories = await clientApi.getStories()
+    return stories.map((story, index) => {
       const {name, kind, parameters} = story;
+      const id = story.id
```
SophieDePaula, line 111: I get the `id` separately from the rest of the properties above, as I am not sure if `id` is mandatory for all previous storybook versions.

js/packages/eyes-storybook/src/browser/getClientAPI.js
```diff
@@ -95,12 +95,20 @@ function getClientAPI() {
...
+              if (clientAPI.storyStore.cacheAllCSFFiles) {
+                await clientAPI.storyStore.cacheAllCSFFiles();
+              }
               return clientAPI.raw();
             },
-            selectStory: async i => {
+            selectStory: async (i, id) => {
+              let storyId  = !clientAPI.storyStore.cacheAllCSFFiles ? clientAPI.raw()[i].id : id
+              if(!storyId){
+                await clientAPI.storyStore.cacheAllCSFFiles();
```
SophieDePaula, line 107: if for some reason we don't have the `id`, then we'll cache all stories and retrieve the story the "old" way. However, if we do have the `id` already, we can improve the performance and not cache all stories.

madiyetov:
Hi @SophieDePaula 👋🏼 , what's the current status of this PR?

js/packages/core/src/make-manager.ts
```diff
@@ -45,23 +45,26 @@ export function makeMakeManager<TDriver, TContext, TElement, TSelector>({
...
-        },
-        abort(options) {
-          const promise = eyes.abort(options)
-          item.promise ??= promise
-          return promise
-        },
-      })
-    })
+    const openEyes = utils.general.wrap(
+      makeOpenEyes({spec, core: typedCore, logger, generatedBatchId: `generated-${utils.general.guid()}`}),
```
kyryloonufriiev, line 49: I would propose not inventing new properties, but simply passing the `batch` object with the `id` property in the `makeOpenEyes`. So this batch will be used if a custom one was not specified.

js/packages/core/src/open-eyes.ts
```diff
@@ -18,15 +18,15 @@ type Options<TDriver, TContext, TElement, TSelector> = {
   core?: BaseCore | ClassicCore<TDriver, TElement, TSelector> | UFGCore<TDriver, TElement, TSelector>
   concurrency?: number
   logger?: Logger
-  generatedBatchId?: string
+  generatedBatch?: Batch
```
kyryloonufriiev, line 21: call it just `batch`, we don't care at this point if it is generated or not, it is just a butch for all eyes that will be open with the created function

js/packages/core/src/open-eyes.ts
```diff
@@ -46,7 +46,7 @@ export function makeOpenEyes<TDriver, TContext, TElement, TSelector>({
     settings.serverUrl ??= utils.general.getEnvValue('SERVER_URL') ?? 'https://eyesapi.applitools.com'
     settings.apiKey ??= utils.general.getEnvValue('API_KEY')
     settings.batch ??= {}
-    settings.batch.id ??= utils.general.getEnvValue('BATCH_ID') ?? generatedBatchId
+    settings.batch.id ??= utils.general.getEnvValue('BATCH_ID') ?? generatedBatch.id
```
kyryloonufriiev, line 49: - Since `batch` (`generatedBatch`) could have also other default fields we should merge it with the batch we got from the settings
- Since `batch` (`generatedBatch`) is an optional property we should preserve guide generation.

I recommend this 👇🏼 
```
settings.batch = {...batch, ...settings.batch, properties: [...(batch?.properties ?? []), ...(settings.batch?.properties ?? [])]}
settings.batch.id ??= utils.general.getEnvValue('BATCH_ID') ?? `generated-${utils.general.guid()}`
```

SophieDePaula:
[Heavy storybook run ](https://eyes.applitools.com/app/test-results/00000251731684617407/?accountId=sQgSIcxAzUyaYiw7opq8lw~~) (this is under my account so lmk if you need access)

SophieDePaula:
Heavy storybook:

[Current Branch](https://eyes.applitools.com/app/test-results/00000251731401397934/?accountId=UAujt6tHnEKUivQXIz7G6A~~) - 39:23
[Master](https://eyes.applitools.com/app/test-results/00000251731405347428/?accountId=UAujt6tHnEKUivQXIz7G6A~~) = 39:53

@kyryloonufriiev fyi

js/packages/core/src/ufg/utils/take-dom-snapshots.ts
```diff
@@ -37,12 +37,19 @@ export async function takeDomSnapshots<TDriver extends Driver<unknown, unknown,
...
+      await utils.general.sleep(settings.waitBeforeCapture)
+    }
+  }
   await hooks?.beforeSnapshots?.()
 
   if (!settings.layoutBreakpoints) {
     logger.log(`taking single dom snapshot`)
     await hooks?.beforeEachSnapshot?.()
-    await utils.general.sleep(settings.waitBeforeCapture)
+    await waitBeforeCapture()
```
kyryloonufriiev, line 52: Isn't it possible to simply use the hook here?

js/packages/eyes-storybook/src/browser/getStories.js
```diff
@@ -104,9 +104,11 @@ async function getStories({timeout = DEFAULT_TIMEOUT} = {timeout: DEFAULT_TIMEOU
     }
   }
 
-  function getStoriesThroughClientAPI(clientApi) {
-    return clientApi.getStories().map((story, index) => {
+  async function getStoriesThroughClientAPI(clientApi) {
+    const stories = await clientApi.getStories();
+    return stories.map((story, index) => {
       const {name, kind, parameters} = story;
+      const id = story.id;
```
kyryloonufriiev, line 111: Why not destruct it the same way as other props?

js/packages/eyes-storybook/src/eyesStorybook.js
```diff
@@ -48,26 +57,48 @@ async function eyesStorybook({
...
+    logger,
+    agentId,
+  });
+  const manager = await core.makeManager({
+    spec,
+    concurrency: testConcurrency,
+    agentId,
+    logger,
+    core,
+    legacy: false,
```
kyryloonufriiev, line 76: there is no such property

js/packages/eyes-storybook/src/eyesStorybook.js
```diff
@@ -48,26 +57,48 @@ async function eyesStorybook({
...
+      proxy,
+      concurrency: testConcurrency,
+    },
+    logger,
   });
 
+  const provides = {
+    getChromeEmulationDevices: client.getChromeEmulationDevices,
+    getIOSDevices: client.getIOSDevices,
+  };
```
kyryloonufriiev, line 100: This is not a big deal, but I would prefer us to not create variables when we don't need them. Just because such a code will be much harder to read later. If I will look at the call to `takeDomSnapshots` I will see that it uses your variable, and then I will have to scroll up to see what this variable means, on my opinion it will be much more  easy to read if it was simply used where it is needed this way we can avoid jumping up and down while reading the code. I hope you agree so please keep it in mind

js/packages/eyes-storybook/src/generateConfig.js
```diff
@@ -44,12 +44,35 @@ function generateConfig({argv = {}, defaultConfig = {}, externalConfigParams = [
...
+    result.batch = {id: result.batchId, ...result.batch};
   }
 
   if (result.storyDataGap === undefined) {
-    result.storyDataGap = result.testConcurrency;
+    if (result.testConcurrency * 2 >= MAX_DATA_GAP) {
+      result.storyDataGap = MAX_DATA_GAP;
+    } else {
+      result.storyDataGap = result.testConcurrency * 2;
+    }
```
kyryloonufriiev, line 75: What if I have `61` concurrency?
During our discussion, I was envisioning something like this 
```suggestion
    result.storyDataGap = Math.max(Math.min(result.testConcurrency * 2, MAX_DATA_GAP), result.testConcurrency);
```
or
```suggestion
    result.storyDataGap = Math.min(result.testConcurrency * 2, Math.max(MAX_DATA_GAP, result.testConcurrency));
```

js/packages/eyes-storybook/src/processResults.js
```diff
@@ -124,16 +124,14 @@ function processResults({results = [], totalTime, testConcurrency, saveNewTests
...
-  const sortedTestResults = results.sort((a, b) => a.getName().localeCompare(b.getName()));
+  const sortedTestResults = results.sort((a, b) => a.result.name.localeCompare(b.result.name));
   sortedTestResults.forEach(result => {
-    const storyTitle = `${result.getName()} [${result.getHostApp()}] [${result
-      .getHostDisplaySize()
-      .toString()}] - `;
+    const storyTitle = `${result.result.name} [${result.result.hostApp}] [${result.result.hostDisplaySize.width}x${result.result.hostDisplaySize.height}] - `;
 
-    if (result.getIsNew()) {
+    if (result.result.isNew) {
```
kyryloonufriiev, line 131: Why is that `result.result`?

js/packages/eyes-storybook/src/renderStory.js
```diff
@@ -62,39 +79,70 @@ function makeRenderStory({logger, testWindow, performance, timeItAsync}) {
...
-      return testWindow({openParams, checkParams, throwEx: false});
+      // need to check what open expects and what I send
+      const {checkAndClose} = await openEyes({settings: openParams});
+      return new Promise((resolve, reject) => {
+        throttle(async () => {
+          const checkResults = await checkAndClose({
+            settings: {...checkParams, throwEx: false},
+            target: snapshots,
+          });
+          resolve(checkResults);
```
kyryloonufriiev, line 140: Does it mean that if `checkAndClose` will throw the promise will just hand?

js/packages/eyes-storybook/src/renderStory.js
```diff
@@ -62,39 +79,70 @@ function makeRenderStory({logger, testWindow, performance, timeItAsync}) {
...
+        });
+      });
     }).then(onDoneStory);
 
-    function onDoneStory(results) {
+    async function onDoneStory(results) {
```
kyryloonufriiev, line 145: why?

js/packages/eyes-storybook/test/e2e/TestAccessibility.e2e.test.js
```diff
@@ -53,17 +53,12 @@ describe('eyes-storybook accessibility', () => {
     });
 
     const expectedTitles = ['Single category: Story with local accessibility region'];
-    expect(results.map(e => e.title).sort()).to.eql(expectedTitles.sort());
-    results = flatten(results.map(r => r.resultsOrErr));
+    expect(results.results.map(e => e.title).sort()).to.eql(expectedTitles.sort());
+    results = flatten(results.results.map(r => r.resultsOrErr));
```
kyryloonufriiev, line 57: you can simply use `flatMap` here.

js/packages/eyes-storybook/src/renderStory.js
```diff
@@ -62,39 +79,70 @@ function makeRenderStory({logger, testWindow, performance, timeItAsync}) {
...
-      snapshot,
-      ignore: ignoreRegionsBackCompat,
-      floating: floatingRegions,
-      layout: layoutRegions,
-      strict: strictRegions,
-      content: contentRegions,
-      accessibility: accessibilityRegions,
-      scriptHooks,
+      ignoreRegions: ignoreRegionsBackCompat,
+      floatingRegions,
```
kyryloonufriiev, line 95: I think floating regions are also require some transformation

js/packages/core/src/ufg/utils/take-dom-snapshots.ts
SophieDePaula: @kyryloonufriiev because I would need to check if `waitBeforeCapture` is a `number` or `function` and I didn't want to write the same condition multiple times so I wrote the condition only one time at the start of the file.

js/packages/eyes-storybook/src/browser/getStories.js
SophieDePaula: I can, but I am unsure if `id` is a mandatory property in all versions of `storybook` and that's the reason I decided to separate it.

js/packages/eyes-storybook/src/eyesStorybook.js
SophieDePaula: Makes sense

js/packages/eyes-storybook/src/processResults.js
SophieDePaula: I pass an object that includes `results` and `TestResultsSummary` to `processResults`. So `result.result` is the `object` that we process in `renderStories`

js/packages/eyes-storybook/src/renderStory.js
SophieDePaula: I added `try` and `catch` blocks with `reject`, thanks for catching this one

js/packages/eyes-storybook/src/renderStory.js
SophieDePaula: Forgot to remove this one

js/packages/eyes-storybook/src/browser/getStories.js
kyryloonufriiev: I don't understand how is that different. If there is not id it simply will be undefined the same as in your case

kyryloonufriiev (approved):

pending-changes.yaml
```diff
@@ -108,6 +108,7 @@ Apollo SDK:
 "@applitools/eyes-storybook":
   feature:
     - Deprecated "Content" match level value in favor of "IgnoreColors"
+    - Add Support storyStoreV7 flag
```
SophieDePaula, line 111: @benbabayoff8

pending-changes.yaml
```diff
@@ -108,6 +108,7 @@ Apollo SDK:
 "@applitools/eyes-storybook":
   feature:
     - Deprecated "Content" match level value in favor of "IgnoreColors"
+    - Add Support for storyStoreV7 flag
```
SophieDePaula, line 111: @benbabayoff8

madiyetov:
@SophieDePaula, as there is no Issues section, writing here:
`v3.30.0` brought an error which happens only with passed config for mobile:
```
browsers: [
      {
        deviceName: 'iPhone 11',
        deviceScaleFactor: 1,
        mobile: true,
      },
    ]
```
The error: `Error in executeRenders: resultsOrErr.every is not a function`. It's failing in [didTestPass](https://github.com/applitools/eyes.sdk.javascript1/blob/fe208449189e6b759c7fdd495589097a7e939260/js/packages/eyes-storybook/src/renderStories.js#L129) function

kyryloonufriiev:
@madiyetov you can try to pass this in a different way

```
browsers: [
      {
        chromeEmulationInfo: {
            deviceName: 'iPhone 11',
            deviceScaleFactor: 1,
            mobile: true,
        }
      },
    ]
```

PR #1144, opened 2022-10-31
Fix webview check

tourdedave:
re: [Trello](https://trello.com/c/vDJEIGCx)

Even though [the contract states that an Array of strings should be returned](https://appium.io/docs/en/commands/context/get-contexts/#response) when fetching the app contexts (a.k.a. worlds) the Java Appium client returns an array of objects (!). This causes the driver to misclassify the state of the driver (e.g., it thinks it is in a webview, so it thinks it is running in a web context when it is actually running in a native app context). This causes the SDK to attempt to run JS snippets, causing the SDK to terminate with `Method is not implemented` errors.

This PR adds support for this additional return type.


kyryloonufriiev (changes requested):
I actually already saw this. This is how we should fix it, on the level of the spec driver. To not create an ambiguous return value type - https://github.com/applitools/eyes.sdk.javascript1/blob/cf6aeaa68f592d8998fa6ff53189fc419848d287/js/packages/spec-driver-webdriverio/src/spec-driver.ts#L379-L382

tourdedave:
> I actually already saw this. This is how we should fix it, on the level of the spec driver. To not create an ambiguous return value type -

If that's true, then shouldn't the change be in the spec driver in eyes-universal (instead of sd-wdio)?

https://github.com/applitools/eyes.sdk.javascript1/blob/master/js/packages/eyes-universal/src/spec-driver/webdriver.ts#L330-L332

kyryloonufriiev:
> Oh I see what you meant. You've already ran into this issue and solved it in sd-wdio. Begs the question - why are we duplicating our effort?

I faced this issue when I was doing something with wdio spec driver, so I wasn't able to build it, because of changes in wdio types, didn't give much thought to the cause of the change in types. Looks like the reason is appium, not just a framework decision to change the return value format.

kyryloonufriiev (approved):
💯

PR #1145, opened 2022-11-01
[driver] Add fallback and child selectors support

kyryloonufriiev:
[Trello](https://trello.com/c/ZYQdtJQN)

New selector formats:
A selector to look for elements within other elements (child selector):
```
{
  "type": "id",
  "selector": "my_id",
  "child": {
    "type": "class name",
    "selector": "my_class_name",
    "child": {
      "type": "xpath",
      "selector: "//my/xpath" 
    }
  }
}
```
A selector that looks for elements with fallback if they were not found (fallback selector):
```
{
  "type": "id",
  "selector": "my_id",
  "fallback": {
    "type": "class name",
    "selector": "my_class_name",
    "fallback": {
      "type": "xpath",
      "selector: "//my/xpath" 
    }
  }
}
```

Just an example of how a common selector could look like with all new capabilities:
```
{
  selector: 'frame1',
  frame: {
    selector: 'frame1-shadow1',
    shadow: {
      selector: 'frame1-shadow1-child1',
      child: {
        selector: 'frame1-shadow1-child1-frame1',
        frame: {
          selector: 'frame1-shadow1-child1-child1',
          fallback: {
            selector: 'frame1-shadow1-child1-frame1',
            frame: 'frame1-shadow1-child1-frame1-child1'
          }
        },
        fallback: 'frame1-shadow1-child1-child1'
      },
      fallback: {
        selector: 'frame1-shadow1-shadow1',
        shadow: {
          selector: 'frame1-shadow1-shadow1-child1',
          fallback: 'frame1-shadow1-shadow1-child2'
        },
        fallback: 'frame1-shadow1-child1'
      }
    },
    fallback: {
      selector: 'frame1-shadow2',
      shadow: {
        selector: 'frame1-shadow1-child1',
        child: {
          selector: 'frame1-shadow1-child1-frame1',
          frame: {
            selector: 'frame1-shadow1-child1-child1',
            fallback: {
              selector: 'frame1-shadow1-child1-frame1',
              frame: 'frame1-shadow1-child1-frame1-child1'
            }
          },
          fallback: 'frame1-shadow1-child1-child1'
        },
      },
    }
  },
  fallback: 'child1'
}
```

Things to add/check:

- [ ] Add test for multiple elements lookup
- [ ] Check how testcafe behaves after dropping shadow selector transformation


js/packages/driver/src/context.ts
```diff
@@ -137,6 +137,81 @@ export class Context<TDriver, TContext, TElement, TSelector> {
...
+            parent,
+          )
+          if (element) {
+            clearTimeout(timeout)
+            elements = [element]
+          }
+          await utils.general.sleep(wait.interval)
+        }
+      }
+    } else if (all) {
```
tourdedave, line 175: Do we want to support when `wait` && `all` are both provided?

tourdedave (changes requested):
Looks good overall. But how are we planning to test this e2e? Is it just for Java or are other languages include too (like dotnet)? A GCT that's just for affected languages maybe? Let me know what you think.

Also, you're missing a pending changes entry.

js/packages/driver/src/context.ts
kyryloonufriiev: `waitFor` method always returns only one element.

kyryloonufriiev:
I don't think the generic test is worth the effort if it is available only in Java.

tourdedave:
> I don't think the generic test is worth the effort if it is available only in Java.

Okay, but are they writing a test to verify that this works?

js/packages/driver/src/context.ts
tourdedave: What `waitFor` method? Do you mean `waitForSelector`? I don't see it implemented on any of the specs, so the fallback condition is likely used, which is a hand-rolled wait that, and yes, it returns just one element but I don't see how that's a relevant detail to what I was thinking. Isn't there a case where we would need to wait for a selector to be present before performing a `findElements`?

IdosApplitools:
> > I don't think the generic test is worth the effort if it is available only in Java.
> 
> Okay, but are they (@IdosApplitools & @danielputerman) writing a test to verify that this works?

Yes, I am currently working on adding e2e tests for that on our end :)

PR #1147, opened 2022-11-03
Handle reduced user agent

kyryloonufriiev:


SophieDePaula (commented):
The changes around the `userAgent` vs the client hints make sense to me.
One question, why did you change executePoll?
Also, it would have helped me if you would have added a description of what you changed and why (even a few words about every change would have been helpful).

tourdedave (changes requested):
+1 on @SophieDePaula's comments.

Also, I'm surprised to see that there isn't test coverage in snippets for `getBrowserInfo` to cover the different browser/OS combinations and each of the code paths.

kyryloonufriiev:
> One question, why did you change executePoll?
@SophieDePaula Because I need this to get browser info, so since we don't really want to implement `executeAsync` (which would mean a new spec method), our polling interface should be standardized in the driver package, and this is what I did.

js/packages/driver/src/context.ts
```diff
@@ -405,6 +405,57 @@ export class Context<TDriver, TContext, TElement, TSelector> {
...
+        } else if (response.status === 'SUCCESS') {
+          return response.value
+        } else if (response.status === 'SUCCESS_CHUNKED') {
+          chunks += response.value
+          if (response.done) return deserialize(chunks)
+        } else if (response.status === 'WIP') {
+          await utils.general.sleep(pollTimeout)
+        }
+        this._logger.log('Polling...')
+        response = deserialize(await this.execute(pollScript, pollArg))
```
SophieDePaula, line 439: @kyryloonufriiev what happens if no `pollScript` is sent to `executePoll` like with `lazyLoading` and the `executionTimeOut` is not up?

SophieDePaula (commented):
@kyryloonufriiev I left one comment, other than that, it looks good to me.

js/packages/driver/src/context.ts
kyryloonufriiev: You can see it from the implementation, you specified just one script which will be the user from both operations start and polling. Previously you always had to provide two scripts implicitly so in the case of lazyLoad we had to pass the same script for both operations. In huge scripts like dom-snapshot it made sense to create a separate script for polling to make requests smaller. But as we saw with Dave while he was implementing the lazy load feature for small scripts it makes no sense to do something like that.

SophieDePaula (approved):

pending-changes.yaml
```diff
@@ -10,14 +10,15 @@ Apollo SDK:
       An example of how to inline `markdown code highlights`
 "@applitools/eyes-universal":
   feature:
-    - 
+    - Use user agent metadata to improve browser environment detection logic
```
kyryloonufriiev, line 13: @benbabayoff8

benbabayoff8: @kyryloonufriiev Looks good.

PR #1149, opened 2022-11-04
Fix/appium full page classic

tourdedave:
[Trello](https://trello.com/c/xPatADrO)


js/packages/driver/src/element.ts
```diff
@@ -196,9 +196,7 @@ export class Element<TDriver, TContext, TElement, TSelector> {
         x: contentSize.left,
         y: contentSize.top,
         width: contentSize.width,
-        height: this.driver.isIOS
-          ? Math.max(contentSize.height, contentSize.scrollableOffset)
-          : contentSize.height + contentSize.scrollableOffset,
+        height: contentSize.height + contentSize.scrollableOffset,
```
kyryloonufriiev, line 199: The previous implementation definitely was there because `scrollableOffset` in iOS is the whole scroll height, not just an offset. Maybe it did change in the newer appium (or in some other scenario), we should verify that.

kyryloonufriiev (changes requested):

js/packages/driver/src/element.ts
```diff
@@ -562,6 +560,7 @@ export class Element<TDriver, TContext, TElement, TSelector> {
           y: scrollableRegion.y - actualScrollableRegion.y,
         })
 
+        if (this.driver.isIOS) await new Promise(res => setTimeout(res, 2000)) // to allow the scrollbar to dissipate
```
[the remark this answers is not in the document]
tourdedave: @kyryloonufriiev Okay. Two things.

1)
I see the property, but I don't see how it's used. Is it supposed to be passed to `takeScreenshot`? If so, then `wait` is the argument on `takeScreenshot` & `takeStitchedScreenshot` that fits. So this is currently not working, and we don't have a test for it that I can see. I'm happy to fix it, just want to make sure I'm not misunderstanding.

2)
There is a sensible default of `100` ms. If I want to set a different sensible default for native iOS, is check aware of this state?

js/packages/driver/src/element.ts
tourdedave: @kyryloonufriiev Okay, was there a specific test that verified this?

kyryloonufriiev: There was no specific test. Because this was the behavior of appium that definitely existed, and as far as I understand all ios tests (that do not involve helper lib) were testing this implicitly.

js/packages/driver/src/element.ts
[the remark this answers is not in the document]
kyryloonufriiev: 1. I think at some point I wrongly use `waitBeforeCapture` instead `waitBetweenStitches` (it called `waitBeforeScreenshots` before redesign) to init `wait` prop of the takeScreenshot. You can fix that. There is no test for that. Would be nice to refactor it a bit and make screenshoter accept both (`waitBeforeCapture` and `waitBetweenStitches`) instead of `wait` and use them in the proper places/moments.
2. If we are going to set a different default for ios, it should happen in classic check, we have the driver object there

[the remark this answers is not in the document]
kyryloonufriiev: 1. I think at some point I wrongly use `waitBeforeCapture` instead `waitBetweenStitches` (it called `waitBeforeScreenshots` before redesign) to init `wait` prop of the takeScreenshot. You can fix that. There is no test for that. Would be nice to refactor it a bit and make screenshoter accept both (`waitBeforeCapture` and `waitBetweenStitches`) instead of `wait` and use them in the proper places/moments.
2. If we are going to set a different default for ios, it should happen in classic check, we have the driver object there

js/packages/driver/src/element.ts
tourdedave: I didn't run all of the iOS screenshoter tests. I'll run them to see.

tourdedave: Yeah, it looks like the following tests implicitly check this (and they fail with my change):

- `test/e2e/ios/app-fully-collection.spec.js`
- `test/e2e/ios/app-fully-collapsing.spec.js`
- `test/e2e/ios/app-fully-overlapped.spec.js`
- `test/e2e/ios/app-fully-superview.spec.js`
- `test/e2e/ios/app-fully-table.spec.js`
- `test/e2e/ios/element-fully.spec.js`

Not sure of a good fix then. Need to take a closer look.

tourdedave: I re-ran those tests, along with the customer repro provided in the Trello ticket, both with and without my change and on both the version of iOS and Appium that we've been using (e.g., iOS 14.5 with Appium 1.22.3) along with the latest available in Sauce (e.g., iOS 16.0 with Appium 2.0.0-beta44).

Here are the results.

### without the fix

|test|iOS 14.5 (Appium 1.22.3)|iOS 16.0 (Appium 2.0.0-beta44)|
|-|-|-|
|customer repro|x|x|
|app-fully-collection|✓|x|
|app-fully-collapsing|✓|x|
|app-fully-overlapped|x|x|
|app-fully-superview|x|x|
|app-fully-table|x|x|
|element-fully|✓|✓|

### with the fix

|test|iOS 14.5 (Appium 1.22.3)|iOS 16.0 (Appium 2.0.0-beta44)|
|-|-|-|
|customer repro|✓|✓|
|app-fully-collection|x|x|
|app-fully-collapsing|x|x|
|app-fully-overlapped|x|x|
|app-fully-superview|x|x|
|app-fully-table|x|x|
|element-fully|x|x|

In summary - the fix is no-go, and there are issues that need to be addressed in newer versions of iOS/Appium regardless of this customer issue. Also, some of the tests on iOS 14.5 are failing consistently now but they were passing last time I ran them. I'm not sure why that is - would need to compare against `master`.

So, as far as next steps, we would need to either
1) find how to make the helper lib element clickable (maybe Anton can have it render at a different `x` coordinate? Just a hunch that that might help since the `x` coordinate was the issue with why the app couldn't be scrolled initially), or
2) find why the customer's app is returning an incorrect scroll height for the scrollable element

kyrylodolynskyi:
This PR refers to the same Trello as this one - https://github.com/applitools/eyes.sdk.javascript1/pull/1176. So I am going to close this PR, as it seems obsolete to me. Feel free to re-open if you think differently.

PR #1153, opened 2022-11-08
[Cypress] coded regions, replace selector with region

SophieDePaula:
[Trello card](https://trello.com/c/uZxC9grz/1615-autodesk-incorrect-region-placement)
This PR purpose is to simplify the experience around coded regions. Currently, we support sending`selectore` and `region` properties, however, we support other options like `padding` and `regionId` only with `region` which is confusing for the customer.

TODO:
- [x] look at accessibility and floating regions - this is currently a limitation for elements, and it works when sending a selector.
- [x] Make sure we have tests to cover the changes and the added scenarios


js/packages/driver/src/element.ts
```diff
@@ -59,6 +59,8 @@ export class Element<TDriver, TContext, TElement, TSelector> {
 
     if (specUtils.isSimpleCommonSelector(this._selector)) {
       this._commonSelector = this._selector
+    } else if (typeof this._selector === 'string') {
```
kyryloonufriiev, line 62: use `utils.types.isString(...)`

kyryloonufriiev (changes requested):

js/packages/eyes-cypress/test/coverage/generic/package.json
```diff
@@ -12,7 +12,7 @@
     "node": ">=8.0.0"
   },
   "devDependencies": {
-    "cypress": "latest",
+    "cypress": "^10.0.0",
```
kyryloonufriiev, line 15: You can return the `latest` here after our solution for cypress@11 right?

SophieDePaula: @kyryloonufriiev I noticed that I get a webpack error when I try to run our tests with Cypress 11 that I need to investigate and I didn't want to delay the release of this and the other PR I am working on.
I'll upgrade our tests to run with cypress 11 after this is done.

pending-changes.yaml
```diff
@@ -91,7 +91,7 @@ Apollo SDK:
   feature:
     - Added new selector extensions `child` and `fallback`
   bug-fix:
-    - 
+    - Add support for regionId and padding for accessibility and floating regions
```
SophieDePaula, line 94: @benbabayoff8

js/packages/eyes-cypress/test/coverage/generic/package.json
SophieDePaula: @kyryloonufriiev it seems that the webpack error no longer occurs with cypress 11.1.0

kyryloonufriiev (approved):
Looks good to me

PR #1155, opened 2022-11-09
[driver] fb issue with contet size ios

zamboney:
Fix issue with the fallback of contentsize

check if `isIOS` before go to the `IOS` fallback

[Trello](https://trello.com/c/4Tr5Vitc/1623-java-jpmc-classic-runner-saucelabs-failures-5340)


kyryloonufriiev (changes requested):
A test is absolutely required for this. I shared [here](https://trello.com/c/4Tr5Vitc/1623-java-jpmc-classic-runner-saucelabs-failures-5340) what it could be

zamboney:
i could not reproduce this issue by myself as I work in the Terrlo card link in the description.
i can write a unit test, but i don't see the point because it could not be reproduced.

What i did was to run all of the `android` tests in `screenshoter` with the `disabledHelper` flag in the `customOptions`, and it didn't get to fallback in all of the cases.

kyryloonufriiev:
You need to use a screen that will throw when you try to extract `contentSize` attr. Not all of them throw when you do that, but I believe there is a screen in the app that will throw. You have to experiment with different screens, to find the right one.
I agree that a unit test for that makes less sense. Because we are not that much interested in testing this particular case (that we don't use ios fallback for android), we actually want to have a test that will demonstrate what is expected behavior for this case when extracting `contentSize` is not possible and helper lib is not applied.

zamboney:
I just ran all of the tests without the `android-helper`

on the local environment. none of the tests threw an error
on `sauce-lab` there was 1 test that threw an error (`test/e2e/android/app-fully-collapsing.spec.js`) that threw the error, but when I ran it again, it worked as expected.

I think this is a flakiness problem that arises because of the `isIOS` condition that I miss. I don't think it will solve the issue, and I also don't think that there is a test that can test this code branch (without creating a glitch in the `sauce-lab` `appium`

@kyryloonufriiev let me know if this is fine and if not, please suggest a solution 10x.

kyryloonufriiev:
This was not flakiness. There is no such behavior when `contentSize` attr throws from time to time. You can ask for sauce labs logs for this run from the customer to verify it.
I saw with my own eyes this behavior - `contentSize` throws due to some specific stuff on the screen (I don't know what exactly). In this case, we recommend using helper lib, because there is no way we can get content size somehow else. But I want a test to verify that in this case android will not blow up, but generate a viewport screenshot as it should.

zamboney:
Maybe we should consider doing a system test for that code. As I explained in the previous comments, I could not find a similar (consistent) behavior in any of our tests without the `helper`.

js/packages/screenshoter/test/e2e/e2e.js
```diff
@@ -95,7 +95,7 @@ exports.makeDriver = async function makeDriver({
   console.log(`makeDriver called for worker #${process.env.MOCHA_WORKER_ID}`, workerId)
   const androidEmulatorIds = process.env.ANDROID_EMULATOR_UDID
     ? process.env.ANDROID_EMULATOR_UDID.split(',')
-    : ['emulator-5554']
+    : ['emulator-5555']
```
zamboney, line 98: i don't think its was intentional

zamboney (commented):
@kyryloonufriiev looking good

js/packages/screenshoter/test/e2e/e2e.js
kyryloonufriiev: It was intentional!

PR #1158, opened 2022-11-14
Support multiple cypress instances

SophieDePaula:
Trello card:
[Parallel run of multiple cypress instances](https://trello.com/c/HCM1qXa1/1620-kpn-cypress-test-are-stuck-when-running-in-parallel)

To solve this issue, we will now create a universal server per cypress process using `singleton: false`, so that once one cypress instance is done, it will close its own universal instance and will not create a problem for other cypress instances that are running in parallel.

TODO:

- [x] Add tests for parallel runs
- [ ] [Complete this PR](https://github.com/applitools/eyes.sdk.javascript1/pull/1153)
- [x] Update pending changes

note: this PR should also solve the[ issue with the self-signed cert for cypress 11](https://trello.com/c/Gm5D8UpH/1625-cypress-1100-self-signed-cert-error)


pending-changes.yaml
```diff
@@ -99,7 +99,7 @@ Apollo SDK:
     - Added new selector extensions `child` and `fallback`
     - Use user agent metadata to improve browser environment detection logic
   bug-fix:
-    - 
+    - Support parallel run of multiple cypress instances
```
SophieDePaula, line 102: @benbabayoff8

pending-changes.yaml
```diff
@@ -99,7 +99,8 @@ Apollo SDK:
     - Added new selector extensions `child` and `fallback`
     - Use user agent metadata to improve browser environment detection logic
   bug-fix:
-    - 
+    - |
+      Support parallel run of multiple `Cypress` instances
```
SophieDePaula, line 103: @benbabayoff8

js/packages/eyes-cypress/src/plugin/server.js
```diff
@@ -46,18 +46,18 @@ function makeStartServer({logger}) {
     }
 
     const {port: universalPort, close: closeUniversalServer} = await makeServerProcess({
-      key: path.resolve(__dirname, '../pem/server.key'),
-      cert: path.resolve(__dirname, '../pem/server.cert'),
       idleTimeout: 0,
       shutdownMode: 'stdin',
       forkOptions,
+      singleton: false,
+      portResolutionMode: 'random',
```
kyryloonufriiev, line 53: I think you can add `port: 0` instead of `portResolutionMode`. Then you will get a random port on the first try.

kyryloonufriiev (approved):
Just a small comment from me. Other then that it is good to go

pending-changes.yaml
benbabayoff8: @SophieDePaula Looks good.

js/packages/eyes-cypress/src/plugin/server.js
SophieDePaula: @kyryloonufriiev I actually prefer to have the first instance of universal running on a known port number as most users use only one instance of cypress when running their tests and this way it's easier to debug.

kyryloonufriiev: Ok, makes sense

PR #1163, opened 2022-11-17
[Cypress] throw diff exception from the browser

SophieDePaula:
[Trello card](https://trello.com/c/EFt8up9W/1582-cypress-tests-not-failing-with-diff-on-latest-version)
Execute browser hooks even if global hooks are enabled so that cypress reports show that the test failed and to print the diffs exception per `spec` file.

TODO:
- [x] Write test
- [x] update pending changes


js/packages/eyes-cypress/src/plugin/hooks.js
```diff
@@ -36,8 +31,6 @@ function makeGlobalRunHooks({closeManager, closeBatches, closeUniversalServer})
             tapFileName: config.appliConfFile.tapFileName,
           });
         }
-
-        handleTestResults.printTestResults({testResults, resultConfig});
```
SophieDePaula, line 40: This is deleted to now print the diff exception twice.

pending-changes.yaml
```diff
@@ -114,6 +114,7 @@ Apollo SDK:
     - |
       Support parallel run of multiple `Cypress` instances
     - Add support for regionId and padding for accessibility and floating regions
+    - Make sure to fail tests when diffrences are found
```
SophieDePaula, line 117: @benbabayoff8

benbabayoff8: @SophieDePaula I would write it as explicitly as it is :) 
Maybe something like "Handle cases where tests weren't failed when differences are found"
Let me know what you think

SophieDePaula: @benbabayoff8 I thought I was explicit :)
In any case, I fixed, lmk if it's OK

tourdedave (approved):
LGTM. Not sure about the test failure in the automated check though. Good luck!

js/packages/eyes-cypress/src/browser/commands.js
```diff
@@ -58,7 +58,7 @@ Cypress.Commands.add('eyesGetAllTestResults', () => {
   });
 });
 
-if (shouldUseBrowserHooks) {
+if (shouldUseBrowserHooks || Cypress.config('eyesFailCypressOnDiff')) {
```
amitzur, line 61: @SophieDePaula I think a better way to do this is to include the `|| Cypress.config('eyesFailCypressOnDiff')` in the variable `shouldUseBrowserHooks`. This leads to clearer code, and also this variable is used in 2 more places (`shouldCreateTapFile` and `eyesOpenMapValues`) and I believe you want its value to be true in case of `Cypress.config('eyesFailCypressOnDiff')`

SophieDePaula: @amitzur that's actually not accurate, if we support global hooks and `eyesFailCypressOnDiff` is `true`, then we still want to create the tap file only once in the global hooks.

PR #1169, opened 2022-11-18
Extension to capture screenshots with the SDK

tourdedave:
[Overview](https://docs.google.com/document/d/1Y9f1xaPLb7wo69_wH4Cljw4TUO7tC_BLK4UqjWhhSF0/edit?usp=sharing)

### What's left

- [x] refactor `screenshoter-browser-extension` as needed (needs code review)
- [x] get tests working in `spec-driver-browser-extension` (port from `eyes-browser-extension`) -- needs to run inside of a browser extension
- [x] update the extension with the final name, images and copy changes when provided by the design team
- [x] show alternate icon (to signify that the extension is disabled) on non-local http sites

### Once merged

- publish `spec-driver-browser-extension`, `screenshot-browser-extension`, and `eyes-browser-extension` to npm
- stand up new Chrome web store page for the screenshot browser extension using the existing group access controls for our other Chrome extensions
- provide web store page edit access to the product owner (Justin)

### Deferred
- add way for extension to be detected when installed/active (and provide JS snippet to use it to the front-end dev)


js/packages/spec-driver-browser-extension/src/spec-driver.ts
```diff
@@ -3,19 +3,19 @@ import * as utils from '@applitools/utils'
 
 // #region UTILITY
 
-export function isDriver(driver) {
+export function isDriver(driver: any) {
```
kyryloonufriiev, line 6: I hope you don't want to leave those `any`s here :)

js/packages/spec-driver-selenium/package.json
```diff
@@ -89,7 +89,8 @@
     "typescript": "^4.7.2"
   },
   "peerDependencies": {
-    "selenium-webdriver": ">=3.6.0"
+    "selenium-webdriver": ">=3.6.0",
+    "@applitools/test-utils": "1.5.2"
```
kyryloonufriiev, line 93: Why?

tourdedave: Because of [the `require` in `build`](https://github.com/applitools/eyes.sdk.javascript1/blob/master/js/packages/spec-driver-selenium/src/spec-driver.ts#L327). While `build` is only used internally for testing, it's not clear that this is a necessary dependency when calling `spec.build` in an SDK.

js/packages/spec-driver-browser-extension/src/spec-driver.ts
tourdedave: https://github.com/applitools/eyes.sdk.javascript1/pull/1169/commits/e6a73aa1a85f24ec4b4795a2198ad4544bd04b6b

js/packages/spec-driver-selenium/package.json
kyryloonufriiev: Because `spec.build` is an internal function. But if you put it in the `peerDeps` all of the customers will have to download this module.

js/packages/spec-driver-browser-extension/src/spec-driver.ts
```diff
@@ -1,21 +1,27 @@
 import browser from 'webextension-polyfill'
 import * as utils from '@applitools/utils'
 
+type CommonSelector<TSelector = never> = string | {selector: TSelector | string; type?: string}
+type Driver = {windowId: number, tabId: number}
+type Context = {windowId: number, tabId: number, frameId: number}
+type El = Element | {'applitools-ref-id': string}
+type Size = {width: number, height: number}
```
kyryloonufriiev, line 8: Import that from `@applitools/utils`

js/packages/spec-driver-browser-extension/src/spec-driver.ts
```diff
@@ -1,21 +1,27 @@
 import browser from 'webextension-polyfill'
 import * as utils from '@applitools/utils'
 
+type CommonSelector<TSelector = never> = string | {selector: TSelector | string; type?: string}
+type Driver = {windowId: number, tabId: number}
+type Context = {windowId: number, tabId: number, frameId: number}
+type El = Element | {'applitools-ref-id': string}
```
kyryloonufriiev, line 7: You should call it `Element`. To get access to the global `Element` type you should use `globalThis.Element`

js/packages/spec-driver-browser-extension/src/spec-driver.ts
```diff
@@ -37,8 +43,9 @@ export function isStaleElementError(error) {
 
 // #region COMMANDS
 
-export async function executeScript(context, script, arg) {
-  const [{result, error}] = await browser.scripting.executeScript({
+export async function executeScript(context: Context, script: any, arg: any) {
```
kyryloonufriiev, line 46: `script` is not an `any` it is always a function. You can see how it is declared in SpecDriver type or other impls of spec driver.

js/packages/spec-driver-browser-extension/src/spec-driver.ts
```diff
@@ -81,16 +89,20 @@ export async function childContext(context, element) {
 
   return {...context, frameId: childFrameId}
 }
-export async function findElement(context, selector, parent) {
+export async function findElement(context: Context, selector: CommonSelector, parent: any) {
```
kyryloonufriiev, line 92: `parent` is optional argument and it should be an Element (or Node in case of shadow root)

js/packages/spec-driver-browser-extension/src/spec-driver.ts
```diff
@@ -81,16 +89,20 @@ export async function childContext(context, element) {
...
   const [{result}] = await browser.scripting.executeScript({
     target: {tabId: context.tabId, frameIds: [context.frameId || 0]},
     /* eslint-disable no-undef */
     func: (selector, parent) => {
       if (selector.type === 'css') {
+        // @ts-ignore
```
kyryloonufriiev, line 98: There should be a better way to solve this problem 🤔

js/packages/spec-driver-selenium/package.json
tourdedave: Right, okay. But how do we make this requirement more concrete for the SDK? We just have to know that the SDK needs to have test-utils installed in order to use spec.build?

js/packages/spec-driver-browser-extension/src/spec-driver.ts
tourdedave: I'm open to suggestions.

js/packages/spec-driver-browser-extension/src/spec-driver.ts
tourdedave: It couldn't also be a string?

tourdedave: Ah, I see [in the docs](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/scripting/executeScript#func) it's only a function

js/packages/spec-driver-browser-extension/src/spec-driver.ts
tourdedave: I think I've got a better solution --> https://github.com/applitools/eyes.sdk.javascript1/pull/1169/commits/d839bac31e4576ae13055f20565c745c8a2b2c96

js/packages/spec-driver-selenium/package.json
kyryloonufriiev: When you try to call it you will know :) I understand and relate to the wish to make it clearer. If it was a public function we would simply have `test-utils` in dependencies, but since it is not a public function we put it in the devDeps to avoid actual customers to install it. So we have to remember to install it manually if we want to run `spec.build`.

kyryloonufriiev: I absolutely agree. It is suboptimal. The "right" way to do it is to put it in a separate module "internal" or something like this, or refactor this function to not depend on `parseEnv` (it was implemented like this initially but I want it to be smarter, so I put it inside, so the usage of `spec.build` would be easier). All of this will be even worse than small sub-optimality, in my opinion. We can document it somewhere if we need.

js/packages/spec-driver-webdriverio/package.json
```diff
@@ -130,7 +130,8 @@
     "webdriverio": "^7.19.7"
   },
   "peerDependencies": {
-    "webdriverio": ">=5.0.0"
+    "webdriverio": ">=5.0.0",
+    "chromedriver": "^101.0.0"
```
kyryloonufriiev, line 134: `chromedriver` also could not be a `peerDependency` for the same reason as `test-utils`

js/packages/spec-driver-browser-extension/package.json
```diff
@@ -0,0 +1,86 @@
...
+    "mocha": "^9.2.2",
+    "playwright": "^1.22.2",
+    "prettier": "^2.6.2",
+    "ts-node": "^10.8.0",
+    "ttypescript": "^1.5.13",
+    "typescript": "^4.8.4"
+  },
+  "engines": {
+    "node": ">=12.13.0"
+  }
```
kyryloonufriiev, line 85: Do we need a node engine if it is not even meant to be executed in node environment?

js/packages/spec-driver-browser-extension/src/spec-driver.ts
```diff
@@ -1,21 +1,33 @@
 import browser from 'webextension-polyfill'
 import * as utils from '@applitools/utils'
+import type {Size} from '@applitools/utils'
+
+type CommonSelector<TSelector = never> = string | {selector: TSelector | string; type?: string}
+type Driver = {windowId: number, tabId: number}
+type Context = {windowId: number, tabId: number, frameId: number}
+type El = globalThis.Element | {'applitools-ref-id': string}
```
kyryloonufriiev, line 8: This should be called `Element`. Also, I am not sure if it could be `globalThis.Element`, rather just a `Ref` type, which BTW should be exported from the refer implementation file, since there this type should be defined.

js/packages/spec-driver-browser-extension/test/ext/dist/content.js
```diff
@@ -0,0 +1,7362 @@
+(function webpackUniversalModuleDefinition(root, factory) {
```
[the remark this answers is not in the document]
tourdedave: I assumed our top-level gitignore would have prevented it. I didn't confirm.

The source of the extension is in `test/ext/src`

js/packages/spec-driver-browser-extension/test/e2e/spec-driver.spec.ts
```diff
@@ -264,6 +274,7 @@ describe('spec driver', async () => {
       const elementKey = await expectedElement.evaluate(element => (element.dataset.key = 'element-key'))
       const isCorrectElement = await backgroundPage.evaluate(
         async ([context, element, elementKey]) => {
+          // @ts-ignore
```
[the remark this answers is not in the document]
tourdedave: I added it because of intellisense when reviewing the file, but it works when removing it.

js/packages/spec-driver-browser-extension/package.json
tourdedave: You're right, we don't need it.

js/packages/spec-driver-browser-extension/src/spec-driver.ts
tourdedave: 👍

js/packages/spec-driver-selenium/package.json
tourdedave: Okay. Thanks for talking that through. I added a code comment to the `build` function in each of the spec-drivers that have it/need it.

kyryloonufriiev (approved):
The PR is good to merge. But it would be nice if you update the new packages to the new way to do linting and ts. You can check out any other package to see `tsconfig` files, `.eslintrc`, and `devDependencies`

PR #1170, opened 2022-11-18
make sure to create display when running only test from a cypress PR

SophieDePaula:
We need to execute `Xvfb setup` step even when we just run the test without the publish step for cypress that's the reason the test from the PR did not work so far.
Also, we use the SDK key and we need the JS key for some of that tests so that was updated as well.


tourdedave:
@SophieDePaula why did you close this?

SophieDePaula:
@tourdedave because I used this PR to test a few unrelated things and, I'll open a new PR with the relevant changes

PR #1171, opened 2022-11-20
[eyes-cypress] add type definition for eyes plugin

zamboney:
## Trello

1. [ticekt-1](https://trello.com/c/EiZummQg/1529-cypress-typescript-support-improvement-island)
2. [ticket-2](https://trello.com/c/ojUJMxwN/1711-adding-cypress-sdk-to-a-project-with-vite-breaks-the-package-imports)
3. [ticket-3](https://trello.com/c/JPRbJ7aw/1723-cypress-install-challenges-google-poc)

- [x] add support for `ESM` and `TS`
- [x] add support to `xvfb` using the `github action` `test`
- [x] change the `plugin` to `typescript`
- [x] tests to `cypress` `9`, `10`, `11`, `12`
- [x] add support to `moduleResolution:node` using `ttsc`


pending-changes.yaml
```diff
@@ -104,7 +104,7 @@ Apollo SDK:
   feature:
     - 
   bug-fix:
-    - 
+    - Add typescript definition for eyes plugin
```
amitzur, line 107: I would move this from `bug-fix` to `feature` and write `Support ESM and TypeScript in Applitools' config file`. I hope this is indeed the nature of this PR.

amitzur (changes requested):
Please add e2e test.
Also, I'm not familiar with this syntax: `export = _exports`. Can you explain it or share a link to TS docs? Is it a feature of TS or just ESM?

pending-changes.yaml
zamboney: done

zamboney:
> Please add e2e test.

not sure how to add `e2e` test but i will change the `test/e2e/ts/cypress/plugin/index.js` to `test/e2e/ts/cypress/index.ts` 

>  Also, I'm not familiar with this syntax: `export = _exports`. Can you explain it or share a link to TS docs? Is it a feature of TS or just ESM?

the syntax is what auto-generated by `tsc --declaration --allowJs --emitDeclarationOnly --outDir types`. i just took the product of `types/plugin/startPlugin.d.ts` and put it in `index.d.ts` so it will not throw a `typescript` error once using typescript

amitzur:
@zamboney an e2e test for this is taking the snippet you put in the description, and actually running it in an automated test. This is how to prove that the code actually solves this case.
I think it's important to add such a test and not just change `index.js` to `index.ts`.

zamboney:
@amitzur this is what i did, i changed the `plugin.js` to `plugin.ts` and changed the code so it will support the new type definitions

kyryloonufriiev:
I don't think this is the right solution in general. Why would we have to wrap the whole config object if we want to modify just one property of it?

kyryloonufriiev:
This is a random cypress plugin that I found by googling "cypress plugins" - https://github.com/Shelex/cypress-allure-plugin. It has a more clear forward setup logic, I think this is what we have to provide for the people who wanna take a manual step to setup applitools.

zamboney:
@kyryloonufriiev I'm not sure what exactly you want me to change?
It's just for the setting of the plugin, not for the `eyes` usage, all the customer wants, in the ticket is to be able to set the plugin in `typescript` without the right definition and without using `@ts-ignore`.
If you understand it differently please care to explain

kyryloonufriiev:
Do you see how plugins setup works in this package I shared? We should do the same. Because this is more transparent and clearer API for the end user.

kyryloonufriiev (changes requested):
I think the current implementation is trying to make something work on top of what we have. While deeper understanding is definitely required here. I see that the `eyesPlugin` method still has the logic that works with `module.exports` stuff, this could not be right. The `eyesPlugin` should do only what is required to do - run the universal server/setup hooks/modify cypress config, and that's it everything else (module.exports) related to a different function wrapper that is created around this function to work with the module object.

js/packages/eyes-cypress/src/plugin/pluginExport.js
```diff
@@ -4,8 +4,60 @@ const {presult} = require('@applitools/functional-commons');
...
+    const {server, port, closeManager, closeBatches, closeUniversalServer} = await startServer();
+    eyesServer = server;
+
+    const globalHooks = makeGlobalRunHooks({closeManager, closeBatches, closeUniversalServer});
+
+    const [origOn, config] = args;
+
+    if (!pluginModuleExports) {
+      pluginModuleExports =
+        config.testingType === 'e2e' ? pluginExportsE2E : pluginExportsComponent;
```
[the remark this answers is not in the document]
zamboney: it's should not be populated. this is for the `setupNodeEvents`. if there is  a `e2e` or `component` then it will use the `setupNodeEvents` directly like so 

```typescript
import { defineConfig } from 'cypress'
import { eyesPlugin } from 'eyes-cypress'

export default defineConfig({
  // setupNodeEvents can be defined in either
  // the e2e or component configuration
  e2e: {
    setupNodeEvents(on, config) {
       eyesPlugin(on, config)
    }
  }
})
```

zamboney (commented):
@kyryloonufriiev i don't want to make a huge change base on a ticket of something else.

i sure that there is a task on it and once we will have time to do it we will. in the meanwhile we are not fix a typescript error that just look bad...

kyryloonufriiev:
If we would wait for tickets for everything we do we would be in a much worse position. I think we should be in a mindset of improving everything we touch, not just rush with addressing exactly what bothering this particular customer. Tomorrow there will be another one that would want to use it with esm and it won't be possible.
I see this card as an opportunity to improve the plugin setup process that we have, so not just address an issue but improve the area around it.

js/packages/eyes-cypress/src/plugin/pluginExport.js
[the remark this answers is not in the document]
kyryloonufriiev: If I do what you wrote in the example it still will not be populated. Or am I missing something?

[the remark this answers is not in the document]
zamboney: I just tested it, and it's working (you need to add the eyes `command`, just like in the example your share)

But it still needs small changes


```js
import { defineConfig } from 'cypress'
import { eyesPlugin } from 'eyes-cypress'

export default defineConfig({
  // setupNodeEvents can be defined in either
  // the e2e or component configuration
  e2e: {
    setupNodeEvents(on, config) {
       const eyesConfig = await eyesPlugin(on, config);
       return { ...config, ...eyesConfig};
    }
  }
})
```

i think that this is a great approach @amitzur what do you think?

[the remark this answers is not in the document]
kyryloonufriiev: I wasn't saying it is not going to work, but did you check if `pluginExportsE2E`/`pluginExportsComponent` were populated? I think they didn't and this is correct, but this raise a question for me, should it even be written in the place it is right now?
Feel free to correct me if I am missing something.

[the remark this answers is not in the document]
zamboney: if you are not passing the `pluginExports` then of course it will not populate in the configuration above. but way it should be? if you use the configuration above then you are not using the old `require('@applitools/eyes-cypress')(module)` and therefor you need to setup each `e2e` and `component` on his own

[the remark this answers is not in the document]
kyryloonufriiev: If it will not be used then why should it be there? My point is that if I use `eyesPlugin(on, config)`, then I don't need this code, which means this code should not even be there.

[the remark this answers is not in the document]
zamboney: i could not figure out an easy way to remove it without breaking everything... i think in this point of the pr is ok to leave it like so

amitzur:
> > Please add e2e test.
> 
> not sure how to add `e2e` test but i will change the `test/e2e/ts/cypress/plugin/index.js` to `test/e2e/ts/cypress/index.ts`
> 
> > Also, I'm not familiar with this syntax: `export = _exports`. Can you explain it or share a link to TS docs? Is it a feature of TS or just ESM?
> 
> the syntax is what auto-generated by `tsc --declaration --allowJs --emitDeclarationOnly --outDir types`. i just took the product of `types/plugin/startPlugin.d.ts` and put it in `index.d.ts` so it will not throw a `typescript` error once using typescript

I was looking for this reference:
https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require

js/packages/eyes-cypress/src/plugin/pluginExport.js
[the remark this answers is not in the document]
zamboney: @kyryloonufriiev can you please replay

amitzur:
@zamboney @kyryloonufriiev Let me try to do some order in the next steps:

1. Let's just make it clear what scenarios we wish to support.
2. Let's decide what we want to cover in this PR. IMO if we're already working on it, we should just support all the possible scenarios (there shouldn't be too much) and be done with it.
3. We should cover all these scenarios in e2e tests.
4. Understand the implications on `eyes-setup`, and do adjustments if necessary (including tests)
5. Make sure all scenarios are documented properly.
6. Ship it!


So, here are the scenarios I see and would like to support, and the proposition of how to support them:

1. `pluginsFile` in CJS
```
module.exports = (on, config) => {
  // ...
}
require('@applitools/eyes-cypress')(module)
```

2. `cypress.config.js` in CJS
```
const { defineConfig } = require("cypress");

module.exports = defineConfig({
  e2e: {
    setupNodeEvents(on, config) {
      // implement node event listeners here
    },
  },
});

require('@applitools/eyes-cypress')(module)
```

3. `cypress.config.ts` or `cypress.config.mjs` (ESM)
```
import {defineConfig} from 'cypress';
import {eyesPlugin} from '@applitools/eyes-cypress';

export default defineConfig({
  e2e: {
    setupNodeEvents(on, config) {
      // implement node event listeners here
      eyesPlugin(on, config)
    },
  },
});
```

4. `pluginsFile` in ESM
```
import {eyesPlugin} from '@applitools/eyes-cypress';
export default (on, config) => {
  // ...
  eyesPlugin(on, config)
}
```

In short, the new thing we aim to provide in this PR is a new API for ESM (and TS, wow so many acronyms...).
Since the proposed ESM support is harder to generate automatically in the `eyes-setup` script, we need to instruct the user to do it manually, preferably by detecting this situation in `eyes-setup` and then recommending the manual solution to the user.

I hope this makes it clearer, and also that my proposition makes sense to you.

amitzur:
@zamboney yep

amitzur:
@kyryloonufriiev @zamboney I tend to agree. In my words - the `pluginExport.js` file has become a little monster, much more than it was planned, with patches added to it over time. I tend to agree that we have already spent a lot digging into this PR, so now it's a good time to sort it out.

To put it in technical terms - I would refactor it into TS, and separate wrapper function (which is what `pluginExport.js` was meant to be) with the actual **plugin** (currently named `setupNodeEvents`). This way we could also unit test many scenarios without the need for e2e for **all** combinations (e.g. both `e2e` and `component`, AND each of them alone). I do think that e2e tests should exist for all 4 scenarios that I described above.

@zamboney - are you good with this?

js/packages/eyes-cypress/test/e2e/ts/cypress/tsconfig.json
```diff
@@ -1,12 +1,14 @@
...
     "moduleResolution": "node16",
     "lib": ["es6", "dom"],
     "types": ["cypress", "../../../.."],
     "noImplicitAny": true,
     "strictNullChecks": true,
     "noEmit": true,
-  },
-}
\ No newline at end of file
+    "esModuleInterop": true
```
amitzur, line 12: Does it mean that our users need to have `esModuleInterop:true` and `allowSyntheticDefaultImports:true` in their tsconfig.json as well?

amitzur (commented):
I reviewed about 30% of the PR so far. I thought I would sumbit a review and then continue.
See my inline comments, and also `yarn test:e2e` doesn't even run because of the missing `@applitools/test-server` depenedency. Did you verify that all tests pass?

js/packages/eyes-cypress/test/fixtures/testAppCypress10/cypress.config.js
```diff
@@ -1,4 +1,4 @@
-const {defineConfig} = require('cypress');
+const {defineConfig} = require('./node_modules/cypress');
```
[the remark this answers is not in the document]
zamboney: I had an issue with typescript, i solve it without using `./node_modules` but i forgot to remove this line. I will remove it. Thanks for the feedback

js/packages/eyes-cypress/test/e2e/ts/cypress/tsconfig.json
zamboney: I'm not sure how it's work before before my changes but this isn't work now without it... we need to mention it and maybe upgrade a major version?

js/packages/eyes-cypress/README.md
```diff
@@ -45,12 +45,40 @@ Normally, this is `cypress/plugins/index.js`. You can read more about it in Cypr
 
  #### Cypress version >= 10:
 
- Add the following code to your `cypress.config.js` file after `module.exports`:
```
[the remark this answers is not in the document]
zamboney: but this explain in the section above? isn't it enough?

zamboney:
@amitzur about verify that it's work [this action](https://github.com/applitools/eyes.sdk.javascript1/actions/runs/3817382577) run on this branch and pass

js/packages/eyes-cypress/src/plugin/config.ts
```diff
@@ -1,19 +1,12 @@
-'use strict'
-const utils = require('@applitools/utils')
-const {configParams} = require('./configParams')
+import * as utils from '@applitools/utils'
+import configParams from './configParams'
 const DEFAULT_TEST_CONCURRENCY = 5
-const uuid = require('uuid')
+import * as uuid from 'uuid'
+import {type EyesConfig} from './../expose'
```
amitzur, line 5: no need for `./../`, just `../`

js/packages/eyes-cypress/src/plugin/pluginExport.ts
```diff
@@ -1,11 +1,12 @@
-'use strict'
-const isGlobalHooksSupported = require('./isGlobalHooksSupported')
-const {presult} = require('@applitools/functional-commons')
-const makeGlobalRunHooks = require('./hooks')
+import isGlobalHooksSupported from './isGlobalHooksSupported'
+// @ts-ignore
+import {presult} from '@applitools/functional-commons'
+import makeGlobalRunHooks from './hooks'
+import {type EyesConfig} from './../expose'
```
amitzur, line 5: no need for `./../`, just `../`

js/packages/eyes-cypress/src/plugin/setupNodeEvents.ts
```diff
@@ -0,0 +1,18 @@
+import makeGlobalRunHooks from './hooks'
+import {type EyesConfig} from './../expose'
```
amitzur, line 2: no need for `./../`, just `../`

js/packages/eyes-cypress/src/plugin/index.ts
```diff
@@ -0,0 +1,27 @@
+'use strict'
+import makePluginExport from './pluginExport'
+import setupNodeEvents from './setupNodeEvents'
+import makeConfig from './config'
+import makeStartServer from './server'
+import {makeLogger} from '@applitools/logger'
+
+// DON'T REMOVE
+// if remove the ttsc will compile the absolute path
+import {type EyesConfig} from '../expose' // eslint-disable-line @typescript-eslint/no-unused-vars
```
amitzur, line 10: You can also define the type and export it from here, instead of in `expose.ts`. It will do the same thing. I would leave the comment, though, it's great.
I would also add to the comment that the absolute path is added because TS (or api-extractor?) goes over the `eyesPlugin` declaration before it goes over the `EyesConfig` definition, and this is why it's importat to reverse the order.

js/packages/eyes-cypress/src/plugin/setupNodeEvents.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export default function ({startServer, eyesConfig}: any) {
+  return async function (origOn: Cypress.PluginEvents, cypressConfig: Cypress.PluginConfigOptions): Promise<EyesConfig> {
```
amitzur, line 5: The parameter name is important here, since it shows up to the user. So I would change `origOn` to `on` and `cypressConfig` to `config`. And add a comment about why the parameter names are important :)

js/packages/eyes-cypress/src/plugin/setupNodeEvents.ts
```diff
@@ -0,0 +1,18 @@
+import makeGlobalRunHooks from './hooks'
```
amitzur, line 1: I would rename this file from `setupNodeEvents` to something like `eyesPlugin` or `startPlugin`. And I would have a named export called `makeEyesPlugin`, because that's essentially what it's doing.

js/packages/eyes-cypress/src/plugin/setupNodeEvents.ts
```diff
@@ -0,0 +1,18 @@
...
+import {type EyesConfig} from './../expose'
+
+export default function ({startServer, eyesConfig}: any) {
+  return async function (origOn: Cypress.PluginEvents, cypressConfig: Cypress.PluginConfigOptions): Promise<EyesConfig> {
+    const {port, closeManager, closeBatches, closeUniversalServer} = await startServer()
+
+    const globalHooks = makeGlobalRunHooks({closeManager, closeBatches, closeUniversalServer})
+
+    if (eyesConfig.eyesIsGlobalHooksSupported) {
```
amitzur, line 10: This will never be true, because `eyesIsGlobalHooksSupported` is assigned a value in `pluginExport`. I would like this bug to be covered in a test. This is exactly the type of regressions we keep having in Cypress because this is a fragile area that is sensitive to how the user uses the SDK.

The test should check that the functionality that is done in the global hooks is actually performed when using the `eyesPlugin` directly.

I can also argue that this if statement is not necessary at all, because this is the new way to do things, and only new Cypress versions would use it.

js/packages/eyes-cypress/src/plugin/pluginExport.ts
```diff
@@ -19,24 +20,29 @@ function makePluginExport({startServer, eyesConfig}) {
...
+      origOn: Cypress.PluginEvents,
+      cypressConfig: Cypress.PluginConfigOptions,
+    ): Promise<EyesConfig> {
       const {server, port, closeManager, closeBatches, closeUniversalServer} = await startServer()
       eyesServer = server
 
-      const globalHooks = makeGlobalRunHooks({closeManager, closeBatches, closeUniversalServer})
-
-      const [origOn, config] = args
+      const globalHooks: any = makeGlobalRunHooks({
```
amitzur, line 30: It's a shame that we can't call the code in `setupNodeEvents` (which I think should be renamed) here. There's risky code duplication here. I understand why, and I don't have a solution for it :(

amitzur (changes requested):
General notes:
1. Most of the code that was changed from JS to TS has types that are just `any`. I think it's fine not to add all types right now, but then let's not call it "refactor JS to TS", since there is no added value in it. So we should be aware that there is a task waiting to actually write types in the Eyes-Cypress SDK codebase.
2. I think the solution with the `ttsc` is clever and valid. Although I am curious to know how to fix `api-extractor`, I will hold my curiousity in favor of focusing on what matters :)
3. I haven't gone over the CI files yet. Those are the only ones left for me to review.

js/packages/eyes-cypress/src/plugin/setupNodeEvents.ts
zamboney: done

js/packages/eyes-cypress/src/plugin/setupNodeEvents.ts
zamboney: done

js/packages/eyes-cypress/src/plugin/setupNodeEvents.ts
zamboney: you right i miss that 🤦, i will add a test for that

js/packages/eyes-cypress/src/plugin/webSocket.ts
```diff
@@ -75,29 +87,33 @@ function connectSocket(url) {
...
+    // @ts-ignore
     const command = () => socket._socket.ref()
     if (isReady) command()
     else queue.add(command)
     return () => queue.delete(command)
   }
 
   function unref() {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
```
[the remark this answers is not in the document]
zamboney: i will 10x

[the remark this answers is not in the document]
zamboney: i will 10x

js/packages/eyes-cypress/test/e2e/ts/cypress-12.config.ts
```diff
@@ -0,0 +1,13 @@
+import {defineConfig} from 'cypress12'
```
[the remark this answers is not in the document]
zamboney: you think that but you need to `import` the right `cypress` version and then `export` this `defineConfig`. it can be done if you create the `cypress.config.ts` dynamically is it worth it? I think it's better to have 3 files than to have a dynamic file.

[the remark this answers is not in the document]
zamboney: you think that but you need to `import` the right `cypress` version and then `export` this `defineConfig`. it can be done if you create the `cypress.config.ts` dynamically is it worth it? I think it's better to have 3 files than to have a dynamic file.

js/packages/eyes-cypress/test/e2e/ts/cypress/plugins/index.js
```diff
@@ -1,4 +0,0 @@
-module.exports = (_on, _config) => {}
```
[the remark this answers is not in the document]
zamboney: no... its used for testing

js/packages/eyes-cypress/test/fixtures/cypressConfig-global-hooks-overrides-config-file.js
```diff
@@ -1,4 +1,4 @@
-const {defineConfig} = require('cypress');
+const {defineConfig} = require('./node_modules/cypress');
```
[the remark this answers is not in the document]
zamboney: sure fixed

js/packages/eyes-cypress/types/index.d.ts
```diff
@@ -0,0 +1,53 @@
+/// <reference types="cypress" />
```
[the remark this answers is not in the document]
zamboney: `cypress` browser commands definitions read `types/index.d.ts` by default

js/packages/eyes-cypress/types/index.d.ts
```diff
@@ -0,0 +1,53 @@
...
+       *  })
+       * })
+       */
+      eyesGetAllTestResults(): Chainable<CypressTestResultsSummary>
+    }
+  }
+}
+
+import exposeDefault from './expose'
+import {eyesPlugin} from './expose'
```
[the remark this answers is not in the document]
zamboney: done

zamboney (commented):
@amitzur thank you for the great review 🎉
i still need to test the `eyesIsGlobalHooksSupported` issue but i think remove it is the good approach and so i did

js/packages/eyes-cypress/src/browser/spec-driver.ts
```diff
@@ -30,65 +25,63 @@ export function parentContext(context: Context): Context {
...
-  if(element.contentDocument)
-    return element.contentDocument
+  if (element.contentDocument) return element.contentDocument
   else {
     throw new Error('Context is not accessible')
   }
 }
 
-export function getViewportSize(): Object {
+export function getViewportSize(): any {
```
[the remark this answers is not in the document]
amitzur: @zamboney ping

js/packages/eyes-cypress/README.md
```diff
@@ -45,12 +45,40 @@ Normally, this is `cypress/plugins/index.js`. You can read more about it in Cypr
...
+ Add the following code to your:
+
+ ##### `cypress.config.js`
```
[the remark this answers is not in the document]
amitzur: Good point. Well, they have this interactive widget that makes it easier to see the difference. I respect your approach :)

js/packages/eyes-cypress/src/plugin/errorDigest.ts
```diff
@@ -8,25 +7,50 @@ const colors = {
...
   Unresolved: {
     color: 'orange',
     symbol: '\u26A0',
-    title: tests => `Diffs detected - ${tests} tests`,
+    title: (tests: any) => `Diffs detected - ${tests} tests`,
   },
 }
 
-function errorDigest({passed, failed, diffs, logger, isInteractive}) {
+function stringifyTestResults(testResults: any) {
```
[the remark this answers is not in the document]
amitzur: I guess using `noImplicitAny: false` in `tsconfig` would remove it. I don't see great value in these TS errors, as the solution for them is just adding `any` everywhere. WDYT?

js/packages/eyes-cypress/src/plugin/server.ts
```diff
@@ -1,33 +1,27 @@
...
-      if (err) {
-        logger.log('error starting plugin server', err)
-      } else {
-        logger.log(`plugin server running at port: ${https.address().port}`)
-        port = https.address().port
-      }
-    })
-
+    https.listen()
+    const port = (https.address() as AddressInfo).port
```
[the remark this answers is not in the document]
amitzur: You are right that `server.listen` doesn't return a Promise, and the `await` keyword that was there is a mistake. But the more important part is that the `port` is set in the callback. So I don't think it's just good measure. I think it's good that you promisify it. Let's keep sending `0` as the port.

js/packages/eyes-cypress/src/plugin/server.ts
```diff
@@ -134,13 +130,11 @@ function makeStartServer({logger}) {
         ),
       )
     }
-    function closeBatches(settings) {
+    function closeBatches(settings: any) {
       if (socketWithUniversal)
-        return socketWithUniversal.request('Core.closeBatches', {settings}).catch(err => {
+        return socketWithUniversal.request('Core.closeBatches', {settings}).catch((err: Error) => {
           logger.log('@@@', err)
```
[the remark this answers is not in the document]
amitzur: Yes, that's what I meant. I added this a while back.

js/packages/eyes-cypress/test/e2e/ts/cypress-12.config.ts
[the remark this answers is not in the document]
amitzur: Oh, I didn't notice it. Sorry.

js/packages/eyes-cypress/test/e2e/ts/cypress/plugins/index.js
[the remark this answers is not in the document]
amitzur: This file was deleted. I guess github just doesn't shows it properly.

js/packages/eyes-cypress/types/index.d.ts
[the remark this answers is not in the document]
amitzur: Do you have a reference for it? (I mean I believe you, I'm just curious to learn more)

js/packages/eyes-cypress/test/e2e/ts/cypress/plugins/index.ts
```diff
@@ -0,0 +1,2 @@
+import {eyesPlugin} from '../../../../../'
+export default eyesPlugin
```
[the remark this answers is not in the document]
amitzur: @zamboney :point_up:

js/packages/eyes-cypress/test/e2e/ts/cypress/tsconfig.json
```diff
@@ -1,12 +1,9 @@
...
-  "include": ["integration/typescript.spec.ts"],
-    "compilerOptions": {
+  "include": ["**/*.ts"],
+  "compilerOptions": {
     "target": "es6",
-    "moduleResolution": "node16",
+    "moduleResolution": "node",
     "lib": ["es6", "dom"],
-    "types": ["cypress", "../../../.."],
-    "noImplicitAny": true,
```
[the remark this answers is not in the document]
amitzur: @zamboney :point_up:

js/packages/eyes-cypress/src/plugin/server.ts
```diff
@@ -53,13 +49,13 @@ function makeStartServer({logger}) {
       portResolutionMode: 'random',
     })
 
-    const managers = []
-    let socketWithUniversal
+    const managers: {manager: string; socketWithUniversal: SocketWithUniversal}[] = []
```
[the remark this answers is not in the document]
amitzur: @zamboney you changed it to `object`. I think `any` is fine. And shouldn't it be `Object` (uppercase O)?

zamboney:
@amitzur please take another look 👀 . I added `userGlobalHook` support!

js/packages/eyes-cypress/src/plugin/eyesPlugin.ts
```diff
@@ -0,0 +1,69 @@
...
+  return hooks.reduce(
+    (acc, hook) => ({
+      ...acc,
+      [hook]: content
+        .split('\n')
+        .some(line => new RegExp(`on\\s?\\(\\s?[\`|"|']${hook}[\`|"|']\\s?,`).test(line) && !isCommented(line)),
+    }),
+    results,
+  )
+}
```
zamboney, line 69: This idea to check if plugins exist is from [this](https://github.com/Shelex/cypress-allure-plugin/blob/master/writer/checkPluginsFile.js) plugin

@amitzur ☝️

js/packages/eyes-cypress/test/e2e/ts/cypress/tsconfig.json
```diff
@@ -1,12 +1,14 @@
...
-    "compilerOptions": {
+  "include": ["**/*.ts"],
+  "compilerOptions": {
+    "allowSyntheticDefaultImports": true,
     "target": "es6",
-    "moduleResolution": "node16",
+    "moduleResolution": "node",
     "lib": ["es6", "dom"],
-    "types": ["cypress", "../../../.."],
+    "types": ["cypress", "../../../..", "node"],
```
zamboney, line 8: @amitzur, the `"node"` type is required for using the `NodeJS.Module` type. if you think we could drop this and use custom type instead

js/packages/eyes-cypress/test/e2e/ts/cypress/plugins/index.ts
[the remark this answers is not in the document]
zamboney: this is the syntax for [cypress < 10](https://docs.cypress.io/guides/tooling/typescript-support#Types-for-plugins)

js/packages/eyes-cypress/types/index.d.ts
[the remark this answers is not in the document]
zamboney: No, but I try it in other ways and it would not work...

zamboney (commented):
@amitzur the `e2e` test [pass](https://github.com/applitools/eyes.sdk.javascript1/actions/runs/4044339065/jobs/6954394238) `node16` (the other tests just stack) so i thing we are good to go. 

i'm running the `e2e` test again (just in case) once i submit this comment so i think after you refer to my replay we are g2g

js/packages/eyes-cypress/index.js
```diff
@@ -1,2 +1,5 @@
-'use strict'
-module.exports = require('./src/plugin/startPlugin')
+// This is the main entry point of our application
```
amitzur, line 1: @zamboney why do you still need this file?

js/packages/eyes-cypress/package.json
```diff
@@ -11,8 +11,8 @@
     "cypress",
     "cy"
   ],
-  "main": "index.js",
-  "types": "./index.d.ts",
+  "main": "./index",
```
amitzur, line 14: @zamboney why not `dist/plugin/index.js`? And why not specify the file suffix?

js/packages/eyes-cypress/index.js
zamboney: only for good measure, i can remove it if you think its unnecessary

js/packages/eyes-cypress/package.json
zamboney: same as last comment

js/packages/eyes-cypress/src/expose.ts
```diff
@@ -0,0 +1,72 @@
...
+  envName?: CypressEyesConfig['environmentName']
+
+  accessibilitySettings?: NonNullable<CypressEyesConfig['defaultMatchSettings']>['accessibilitySettings']
+}
+
+export type CypressTestResultsSummary = api.TestResultsSummary
+
+import {type EyesPluginConfig} from './plugin' // eslint-disable-line @typescript-eslint/no-unused-vars
+
+export * from './plugin'
```
amitzur, line 70: @zamboney change this to `export {type EyesPluginConfig} from './plugin'`

js/packages/eyes-cypress/src/plugin/errorDigest.ts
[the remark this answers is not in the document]
amitzur: how is it going to help you? you will search for `: any` in the code and start fixing? anyway, here I don't have a strong opinion so go with however you feel is right.

js/packages/eyes-cypress/src/plugin/index.ts
```diff
@@ -0,0 +1,44 @@
...
+const logger = makeLogger({level: config.showLogs ? 'info' : 'silent', label: 'eyes'})
+
+const startServer = makeStartServer({logger})
+
+const pluginExport = makePluginExport({
+  startServer,
+  eyesConfig: Object.assign({}, eyesConfig, {appliConfFile: config}),
+})
+
+export const eyesPlugin = makeEyesPlugin({
```
amitzur, line 38: @zamboney You no longer need this in the new proposal

js/packages/eyes-cypress/src/expose.ts
```diff
@@ -0,0 +1,72 @@
...
+  notifyOnCompletion?: NonNullable<CypressEyesConfig['batch']>['notifyOnCompletion']
+
+  envName?: CypressEyesConfig['environmentName']
+
+  accessibilitySettings?: NonNullable<CypressEyesConfig['defaultMatchSettings']>['accessibilitySettings']
+}
+
+export type CypressTestResultsSummary = api.TestResultsSummary
+
+import {type EyesPluginConfig} from './plugin' // eslint-disable-line @typescript-eslint/no-unused-vars
```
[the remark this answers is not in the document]
zamboney: i can remove the `EyesPluginConfig` but `CypressTestResultsSummary` will not work with for `yarn test:ts` it's needed [here](https://github.com/applitools/eyes.sdk.javascript1/blob/ae0eb70937c7a65042057fa1fff7b326324bbf48/js/packages/eyes-cypress/types/index.d.ts#L45)

```
yarn run v1.22.19
$ yarn test:ts:compile && yarn test:ts:run
$ tsc --project test/e2e/ts/cypress
types/index.d.ts:2:65 - error TS2614: Module '"./expose"' has no exported member 'CypressTestResultsSummary'. Did you mean to use 'import CypressTestResultsSummary from "./expose"' instead?

2 import {type CypressCheckSettings, type CypressEyesConfig, type CypressTestResultsSummary} from './expose'
                                                                  ~~~~~~~~~~~~~~~~~~~~~~~~~
```

js/packages/eyes-cypress/src/setup/handlePlugin.js
```diff
@@ -1,16 +1,38 @@
...
+  if (!isESMOrTS && !isPluginDefined(fileContent)) {
     fs.writeFileSync(pluginsFilePath, addEyesCypressPlugin(fileContent))
     console.log(chalk.cyan('Plugins defined.'))
+  } else if (isESMOrTS && !isPluginDefinedESM(fileContent)) {
+    console.log(
+      boxen(
+        `
+We detected that you are using TS or ESM syntax. Please configure the plugin as follows:
+
+${chalk.green.bold('import eyesPlugin from "@appliitools/eyes-cypress"')}
```
amitzur, line 25: @zamboney typo! `appliitools`. This is user facing, please proof read it 10 times!

js/packages/eyes-cypress/src/plugin/index.ts
zamboney: sure

js/packages/eyes-cypress/src/setup/handlePlugin.js
```diff
@@ -1,16 +1,38 @@
 'use strict'
 
 const chalk = require('chalk')
+const boxen = require('boxen')
+const {parse, first} = require('abstract-syntax-tree')
```
amitzur, line 5: @zamboney I think the use of AST here is cool, but you introduce more noise than value, and we really want to converge in this PR.
For example, how does this behave if the user's file is not a valid JS/TS file? Does this library support TS? Is there an advantage to using this library over anything else from https://astexplorer.net/ ?
The usage we need is really basic. You might as well just look for a regex just like we do in commonJS. And detecting commonJS in the plugins file is IMO just a matter of looking for `module.exports`.

I think it's fine to push this route in the future, but if you do, I expect a rich suite of tests.
Please, let's focus on the core functionality in this PR.

js/packages/eyes-cypress/types/index.d.ts
```diff
@@ -0,0 +1,52 @@
...
+       * })
+       */
+      eyesGetAllTestResults(): Chainable<CypressTestResultsSummary>
+    }
+  }
+}
+
+import exposeDefault from './expose'
+export default exposeDefault
+export {eyesPlugin} from './expose'
```
amitzur, line 52: @zamboney no need for this with the new proposal

js/packages/eyes-cypress/types/index.d.ts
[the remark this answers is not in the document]
amitzur: ok. we need to make sure it doesn't brake in future versions

js/packages/eyes-cypress/test/e2e/ts/cypress/tsconfig.json
amitzur: no, no - if this requires the user to add these flags then we need to change how we do things. We can't upgrade a major version.

js/packages/eyes-cypress/test/e2e/ts/cypress/tsconfig.json
amitzur: @zamboney This is not good. It means the user needs to do this as well.
Use `import type Module from 'module'` in order to use `Module`, instead of requiring the user to import the node types.

js/packages/eyes-cypress/test/e2e/ts/cypress/plugins/index.ts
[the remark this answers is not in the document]
amitzur: @zamboney I honesty don't understand where you see this sort of syntax used, including the link you referenced.

js/packages/eyes-cypress/src/plugin/eyesPlugin.ts
amitzur: @zamboney I don't undertand

js/packages/eyes-cypress/src/plugin/eyesPlugin.ts
```diff
@@ -0,0 +1,69 @@
+import fs from 'fs'
+import makeGlobalRunHooks from './hooks'
+import {type EyesCypressAction} from './hooks'
+import {type EyesPluginConfig} from './'
+
+export function makeEyesPlugin({startServer, eyesConfig, logger}: any) {
```
amitzur, line 6: @zamboney I don't know why you didn't touch this file, but it should be completely changed. Now **in the new API** you have access to the `on` method in both ESM and CJS.
The separation I talk about is that `pluginExport` is responsible to find the `on` and `config`, and then pass them to `eyesPlugin`, which should be in this file, and act identically in all cases.
Currently `pluginExport.ts` has a function called `setupNodeEvents` inside, which is essentially the `eyesPlugin`. Please refactor that function out of `pluginExport` and put it in this file.
There is no more need to have a named export called `eyesPlugin`. Only a default export.

js/packages/eyes-cypress/src/plugin/pluginExport.ts
```diff
@@ -0,0 +1,118 @@
+import isGlobalHooksSupported from './isGlobalHooksSupported'
+// @ts-ignore
+import {presult} from '@applitools/functional-commons'
+import makeGlobalRunHooks from './hooks'
+import {type EyesPluginConfig} from './'
+
+export default function makePluginExport({startServer, eyesConfig}: any) {
+  return function pluginExport(pluginInitArgs: Cypress.ConfigOptions | NodeJS.Module) {
```
amitzur, line 8:
```
import type Module from 'module'
//...
return function pluginExport(pluginInitArgs: Cypress.ConfigOptions | Module) {
```

This way you won't require your users to import the node builtin types.

js/packages/eyes-cypress/src/plugin/eyesPlugin.ts
zamboney: i remove this file... no need it anymore

amitzur:
@zamboney I think given the magnitude of this PR, we can't settle for "tests are stuck on Node 12 and 14".
We can remove Node 12 since it's end of life. Let's switch that to Node 18. But everything needs to be green in order to release this PR. I would add that we should also safe-proof it with some manual testing :scream:

js/packages/eyes-cypress/src/plugin/pluginExport.ts
zamboney: @amitzur it's not working...

zamboney: not working...

zamboney:
@amitzur after running the "cypress publish" `github` action, i will try to explain why we need `index.js` in the root of the project and how it solves the `"esModuleInterop": true` problem.

once set to  `"esModuleInterop": true` it's try (among other stuff it does) to get the `default` value of an object once `require` it (for example, in the legacy cypress [here](https://github.com/applitools/eyes.sdk.javascript1/blob/1529-cypress-typescript/js/packages/eyes-cypress/test/e2e/ts/cypress/plugins/index.ts)). this works once we remove the root `index.js` and move it to `./dist/plugin/index.js` because there `tsc` (using `ttsc`) expose the code like so:`exports.default = pluginExport;`.

this will all work fine but we **still need to support** the legacy `cypress@9`. and once I run the `e2e` tests it start throwing error because we are `require` it without the `default` like [here](https://github.com/applitools/eyes.sdk.javascript1/blob/fc3467a216ae8b2fbfbbe7dd5319437e595265ce/js/packages/eyes-cypress/test/fixtures/testApp/cypress/plugins/index-run.js#L4).

To fix the `"esModuleInterop": true` and keep the legacy code without breaking changes, i added the `index.js` file that export, by default, the `module.exports = expose.default` also `Object.defineProperty(module.exports, 'default', {value: expose.default})` so that if the client does not set `"esModuleInterop": true` it will work

I hope it's makes sense ✌️

amitzur (approved):
This is a great milestone!

I'm glad the user doesn't need to add `esModuleInterop` to their TS config.
Please remove the ESM detection with AST part from eyes-setup.

So to recap
- There **is** a requirement to have `"node"` types in tsconfig. But we are fine with it since it's also documented by Cypress itself: https://docs.cypress.io/guides/tooling/typescript-support#Configure-tsconfigjson
- There **is** a requirement to have `moduleResolution` set to `node` or `node16`, and we are fine with it since `node` is a recommended practice in Cypress. We will have motivation to solve this only if we hear about it from a customer/PoC.
- This PR adds the ability to test Cypress with the `Test packages` workflow.
- Leftovers:
  a) Reduce test time
  b) Refactor `eyesPlugin` out of `pluginExport`
  c) Improve setup script with AST analysis
  d) Test on future version of Cypress
  e) Add proper types everywhere there's `any`
 
Anything I forgot?

Great work @zamboney ! This was a long one :)

.gitignore
```diff
@@ -3,3 +3,4 @@
 .history
 .DS_Store
 Session.vim
+js/packages/@applitools/
```
kyryloonufriiev, line 6: If you need something like that, you should add it in js' `.gitignore`

PR #1177, opened 2022-11-24
update tests results server url

Artem-applitools:
In the upcoming month we are going to update test results server url 
Here are updates for the python and JS SDKs 
@kyryloonufriiev Update need new release for the sdk.coverage.tests package as well.

### DIFF
--- a/js/packages/bongo/src/send-report/send-report-util.js
+++ b/js/packages/bongo/src/send-report/send-report-util.js
@@ -10,11 +10,11 @@ async function _send({uri, payload}) {
 }
 
 async function sendReport(payload) {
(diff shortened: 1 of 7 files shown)


kyryloonufriiev (approved):
Looks good to me

kyryloonufriiev (commented):
BTW @ikseek python sdks have its own reporting implementation? Do you have any objections to using the same js tool you already have to generate your tests?

ikseek (approved):

ikseek:
@kyryloonufriiev I don't really know why we have two reporting tools and when they're used. Reporting hook implemented in pytest_reporting.py and test_integration.py is used to report non-generated test run results. But I have no idea if that data is actually used or where.

PR #1179, opened 2022-11-30
[driver] Replace spec driver method to set value of the input

kyryloonufriiev:
## Problem
In this [Trello](https://trello.com/c/dQqN02TN/1451-suncorp-unable-to-take-full-page-screenshot-for-android-app-using-appium) we faced an issue of something working fine on wdio but working poorly with Java (means universal). After inspecting the logs I found out we have a critical problem in spec driver implementation, the `type` method that we use to set a string value in the input element works differently in wdio and webdriver (the one we use in universal) spec drivers. Wdio implementation by default always clears input before setting a new string and webdriver implementation simply appends the new value to whatever is already in input.

## Solution
With this PR I removed `type` method from the spec driver and added a new method called `setElementText` (nice symmetry with already existing `getElementText`). This method by definition should clear the input when the new value is given. 
The new logic is covered with spec driver tests in respective framework implementations.


js/packages/spec-driver-puppeteer/src/spec-driver.ts
```diff
@@ -156,6 +156,11 @@ export async function findElements(frame: Context, selector: Selector, parent?:
   const root = parent ?? frame
   return (isXpathSelector(selector) ? root.$x(selector) : root.$$(selector)) as Promise<Element[]>
 }
+export async function setElementText(frame: Context, element: Element | Selector, text: string): Promise<void> {
+  if (isSelector(element)) element = await findElement(frame, element)
+  await element.evaluate(element => ((element as any).value = ''))
```
zamboney, line 161: why `any`?

js/packages/eyes-universal/src/spec-driver/webdriver.ts
```diff
@@ -291,9 +298,6 @@ export async function click(driver: Driver, element: Element | Selector): Promis
   if (isSelector(element)) element = await findElement(driver, element)
   await driver.elementClick(extractElementId(element))
 }
-export async function type(driver: Driver, element: Element, value: string): Promise<void> {
-  await driver.elementSendKeys(extractElementId(element), value)
-}
```
zamboney, line 296: you sure about it? maybe keep it but add a deprecation warning?

zamboney (approved):
i understand the issue but i will keep the `type` method. still approving it because this not that important. let me know what you think

js/packages/spec-driver-puppeteer/src/spec-driver.ts
kyryloonufriiev: Makes sense. Addressed!

zamboney:
and what about the `selenium` and `universal` tests? you think we should fix them?

kyryloonufriiev:
@zamboney What do you mean?

kyryloonufriiev:
Get what you meant. I did fix for universal types. Everything else is just a flakiness, and changes in selenium I made unintentionally (just updated the deps there)

PR #1181, opened 2022-12-01
[core] fix switch context before `openEyes`

zamboney:
On the ingeration with `testim` we found that they change the frame before
`openEyes` and then try to capture region inside the frame.

Before the `@applitools/core` change its work

After the `@applitools/core` change its thrown an error

```
StaleElementReferenceError: stale element reference: element is not attached to the page document
  (Session info: headless chrome=107.0.5304.87)
      at Object.throwDecodedError (node_modules/selenium-webdriver/lib/error.js:522:15)
      at parseHttpResponse (node_modules/selenium-webdriver/lib/http.js:589:13)
      at Executor.execute (node_modules/selenium-webdriver/lib/http.js:514:28)
```

We found that this cause because of `driver.getViewportSize`([in classic](https://github.com/applitools/eyes.sdk.javascript1/blob/101ba54b38385e15428ceb7486c7fc2448067503/js/packages/core/src/classic/open-eyes.ts#L41) and [in ufg](https://github.com/applitools/eyes.sdk.javascript1/blob/101ba54b38385e15428ceb7486c7fc2448067503/js/packages/core/src/ufg/open-eyes.ts#L43)). This PR will fix it by saving the `driver.currentContext` in the beginning of `openEyes` and then `focus` this context after.

### DIFF
--- a/js/packages/core/test/e2e/classic/swtich-context-before-openEyes.spec.ts
+++ b/js/packages/core/test/e2e/classic/swtich-context-before-openEyes.spec.ts
@@ -1,0 +1,50 @@
+import {makeCore} from '../../../src/classic/core'
+import {By} from 'selenium-webdriver'
+import * as spec from '@applitools/spec-driver-selenium'
+import assert from 'assert'
+
+describe('switch context before openEyes', () => {
+  let driver, destroyDriver
+
+  before(async () => {
+    ;[driver, destroyDriver] = await spec.build({browser: 'chrome'})
+  })
+
+  after(async () => {
+    await destroyDriver?.()
+  })
+
+  it('works', async () => {
+    const core = makeCore<spec.Driver, spec.Driver, spec.Element, spec.Selector>({spec})
+
+    await driver.get('https://applitools.github.io/demo/TestPages/FramesTestPage/')
+    const frame1 = await driver.findElement(By.css('[name="frame1"]'))
+    await driver.switchTo().frame(frame1)
+    const innerFrameDiv = await driver.findElement(By.css('#inner-frame-div'))
+
+    const eyes = await core.openEyes({
+      target: driver,
+      settings: {
+        serverUrl: 'https://eyesapi.applitools.com',
+        apiKey: process.env.APPLITOOLS_API_KEY,
+        appName: 'core e2e',
+        testName: 'switch context before openEyes',
+        environment: {viewportSize: {width: 700, height: 460}},
+      },
+    })
+
+    await eyes.check({
+      settings: {
+        name: 'layout region screenshot',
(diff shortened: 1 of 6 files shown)


js/packages/core/test/e2e/classic/swtich-context-before-openEyes.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+    await eyes.check({
+      settings: {
+        name: 'layout region screenshot',
+        region: innerFrameDiv,
+        fully: true,
+        ignoreCaret: true,
+        layoutRegions: ['#inner'],
+        matchLevel: 'Strict',
+      },
+    })
```
kyryloonufriiev, line 45: No reason to call `check` in this test. The only thing you care about is that after calling open your context/frame didn't change

kyryloonufriiev (changes requested):
Please rename tests to something more meaningful. I try to describe what, I think is wrong with the current naming. Your test is called `switch context before openEyes`, but this does not describe what you are testing, what the name should be `should preserve original frame after opening`. But I would recommend renaming the file to `open-eyes.spec.ts` (to make a general file for e2e testing in open), and then the test could be called `preserves original frame` or `should preserve original frame`.

js/packages/core/test/e2e/classic/swtich-context-before-openEyes.spec.ts
zamboney: I disagree with you, this is exactly what `testim` try to do and gets the error in the description. Furthermore, if you remove the fix an exact error that `testim` get will throw.

i think this reproduces the issue and only for that, i will keep the test. i could agree that maybe `core` is not the place to test it but i don't minded to keep this test here

kyryloonufriiev: We have to locate an issue and test it in a way that makes sense. Testing checks and close to detect an issue that is happening in open is obsolete.

kyryloonufriiev: Check the current frame before open and after open, they should be the same

zamboney: not sure how to do it, i could not find a way to do it using the `driver` api. i could only think to check the `href` using `driver.execute`  but it feels hacky... do you think of a different way to do it directly from the `driver`?

kyryloonufriiev: This is good enough and it is not hacky. You can also verify if the HTML element before opening and after opening is the same one.

PR #1182, opened 2022-12-01
check that we get a real shadowRoot in add and cleanup ElementIds 

SophieDePaula:
[Trello card](https://trello.com/c/kQT5e0SL/1607-mfs-java-error-when-taking-region-screenshot-or-using-coded-region)

In cases where an element has a fake shadowRoot, we should not construct the path to the element in `addElementIds` and `clearupElementIds`, but instead, only mark the element itself. 
To determine if a document fragment is a real shadowRoot, we will check the constructor name like we do in `dom-snapshot`

TO DO:
- [x] update pending changes for all SDKs


pending-changes.yaml
```diff
@@ -14,6 +14,7 @@ Apollo SDK:
   bug-fix:
     - Fixed issue when helper lib inputs were not cleaned up before calling new command
     - Fixed issue with element scroll position not being restored after screenshot is taken on native platforms
+    - Handle fake shadowRoot with UFG
```
SophieDePaula, line 17: @benbabayoff8

benbabayoff8: @SophieDePaula looks good to me.

js/packages/snippets/src/browser/addElementIds.js
```diff
@@ -1,8 +1,16 @@
 function addElementIds([elements, ids]) {
   return elements.map((element, index) => {
     const path = [element]
-    if (element.getRootNode) {
-      for (let root = element.getRootNode(); root !== document; root = root.host.getRootNode()) {
+    if (
+      element.getRootNode &&
+      element.getRootNode().constructor &&
+      element.getRootNode().constructor.name === 'ShadowRoot'
```
kyryloonufriiev, line 7: You do this check twice. In this if-statemen and in the for-loop. I would remove the one in if-statement, since here you just need to check that this method even exists, it will not exist in ie or other old browsers.

kyryloonufriiev (changes requested):

PR #1187, opened 2022-12-05
[core] fix issue with backward compatibility

zamboney:
If the `concurrency` is `string` that can convert to `number` it's should convert to `number` and not throw an error

[Trello](https://trello.com/c/8aSiNw0F/1330-ufg-browser-extension-error)


kyryloonufriiev:
We are not going to support this. The explanation is in [Trello](https://trello.com/c/8aSiNw0F/1330-ufg-browser-extension-error)

PR #1195, opened 2022-12-07
[core] add support to lazy load with SRE

zamboney:
Passing the scroll root element to the `lazyLoad` snippet.

[Trello](https://trello.com/c/xXhsxgHX/1649-mastercard-lazyload-options-doesnt-work-with-scrollrootelement)

## TODO
- [x] add generic tests [done](https://github.com/applitools/sdk.coverage.tests/pull/174)

## testing

i want to to an `e2e` test and it's can't be done in `screenshoter` because the `lazyLoad` execution don in the `core` package, therefore i add here the [eyes](https://eyes.applitools.com/app/test-results/00000251731872670466) test.


js/packages/snippets/src/browser/lazyLoad.js
```diff
@@ -7,7 +7,7 @@ const scrollTo = require('./scrollTo')
 // NOTE:
 // This snippet uses `scrollTo`, which sets `scroll-behavior` to `auto`,
 // and is called on each scrolling iteration.
-function lazyLoad([{scrollLength, waitingTime, maxAmountToScroll} = {}] = []) {
+function lazyLoad([{scrollLength, waitingTime, maxAmountToScroll, rootElement} = {}] = []) {
```
kyryloonufriiev, line 10: You better don't pass elements like that, it should be a separate element in the array otherwise it could cause problems on some platforms. Also, I would prefer the element to be the first item in the arg array, it is like a convention in the snippets.

js/packages/core/test/e2e/classic/lazy-load.spec.ts
```diff
@@ -32,8 +32,22 @@ describe('lazy load', () => {
...
+        fully: true,
+        lazyLoad: true,
+        hideScrollbars: true,
+        scrollRootElement: {type: 'css', selector: '#sre'},
+        stitchMode: 'Scroll',
+        region: '#sre',
+      },
+    })
+
+    const [result] = await eyes.close({settings: {updateBaselineIfNew: false}})
```
kyryloonufriiev, line 50: We need it for ufg as well

js/packages/snippets/src/browser/lazyLoad.js
```diff
@@ -30,7 +30,7 @@ function lazyLoad([{scrollLength, waitingTime, maxAmountToScroll} = {}] = []) {
         setTimeout(() => {
           try {
             if (doneScrolling) {
-              const {x, y} = scrollTo([undefined, startingScrollPosition])
+              const {x, y} = scrollTo([rootElement, startingScrollPosition])
```
kyryloonufriiev, line 33: We need snippet test for that

kyryloonufriiev (changes requested):
We would also need a generic coverage test for that

js/packages/snippets/test/lazyLoad.spec.js
```diff
@@ -62,19 +62,41 @@ describe('lazyLoad', () => {
...
       const transactionHistory = result.value
-      console.log(transactionHistory)
       const scrolledHeight = transactionHistory[transactionHistory.length - 2].y
       const resetScroll = transactionHistory[transactionHistory.length - 1]
       assert.deepStrictEqual(scrolledHeight, options.maxAmountToScroll)
       assert.deepStrictEqual(resetScroll.x, 0)
       assert.deepStrictEqual(resetScroll.y, 0)
     })
+
+    it('works on a page with a innser scroll root element', async () => {
```
kyryloonufriiev, line 78:
```suggestion
it('works on a page with a custom scrolling element', async () => {
```

js/packages/core/src/utils/wait-for-lazy-load.ts
```diff
@@ -20,14 +20,15 @@ export async function waitForLazyLoad<TDriver extends Driver<unknown, unknown, u
   logger: Logger
 }) {
   logger.log('lazy loading the page before capturing a screenshot')
+  const sre = await driver.currentContext.getScrollingElement()
```
kyryloonufriiev, line 23: `sre` - stands for `scroll root element` this terminology is not used in our core or the driver for that matter. It should be `scrollingElement` - this is the terminology we use here. Also, abbreviations are less readable.

js/packages/core/test/e2e/classic/lazy-load.spec.ts
kyryloonufriiev: But this is exactly the first step. This feature could be beneficial for ufg.

kyryloonufriiev (changes requested):

js/packages/core/test/e2e/classic/lazy-load.spec.ts
zamboney: [./src/ufg/check.ts](https://github.com/applitools/eyes.sdk.javascript1/blob/1649-mastercard-lazyload-options/js/packages/core/src/ufg/check.ts) not set sre like [classic](https://github.com/applitools/eyes.sdk.javascript1/blob/ad4562ced7f0ca0dd25134dbff95461f287980bf/js/packages/core/src/classic/check.ts#L39) do... so it's not [working](https://eyes.applitools.com/app/test-results/00000251731793713511?accountId=UAujt6tHnEKUivQXIz7G6A~~&display=details&top=00000251731793713511%2830%29) as expected

PR #1196, opened 2022-12-08
[eyes-universal] add support to `node14-linux-arm64`

zamboney:
Adding new env for `arm64`

## TODO

- [x] revert REVERT IT!!!!


kyryloonufriiev (approved):

kyryloonufriiev (changes requested):
I did approve it. But on second thought I understand that we don't have a test for that. We usually test new binaries with a gh action matrix, I'm not sure if you can run arm + Linux there, but worth an investigation. Also, I don't get how the newly generated file will be named. `eyes-universal-linux-arm64`?

zamboney:
About the name, it's generated by `pkg` and it's `eyes-universal-linux-arm64` under the `bin` folder. I will check the `gh action` and get back to you. Thanks for the feedback.

zamboney:
@kyrylo 

there are no `arm64` image as part of the [running-images](https://github.com/actions/runner-images) that `github action` support, so as far as i can see it there are 2 options to add the test you want:

  1. add a new ["self-hosted" runner](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners) that will run on `os: arm64`
  2. add a new `gh action` that will run on `os: ubuntu-latest` and `container: amd64/node` [link](https://hub.docker.com/r/amd64/node/)

The first options not recommenced as they write in there docs:
> Warning: We recommend that you only use self-hosted runners with private repositories. This is because forks of your public repository can potentially run dangerous code on your self-hosted runner machine by creating a pull request that executes the code in a workflow.

The second options is more workaround as it will require to build the docker image with the `arm64` binary and run it on **none** `amd64` machine. This is exactly what `pkg` done so i think is fine

WDYT?

kyryloonufriiev:
So the second option looks like exactly what we are doing today with alpine test, no? If so please add it to the workflow

zamboney:
So the second options not working, `github action` [not supporting](https://github.community/t/running-docker-container-on-arm64-host/128624/2) running `amd64` container on `arm64` host. I will try to find another way to test it. Thanks for the feedback.

kyryloonufriiev:
I don't think you need to spend any more time on that. We can skip the testing for now. The assumption was that it might be very easy to add a test but since it takes more time that I was expecting we can skip the testing

zamboney:
i finally figured out how to run the tests! If you take a look at the `github action` `workflow` i change and let me know if this is a good fit...

js/packages/eyes-universal/package.json
```diff
@@ -46,10 +46,14 @@
   "scripts": {
     "lint": "eslint '**/*.ts'",
     "build": "yarn build:dist && yarn build:bin",
+    "build:x64": "yarn build",
+    "build:arm64": "yarn build:dist && yarn build:bin:arm64",
     "build:dist": "tsc",
     "build:bin": "pkg .",
-    "build:zip": "zip -j ./bin/eyes-universal.zip ./bin/eyes-universal-alpine ./bin/eyes-universal-linux ./bin/eyes-universal-macos ./bin/eyes-universal-win.exe",
-    "build:tgz": "tar -czf ./bin/eyes-universal.tar.gz -C ./bin ./eyes-universal-alpine ./eyes-universal-linux ./eyes-universal-macos ./eyes-universal-win.exe",
+    "build:bin:arm64": "pkg -t node14-linux-arm64 -o bin/eyes-universal-linux-arm64 .",
```
kyryloonufriiev, line 53: I think we should re-name stuff here. `yarn build` is always should be something complete. It was dist + bin before, but now you cannot get complete build of binaries at all. So I would recommend `yarn build` being simply `yarn build:dist`, then you have `yarn build:bin:x64` and `yarn build:bin:arm64` that build binaries for specific platforms (BTW you can remove config for `pkg`, since there two build modes it is just confusing at this point), then you can have `yarn build:x64` and `yarn build:arm64` just to combine dist + bin:{arch}.

.github/workflows/publish-universal.yml
```diff
@@ -130,18 +209,88 @@ jobs:
...
+          name: eyes-universal-bin-${{needs.prepare.outputs.version}}-arm64
+          path: ./eyes-universal-arm64
+      - name: Bundle artifacts
+        shell: bash
+        working-directory: ./eyes-universal
+        run: |
+          find ../eyes-universal-arm64 -type f -not -name '*.zip' -not -name '*.tar.gz' -exec bash -c 'cp ${0} ./bin/$(basename ${0})' {} \;
+          rm -rf ./bin/*.tar.gz ./bin/*.zip
+          zip -j ./bin/eyes-universal.zip $(find ./bin -type f -not -name '*.zip' -not -name '*.tar.gz' | xargs)
+          tar -czf ./bin/eyes-universal.tar.gz $(find ./bin -type f -not -name '*.zip' -not -name '*.tar.gz' | xargs)
```
kyryloonufriiev, line 283: why you don't use `yarn` commands? The point of this job is also lost to me. Why you cannot create archives at the point you really need them, in publish job?

.github/workflows/publish-universal.yml
```diff
@@ -130,18 +209,88 @@ jobs:
...
+            esac
+          run: |
+            cd /src
+            echo //registry.npmjs.org/:_authToken=${{secrets.NPM_TOKEN}} > $HOME/.npmrc && yarn install && rm -rf $HOME/.npmrc
+            yarn test:e2e
+
       - name: Run tests
+        env:
+            arch: ${{matrix.arch}}
+        if: ${{env.arch == ''}}
```
kyryloonufriiev, line 253: Isn't it possible to simply check for `matrix.arch` in `if`?

kyryloonufriiev: Also `whatever.arch == ''` is a bit confusing you can simply add `x64` where you use this arch, so it will be more self-explanatory.

.github/workflows/publish-universal.yml
```diff
@@ -130,18 +209,88 @@ jobs:
...
+                ;;
+              fedora*)
+                echo not supported
+                exit 1
+                ;;
+              alpine*)
+                echo not supported
+                exit 1
+                ;;
+            esac
```
kyryloonufriiev, line 244: I understand you copied it from the example for that action, but can we simplify this thing for our use? I don't think we care about anything that is not ubuntu.

.github/workflows/publish-universal.yml
kyryloonufriiev: Also if you have `arch` for everything you will be able to avoid having `arch-group` since it basically always should be the same.

.github/workflows/publish-universal.yml
```diff
@@ -56,20 +56,86 @@ jobs:
...
+        run: |
+          echo //registry.npmjs.org/:_authToken=${{secrets.NPM_TOKEN}} > $HOME/.npmrc && yarn install && rm -rf $HOME/.npmrc
+          yarn build:${{matrix.name}} 
+          yarn build:zip && yarn build:tgz
+      - uses: uraimo/run-on-arch-action@v2
+        if: ${{matrix.arch}}
+        name: Build ARM
+        with:
+          env: |
+            artifact_name: universal-${{ matrix.distro }}_${{ matrix.arch }}
```
kyryloonufriiev, line 100: Do you use it?

kyryloonufriiev (changes requested):
Well done. I wrote some small comments, but I like the idea, I would love to reuse our action for testing but this solution is also good enough.

.github/workflows/publish-universal.yml
zamboney: yes...

.github/workflows/publish-universal.yml
zamboney: np i will

.github/workflows/publish-universal.yml
zamboney: i don't want to have a long command in the `package.json` `scripts` but i can move it to `sh` file in the `workflow` if it's suite your need

js/packages/eyes-universal/package.json
zamboney: i don't understand what you are want me to do here please write it as snippet

js/packages/eyes-universal/package.json
```diff
@@ -59,21 +63,18 @@
     "docker:setup": "node ../scripts/scripts/generate-docker-compose-config.js && docker-compose up -d",
     "docker:teardown": "docker-compose down",
     "deps": "bongo deps",
-    "gh:publish": "gh workflow run publish-universal.yml --ref $(git rev-parse --abbrev-ref HEAD)",
-    "preversion": "bongo preversion --verify-pending-changes --skip-deps",
-    "version": "bongo version --with-pending-changes",
-    "postversion": "bongo postversion --skip-release-notification"
```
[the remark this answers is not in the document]
zamboney: my bad i will revert it

.github/workflows/publish-universal.yml
```diff
@@ -113,6 +87,10 @@ jobs:
           - name: macos
             os: macos-latest
             shell: bash
+          - name: linux-arm64
+            os: macos-latest
+            container: arm64v8/node
+            shell: bash
```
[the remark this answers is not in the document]
zamboney: no reason i will change it back

.github/workflows/publish-universal.yml
zamboney: you need to unify the `x64` and the `arch64` artifact into one artifact and i want to keep the `publish` step as it was. I think of the "publish" step as a step to publish resources to create or modify them...

.github/workflows/publish-universal.yml
```diff
@@ -68,32 +68,6 @@ jobs:
...
-        uses: actions/download-artifact@v3
-        with:
-          name: eyes-universal-bin-${{needs.prepare.outputs.version}}
-          path: ./js/packages/eyes-universal
-      - name: Run tests
-        uses: ./.github/actions/tests
-        with:
-          working-directory: ./js/packages/eyes-universal
-          build-type: none
-          test-type: it
```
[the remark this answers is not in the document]
zamboney: ?

.github/workflows/publish-universal.yml
kyryloonufriiev: You already have yarn commands to create archives, and 1st and 2nd commands are looks redundant to me. I understand the intent to separate bundling from publishing, it would be nice if it cost us nothing, but in this case, it causes another job to spin up for no actual reason. I don't see anything wrong with creating archives from artifacts in the separate step of the publishing job.

zamboney: i need the first line to move the `arm64` artifact to the `x64` files and the 2nd line to clean old `zip` and `tar.gz` files that were created in the `x64` bundle.

About the time it takes, take a look [here](https://github.com/applitools/eyes.sdk.javascript1/actions/runs/3706224633/jobs/6281322052) it took 1s to set up the job. all of the steps, if it is done on `bundle` or in `publish`, will take the same time, so i don't see any reason to change its location

js/packages/eyes-universal/package.json
kyryloonufriiev: something like that
```suggestion
    "build": "yarn build:dist",
    "build:dist": "tsc",
    "build:x64": "yarn build:dist && yarn build:bin:x64",
    "build:arm64": "yarn build:dist && yarn build:bin:arm64",
    "build:bin:x64": "pkg . --target node14-linux-x64,node14-alpine-x64,node14-macos-x64,node14-win-x64",
    "build:bin:arm64": "pkg . --target node14-linux-arm64 --output ./bin/eyes-universal-linux-arm64",
```

zamboney: i don't see any reason to change it this way... i think that the old way (a.k.a. using the `pkg` section) still more straight forward

kyryloonufriiev: I think it is confusing. When someone will find pkg config that has `targets` by default you are thinking that this is all of the targets we are going to use.

zamboney: i don't think so... i think that this is the `x64` config and you can understand it by the naming of the script

PR #1198, opened 2022-12-09
Python tests integration into js workflow

ikseek:


kyryloonufriiev (commented):
Can you use this wf in your publishing workflow? I am asking because I don't see that a `test-type` argument is used in any way. Is it possible to create test types in python? Also, it looks to me that using this action I cannot run tests in images SDK or upcoming playwright SDK.
It is meant to be flexible. Just the problems I spotted so far:
1. No way to specify a testing type
2. No way to run in a different working directory in order to test a different sdk
3. It always builds universal and uses those binaries, while it also should be configured.

ikseek:
@kyrylodolynskyi I apologize for missing your comment! I fixed my mail filters not to miss PR comments again.

0. I can't use this action in my publishing workflow because I don't really need to build any JS packages in python workflow, it will just add complexity without solving any problems.
1. I used the test type in the script to select one of the test types python SDK has.
2. Switching directories is unnecessary when you run python tests. I left those additional github action arguments because your package building code uses them, and I don't fully understand if and why they are needed.
3. What configuration options do you see regarding the use of built binaries?

Thinking about what we're doing here I see this approach as troubling for these reasons:
- You want me to write (and maintain?) the code that builds your packages in the environment you control. I have no idea how to do that in any predictable way.
- Even if we end up with working GitHub actions script that will build both JS and then python packages, what will you do when JS code regression will trigger a failure in python code?
You will need to be able to get logs and I'm sure you'd like to run it locally to attach a debugger to see what happens in runtime. You will have to dig CI scripts looking for the sequence of commands needed to be executed to replicate the failure. Good thing in python it's just one 'make' command (which still lacks particular test selection option, I will add it). But what about the next SDKs that will move here?
I believe that code that knows how to execute test suites and build their dependencies SHOULD NOT be part of CI configuration. We should instead aim for scripts (or a tool) that can be executed locally with just a single command that will build everything needed to build and run the needed test using it's testing framework.

ikseek:
I changed the test action so it doesn't do any binary building but expects them to be built and provided by the caller.

PR #1199, opened 2022-12-11
[core] fix default `overlap`

zamboney:
Once pass an empty object to `overlap`, the default value of `top` and `bottom` are not set. This pr will set a default `top` and `bottom` for `overlap` even if an empty `object` is passed to the `overlap` settings

## Testing

Add new `unit` tests in `test/unit/check.spec.ts` test.


js/packages/core/src/check.ts
```diff
@@ -10,6 +10,8 @@ type Options<TDriver, TElement, TSelector> = {
   logger?: Logger
 }
 
+export const defaultOverlap = {top: 10, bottom: 50}
```
kyryloonufriiev, line 13: No need for this variable. Just put those values inline in the only place you need them

js/packages/core/test/unit/check.spec.ts
```diff
@@ -41,4 +45,42 @@ describe('check', () => {
...
+    })
+
+    it('should not set default overlap when pass an object with top and bottom', async () => {
+      const bottom = parseInt(Math.random() * 100 + defaultOverlap.bottom + '')
+      const top = parseInt(Math.random() * 100 + defaultOverlap.top + '')
+      await test({overlap: {bottom, top}}, ({settings}) => {
+        assert.deepStrictEqual(settings.overlap, {...defaultOverlap, bottom, top})
+      })
+    })
+  })
```
kyryloonufriiev, line 85: Please do not overcomplicate the tests. I saw this problem too often, those tests that write "smart" is tough to maintain later. So
- Do not invent type aliases without a real need
- No need to calculate offsets simply set some custom values.
- Hardcode defaults in the tests not export them from somewhere

In addition, I'm not against the `test` function in general, but introducing it just for a small subgroup of tests just causes problems in long term, so I would prefer to not have it at all. Because toy build it for your needs in this particular scenario without thinking about other scenarios, and tomorrow someone else will do the same but for his/her needs and we will end up with code that is written with personal preferences of everyone and no one wants to maintain.
I also think that everything here could be placed in one simple test that will check three different ways to pass overlap, the same as it is done in the test `should calculate default value for sendDom using other settings`, this way you will eliminate the need of test function and the test itself still will be pretty clear and the most important **small**.

kyryloonufriiev (changes requested):

js/packages/core/test/unit/check.spec.ts
zamboney: i don't like this approach but if this is important to you then i will change it

js/packages/core/src/check.ts
zamboney: I need this for testing [here](https://github.com/applitools/eyes.sdk.javascript1/blob/558e6f3a4ad103fa15034961ba2f4b676abf3c0b/js/packages/core/src/check.ts#L13)

kyryloonufriiev: Ok, so why do you export this one but not other default values? I don't like the direction where we decide on exporting whatever we think we need, this way we will end up with a messy code with no style

js/packages/core/test/unit/check.spec.ts
kyryloonufriiev: I understood your point just by looking at the implementation, but this is not the way to do it. You either do it in a generic way (it might be partial) or don't do it at all. I don't see us having a problem with maintenance overhead due to lots of changes in default values ever. But having a source where you can read a default value for a property seems nice, but again it should be a generic design with all of the methods and properties in mind.

js/packages/core/src/check.ts
zamboney: I export this one and not other default values because I need it in the test. When i will need anther default value in other test i will expose it as well.

kyryloonufriiev: This is how you end up with a very bad design and the need to refactor it. If we agree on the need to have defaults statically readable we should design it first. Your approach may be a proposed design, but this is not the only one and it is not even the one that will meet your needs the best (I mean the needs of the testing).

zamboney: Just by `export` default value form where it's used, it's not a bad design. Maybe in the future, you can tell that on every exporting that export something internal. Our job is to force that its use for good practice. 

in general, every `export` that not use for external use can become a "bad design" but in this case, I could not figure out a way to use a default value within the code without duplicating the value that, as far as I can see, it is bad practice. Export the default value still seems reasonable to me.

kyryloonufriiev: Let me draw a line here. This is what we are not going to do. 
> I export this one and not other default values because I need it in the test. When i will need anther default value in other test i will expose it as well.

Because this will be a mess. Right now we have defaults set within a function with its internal logic (not saying this is ideal). You are trying to bring order in this area, and I would love that, but please if you going to do this should be a solution for all of the defaults with types, naming, and maintenance in mind. The solution that you propose is not good because it has no structure. Let me give you an example, you saying you will export `defaultOverlap` with the default value for overlap. But how would I know how to scale it for another property? What if for `sendDom` I will export `sendDomDefault`, nothing will prevent me from doing that, then we will end up with random naming for random properties, also it is not clear on which level I should create the "default" variable, in other words, why you didn't create `defaultOverlapTop` and `defaultOverlapBottom`.

The much better design would be to accept a typed object with default properties in the constructor of the function. This will eliminate the problem with tests because you will be able to configure the default value in the test without a need to export something from somewhere. This will also eliminate the problem of unclear naming and tons of exports because you will have no additional exports, and the default values are structured in statically typed object.

zamboney:
@kyryloonufriiev, to be clear, you prefer to have **the same value** in **two default locations**, one in the code and the other in the test. To me, it doesn't make sense, but because you insist and you think this is "bad practice" i will do it as you please

zamboney:
@kyryloonufriiev i did what you ask, can you review it again?

PR #1201, opened 2022-12-15
Fix extraneous imports in core

amitzur:
The rule `node/no-extraneous-import` should probably be added to all packages


zamboney (approved):
👍 good catch

amitzur:
I added the same version number as other packages that probably cover for this bug:
`chalk` --> imported by `logger`
`node-fetch` --> imported by `req`

PR #1205, opened 2022-12-16
Self-healing support

tourdedave:
In the SDK & eg-client.

[Trello](https://trello.com/c/6LXTxGof)

### TODO
- [x] eg-client: pass the capability to enable self healing
- [x] eg-client: parse, store, and make available the self healing data returned from the EG (e.g., as a new driver command)
- [x] core-base: transform data from EG and add support for new request to Eyes for self healing data
- [x] driver & relevant spec-drivers: add support for new driver command to export self healing data from the eg-client (a.k.a. session metadata)
- [x] core: glue everything together (e.g., call to driver for the session metadata, and passing it to core-base)
- [x] add e2e test coverage for full user flow & verify (classic)
- [x] add e2e test for classic abort
- [x] add e2e test coverage for full user flow & verify (ufg)
- [x] see if the self healing report can be sent after close/abort
- [ ] determine if the driver is running on the eg, and populate the egSessionId in the Eyes request


js/packages/core-base/src/types.ts
```diff
@@ -122,6 +124,7 @@ export interface OpenSettings extends ServerSettings {
   abortIdleTestTimeout?: number
   connectionTimeout?: number
   removeSession?: boolean
+  selfHealingEnabled?: boolean
```
kyryloonufriiev, line 127: @tourdedave What is the meaning of this flag? Is it sdk telling the server that self-healing is enabled in the current session?

js/packages/execution-grid-client/src/proxy-server.ts
```diff
@@ -67,6 +70,26 @@ export function makeServer({
         return await createSession({request, response, logger: requestLogger})
       } else if (request.method === 'DELETE' && /^\/session\/[^\/]+\/?$/.test(request.url)) {
         return await deleteSession({request, response, logger: requestLogger})
+      } else if (useSelfHealing && request.method === 'POST' && /element/.test(request.url)) {
```
kyryloonufriiev, line 73: Do you think this is the right regexp? there are lots of requests that have `element` in them. I believe we should use regexp for full request `^\/session\/[^\/]+\/element\/?` Also what about `elements`?

js/packages/execution-grid-client/src/proxy-server.ts
```diff
@@ -67,6 +70,26 @@ export function makeServer({
...
+        const {appliCustomData} = await proxyResponse.json()
+        if (appliCustomData?.selfHealing) {
+          requestLogger.log('Self-healed locators detected', appliCustomData.selfHealing)
+          metadata.push(appliCustomData.selfHealing)
+        } else {
+          requestLogger.log('No self-healing metadata found')
+        }
+        proxyResponse.pipe(response)
+        return
+      } else if (useSelfHealing && request.method === 'GET' && /metadata/.test(request.url)) {
```
kyryloonufriiev, line 90: I think the regexp should also be full here. Also to follow the webdriver spec we better call it `/session/:sessionId/applitools/metadata`, this way we will never mess up with some standard commands also this is a very popular practice.

js/packages/core-base/src/types.ts
tourdedave: No, I think I added it here by mistake.

js/packages/execution-grid-client/src/proxy-server.ts
tourdedave: Agreed, I was just trying to get something working to demonstrate the flow. I intended to improve it to something more precise before taking the PR out of draft. Happy to go with your recommendation.

Supporting `elements` is out of scope.

js/packages/execution-grid-client/src/proxy-server.ts
tourdedave: 💯

js/packages/execution-grid-client/src/proxy-server.ts
kyryloonufriiev: Really? I was under impression that we are going to support `elements`.

js/packages/core-base/src/open-eyes.ts
```diff
@@ -62,6 +62,7 @@ export function makeOpenEyes({requests, cwd = process.cwd(), logger: defaultLogg
         aborted.reject(new Error('Command "check" was aborted due to possible error in previous step'))
         return abort(options)
       }),
+      reportSelfHealing: eyesRequests.reportSelfHealing,
```
kyryloonufriiev, line 65: no need to do that. As you can see here we already extending the `eyesRequests` object so you will get this method "for free"

js/packages/core-base/src/server/requests.ts
```diff
@@ -55,6 +55,7 @@ export interface EyesRequests extends Eyes {
   extractText(options: {target: ImageTarget; settings: ExtractTextSettings; logger?: Logger}): Promise<string[]>
   close(options?: {settings?: CloseSettings; logger?: Logger}): Promise<TestResult[]>
   abort(options?: {logger?: Logger}): Promise<TestResult[]>
+  reportSelfHealing(options: {report: any; logger?: Logger}): Promise<void>
```
kyryloonufriiev, line 58: You should add this method to Eyes interface

js/packages/core/src/automation/types.ts
```diff
@@ -54,6 +54,7 @@ export interface Eyes<TDriver, TContext, TElement, TSelector, TTarget = DriverTa
     settings: MaybeArray<ExtractTextSettings<TElement, TSelector>>
     logger?: Logger
   }): Promise<string[]>
+  reportSelfHealing?(options: {report: any; logger?: Logger}): Promise<void>
```
kyryloonufriiev, line 57: It should be in BaseCore.Eyes interface, then you will not have to add it here

js/packages/core/src/close.ts
```diff
@@ -23,6 +28,11 @@ export function makeClose<TDriver, TContext, TElement, TSelector, TType extends
     settings = {...config?.close, ...settings}
     settings.updateBaselineIfNew ??= true
     const typedEyes = await eyes.getTypedEyes({logger})
+
+    const driver = isDriver(target, spec) ? await makeDriver({spec, driver: target, logger}) : null
+    const report = await driver.getSessionMetadata()
+    await typedEyes.reportSelfHealing(report)
```
kyryloonufriiev, line 34: I don't think doing it here is the right way to go for a few reasons. First of all, on this level, we should not have any actual business logic. This feature is pretty much a feature of classic and ufg eyes separately. Also this way you will have to create a custom implementation for ufg, since it has multiple eyes sessions running and you will have to do it for each of them.

kyryloonufriiev (commented):
@tourdedave Let's sync on it today, I have some thoughts to propose, and I believe it could simplify the implementation.

js/packages/core-base/src/types.ts
```diff
@@ -247,16 +249,26 @@ export interface ExtractTextSettings<TRegion = Region> extends ImageSettings<TRe
   userCommandId?: string
 }
 
+export type SelfHealingReport = {
+  operations: [
+    timestamp: string,
+    old: string,
+    new: string
+  ]
```
kyryloonufriiev, line 257: Is it indeed an array, and not an object?

js/packages/core-base/src/server/requests.ts
```diff
@@ -503,6 +505,7 @@ export function makeEyesRequests({
       logger.log(`Request "close" called for test ${test.testId} that was already stopped`)
       return null
     }
+    await reportSelfHealing(settings)
```
kyryloonufriiev, line 508: I think you should preserve the argument structure of other functions here -> `{settings, logger}`. I also see that you expect to get a `logger` in the `reportSelfHealing` function, but this way you will never pass it.

js/packages/core/src/classic/close.ts
```diff
@@ -0,0 +1,33 @@
...
+  logger: defaultLogger
+}: Options<TDriver, TContext, TElement, TSelector>) {
+  return async function ({
+    settings,
+    logger = defaultLogger,
+  }: {
+    settings?: CloseSettings
+    logger?: Logger
+  } = {}): Promise<TestResult[]> {
+    const driver = await makeDriver({spec, driver: target, logger})
```
kyryloonufriiev, line 27: If this is an image session there will be no driver. You should use `isDriver` function to check and then `makeDriver` only if the target is indeed a driver.

js/packages/core-base/src/types.ts
tourdedave: An array of objects, based on [this](https://trello.com/c/6LXTxGof/107-self-healing-in-eyes#comment-631489149dbf72031737a5e9).

js/packages/core-base/src/types.ts
```diff
@@ -247,16 +249,28 @@ export interface ExtractTextSettings<TRegion = Region> extends ImageSettings<TRe
...
 }
 
+export type SelfHealingReport = {
+  operations: [
+    {
+      timestamp: string,
+      old: string,
+      new: string
+    }
+  ]
```
kyryloonufriiev, line 259: This is still not what you want. Here is
```suggestion
  operations: {timestamp: string; old: string; new: string}[]
```

tourdedave: Ah 👏

js/packages/core-base/src/server/requests.ts
```diff
@@ -524,12 +527,13 @@ export function makeEyesRequests({
...
   }
 
-  async function abort({logger = defaultLogger}: {logger?: Logger} = {}): Promise<TestResult[]> {
+  async function abort({settings, logger = defaultLogger}: {settings?: AbortSettings, logger?: Logger} = {}): Promise<TestResult[]> {
     logger.log(`Request "abort" called for test ${test.testId}`)
     if (aborted || closed) {
       logger.log(`Request "abort" called for test ${test.testId} that was already stopped`)
       return null
     }
+    await reportSelfHealing({settings, logger})
```
kyryloonufriiev, line 536: What about making this run in the background? Something like we do in `openEyes` with `getAccountInfoWithCache`, call it without awaiting and then wait for a promise result later after the important part is done (close/abort request). I am saying it because I believe we don't want to be in a situation where this request failed and we didn't even close/abort a test.
Another question do we even wanna throw an error when a metadata request failed for any reason, as for me this is such a minor thing that log would be enough.

js/packages/core/src/automation/types.ts
kyryloonufriiev, line 57: I think you don't need it here anymore

js/packages/core-base/src/types.ts
```diff
@@ -247,16 +249,22 @@ export interface ExtractTextSettings<TRegion = Region> extends ImageSettings<TRe
...
+export type SelfHealingReport = {
+  operations: {timestamp: string; old: string; new: string}[]
+}
+
 export interface CloseSettings {
   updateBaselineIfNew?: boolean
   updateBaselineIfDifferent?: boolean
   /** @internal */
   userCommandId?: string
+  driverSessionMetadata?: []
```
kyryloonufriiev, line 261: why `[]` and not `SelfHealingReport`?

tourdedave: Because the EG returns objects of a different shape than what is needed for the Eyes request (e.g., EG returns `{'old-selector': 'new-selector'}` and Eyes needs `{timestamp: 'timestamp', old: 'old-selector', new: 'new-selector'}`. I was planning to just return what is provided by the EG (collected into an array by the egc) and handle the transform in core-base as part of the request function.

Based on the spec I've seen so far, the EG doesn't return a timestamp (waiting for e2e testing to verify). And I'm assuming the timestamp is meant to reflect when the selector was self-healed. So we'd probably want to have the egc add it to the metadata. Which begs the question, do we just want the egc to handle the transform too? Then we can just work with the correct thing. I figured that we didn't want to go this direction since the egc shouldn't know anything about what is needed for Eyes.

Alternatively, we could ask Or and Roni if they could return the same shape as what is needed for the Eyes request.

Thoughts @kyryloonufriiev?

js/packages/core/src/automation/types.ts
tourdedave: Ah, nice catch.

js/packages/core-base/src/server/requests.ts
tourdedave: I agree that if this request fails that it shouldn't prevent close/abort from happening. But the report request needs to complete before the close/abort request, no?

js/packages/core-base/src/types.ts
tourdedave: From Adam:

```
The self healing timestamp is used on the server side to correlate self healing events with checkpoint events. So it is ok for the sdk to add the timestamps according to its clock.
No real need for the eg to add them
```

tourdedave: I added the transform in core-base and added the timestamp as part of it. See https://github.com/applitools/eyes.sdk.javascript1/pull/1205/commits/37c2dc82f186a8dab2d7a8ed0b814089c8457852.

tourdedave:
I tried e2e testing through the EG to see if Sentient is available yet. No luck. I'll try again early next week.

I also need to add test coverage for the Eyes integration. Not sure how to verify that the data was successfully sent (maybe it will be available as part of test info?).

There's also the need to add a proper e2e test for the full flow (e.g., egc w/ self healing enabled healing some selectors and sending the data to Eyes). It will be easier to do this once the egc can be launched programmatically from core.

js/packages/core-base/src/server/requests.ts
kyryloonufriiev: I don't know about that, you can try it to verify.

tourdedave: It can happen after.

js/packages/core/src/classic/close.ts
```diff
@@ -0,0 +1,33 @@
...
+    settings,
+    logger = defaultLogger,
+  }: {
+    settings?: CloseSettings
+    logger?: Logger
+  } = {}): Promise<TestResult[]> {
+    const driver = isDriver(target, spec) ? await makeDriver({spec, driver: target, logger}) : null
+    const driverSessionMetadata = await driver?.getSessionMetadata()
+
+    const [baseEyes] = await eyes.getBaseEyes()
```
amitzur, line 30: Probably better to do the following, like it's done in `check.ts`:
```
const baseEyes = await eyes.getBaseEyes()
    const result = (
      await Promise.all(baseEyes.map(baseEyes => baseEyes.close({settings: {...settings, driverSessionMetadata}, logger})))
    ).flat()
```
I'm still not super proficient in the new core. Why was there no need for this file before?

tourdedave: You're referring to the `check.ts` for the ufg? It's my understanding that for ufg there can/will be multiple eyes sessions, but for classic there would be just one. Is that not correct?

The file wasn't needed before because the default behavior for close that is defined in core-base was used and it was enough until now.

amitzur: Got it. Good reasoning for both questions, thanks!
I believe it was the original intent to be able to support OOTB classic screenshots inside "UFG tests", as well as the other way around - UFG screenshots inside "classic tests". So suppose there was a UFG screenshot in a classic eyes instance - then there would be multiple `baseEyes`es. But I'm not sure about this.

tourdedave: 👌

tourdedave:
A couple of things to point out:

1) We still need to pass `egSessionId` in the eyes.open request. I've exposed the driver session id, but wasn't sure on the best way to determine if we're running on the eg since our vendor specific capabilities get stripped out by the EG. We could expose something on the driver that would help us determine that it is running in the eg, but I figured this was unnecessary if the SDK was going to take control of spawning the eg-client (e.g, it would already know that it is running a driver on the eg). Figured it was worth a discussion.

2) I changed the behavior in `test-utils/parse-env`. See https://github.com/applitools/eyes.sdk.javascript1/pull/1205/commits/77474037e2568987b2aa02cc6e88e5aec83311e2. Up until now, if you pass `url` it would be ignored if there was already a matching preset that had a URL. This doesn't seem like the intended behavior. It seems to me that if we provide a URL in the build command of the test setup, that this should be what is used. What do y'all think?

amitzur:
@tourdedave
1. I was thinking the same thing. There is a guiding line that says that we should rely on eg-client as little as possible, and make it possible to use EG without it, given that you don't care about concurrency etc. I realize it's not exactly the situation ATM, but we should keep it in the back of our heads. So if at some point it will be really important, we can always have logic that adds `egSessionId` if the driver URL is the EG URL. But that's not important right now and I agree with your logic.
2. I saw that! Lovely! I needed it and it was a great surprise :)

js/packages/spec-driver-webdriverio/src/spec-driver.ts
```diff
@@ -245,6 +247,22 @@ export async function setWindowSize(browser: Driver, size: Size): Promise<void>
...
+      ref: '',
+      parameters: [],
+    })
+    const result = await cmd.call(driver)
+    return result as unknown as [] | void
+  } catch (error: any) {
+    if (/unknown command/.test(error.message)) return
+    throw error
+  }
+}
```
kyryloonufriiev, line 265: I have a few comments on this, first of all, you don't have to create this command and call it here, you should simply add it in the `transformDriver` together with other commands. Also, I don't think that you should try to catch `unknown command` error, instead, you should call this spec method only when you are sure that this is an EG client-based driver.

js/packages/core-base/src/types.ts
```diff
@@ -247,16 +249,28 @@ export interface ExtractTextSettings<TRegion = Region> extends ImageSettings<TRe
   userCommandId?: string
 }
 
+export type SelfHealingReport = {
+  operations: {timestamp: string; old: string; new: string}[]
+}
+
+export type DriverSessionMetadata = any[]
```
kyryloonufriiev, line 256: I kinda don't like that we have both... I understand why you made it like this. You have this `toSelfHealingReport` function to prepare input for the server, this way you can test it and this is why you need this type.
The original idea was that we don't want to declare server types, this is why all of the input transformations for the server are done in place when the request is formed. The unit testing for such things is done in `test/unit/request.spec.ts` using nock. This way I intended to reduce the number of types and avoid potential confusion that could happen when two types are describing the same thing.

I think your solution is also fine and not making you change it, just sharing my thoughts about it.

kyryloonufriiev, line 256: BTW I think `DriverSessionMetadata` is a bad naming. Since there is no notion of the driver in `core-base`, I would prefer it to be called `TestMetadata`.
I also think it is safe to say that the type is not an array of any but `Record<string, any>[]` (array of objects).

js/packages/core-base/src/types.ts
```diff
@@ -247,16 +249,28 @@ export interface ExtractTextSettings<TRegion = Region> extends ImageSettings<TRe
...
   /** @internal */
   userCommandId?: string
+  driverSessionMetadata?: DriverSessionMetadata
 }
 
 export interface AbortSettings {
   /** @internal */
   userCommandId?: string
+  driverSessionMetadata?: DriverSessionMetadata
+}
```
kyryloonufriiev, line 270: I believe it makes sense to extend those from `ReportSelfHealingSettings`, instead of defining properties

js/packages/core-base/src/types.ts
```diff
@@ -247,16 +249,28 @@ export interface ExtractTextSettings<TRegion = Region> extends ImageSettings<TRe
...
+
+export interface ReportSelfHealingSettings {
+  driverSessionMetadata?: DriverSessionMetadata
```
kyryloonufriiev, line 273: Following my previous comment about the type name this also should not use `driverSession` terminology, so I suggest `metadata` or `testMetadata`

kyryloonufriiev (changes requested):
Looks great 👍 Just a few small comments, I hope it is not a big deal.

js/packages/spec-driver-webdriverio/src/spec-driver.ts
tourdedave: I initially tried adding it in `transformDriver` but simply appending the new command to the additional command definitions doesn't work because the new command isn't included in the existing type definitions for `webdriver`, `wdio`, etc. Based on what I could tell, the additional commands that we've been using up until this point are for commands that already exist. I tried extending the type definition to include the new command but wasn't able to get it working. So this was the next best option I could find. If you feel strongly about the direction then we should sync on it to see about getting it working.

Only calling it when we know the driver is an EG driver is fine with me. But we'll want to make this explicit (e.g., a code comment should be enough).

kyryloonufriiev: Hmmm. Since this is custom command it is fine to do something like `(driver as any).yourCommand()`, or you can add it to the types in `types/v*.d.ts` files

js/packages/core-base/src/types.ts
tourdedave: Thanks for the explanation! I like the idea of going with the existing strategy that is there, but for now I'm going to leave what I did. We can change it later.

js/packages/spec-driver-webdriverio/src/spec-driver.ts
tourdedave: Something like this?

```
diff --git a/js/packages/spec-driver-webdriverio/src/spec-driver.ts b/js/packages/spec-driver-webdriverio/src/spec-driver.ts
index f5297c0fd..ee3160590 100644
--- a/js/packages/spec-driver-webdriverio/src/spec-driver.ts
+++ b/js/packages/spec-driver-webdriverio/src/spec-driver.ts
@@ -129,6 +129,12 @@ export function transformDriver(driver: Driver): Driver {
         {name: 'y', type: 'number', required: true},
       ],
     }),
+    getSessionMetadata: command('GET', '/session/:sessionId/applitools/metadata', {
+      command: 'getSessionMetadata',
+      description: '',
+      ref: '',
+      parameters: [],
+    }),
   }
   Object.entries(additionalCommands).forEach(([name, cmd]) => driver.addCommand(name, cmd))
   return driver
@@ -250,15 +256,7 @@ export async function setWindowSize(browser: Driver, size: Size): Promise<void>
 // NOTE: this command is meant to be called when running with the eg-client
 // otherwise it will not be implemented on the driver and throw
 export async function getSessionMetadata(driver: Driver): Promise<[]|void> {
-  const command = loadCommand()
-  const cmd = command('GET', '/session/:sessionId/applitools/metadata', {
-    command: 'getSessionMetadata',
-    description: '',
-    ref: '',
-    parameters: [],
-  })
-  const result = await cmd.call(driver)
-  return result as unknown as [] | void
+  return await (driver as any).getSessionMetadata()
 }
 export async function getCookies(browser: Driver, context?: boolean): Promise<Cookie[]> {
   if (context) return browser.getCookies()
```

It throws `driver.getSessionMetadata is not a function`

tourdedave: We can just stick with the current implementation for now and revisit later if need-be. Sound good?

tourdedave:
Two things I thought of:

1. The way the implementation is currently written, there is a case where `spec.getSessionMetadata` could be called when not running with the eg-client (e.g., when running on the EG w/o the eg-client). This would throw an exception.
2. It's simple enough to fix, but it begged the question, would it be helpful to throw a custom error message from `spec.getSessionMetadata` if called when it shouldn't be (e.g., something helpful to inform the user that they're running a driver in an unsupported configuration)?

js/packages/driver/src/capabilities.ts
```diff
@@ -18,6 +18,7 @@ export function parseCapabilities(capabilities: Capabilities): DriverInfo {
     isW3C: isW3C(capabilities),
     isMobile: isMobile(capabilities),
     isChrome: isChrome(capabilities),
+    isExecutionGrid: capabilities.applitools,
```
kyryloonufriiev, line 21: If this capability is present you can be sure that it is not just eg but also eg client, this is why I would call this property `isExecutionGridClient`

js/packages/driver/src/driver.ts
```diff
@@ -405,6 +412,10 @@ export class Driver<TDriver, TContext, TElement, TSelector> {
   }
   // end world
 
+  async getSessionMetadata(): Promise<any> {
+    if (this.isExecutionGrid) return await this._spec?.getSessionMetadata(this.target)
```
kyryloonufriiev, line 416: You can call this method only if it is eg-client, if this is just eg you will have a problem. I would recommend creating a separate getter - `isExecutionGridClient`.

kyryloonufriiev (changes requested):
The last change request:)

tourdedave:
FYI - self healing data isn’t showing up correctly when running the core e2e tests. The request fires and Eyes returns a 200, but the info doesn’t show up in testInfo. This was working previously on production. I've asked about it [here in Trello](https://trello.com/c/6LXTxGof/107-self-healing-in-eyes#comment-63c03769d3f9ed03556574c0). Waiting to hear back from Yotam and Itai.

kyryloonufriiev (approved):

PR #1206, opened 2022-12-18
[eyes-webdriverio-5] bad version `webdriverio@7`

zamboney:
`webdriverio@7.28.0` [version](https://github.com/webdriverio/webdriverio/releases/tag/v7.28.0) throw an `tsc` error ([diff](https://github.com/webdriverio/webdriverio/compare/v7.27.0...v7.28.0))

once downgrade to `7.27.0` version the `tsc`  work as expected

This change will allow to fix version for `webdriverio`

### Test

This changes pass the [`update:framework`](https://github.com/applitools/eyes.sdk.javascript1/blob/615640abc2b82161bcedee01b715d81ea063f6c7/.github/actions/tests/action.yml#L62) stage and [`build`](https://github.com/applitools/eyes.sdk.javascript1/blob/615640abc2b82161bcedee01b715d81ea063f6c7/.github/actions/tests/action.yml#L76-L84) stage [here](https://github.com/applitools/eyes.sdk.javascript1/actions/runs/3730139473)


kyryloonufriiev:
Why do this and not fix whatever doesn't work in the latest version?

zamboney:
This is an internal issue in webdriberio.

kyryloonufriiev:
@zamboney Ok. Wdio has a buggy version, I still don't think it is worth a separate script to maintain. I think that instead of testing on latest 7, we should test on 7.27, this should be specified in the workflow. Take a look at how we already doing something similar for playwright.

zamboney:
@kyryloonufriiev can you please show me where you doing it in playwright?

kyryloonufriiev:
@zamboney please check out publish -playwright.yml workflow file. You will see that for node 12 we testing it on a specific minor version of the framework, since this is the last version that supports node 12. You will see that this version (major.minor) set in PLAYWRIGHT_VERSION env variable. Then this env variable used in the package.json to upgrade the framework. In wdio WEBDRIVERIO_MAJOR_VERSION used not only  to install framework, so some additional work should be done

zamboney:
@kyryloonufriiev i done what you ask, please review again 10x

.github/workflows/publish-webdriverio.yml
```diff
@@ -106,6 +106,7 @@ jobs:
             ./js/packages/eyes-webdriverio-5/coverage-tests-metadata.json
             ./js/packages/eyes-webdriverio-5/coverage-test-report.xml
   publish:
+    if: true
```
kyryloonufriiev, line 109: I'm not sure what is that

kyryloonufriiev (changes requested):
@zamboney This is not a complete solution. Please take a better look at how and where `APPLITOOLS_WEBDRIVERIO_MAJOR_VERSION` is used.

You also should change the name of the variable you provide with an action to `APPLITOOLS_WEBDRIVERIO_MAJOR_VERSION`, no `MAJOR` word since it is simply not true anymore.

I would recommend calculating `APPLITOOLS_WEBDRIVERIO_MAJOR_VERSION` from `APPLITOOLS_WEBDRIVERIO_VERSION` in the same place we are extracting the actual major version today.

BTW the same thing should be done for `spec-driver-wdio` package as well.

.github/workflows/publish-webdriverio.yml
zamboney: for testing

zamboney:
@kyryloonufriiev i will change the name of the it, but can you explain why you think this is not a complete solution?

.github/workflows/publish-spec-webdriverio.yml
```diff
@@ -37,7 +37,7 @@ jobs:
     name: 'Test (wdio: ${{matrix.framework-version}}, node: ${{matrix.node-version}})'
     runs-on: ubuntu-latest
     env:
-      APPLITOOLS_WEBDRIVERIO_MAJOR_VERSION: ${{matrix.framework-version}}
+      APPLITOOLS_WEBDRIVERIO_VERSION: ${{matrix.framework-version}}
```
kyryloonufriiev, line 40: This is not the only thing to change in this file. Take a look at line 35, it says `framework-version: 7`

js/packages/bongo/test/tree/fixtures/packages/eyes-webdriverio-5/package.json
```diff
@@ -84,7 +84,7 @@
   ],
   "scripts": {
     "lint": "eslint '**/*.ts'",
-    "build": "ttsc --project tsconfig$([ ! -z $APPLITOOLS_WEBDRIVERIO_MAJOR_VERSION ] && echo .v$APPLITOOLS_WEBDRIVERIO_MAJOR_VERSION).json",
+    "build": "ttsc --project tsconfig$([ ! -z $APPLITOOLS_WEBDRIVERIO_VERSION ] && echo .v${APPLITOOLS_WEBDRIVERIO_VERSION%%.*}).json",
```
kyryloonufriiev, line 87: No need for the change in `bongo/test/tree`

js/packages/eyes-webdriverio-5/src/service.ts
```diff
@@ -2,11 +2,11 @@ import {makeCore} from '@applitools/core'
...
-if (!process.env.APPLITOOLS_WEBDRIVERIO_MAJOR_VERSION) {
+if (!process.env.APPLITOOLS_WEBDRIVERIO_VERSION) {
   try {
     const {version} = require('webdriverio/package.json')
     const [major] = version.split('.', 1)
-    process.env.APPLITOOLS_WEBDRIVERIO_MAJOR_VERSION = major
+    process.env.APPLITOOLS_WEBDRIVERIO_VERSION = major
   } catch {
     // NOTE: ignore error
   }
```
kyryloonufriiev, line 12: I would prefer to leave the calculated name as `APPLITOOLS_WEBDRIVERIO_MAJOR_VERSION`, but in case `APPLITOOLS_WEBDRIVERIO_VERSION` is specified we can take a major from there.

js/packages/spec-driver-webdriverio/src/index.ts
```diff
@@ -1,11 +1,11 @@
...
-if (!process.env.APPLITOOLS_WEBDRIVERIO_MAJOR_VERSION) {
+if (!process.env.APPLITOOLS_WEBDRIVERIO_VERSION) {
   try {
     const {version} = require('webdriverio/package.json')
     const [major] = version.split('.', 1)
-    process.env.APPLITOOLS_WEBDRIVERIO_MAJOR_VERSION = major
+    process.env.APPLITOOLS_WEBDRIVERIO_VERSION = major
   } catch {
     // NOTE: ignore error
   }
```
kyryloonufriiev, line 11: Same here

kyryloonufriiev (changes requested):

.github/workflows/publish-spec-webdriverio.yml
zamboney: why do we need to change it?

js/packages/spec-driver-webdriverio/test/e2e/spec-driver.spec.ts
```diff
@@ -18,7 +18,9 @@ describe('spec driver', async () => {
 
   describe('headless desktop (@puppeteer)', async () => {
     before(function () {
-      if (Number(process.env.APPLITOOLS_WEBDRIVERIO_MAJOR_VERSION) < 7) this.skip()
+      const version =
+        process.env.APPLITOOLS_WEBDRIVERIO_VERSION && Number(process.env.APPLITOOLS_WEBDRIVERIO_VERSION.split('.')[0])
+      if (typeof version === 'undefined' || version < 7) this.skip()
```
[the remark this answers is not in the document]
zamboney: can you explain what you want me to do?

js/packages/eyes-webdriverio-5/src/service.ts
zamboney: can you explain what you want me to do?

kyryloonufriiev: I think you should not simply rename the env var here. But extend a logic like this:
 - if there is nor `APPLITOOLS_WEBDRIVERIO_VERSION`, nor `APPLITOOLS_WEBDRIVERIO_MAJOR_VERSION` then extract the major version of the framework and put it in `APPLITOOLS_WEBDRIVERIO_MAJOR_VERSION`
 - if there is `APPLITOOLS_WEBDRIVERIO_VERSION` and no `APPLITOOLS_WEBDRIVERIO_MAJOR_VERSION` then extract the major version from `APPLITOOLS_WEBDRIVERIO_VERSION` and put it in `APPLITOOLS_WEBDRIVERIO_MAJOR_VERSION`
 - if there is `APPLITOOLS_WEBDRIVERIO_MAJOR_VERSION` do nothing

js/packages/spec-driver-webdriverio/test/e2e/spec-driver.spec.ts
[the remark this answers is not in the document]
kyryloonufriiev: You should not do any change here if you populate `APPLITOOLS_WEBDRIVERIO_MAJOR_VERSION` as I wrote above

js/packages/eyes-webdriverio-5/src/service.ts
zamboney: not sure i understand

js/packages/eyes-webdriverio-5/package.json
```diff
@@ -89,7 +89,7 @@
   ],
   "scripts": {
     "lint": "eslint '**/*.ts'",
-    "build": "ttsc --project tsconfig$([ ! -z $APPLITOOLS_WEBDRIVERIO_MAJOR_VERSION ] && echo .v$APPLITOOLS_WEBDRIVERIO_MAJOR_VERSION).json",
+    "build": "ttsc --project tsconfig$([ ! -z $APPLITOOLS_WEBDRIVERIO_VERSION ] && echo .v${APPLITOOLS_WEBDRIVERIO_VERSION%%.*}).json",
```
kyryloonufriiev, line 92: BTW This is super nice. Didn't know you we can do that!

kyryloonufriiev (approved):
@zamboney I did pushed what I tried to explain about the major version.

PR #1207, opened 2022-12-19
[eyes-webdriverio-5] upgrade to `wdio@8`

zamboney:
This is an attempt to make `wdio@8`

What it solve

```typescript
  const command = require('webdriver/build/command').default
```
[ref](https://github.com/applitools/eyes.sdk.javascript1/blob/8bbd3bb7fbb75ccc33e57c6bb6660447622622c0/js/packages/spec-driver-webdriverio/src/spec-driver.ts)

Throw this error

```
Uncaught:                                                                                                                                                           Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './build/command' is not defined by "exports" in ./node_modules/webdriver/package.json
  at __node_internal_captureLargerStackTrace (node:internal/errors:465:5)
  at new NodeError (node:internal/errors:372:5)
  at throwExportsNotFound (node:internal/modules/esm/resolve:472:9)
  at packageExportsResolve (node:internal/modules/esm/resolve:753:3)
  at resolveExports (node:internal/modules/cjs/loader:482:36)
  at Function.Module._findPath (node:internal/modules/cjs/loader:522:31)
  at Function.Module._resolveFilename (node:internal/modules/cjs/loader:919:27)
  at Function.Module._load (node:internal/modules/cjs/loader:778:27)
  at Module.require (node:internal/modules/cjs/loader:1005:19)
  at require (node:internal/modules/cjs/helpers:102:18) {                                                                                                                                       code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
```

this pr solve it

## TODO

- [ ] figure a way to test it
- [ ] remove the code below from `tsconfig`

```
"skipLibCheck": true,                     /* Skip type checking of declaration files. */
"forceConsistentCasingInFileNames": true,  /* Disallow inconsistently-cased references to the same file. */
```

### `tsc` errors without `skipLibCheck`

```
Found 181 errors in 16 files.

Errors  Files
     6  node_modules/@types/mocha/index.d.ts:18
     3  node_modules/@wdio/mocha-framework/build/index.d.ts:3
     6  node_modules/@wdio/mocha-framework/node_modules/@types/mocha/index.d.ts:18
    10  node_modules/@wdio/protocols/build/index.d.ts:1
     7  node_modules/@wdio/types/build/index.d.ts:1
     1  node_modules/devtools/build/index.d.ts:3
     2  node_modules/devtools/build/types.d.ts:2
     1  node_modules/webdriver/build/bidi.d.ts:2
     2  node_modules/webdriver/build/command.d.ts:3
     1  node_modules/webdriver/build/index.d.ts:5
     2  node_modules/webdriver/build/request/index.d.ts:3
    11  node_modules/webdriver/build/types.d.ts:2
     1  node_modules/webdriver/build/utils.d.ts:3
     1  node_modules/webdriverio/build/index.d.ts:3
   126  node_modules/webdriverio/build/types.d.ts:2
     1  node_modules/webdriverio/build/utils/interception/types.d.ts:2
error Command failed with exit code 1.
```


pending-changes.yaml
```diff
@@ -128,6 +128,7 @@ Apollo SDK:
     - Drop support for Android 9 and Android 10 specific devices
     - Added new iOS device - 'iPad Pro (11-inch) (4th generation)'
     - Use user agent metadata to improve browser environment detection logic
+    - Upgrade to `wdio@8`
```
zamboney, line 131: @benbabayoff8

benbabayoff8: @zamboney Looks great, thank you

kyryloonufriiev:
`skipLibCheck` is not a solution. I already did my own investigation, and it happens because of the wrong combination of `modue` and `moduleResolution` for wdio 8 options.

zamboney:
@kyryloonufriiev ok so what you want to do about it

kyryloonufriiev:
I just wrote in the previous comment. With different module and moduleResolution you will have no errors from wdio types. But you still have to preserve whatever we had for previous versions of wdio since they are not esm modules.

the right combination would be module->nodenext and moduleResolution->node

zamboney:
@kyryloonufriiev even with the combination that you suggested, there are still lots of errors (i think it relates to t `"moduleResolution": "node")

```

node_modules/@wdio/types/build/Options.d.ts:5:23 - error TS2307: Cannot find module 'node:http' or its corresponding type declarations.

5 import * as http from 'node:http';
                        ~~~~~~~~~~~

node_modules/@wdio/types/build/Options.d.ts:6:24 - error TS2307: Cannot find module 'node:https' or its corresponding type declarations.

6 import * as https from 'node:https';
                         ~~~~~~~~~~~~

node_modules/@wdio/types/build/Options.d.ts:7:26 - error TS2307: Cannot find module 'node:url' or its corresponding type declarations.

7 import type { URL } from 'node:url';
                           ~~~~~~~~~~

node_modules/@wdio/types/build/Reporters.d.ts:3:34 - error TS2307: Cannot find module 'node:fs' or its corresponding type declarations.

3 import type { WriteStream } from 'node:fs';
                                   ~~~~~~~~~

node_modules/@wdio/types/build/Reporters.d.ts:4:35 - error TS2307: Cannot find module 'node:events' or its corresponding type declarations.

4 import type { EventEmitter } from 'node:events';
                                    ~~~~~~~~~~~~~

node_modules/@wdio/types/build/Workers.d.ts:2:35 - error TS2307: Cannot find module 'node:events' or its corresponding type declarations.

2 import type { EventEmitter } from 'node:events';
                                    ~~~~~~~~~~~~~

node_modules/devtools/build/types.d.ts:2:30 - error TS2307: Cannot find module 'node:events' or its corresponding type declarations.

2 import { EventEmitter } from 'node:events';
                               ~~~~~~~~~~~~~

node_modules/webdriver/build/bidi.d.ts:2:30 - error TS2307: Cannot find module 'node:events' or its corresponding type declarations.

2 import { EventEmitter } from 'node:events';
                               ~~~~~~~~~~~~~

node_modules/webdriver/build/request/index.d.ts:3:30 - error TS2307: Cannot find module 'node:events' or its corresponding type declarations.

3 import { EventEmitter } from 'node:events';
                               ~~~~~~~~~~~~~

node_modules/webdriver/build/request/index.d.ts:4:26 - error TS2307: Cannot find module 'node:url' or its corresponding type declarations.

4 import type { URL } from 'node:url';
                           ~~~~~~~~~~

node_modules/webdriver/build/types.d.ts:2:30 - error TS2307: Cannot find module 'node:events' or its corresponding type declarations.

2 import { EventEmitter } from 'node:events';
                               ~~~~~~~~~~~~~

node_modules/webdriverio/build/types.d.ts:2:35 - error TS2307: Cannot find module 'node:events' or its corresponding type declarations.

2 import type { EventEmitter } from 'node:events';
                                    ~~~~~~~~~~~~~

```

kyryloonufriiev:
I tried it and there were no issues during the normal TS build. However, there is an issue with our api-extractor package. In any case, there is a way to do it right, we should not skip type-checking.

zamboney:
Ok i didn't figure a way to handle it like you do... can you check if all of
your dependences of `wdio` are install with `8` version 

```bash
npm ls -s webdriver webdriverio
```


this is what i get

```
@applitools/eyes-webdriverio@5.38.2 ./javascript1-bare/wdio-8-issue/js/packages/eyes-webdriverio-5
├─┬ @applitools/spec-driver-webdriverio@1.4.7
│ └── webdriverio@8.0.13 deduped
├─┬ @wdio/cli@8.0.13
│ ├─┬ @wdio/globals@8.0.13
│ │ ├─┬ expect-webdriverio@4.0.1
│ │ │ └── webdriverio@8.0.13 deduped
│ │ └── webdriverio@8.0.13 deduped
│ └── webdriverio@8.0.13 deduped
├─┬ @wdio/local-runner@8.0.13
│ └─┬ @wdio/runner@8.0.13
│   ├── webdriver@8.0.13 deduped
│   └── webdriverio@8.0.13 deduped
└─┬ webdriverio@8.0.13
  └── webdriver@8.0.13

```

zamboney (commented):
@kyryloonufriiev i figure out a way to remove `skipLib`. but i had to change the `moduleResolution` to `node` in order for that to work...

i know that it's sound counterintuitive but i could not figure out a way to make it work in any other way.

to make it work using `node` i had to change some code in `core` and `core-base` (take a look at the comments i wrote in the code line)

kyryloonufriiev:
@zamboney I see you start applying changes to `core`, this is something I asked you to not rush with. I checked myself and find an actual reason for the error you see, it happens because our submodule declarations are not applied, (this is why it sees two `makeCore` exports from the core). So this means that we don't need to rename stuff in the `core`/`core-base` packages.
Submodules are not applied because `exports` declaration is ignored by ts if the module resolution mode is not  `node16`/`nodenext`. There is also a different way to declare typescript submodules (this good enough for us because we use submodules only for types here).
But it made me think that we actually should not change anything and it should work just fine out of the box, and the reason it is not working is the wdio's types. So I fixed it in wdio with this PR - https://github.com/webdriverio/webdriverio/pull/9474, hopefully, it will solve our problem (I test, so it should).

kyryloonufriiev:
@zamboney I also see that you upgraded `@types/node` to version `14`, this is not completely right, the version should stay `12`, because this is the lowest major we have to support for now, I agree that since `wdio@8` supports node 16 and above we have to upgrade `@types/node` to version `16` when we build sdk for `wdio@8`, this requires us to create a mechanism similar `upgrade:frameworks` to upgrade `@types/node` to match node version in the test runner.

Maybe something like this
```
packagejson=`cat package.json`; yarn upgrade --no-lockfile @types/node@${${${$(node -v)}#v}%%.*}; echo \"$packagejson\" > package.json;
```

zamboney:
@kyryloonufriiev i don't see any reason to upgrade `@types/node` to `14` (i think that it could be even `16`). this is a `devDependencies` so why not?

kyryloonufriiev:
If we have it for node 16, when we still support node 12, what will prevent us from using node 16 features in our code?

kyryloonufriiev:
@zamboney Sure we have to use it in gh actions. But I would love to use it locally as well because if I wanna create a wdio 5 build I want to use `@types/node@12`, but for wdio 8 I have to use `@types/node@16`.

zamboney:
@kyryloonufriiev i think your last statement contradicts your last last statement 😄, if we want to support `node >= 12` we must fix the type to `12`

kyryloonufriiev:
Yes and no. We have node types fixed on version 12, and this was fine since wdio also supported node 12. But with wdio 8 they support only node 16 and above. This is why we have to be flexible here. If the customer uses wdio 8 they have to use node 16, this is why we will never build or test wdio 8 with node 12 or 14. But we still want to be able to build and test wdio 5/6/7 so we have to do the script that I proposed.

zamboney:
@kyryloonufriiev seems like your fix in `@wdio` solve the issue 🥳 🎉 🌮 

I figure out the issue with the `@types/node`. Once it's removed, each package takes its version, and it works.

[this issue](https://github.com/applitools/eyes.sdk.javascript1/actions/runs/3800749433/jobs/6464522457#step:5:201) is because the linking is not working as expected, I tested it locally.

I think we have reached a point where we can push this to master and publish it so that `wdio@8` will work for our client. 

WDYT?

.github/workflows/publish-spec-webdriverio.yml
```diff
@@ -8,6 +8,10 @@ on:
         type: choice
         options: [patch, minor, major]
         default: patch
+      multienv:
+        description: test on multiple environments
+        type: boolean
+        default: true
```
kyryloonufriiev, line 14: No need for that in spec-* packages, this flag was added to SDKs because tests there are flaky and slow, here we have a suite of stable and fast tests, so we basically always want to test it on all of the envs

.github/workflows/publish-webdriverio.yml
```diff
@@ -67,12 +67,14 @@ jobs:
...
         shell: bash
         run: |
+          ENVIRONMENTS=
           if [ ${{github.event.inputs.multienv}} = true ]
           then
-            echo "::set-output name=environments::[{\"framework-version\":7.27,\"node-version\":16}, {\"framework-version\":6,\"node-version\":16}, {\"framework-version\":5,\"node-version\":16}, {\"framework-version\":7.27,\"node-version\":14}, {\"framework-version\":7.27,\"node-version\":12}]"
+            ENVIRONMENTS="[{\"framework-version\":8,\"node-version\":16}, {\"framework-version\":7.27,\"node-version\":16}, {\"framework-version\":6,\"node-version\":16}, {\"framework-version\":5,\"node-version\":16}, {\"framework-version\":7.27,\"node-version\":14}, {\"framework-version\":7.27,\"node-version\":12}]"
           else
-            echo "::set-output name=environments::[{\"framework-version\":7.27,\"node-version\":16}]"
+            ENVIRONMENTS="[{\"framework-version\":8,\"node-version\":16}]"
```
kyryloonufriiev, line 75: The idea behind that was to test all of the supported wdio versions, and all of the node versions at least once. Creating a full matrix is a crazy idea since 3 wdio versions X 3 node versions are 9 jobs. This is why we had all node versions on the latest wdio + all other wdio versions on the latest node. Also since we are already here let's include node 18 here. If we keep doing what we were doing before with that many node and wdio versions it will be too many jobs, so I recommend:

- wdio 8 + node 18
- wdio 8 + node 16
- wdio 7 + node 14
- wdio 6 + node 12
- wdio 5 + node 12

js/packages/spec-driver-webdriverio/src/spec-driver.ts
```diff
@@ -97,7 +97,10 @@ export function isSelector(selector: any): selector is Selector {
   )
 }
 export function transformDriver(driver: Driver): Driver {
-  const command = require('webdriver/build/command').default
+  function command(method: string, url: string, body: any) {
+    // @ts-ignore
+    return async (...args) => (await import('webdriver')).command(method, url, body)(...args)
+  }
```
kyryloonufriiev, line 103: This will blow up on older node versions - before esm support. Also, it will blow up on wdio versions < 8. We need to support both here.

js/packages/eyes-webdriverio-5/package.json
```diff
@@ -128,13 +128,12 @@
     "@applitools/sdk-shared": "0.9.15",
     "@applitools/test-utils": "1.5.4",
     "@types/mocha": "^9.1.1",
-    "@types/node": "12",
```
kyryloonufriiev, line 131: I didn't understand what is your idea for that.

kyryloonufriiev (changes requested):
@zamboney We are almost there.

js/packages/eyes-webdriverio-5/package.json
zamboney: once remove each package takes it's own `@types/node` dependency and then there are not `ts` error

js/packages/spec-driver-webdriverio/src/spec-driver.ts
```diff
@@ -98,8 +98,15 @@ export function isSelector(selector: any): selector is Selector {
...
-    return async (...args) => (await import('webdriver')).command(method, url, body)(...args)
+    try {
+      // @ts-ignore
+      return async (...args) => (await import('webdriver')).command(method, url, body)(...args)
+    } catch (err) {
+      // @ts-ignore
+      return (...args) => {
+        return require('webdriver/build/command').default(method, url, body)(...args)
+      }
+    }
```
kyryloonufriiev, line 109: you should simply check env variable as we do in other places.
Also, older versions of wdio will not have default export. You should simply import `command` using `require('webdriver').command`

.github/workflows/publish-spec-webdriverio.yml
```diff
@@ -43,13 +39,7 @@ jobs:
...
         run: |
-          ENVIRONMENTS=
-          if [ ${{github.event.inputs.multienv}} = true ]
-          then
-            ENVIRONMENTS="[{\"framework-version\":8,\"node-version\":16}, {\"framework-version\":7.27,\"node-version\":16}, {\"framework-version\":6,\"node-version\":16}, {\"framework-version\":5,\"node-version\":16}, {\"framework-version\":7.27,\"node-version\":14}, {\"framework-version\":7.27,\"node-version\":12}]"
-          else
-            ENVIRONMENTS="[{\"framework-version\":8,\"node-version\":16}]"
-          fi
+          ENVIRONMENTS="[{\"framework-version\":8,\"node-version\":18}]"
           echo "environments=${ENVIRONMENTS}" >> $GITHUB_OUTPUT
```
kyryloonufriiev, line 43: There is no logic now, so I believe you can remove this job. Also, you did the opposite of what I was requesting, it basically should be the same as it was before, just with a slightly updated matrix to include new node and wdio versions.

zamboney: Not sure I understand

js/packages/eyes-webdriverio-5/package.json
kyryloonufriiev: Still not what we want. Because this way WE will use an unknown version of `@types/node` during development, which means we are not protected from using some api from a newer node version.

vinothvk:
@kyryloonufriiev may i know when this change will be released to users ?

kyryloonufriiev:
@vinothvk this is already released. But keep in mind that the latest version of wdio still has some bugs that may break your test. Especially if you use appium

vinothvk:
@kyryloonufriiev thanks for the input. Last published version of @applitools/eyes-webdriverio is 5.39.0 which is 14 days ago. Are you sure this version has this wdio 8 support ?
https://www.npmjs.com/package/@applitools/eyes-webdriverio

vinothvk:
It's working thanks

PR #1208, opened 2022-12-20
Add lab provider and CI name to logEvents

SophieDePaula:
[Trello card](https://trello.com/c/ZSN0uuiR/85-splunk-add-lab-provider-url)

notes:
 - currently, this `driverUrl` is only added to universal and wdio spec-driver 

TODO:
- [x] Add tests
- [x] Add pending changes entry


js/packages/core-base/src/types.ts
```diff
@@ -148,6 +148,7 @@ export interface LogEventSettings extends ServerSettings {
   event: {type: string} & Record<string, any>
   level?: 'Info' | 'Notice' | 'Warn' | 'Error'
   timestamp?: string
+  labProvider?: string
```
kyryloonufriiev, line 151: You don't add it here. You add it to `event` object. That has a free structure so no need to declare anything. Also do not call the field `labProvider`, because it is not a lab provider, but a simple URL that may or may not reference a lab provider.

js/packages/core/src/open-eyes.ts
```diff
@@ -81,6 +81,7 @@ export function makeOpenEyes<TDriver, TContext, TElement, TSelector, TDefaultTyp
           concurrentRendersPerTest: (settings as OpenSettings<'ufg'>).renderConcurrency,
           node: {version: process.version, platform: process.platform, arch: process.arch},
         },
+        labProvider: spec.extractHostName ? spec.extractHostName(target) : undefined,
```
kyryloonufriiev, line 84: You can simply do that `spec?.extractHostName?.(target)`. This way you also handle the case when there is no spec at all (it could happen).

js/packages/driver/src/spec-driver.ts
```diff
@@ -101,5 +101,6 @@ export interface SpecDriver<TDriver, TContext, TElement, TSelector> {
   getCurrentWorld?(driver: TDriver): Promise<string>
   getWorlds?(driver: TDriver): Promise<string[]>
   switchWorld?(driver: TDriver, id: string): Promise<void>
+  extractHostName?(driver: TDriver): string
```
kyryloonufriiev, line 104: This is a utility method please put it in the proper place, along with other utility methods

js/packages/eyes-universal/src/types.ts
```diff
@@ -113,6 +113,7 @@ export interface UniversalSpecDriver<TDriver, TContext, TElement, TSelector> {
   getCurrentWorld?(options: {driver: TDriver}): Promise<string>
   getWorlds?(options: {driver: TDriver}): Promise<string[]>
   switchWorld?(options: {driver: TDriver; name: string}): Promise<void>
+  extractHostName?(options: {driver: TDriver}): string
```
kyryloonufriiev, line 116: it should not go there at all. This is a sync method, it could not be done through the web sockets

kyryloonufriiev (changes requested):

pending-changes.yaml
```diff
@@ -36,6 +36,7 @@ Apollo SDK:
 "@applitools/eyes-webdriverio":
   feature:
     - Added `nmgOptions` to `CheckSettings`
+    - Additional internal event logs
```
SophieDePaula, line 39: @benbabayoff8

SophieDePaula:
Here is an example for `driverUrl` when running against saucelabs from splunk:

```
{
  "data": {
    "Site": "Public",
    "Service": "Eyes",
    "HostName": "SDK000002",
    "Timestamp": "2022-12-20T18:16:51.0485027+00:00",
    "Logger": "eyes.client.sessions",
    "Level": "NOTICE",
    "Context": {
      "EyesRequestId": "6fa7bdea-4595-446f-9833-e39a80953de0",
      "EyesClient": "eyes-webdriverio/5.38.2",
      "EyesClientRequestId": "0--47e41e3e-57b0-4913-989f-2b1bc7fb80a8",
      "CallerIP": "73.189.112.142",
      "RequestUrl": "https://eyesapi.applitools.com/api/sessions/log?apiKey=***",
      "RequestMethod": "POST",
      "UserName": null,
      "AccountId": "xIpd7EWjhU6cjJzDGrMcUw~~",
      "AccessKey": "***",
      "IsImpersonated": false,
      "Permissions": [
        "execute"
      ],
      "CreatedAt": "2022-12-20T18:16:51.0329287+00:00"
    },
    "Text": null,
    "Info": {
      "timestamp": "2022-12-20T18:16:50.7720000+00:00",
      "Event": {
        "type": "runnerStarted",
        "node": {
          "version": "v16.1.0",
          "platform": "darwin",
          "arch": "x64"
        },
        "driverUrl": "ondemand.us-west-1.saucelabs.com"
      }
    },
    "Exception": null,
    "ExceptionType": null,
    "ExceptionStackTrace": null
  }
}
```

@benbabayoff8 , @kyryloonufriiev fyi

pending-changes.yaml
benbabayoff8: @SophieDePaula amazing

js/packages/core-base/src/types.ts
```diff
@@ -145,7 +145,7 @@ export interface DeleteTestSettings extends ServerSettings {
 }
 
 export interface LogEventSettings extends ServerSettings {
-  event: {type: string} & Record<string, any>
```
kyryloonufriiev, line 148: as I wrote in my previous comment. `even` could be any object, this is done on purpose so we can log any events. This is why you don't need to add your property here.

js/packages/core/src/open-eyes.ts
```diff
@@ -80,6 +80,8 @@ export function makeOpenEyes<TDriver, TContext, TElement, TSelector, TDefaultTyp
           testConcurrency: concurrency,
           concurrentRendersPerTest: (settings as OpenSettings<'ufg'>).renderConcurrency,
           node: {version: process.version, platform: process.platform, arch: process.arch},
+          driverUrl: spec?.extractHostName?.(target),
+          CIProvider: spec?.extractCIProvider(),
```
kyryloonufriiev, line 84: do not call properties with leading uppercase latter. Also `extractCIProvider` has no relation to the driver so it shouldn’t be there.

js/packages/core/src/open-eyes.ts
```diff
@@ -80,6 +80,8 @@ export function makeOpenEyes<TDriver, TContext, TElement, TSelector, TDefaultTyp
           testConcurrency: concurrency,
           concurrentRendersPerTest: (settings as OpenSettings<'ufg'>).renderConcurrency,
           node: {version: process.version, platform: process.platform, arch: process.arch},
+          driverUrl: spec?.extractHostName?.(target),
```
kyryloonufriiev, line 83: It is also worth checking for the target is not undefined
```suggestion
          driverUrl: target && spec?.extractHostName?.(target),
```

kyryloonufriiev (changes requested):

js/packages/core-base/src/types.ts
SophieDePaula: @kyryloonufriiev I did change it but forgot to push in that commit so I pushed it in the commit after.

SophieDePaula:
@kyryloonufriiev I addressed all your comments.

js/packages/core/src/utils/extract-CI-provider.ts
```diff
@@ -0,0 +1,38 @@
...
+  if (process.env.GITHUB_ACTIONS) {
+    return 'GitHub Actions'
+  }
+  if (process.env.GITLAB_CI) {
+    return 'GitLab'
+  }
+  if (process.env.HEROKU_TEST_RUN_ID) {
+    return 'Heroku'
+  }
+  if (process.env.BUILD_ID) {
```
SophieDePaula, line 29: I don't think that `JENKINS_URL` env var is mandatory, and I read that `Jenkins` is a continuation of `Hudson`, so I decided to put them together.
@kyryloonufriiev , @benbabayoff8 lmk if that is a problem.

benbabayoff8: @SophieDePaula I would consider Hudson as Jenkins as well. I think that is better than letting the field guess.

js/packages/core-base/src/types.ts
```diff
@@ -145,7 +145,7 @@ export interface DeleteTestSettings extends ServerSettings {
...
+  event: {type: string; driverUrl?: string; extractedCIProvider?: string} & Record<string, any>
```
kyryloonufriiev, line 148: Did you understand from my previous message that it should not be here?

js/packages/core/src/open-eyes.ts
```diff
@@ -11,6 +11,7 @@ import {makeLocateText} from './locate-text'
 import {makeExtractText} from './extract-text'
 import {makeClose} from './close'
 import * as utils from '@applitools/utils'
+import {extractCIProvider} from './utils/extract-CI-provider'
```
kyryloonufriiev, line 14: The naming convention for the file is all lowercase + dashes instead of spaces. Please align your file name.

js/packages/driver/src/spec-driver.ts
```diff
@@ -57,6 +57,7 @@ export interface SpecDriver<TDriver, TContext, TElement, TSelector> {
   extractSelector?(element: TElement): TSelector
   isStaleElementError(error: any, selector?: TSelector): boolean
   isEqualElements?(context: TContext, element1: TElement, element2: TElement): Promise<boolean>
+  extractHostName?(driver: TDriver): string
```
kyryloonufriiev, line 60: The proper typing here would be:
```suggestion
  extractHostName?(driver: TDriver): string | null
```

js/packages/spec-driver-webdriverio/test/e2e/spec-driver.spec.ts
```diff
@@ -429,6 +429,15 @@ describe('spec driver', async () => {
...
   })
 
+  describe('browser with saucelabs', async () => {
+    it('extractHostName()', async () => {
+      ;[browser, destroyBrowser] = await spec.build({browser: 'chrome-mac', protocol: 'wd'})
+      await browser.url(url)
+
+      extractHostName({expected: 'ondemand.us-west-1.saucelabs.com'})
+    })
+  })
```
kyryloonufriiev, line 439: What is the point of running a separate browser just to extract a URL. We already run a bunch of browsers you can test this function there.

pending-changes.yaml
```diff
@@ -10,7 +10,7 @@ Apollo SDK:
       An example of how to inline `markdown code highlights`
 "@applitools/eyes-universal":
   feature:
-    - 
+    - Additional internal event logs
```
kyryloonufriiev, line 13: This should go to every sdk

js/packages/core/src/utils/extract-CI-provider.ts
```diff
@@ -0,0 +1,38 @@
+export function extractCIProvider(): string {
```
kyryloonufriiev, line 1: the CI is not always there, I think we should return null if we were not able to find one.
```suggestion
export function extractCIProvider(): string | null {
```

kyryloonufriiev (changes requested):
Great work! I just have some minor comments, feel free to address them and merge.
About the file names, you should change:
extract-CI-provider.ts -> extract-ci-provider.ts
extract-CI-provider.spec.ts -> extract-ci-provider.spec.ts

js/packages/spec-driver-webdriverio/test/e2e/spec-driver.spec.ts
SophieDePaula: But then I have a test that tests two unrelated things. I thought it would be cleaner and less confusing if I just add a test for this method.

pending-changes.yaml
SophieDePaula: But I only added it to wdio and universal

SophieDePaula: I guess the CI provider was added to all

PR #1210, opened 2022-12-20
Wrap types with NonNullable

SophieDePaula:
[Trello card](https://trello.com/c/vd6ALtsI/1667-braze-cypress-strictnullchecks-flag-ts-error)


pending-changes.yaml
```diff
@@ -109,7 +109,7 @@ Apollo SDK:
     - Handed error during polling in long requests to eyes server
 "@applitools/eyes-cypress":
   feature:
-    - 
+    - Wrap types with NonNullable
```
SophieDePaula, line 112: @benbabayoff8

benbabayoff8: @SophieDePaula Is that a feature? It looks like an implementation detail. What is the benefit of that feature?

SophieDePaula: @benbabayoff8 I didn't think it was a bug fix. It was a request from the customer for us to use `NonNullable` so that they won't get an error in their `ts` files.

benbabayoff8: So ok, thanks for the explanation

amitzur:
@SophieDePaula @kyryloonufriiev I adjusted our ts test to actually test that this works, by changing `skipLibCheck` to `strictNullChecks`. Once I did that there were additional failures, so I added more `NonNullable`s.

I'm not sure how our `index.d.ts` file was crafted, but it seems too manual, in the `CypressEyesConfig` and `CypressCheckSettings` part. I am wondering if it has to be this way, or can we inherit more from core?
Another thing I'm wondering is if this issue also exists in core (if users use `strictNullChecks` on other JS SDK's, will they get the same error?)

I am also modifying the release note a bit.

PR #1211, opened 2022-12-26
adjust mobile env structure

SophieDePaula:
TODO
- [ ] Add a test
- [ ] Update pending changes


js/packages/eyes-storybook/src/generateConfig.js
```diff
@@ -75,7 +76,24 @@ function generateConfig({argv = {}, defaultConfig = {}, externalConfigParams = [
...
+    }
+    for (const browser of result.browser) {
+      if (browser.deviceName) {
+        result.renderers.push({
+          chromeEmulationInfo: browser,
+        });
+      } else {
+        result.renderers.push(browser);
+      }
+    }
```
kyryloonufriiev, line 95: What about doing it like this? A little bit more compact
```suggestion
    result.renderers = result.browser.map(browser => {
      return browser.deviceName ? {chromeEmulationInfo: browser} : browser
    })
```

kyryloonufriiev (approved):

pending-changes.yaml
```diff
@@ -103,6 +103,7 @@ Apollo SDK:
 "@applitools/eyes-storybook":
   feature:
     - Additional internal event logs
+    - Fix mobile environment
```
SophieDePaula, line 106: @benbabayoff8

benbabayoff8: @SophieDePaula Why is it called "fix" but under the features section? What are we fixing here?

SophieDePaula: @benbabayoff8 this is a mistake, I'll move it to bug fix

benbabayoff8: @SophieDePaula Thanks, I would maybe call it something like "fixed bug in mobile environment", and not some general

PR #1235, opened 2023-01-03
[eyes-cypress] fix issue with `cypress run` URL

zamboney:
On the `cypress run`, an [`ASCII color`](https://github.com/applitools/eyes.sdk.javascript1/blob/3bf8cd290039ce2dbd4a5f3148abb46a16b0c2b8/js/packages/eyes-cypress/src/plugin/errorDigest.js#L60-L62) has used to colorize the terminal output.

This addition added to the URL. Once clicked, it goes to an unrelated URL.

This PR will fix it by adding a space (`" "`) before and after the URL, and then the `ASCII color` text has not added to the l


pending-changes.yaml
```diff
@@ -83,7 +83,7 @@ Apollo SDK:
   feature:
     - 
   bug-fix:
-    - 
+    - wrong url when running `cypress run`
```
zamboney, line 86: @benbabayoff8

amitzur:
@zamboney I don't like the double space that this creates in the terminal, but I realize the tradeoff (that otherwise it's much more work). It's very subtle and probably won't be noticed by anyone, or if it's noticed it won't bother. Just pointing it out.

pending-changes.yaml
amitzur: @zamboney I thought it's when running `cypress open`. No?

benbabayoff8:
@amitzur We have many more interesting things to solve and discuss. If the double spaces will solve the issue I'm fine with that

pending-changes.yaml
zamboney: TLDR; no

There is a `cypress.config` called `isTextTerminal` that is set to `true` when `cypress run` and `false` on `cypress.open`

We use it [here](https://github.com/applitools/eyes.sdk.javascript1/blob/5f7665e005534e8ce819447ab055b02a35937409/js/packages/eyes-cypress/src/browser/commands.js#L71) to add/remove the `ASCII colorize`

zamboney:
@amitzur i remove one space in the beginning of the URL. I think it's enough. What do you think?

PR #1236, opened 2023-01-05
[core] remove duplicated renderers

zamboney:
When the **same** renderer sent to `settings.renderers` its added as a step to the same renderer in eyes.

This fix will add an `id` property (base on the `settings.renderers` index) to make sure that each renderer is unique

The flow:
```
// step 1
- firefox // same
- firefox-1 // same

// step 2
- firefox // same

// step 3
- firefox-1 // same

// expected result
[
  firefox {
    step1
    step2
  }
  firefox-1 {
    step1
    step3
  }
]
```

What expected

- add index of duplication.
- place it as part of the `Renderer` and `@internal` for it.
- the first duplication isn't set a number

TODO:

- [x] tests


kyryloonufriiev (commented):
What if I would have 
```
eyes.check({
  settings: {
    renderers: [
      {name: 'firefox', width: 100, height: 100},
      {name: 'firefox', width: 100, height: 100},
    ]
  }
})
eyes.check({
  settings: {
    renderers: [
      {name: 'chrome', width: 100, height: 100},
      {name: 'safari', width: 100, height: 100},
      {name: 'firefox', width: 100, height: 100},
    ]
  }
})
```

As far as I understand, the second check's firefox will have an index of 3 that will be recognized as a different renderer when it should re-use one of the previous firefox tests.

amitzur:
@kyryloonufriiev what you describe is an internal use case that is noy currently possible. It breaks the assumption that array of output TestResults is correlated with the input array of browsers, as there is no one such input array, but several.
If this is a functionality we want to maintain, it should be clear and documented how the system behaves, and more tests should cover it.
I'd be happy to discuss f2f.

kyryloonufriiev:
@amitzur Let's discuss this in our sync today. I just want to mention that under my impression no one really cares about the order of test results, because I remember the moment when we were asked to add `browserInfo` (now `renderer`) to the test result object because people didn't know how to correlate them with passed renderers.

amitzur:
@kyryloonufriiev you are right about that. It's also not documented. But it's possible that therr are users out there that do correlate it, and we should be very explicit if we want to change this invariant.

kyryloonufriiev:
@amitzur Hmmm. It is not going to be changed for sdk users. Because in their case they are not going to change `renderers` array at all.
My initial comment was related to the core user (which means us). The core api accepts renderers in `check` this means that there could be different arrays with different order and different sets of browsers, and that means that indexing is not a reliable way to create an id for the renderer.
I agree with you that from the SDK user perspective, it is good enough because they not going to change `renderers` between `check`s.

kyryloonufriiev:
@amitzur, To be honest, I don't even understand what a point to having two tests for exactly the same renderer is. I think if this happens we should simply ignore the duplicated renderer, maybe the console log a warning when we do so. This will save the grid's resources and should not harm SDK users in any way.

kyryloonufriiev (changes requested):

js/packages/core/src/ufg/check.ts
```diff
@@ -222,11 +225,28 @@ export function makeCheck<TDriver, TContext, TElement, TSelector>({
...
+    if (duplicates.has(key)) {
+      index = duplicates.get(key) + 1
+      result.push({...value, id: `${typeof value.id !== 'undefined' ? value.id + '-' : ''}${index}`})
+    } else {
+      result.push(value)
+    }
+    duplicates.set(key, index)
+  }
+  return result
+}
```
kyryloonufriiev, line 252: You can move it to the local utils `./ufg/utils`

kyryloonufriiev: This way the testing will be much clearer

js/packages/core/test/it/ufg/check.spec.ts
```diff
@@ -282,4 +283,127 @@ describe('check', () => {
...
+      input: [
+        {name: 'firefox', width: 100, height: 100, id: 'bla'},
+        {name: 'firefox', width: 100, height: 100, id: 'bla'},
+      ],
+      output: [
+        {name: 'firefox', width: 100, height: 100, id: 'bla'},
+        {name: 'firefox', width: 100, height: 100, id: 'bla-1'},
+      ],
+    },
+  ]
```
kyryloonufriiev, line 378: This could go to the test of your function

kyryloonufriiev (changes requested):

kyryloonufriiev (approved):
Looks great now! Feel free to merge

PR #1238, opened 2023-01-09
Feat/self healing support amit

amitzur:
## THIS PR ISN'T INTENDED TO BE MERGED!

### DIFF
--- a/js/packages/eyes-selenium/example/SelfHealing.spec.ts
+++ b/js/packages/eyes-selenium/example/SelfHealing.spec.ts
@@ -1,3 +1,35 @@
+/**
+ * Instructions:
+ *
+ * Prerequisites:
+ * ==============
+ * 1. Node.js (preferrably 18, but 14 or 16 should work as well)
+ * 2. Yarn 1 (not 2) - https://classic.yarnpkg.com/en/docs/install
+ * 3. Git
+ *
+ * Setup local env:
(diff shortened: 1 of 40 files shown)


js/packages/eyes-selenium/example/SelfHealing.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  before(async () => {
+    log('Getting Execution Cloud URL')
+    const executionCloudUrl = await Eyes.getExecutionCloudUrl({
+      capabilities: {
+        eyesServerUrl: serverUrl,
+        apiKey,
+        useSelfHealing: true,
+      },
+    })
```
kyryloonufriiev, line 60: Why do you want to pass capabilities there? I think it is much nicer to pass them in the intended way. My idea was to allow passing an optional configuration object here, just because a user may already create it for the eyes class and we can extract proxy/serverUrl/apiKey from there.
Or if a user had already created an instance of Eyes, then it is even better because all of the configs are already specified and it will look much nicer.
If you don't like the standard capabilities approach because it is over-verbose with all of the `applitools:blablabla` stuff, I was thinking to add `applitools:options` object where you can specify multiple props without prefixes, like `sauce:options` etc.

kyryloonufriiev:
@amitzur I will close it

PR #1241, opened 2023-01-11
use dom-shared to determine if shadowRoot is not fake

SophieDePaula:
[trello card](https://trello.com/c/9JcRAnua/89-consolidate-iscustomshadowroot-logic)

### DIFF
--- a/js/packages/snippets/package.json
+++ b/js/packages/snippets/package.json
@@ -28,7 +28,9 @@
     "version": "bongo version",
     "postversion": "bongo postversion --skip-release-notification"
   },
-  "dependencies": {},
+  "dependencies": {
+    "@applitools/dom-shared": "^1.0.9"
+  },
   "devDependencies": {
     "@applitools/bongo": "^2.2.0",
     "@babel/core": "^7.11.6",
--- a/js/packages/snippets/src/browser/addElementIds.js
+++ b/js/packages/snippets/src/browser/addElementIds.js
@@ -1,10 +1,12 @@
+const {isShadowRoot} = require('@applitools/dom-shared')
+
 function addElementIds([elements, ids]) {
   return elements.map((element, index) => {
     const path = [element]
     if (element.getRootNode) {
       for (
         let root = element.getRootNode();
-        root !== document && root.constructor && root.constructor.name === 'ShadowRoot';
+        root !== document && isShadowRoot(root);
         root = root.host.getRootNode()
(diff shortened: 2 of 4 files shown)


kyryloonufriiev:
The `dom-shared` package is not something we want to maintain forever. At some point, it was created to share logic between two dom scripts, but it was never supposed to contain a utility logic for other packages. This is why I don't think this is the right move, especially for such a trivial function. I you really have a need to replace this with something shared you can use the `utils.types.istanceOf(maybeShadowRoot, 'ShadowRoot')` this utility function is already located in a shared place and doing exactly what you need.

SophieDePaula:
> The `dom-shared` package is not something we want to maintain forever. At some point, it was created to share logic between two dom scripts, but it was never supposed to contain a utility logic for other packages. This is why I don't think this is the right move, especially for such a trivial function. I you really have a need to replace this with something shared you can use the `utils.types.istanceOf(maybeShadowRoot, 'ShadowRoot')` this utility function is already located in a shared place and doing exactly what you need.

@kyryloonufriiev I am not entirely sure how that would help if I need to send wither the constructor name to it or a function that will be executed. Either way the logic will be duplicated in both places.

kyryloonufriiev:
This is why I said I don't think it is even worth doing something about it. `root.constructor.name === 'ShadowRoot'` is the way to check if the node is shadow root, replacing it with `isShadowRoot(root)` is just looking better but it doesn't release you from knowing what you want to check for (the `ShadowRoot`).
The thing is adding a dependency of `dom-shared` isn't right at all. It is a different repository at the moment, so if we going to add this we will have to manually control the order of releasing. For example. I did a change in `dom-shared` and released `dom-snapshot`, the logical thing I would do today is release `core` that depends on `dom-snapshot` and that's it, still a manual step but a small one, with what you suggest we will have to release everything up the tree beginning from snippets.
I am all for code reuse and generalization, but this way you are creating a precedent that isn't right in the long term or in short term.

SophieDePaula:
@kyryloonufriiev I definitely don't want to add more complication to our release process.
The initial issue for me was that I was not aware of the logic in `dom-snapshot` when I had to deal with a custom issue, so @amitzur suggested that we consolidate the logic both to not duplicate it and to be aware that this logic exists in both places.
The logic is very simple, maybe I can just write a comment to reference the logic in `dom-snapshot`, what do you guys think?

PR #1242, opened 2023-01-12
update types to support coded region by element with region options

SophieDePaula:
[Trello card](https://trello.com/c/xaBaXl3b/1695-island-cypress-ufg-smart-ignore-region-not-working-with-cygetthen)

While working on this I also found that we didn't support sending region options to floating and accessibility regions with elements so this PR also fixes that.


pending-changes.yaml
```diff
@@ -83,7 +83,8 @@ Apollo SDK:
   feature:
     - 
   bug-fix:
-    - 
+    - Fix types to include region options when passing an element
```
SophieDePaula, line 86: @benbabayoff8

pending-changes.yaml
```diff
@@ -83,7 +83,8 @@ Apollo SDK:
...
+    - Add support for region options for coded floating and accessibility when using elements
```
SophieDePaula, line 87: @benbabayoff8 (I added two lines)

kyryloonufriiev (approved):

PR #1243, opened 2023-01-15
[logger] add `debug` console handler

zamboney:
Adding the [`debug`](https://www.npmjs.com/package/debug) console for `@applitools/logger`

Example:

[asciicast](https://asciinema.org/a/551841?speed=1.5)

### TODO:
- [x] tests
- [x] readme


js/packages/logger/src/logger.ts
```diff
@@ -45,6 +46,10 @@ export function makeLogger({
       handler = {type: 'console'}
     }
   }
+  if (process.env.APPLITOOLS_SHOW_LOGS && process.env.DEBUG) {
```
amitzur, line 49: It would be better to play along with the intent of the code above, and define the handler by yourself inside the `if (!handler)`.
I also think there's no need to check `APPLITOOLS_SHOW_LOGS`. I would say that if the user defined `appli:` in the DEBUG env var, then they want logs. Here's my suggestion:

```
if (!handler) {
  if (process.env.APPLITOOLS_LOG_FILE) {
    handler = {type: 'file', filename: process.env.APPLITOOLS_LOG_FILE}
  } else if (process.env.APPLITOOLS_LOG_DIR) {
    handler = {type: 'rolling file', dirname: process.env.APPLITOOLS_LOG_DIR}
  } else if (/appli\:/.test(process.env.DEBUG)) {
    handler = {type: 'debug', label}
    format = (chunks: any[]) => defaultFormat(chunks, {prelude: false})
  } else {
    handler = {type: 'console'}
  }
}

if (!utils.types.isNumber(level)) {
  level =
    level ??
    (process.env.APPLITOOLS_LOG_LEVEL as LogLevelName) ??
    (process.env.APPLITOOLS_SHOW_LOGS === 'true' || (utils.types.has(handler, 'type') && handler.type === 'debug')
      ? 'all'
      : 'silent')
  level = LogLevel[level] ?? LogLevel.silent
}
```

amitzur (changes requested):
I'm all for it

js/packages/logger/src/logger.ts
zamboney: let me address each section by itself:

> define the handler by yourself inside the `if (!handler)`

This is what I first did, but it has a wired behavior like this:

Screen Shot 2023-01-16 at 11 31 59


You **must** set the current `label` to be the same as the logger. I will improve it with some `Map`

>  I also think there's no need to check `APPLITOOLS_SHOW_LOGS`

I disagree with you. The whole point of the approach is that you are logging based on context and not level. If you do not agree, i will change it

js/packages/logger/src/logger.ts
```diff
@@ -41,6 +42,13 @@ export function makeLogger({
       handler = {type: 'file', filename: process.env.APPLITOOLS_LOG_FILE}
     } else if (process.env.APPLITOOLS_LOG_DIR) {
       handler = {type: 'rolling file', dirname: process.env.APPLITOOLS_LOG_DIR}
+    } else if (process.env.APPLITOOLS_SHOW_LOGS && process.env.DEBUG) {
+      handler = {type: 'debug'}
+      level = LogLevel.all
+      format = (chunks: any[], {label}) => {
+        ;(handler as Handler).extend(label)
+        return defaultFormat(chunks, {prelude: false})
```
kyryloonufriiev, line 50: This way you will extend the handler for each `format` call. Also, the fact that you have to do `as Handler` highlights that this maybe not the best solution.

kyryloonufriiev (commented):
I liked the previous iteration more. As for me having `extend` method to set labels on the handler isn't robust enough, but I did like it better as a property of `DebugHandler`.
I understand the problem that you demonstrated in the image. The label of our logger was used together with the `debug`'s label. This means that you simply need a `format` that will return the message without a label (I wouldn't throw away the whole prelude, there is other important info as well).

js/packages/logger/test/e2e/logger.spec.ts
```diff
@@ -184,6 +185,46 @@ describe('logger', () => {
     fs.rmSync(dirname, {recursive: true})
   })
 
+  it('handler debug', async () => {
+    // this is the test function that will be called by spawn to test the debug handler
+    // must use spawn to check the e2e flow
```
kyryloonufriiev, line 190: Why?

zamboney: the `handler debug` is an environment variable base. i want to test that it work as expected

js/packages/logger/src/logger.ts
zamboney: i agree that this is a "Code Smell" but the creation of the handler isn't very customizable. i will revert to the solution that each call to `makeLogger` create a new `DebugHandler` if you prefer that.

zamboney:
> I liked the previous iteration more. As for me having `extend` method to set labels on the handler isn't robust enough, but I did like it better as a property of `DebugHandler`. I understand the problem that you demonstrated in the image. The label of our logger was used together with the `debug`'s label. This means that you simply need a `format` that will return the message without a label (I wouldn't throw away the whole prelude, there is other important info as well).

as mentioned, I will revert to what I did before, but once using `DebugHandler` there is no need to use `prelude`.

js/packages/logger/test/e2e/logger.spec.ts
kyryloonufriiev: @zamboney Why not just set those properties to `process.env`? This way you are testing that nodejs properly transfers env vars to this object, I don't think this is necessary. For nodejs program whatever is in `process.env` is an environment variable. Feel free to point out if the `spawn` is really necessary.

kyryloonufriiev:
@zamboney Are you sure you checked what is in prelude of our logger? What debug can provide instead? label and date?

zamboney:
Debug provides time and label and this is the standard that by Cypress etc

js/packages/logger/test/e2e/logger.spec.ts
zamboney: `debug` package encapsulated the `process.env` once require/import. I can use the require.cache technique but this and what I did was both hacky. If you prefer the latter then I will change it.

kyryloonufriiev:
@zamboney But what about tags? We still want to show them as part of the log line, also the log level might be helpful (I don't care that mush about log level). So debug does not replace the whole prelude

zamboney:
I never see a tag in the log that I review but I can add them as part of the format, the problem with the format is that it forces the label and log level as part of the message.

amitzur:
@kyryloonufriiev can you provide a clear view on what the prelude might include?
Where are tags showing today in the log line?

I'd like to make it clear where I think this PR fits in. It should also be communicated with Product.
1. This is meant for JS SDK's only - since the convetion of `DEBUG` env var exists (AFAIK) only in JS ecosystem.
2. We would like to recommend to JS SDK users to turn on logging with this env var, instead of `ConsoleLogHandler`. BTW I can also see us developing multiple log handlers - the user specified a `FileLogHandler` AND runs with `DEBUG=applitools:*`.
3. This will also be used for development and troubleshooting purposes (so consumers are JS SDK dev team and support).
4. When running with `DEBUG`, there should still be enough information to use by the future log analyzer, and the log analyzer would need to support this format as well. This complicates things a bit, but I think it's worth it.
Is there enough information when running with `DEBUG`? I think so. The absolute timestamp is not that important IMO, and it is replaced in DEBUG by relative time `[+XXXms]`. Anything else that I missed?

Given these points, let's make sure that we meet these requirements, and not forget to include Product in the discussion.

kyryloonufriiev:
@amit debug provides only page and timestamp. But tags are also part of the prelude, and we have to include them for log analyzing. So the debug handler should have a custom formatter that will not remove tags

zamboney:
@kyryloonufriiev as long as we are agree about the custom format, I will add the tags

kyryloonufriiev:
There is nothing to agree on. The current format is `label timestamp loglevel tags message`, debug can add first two, so we should hide them when we use debug handler

zamboney:
And the colorized, so once strip label timespan and colorized what was left from the format?

kyryloonufriiev:
I don't care about log level too, however, if I have logs from the customer with log level info I can easier find all of the errors which is helpful. I don't think that debug log practice has any limitations regarding what we should write in the message, and we say that we should write tags and some custom tags.

zamboney:
@kyryloonufriiev i added the tags and log level as requested please take anther look 10x

js/packages/logger/test/e2e/logger.spec.ts
kyryloonufriiev: Ok. I get what you mean. Then your solution is reasonable enough.

js/packages/logger/test/e2e/logger.spec.ts
```diff
@@ -181,7 +183,63 @@ describe('logger', () => {
       assert.strictEqual(output, expected[index])
     })
 
-    fs.rmSync(dirname, {recursive: true})
+    // adding here ts-ignore undtil we upgrade the @types/node version to 14 where `fs.rm` is available
+    // @ts-ignore
+    await promisify(fs.rm)(dirname, {recursive: true})
```
kyryloonufriiev, line 188: Instead of promisification you can simply use `fs.promises.rm`.

js/packages/logger/src/logger.ts
```diff
@@ -36,6 +37,11 @@ export function makeLogger({
   console = true,
   extended = false,
 }: LoggerOptions & {extended?: boolean} = {}): Logger {
+  if (process.env.APPLITOOLS_SHOW_LOGS && process.env.DEBUG) {
+    handler = {type: 'debug', label}
+    level = LogLevel.all
+    format = (chunks: any[], {tags, level}) => defaultFormat(chunks, {level, prelude: true, tags, timestamp: false})
+  }
```
kyryloonufriiev, line 44: Wasn't it decided that `APPLITOOLS_SHOW_LOGS` shouldn't be used together with `DEBUG`?

js/packages/logger/src/logger.ts
```diff
@@ -36,6 +37,11 @@ export function makeLogger({
   console = true,
   extended = false,
 }: LoggerOptions & {extended?: boolean} = {}): Logger {
+  if (process.env.APPLITOOLS_SHOW_LOGS && process.env.DEBUG) {
+    handler = {type: 'debug', label}
+    level = LogLevel.all
+    format = (chunks: any[], {tags, level}) => defaultFormat(chunks, {level, prelude: true, tags, timestamp: false})
```
kyryloonufriiev, line 43: Instead of doing this, there is an easy configure all of this in `makePrinter`, you can check out what argument we are passing there.

js/packages/logger/src/logger.ts
```diff
@@ -36,6 +37,11 @@ export function makeLogger({
...
+  }
   if (!handler) {
```
kyryloonufriiev, line 45: no need for this check if the first one was triggered
```suggestion
  else if (!handler) {
```

js/packages/logger/README.md
```diff
@@ -0,0 +1,39 @@
+# `@applitools/logger`
```
kyryloonufriiev, line 1: 👍🏼

kyryloonufriiev (changes requested):

js/packages/logger/src/logger.ts
zamboney: the `debug` practice is just with `DEBUG` and not with `APPLITOOLS_SHOW_LOGS` @amitzur WDYT?

js/packages/logger/src/logger.ts
zamboney: done

js/packages/logger/src/logger.ts
zamboney: @kyryloonufriiev  I'm not sure what you want me to do with `makePrinter` here... can you please explain?

js/packages/logger/test/e2e/logger.spec.ts
zamboney: 👍

js/packages/logger/src/logger.ts
kyryloonufriiev: @zamboney Why do you need to override format if everything you can do is could be configured in the printer, just take a look at how it is organized there. You can simply override the arguments you want to be different when debug handler is applied. For example, you want the timestamp to be `false` (you set it in your formatter), so you can just do `timestamp = false`. Then you don't want a label so you can do `label = false`

zamboney: @kyryloonufriiev I don't understand the difference

kyryloonufriiev: You have a way to configure what you need. I don't understand the reason to invent another one.

js/packages/logger/src/logger.ts
kyryloonufriiev: This is what I am saying. If debug is specified we don't care about anything else.
I think Amit shared his point of view above
>I also think there's no need to check APPLITOOLS_SHOW_LOGS. I would say that if the user defined appli: in the DEBUG env var, then they want logs.

js/packages/logger/src/logger.ts
zamboney: Printer handler that has been created (a.k.a. has log method) so using the Printer can't be used at this stage....

kyryloonufriiev: Wouldn't it achieve exactly what you need, but in the way it was intended?
```suggestion
    timestamp = false
    label = null
```

zamboney: used `makePrinter` please take a look 10x

kyryloonufriiev (approved):
👍🏼

js/packages/logger/src/logger.ts
```diff
@@ -36,7 +37,12 @@ export function makeLogger({
   console = true,
   extended = false,
 }: LoggerOptions & {extended?: boolean} = {}): Logger {
-  if (!handler) {
+  if (process.env.DEBUG) {
```
amitzur, line 40: @zamboney @kyryloonufriiev you should check if `/appli:./.test(process.env.DEBUG)` otherwise any DEBUG value would trigger our logs.

zamboney: There are some cases where the logs will trigger and I'm fine with them:

1. `*`: If the user wants to open all of the logs then I'm fine with that
2. `appli`: This is when we want the logs to shown

Any of the other cases the logs will not shown

amitzur: @zamboney you're right. In addition, `DEBUG` supports comma separated lists, so you also want `DEBUG=mocha:*,appli:*`.

zamboney: @amitzur i don't think we should do what you suggested. We should move the condition so that it will first check if the `APPLITOOLS_SHOW_LOGS` is present and then check if `DEBUG` is present. @amitzur WDYT?

PR #1245, opened 2023-01-17
[monorepo] global tooling and linting

kyryloonufriiev:
With this PR I want to move our toolings such as eslint and typescript-related tools to the `js` dir level. This serves mainly 4 reasons:

1. Remove the duplicated annoying stuff from each package
2. Unify the linting rules by using the same base file instead copypasting it everywhere
3. Unify the basic ts configuration to keep the rules the same everywhere
4. Use the same eslint/typescript version in every package

If this PR gets merged everyone in @applitools/js-sdks will have to run `yarn` once in the `js` directory


zamboney:
@kyryloonufriiev I understand what you are doing and think this is an excellent idea! but in my opinion, a better way to do it is using `@applitools/bongo` as the `eslint` `prettier`, and `tsconfig`. I prefer using a **shared** package than using the file structure, but this is my opinion

kyryloonufriiev:
@zamboney I see your point. I was thinking about it, to be honest I don't see using `@applitools/bongo` for it. But it make sense to create a package for our own eslint rules or basic tsconfig.
The reason I didn't took this direction is because I don't want to be in a situation when I would not add some new rule or tsconfig, because after it I will have to release some package, and after that re-release everything. I want it to be as easy as possible.
I absolutely agree that for most of the things such as code, we want to do packages, and not use the repo structure. But for such thing as tooling it is just makes more sense, on my opinion. BTW I did check other projects with mono repos doing (such as playwright, pptr, wdio), so look like it is pretty popular practice.

kyryloonufriiev:
@zamboney Can you elaborate on what you see in react repo?
I understand they have a bunch of scripts in the main root/repo pacakge.json, I think it is great, and we definitely should do something like that. But I don't think that we should have to make a root-level command to lint everything, instead of package-level linting. Don't get me wrong, I like the idea to have such a command, but I don't think it should be a replacement for what we have right now. We have lots of packages, with different maintenance levels, so I don't want to find myself fixing something we don't really care about. What I want to do though is auto-linting for changed packages on commit.

kyryloonufriiev:
Also looking after looking at react repo, made me pretty positive about I would really love to kill the `@applitools/scripts` package and have the `scripts` folder along the `packages` folder in the `js`. I just remind myself of a situation when I was dodging doing a change in the `scrips` package because I didn't want to re-release everything after.

zamboney:
@kyryloonufriiev the idea about `eslint` the whole repo (only the js folder of course) is that once you don't have an `eslint` config for each project and all of the packages are aligned with the lint system, you can just take as an input the file that changes and then lint them.

for example: 

let's say we change a file in `core` and another in `core-base`.

in the current system, both of the projects are been linting, this happens because the lint is a **"project"** base

My suggestion is that the lint system will be **"file"** base system, and only the file that changed will be lint if that makes any sense...

---

> Also looking after looking at react repo, made me pretty positive about I would really love to kill the `@applitools/scripts` package and have the scripts folder along the packages folder in the js. I just remind myself of a situation when I was dodging doing a change in the scrips package because I didn't want to re-release everything after.

this is a great idea!

kyryloonufriiev:
@zamboney it makes perfect sense. But I still want to be able to run `yarn lint` in the package to do what it does today, as well as change the rules on the package level if needed. I don't think it contradicts with what you said.

amitzur:
@kyryloonufriiev @zamboney I am enjoying this discussion and agree with both of you.
Doing things on a per-package basis is important, and having a `scripts` folder instead of a package makes total sense.

Why is there a need for both `tsconfig.build.json` and `tsconfig.json`?
In packages where `.eslint` is just extending the base, don't you think it's fine to just not have the file?

kyryloonufriiev:
We need `tsconfig.json` as the default behavior that IDE uses to understand what's what in the package, the `tsconfig.build.json` extends the default behavior with additional build configuration. The most important part is that I want to include `test` folder so vscode will provide me with the hints there, but I don't want to compile those files to js.
About package-level `.eslintrc` it is much simpler, there is no practical need for that, I just like it explicit. Instead of traversing through the directories structure, I can just immediately see where eslint config for this package is defined.

amitzur:
@kyryloonufriiev got it. I agree about being explicit with `.eslint`. As for `tsconfig`, I understand the need. It's like having a bundler in `dev` or `prod` mode. It would be nice to avoid it if possible, though. Just because it's now already 3 `tsconfig` files that come into play when compiling a package.

kyryloonufriiev:
Thanks, everyone for the input, I am going to merge it!

PR #1246, opened 2023-01-17
Add NML Android support to the SDK

tourdedave:


pending-changes.yaml
```diff
@@ -68,23 +71,26 @@ Apollo SDK:
...
 "@applitools/eyes-cypress":
   feature:
     - Support ESM and TypeScript in Applitools' config file
+    - Add NML support for Android
   bug-fix:
     - 
 "@applitools/eyes-storybook":
   feature:
-    - 
+    - Add NML support for Android
```
kyryloonufriiev, line 93: Doesn't make sense in storybook, cypress, pptr, and playwright

js/packages/core/src/utils/extract-broker-url.ts
```diff
@@ -1,11 +1,16 @@
 import {type Driver} from '@applitools/driver'
 
 export async function extractBrokerUrl(driver: Driver<unknown, unknown, unknown, unknown>): Promise<string> {
-  if (!driver.isIOS) return null
   try {
+    const selector = driver.isIOS
+      ? '//XCUIElementTypeOther[@name="Applitools_View"]'
+      : driver.isAndroid
+      ? "//android.widget.TextView[contains(@content-desc,'Applitools_View')]"
```
kyryloonufriiev, line 8: Why `contains`? Isn't it the exact value?

kyryloonufriiev (approved):
LGTM 🎉 , just two small comments

pending-changes.yaml
tourdedave: Right. I got a little trigger happy.

js/packages/core/src/utils/extract-broker-url.ts
tourdedave: It is. So, this? `//android.widget.TextView[@content-desc="Applitools_View"]`. That appears to be legit based on what I see in the inspector. I'll try it.

PR #1257, opened 2023-01-29
[logger] fix `debug` issue

zamboney:
## The Problem
If `DEBUG` environment variable exists and anther handle set it will always use the `DebugHandler`

## The Solution
This PR fix it by passing the `handler` to `makeLogger` only if the `forceInitHandler` is `false`

---

Also it will always choose `{type: 'console'}` if `process.env.APPLITOOLS_SHOW_LOG` exists


amitzur:
@zamboney I don't like this patch with `forceInit`. Why is it necessary? I'd like the API to stay small and reasonable - either you pass in a `handler` or you don't. And if you don't then a `handler` will be created for you from env vars.

zamboney:
@amitzur i see what you mean, but the `forceInit` is necessary and i will try to explain why

the way that the `logger` works is that it will create a `handler` only if the `handler` is not exists, and if it exists it will use it. this is a problem for the `debug` because the way that the `label` interact with the `format`. the `format` is the once that print the `label` but the `debug` assign the `label` **base on the `handler` [initializing](https://github.com/applitools/eyes.sdk.javascript1/blob/e968b59b54dbe3e84012eacc88600d157ab55e86/js/packages/logger/src/handler-debug.ts#L13-L21)**. so to sort this out, once the `handler` been set with the `label` the `makeLogger` assign the `label` to `undefined`. this is why we **must** create a new `handler` each request to `makeLogger`. 

hope it make sense 🤷‍♂️

but i see what you mean about the whole new flag and i agree that this is not reasonable, so i made a small change so it will be self assign (base on the [`extended`](https://github.com/applitools/eyes.sdk.javascript1/blob/e968b59b54dbe3e84012eacc88600d157ab55e86/js/packages/logger/src/logger.ts#L38)).

amitzur:
@zamboney the most crucial problem right now is the wrong check for `if (process.env.DEBUG)`. If the user passes `DEBUG=mocha:*` then our debug handler would be triggered. This is what you need to solve first and foremost.

Second, I think I understand your point, and I think the problem is that you tied the `handler` to the `label` in the type `DebugHandler`. I think it's better to take the label out of the `handler`. Like this (scattered code):
```
// handler-debug.ts
type DebugHandler = {
  type: 'debug'
}

// logger.ts #1
if (!handler) {
  if (process.env.APPLITOOLS_LOG_FILE) {
    handler = {type: 'file', filename: process.env.APPLITOOLS_LOG_FILE}
  } else if (process.env.APPLITOOLS_LOG_DIR) {
    handler = {type: 'rolling file', dirname: process.env.APPLITOOLS_LOG_DIR}
  } else if (process.env.DEBUG && process.env.DEBUG.split(',').some(s => /appli:.+/.test(s))) {
    handler = {type: 'debug'}
    level = LogLevel.all
    label = null
    timestamp = false
  } else {
    handler = {type: 'console'}
  }
}

// logger.ts #2
if (utils.types.has(handler, 'type')) {
  if (handler.type === 'console') {
    handler = makeConsoleHandler()
  } else if (handler.type === 'debug') {
    handler = makeDebugHandler({label})
  } else if (handler.type === 'file') {
    handler = makeFileHandler(handler)
    colors = undefined
  } else if (handler.type === 'rolling file') {
    handler = makeRollingFileHandler(handler)
    colors = undefined
  }
}
```

zamboney:
> @zamboney the most crucial problem right now is the wrong check for `if (process.env.DEBUG)`. If the user passes `DEBUG=mocha:*` then our debug handler would be triggered. This is what you need to solve first and foremost.

I think that is the point... if you set the `process.env.DEBUG` you expect to see only what it's produced. but if before it overrides the none "default" `handler` now it overrides just the "default" one. if you think it's not ok i will add this condition but i would expect to see just what the `debug` write to the `console` 

> Second, I think I understand your point, and I think the problem is that you tied the `handler` to the `label` in the type `DebugHandler`. I think it's better to take the label out of the `handler`. Like this (scattered code):
> 
> ```
> // handler-debug.ts
> type DebugHandler = {
>   type: 'debug'
> }
> 
> // logger.ts #1
> if (!handler) {
>   if (process.env.APPLITOOLS_LOG_FILE) {
>     handler = {type: 'file', filename: process.env.APPLITOOLS_LOG_FILE}
>   } else if (process.env.APPLITOOLS_LOG_DIR) {
>     handler = {type: 'rolling file', dirname: process.env.APPLITOOLS_LOG_DIR}
>   } else if (process.env.DEBUG && process.env.DEBUG.split(',').some(s => /appli:.+/.test(s))) {
>     handler = {type: 'debug'}
>     level = LogLevel.all
>     label = null
>     timestamp = false
>   } else {
>     handler = {type: 'console'}
>   }
> }
> 
> // logger.ts #2
> if (utils.types.has(handler, 'type')) {
>   if (handler.type === 'console') {
>     handler = makeConsoleHandler()
>   } else if (handler.type === 'debug') {
>     handler = makeDebugHandler({label})
>   } else if (handler.type === 'file') {
>     handler = makeFileHandler(handler)
>     colors = undefined
>   } else if (handler.type === 'rolling file') {
>     handler = makeRollingFileHandler(handler)
>     colors = undefined
>   }
> }
> ```

this code will keep printing the first `label` because once the `DebugHandler` set, the native `debug` package `extend` called  and it just keeps writing the first label 

here is the output

```
      + expected - actual

       appli:label-with-spaces-and-upper-cases [INFO ] log
       appli:label-with-spaces-and-upper-cases [WARN ] warn
       appli:label-with-spaces-and-upper-cases [ERROR] error
       appli:label-with-spaces-and-upper-cases [FATAL] fatal
      -appli:label-with-spaces-and-upper-cases label2    | [INFO ] {"tag":"@@@"} log2
      -appli:label-with-spaces-and-upper-cases label2    | [WARN ] {"tag":"@@@"} warn2
      -appli:label-with-spaces-and-upper-cases label2    | [ERROR] {"tag":"@@@"} error2
      -appli:label-with-spaces-and-upper-cases label2    | [FATAL] {"tag":"@@@"} fatal2
      +appli:label2 [INFO ] {"tag":"@@@"} log2
      +appli:label2 [WARN ] {"tag":"@@@"} warn2
      +appli:label2 [ERROR] {"tag":"@@@"} error2
      +appli:label2 [FATAL] {"tag":"@@@"} fatal2
```

amitzur:
@zamboney see #1258 
I understood your point, and found a compromise to not pollute the API of `makeLogger`.

> I think that is the point... if you set the process.env.DEBUG you expect to see only what it's produced. but if before it overrides the none "default" handler now it overrides just the "default" one. if you think it's not ok i will add this condition but i would expect to see just what the debug write to the console

Ok, I get it now.

kyryloonufriiev (changes requested):
I noticed this feature, not about `forceInit`, but about the fact that `DEBUG` env takes priority over any other way to specify log handler, I thought this was by design. Because I believe we already pass handlers (as objects) on the eyes-api level, in actual SDK there will be no case when the handler is not specified. I may be wrong about that.
In any case, this impl looks even better to me, except for the `forceInit` part of the course. I agree with Amit that it doesn't look good, this is why I want to suggest this instead. So with debug handler, we for the first time find out that handlers also may be extendable, we didn't have that with the file or console. So the smart thing to do, in my opinion, would be to add an optional method  `extend` to the `Handler` interface, and implement it for `DebugHandler`, because this is the only handler at the moment that could be extended. Then when you need to extend the logger, instead of just using the same handler you will use the handler's `extend` method first, if it exists of course. I hope you understand what I meant.

zamboney:
@kyryloonufriiev I think I understand what you mean, but I'm not sure about the implications. I will try to explain what I think you mean, and you can correct me if I'm wrong.
you want to create a new `DebugHandler` with the new `label` using the `extend` method of the `DebugHandler` and then use it to the `makeLogger` function.
if this is what you want then the problem is setting the `options` (`timestamp` and `level`). now it's done using the `forceInit`, this is not passable with `extend` because this setting is in the `makeLogger`.

kyryloonufriiev:
@zamboney You understood me right, but you missed the point that you don't need to mutate `label` and `timestamp` because it is already mutated when `extend` is called. However, the label is a little bit trickier... So I'm ok with the current impl.

kyryloonufriiev (approved):

PR #1270, opened 2023-02-02
NML broker URL retry

tourdedave:
This addresses the problem where the NML element is found, but the payload hasn't been populated yet (e.g., it looks like this `{"error":"","nextPath":null}`).

The field reported this issue when testing the customer's app with NML in advance of releasing it to them.


kyryloonufriiev (changes requested):
What is the point of looking for the element again and again? Isn’t is just a matter of waiting for the element's text to be changed?

js/packages/core/src/utils/extract-broker-url.ts
```diff
@@ -13,7 +13,12 @@ export async function extractBrokerUrl(driver: Driver<unknown, unknown, unknown,
       selector,
     })
     const result = JSON.parse(await element.getText())
-    return result.nextPath
+    if (result.nextPath) return result.nextPath
+    if (!result.error) {
+      if (attempt === 60) return null
+      await new Promise(res => setTimeout(res, 500))
```
kyryloonufriiev, line 19: use `utils.general.sleep(ms)`

tourdedave: > What is the point of looking for the element again and again? Isn’t is just a matter of waiting for the element's text to be changed?

Ah, you're right. We should just poll on `getText` until the text has changed.

kyryloonufriiev (approved):

PR #1271, opened 2023-02-02
Storybook external http headers

SophieDePaula:
[Trello card](https://trello.com/c/plcOa7DK/1724-nzme-storybook-sdk-adding-extra-headers-in-order-to-bypass-cloudflare-access-for-standalone-storybook)

The user would like to send http headers, so we need to pass their value to `Page.setExtraHTTPHeaders()`


js/packages/eyes-storybook/src/eyesStorybook.js
```diff
@@ -189,6 +190,7 @@ async function eyesStorybook({
       logger,
       stream: outputStream,
       pagePool,
+      puppeteerExtraHTTPHeaders: config.puppeteerExtraHTTPHeaders,
```
amitzur, line 193: I would pass this to `initPage` rather than to `renderStories`. That's where new pages are created.

amitzur (changes requested):
I agree with the new config parameter and the readme addition. There is only one change I would make, to pass this configuration to `initPage` instead of `renderStories`.

pending-changes.yaml
```diff
@@ -88,7 +88,7 @@ Apollo SDK:
     - 
 "@applitools/eyes-storybook":
   feature:
-    - 
+    - Add the option to send external http headers to Puppeteer Page object
```
SophieDePaula, line 91: @benbabayoff8

js/packages/eyes-storybook/src/initPage.js
```diff
@@ -2,7 +2,7 @@ const {presult} = require('@applitools/functional-commons');
 const browserLog = require('./browserLog');
 const fakeIE = require('./fakeIE');
 
-function makeInitPage({iframeUrl, config, browser, logger, getTransitiongIntoIE, getRenderIE}) {
+function makeInitPage({iframeUrl, config, browser, logger, getTransitiongIntoIE, getRenderIE, puppeteerExtraHTTPHeader}) {
```
amitzur, line 5: why do you need this extra parameter when you have `config.puppeteerExtraHTTPHeaders` ?

amitzur (changes requested):
1. inline comment about the weird API parameter `puppeteerExtraHTTPHeader`.
2. Need to run `yarn lint --fix`
3. Consider avoiding a whole new storybook app, since you actually just need a new test server middleware to serve it. I see that there is `singleStorybook` being used by e2e tests. It's not a big deal, but I was confused why does this test need a storybook app of its own to test the functionality.

js/packages/eyes-storybook/src/initPage.js
SophieDePaula: You are right, I missed it

SophieDePaula:
@amitzur I actually can't use `singleStorybook` in this case. I need my story to include a page that is hosted by `testServerInProcess` so that I use the middleware between the page and Puppeteer to check the headers

PR #1281, opened 2023-02-06
[core-base] crop image base on `renderinfo`

zamboney:
This is the first stage to crop the image base on `renderinfo`
In this stage, we are cropping the snapshot image base on the `renderinfo` `maxImageHeight` and `maxImageArea`
The logic is as follows:

[](https://mermaid.live/edit#pako:eNpNUEFuwyAQ_Mpqz4l6t6pIsZxDL700t5DD2mwAtYC1xaoix38vGKcqJ3ZmdmZgxiFqxgaN0Gjh3KoA-RwvhhM4T4bhx-lkgYIGy87YdIX9_gDt_FbY114OKW4MRClzXSNhKhNkundmqcZtWX7c6OubH3B67_6jSaYMdpdB4rhlV99i098hWQbvAsRbZQv8F7zmwcuzY_XNrmsG7tCzeHI6v3QunMLs5llhk6-a5FOhCkvWTaOmxCftUhRs1qI7pCnFj3sYsCkdn6LOUf41v6mWX3YWbgE)


zamboney:
## few comment on this pr

is feel better for me to create a new property `limit?: ImageServerLimit` i can move it as part of the `setting` but it's not really a "user" setting so it's feels better to use `limit`. i can use `serverSetting` or inside `setting` add `server` but still is feels like twisting my arm

i can remove the `cropToLimit` function and use it on the `transformTarget` if it's not look good to you

need to logic in `cropToLimit` we want to make sure that the `maxHight` limit first before the `maxArea` limit

i liked the `describe('limit')` but i can drop this if it's not look good

kyryloonufriiev (changes requested):

js/packages/core-base/src/types.ts
```diff
@@ -190,6 +190,11 @@ export interface ImageSettings<TRegion = Region> {
   debugImages?: {path: string; prefix?: string}
 }
 
+export interface ImageLimit {
+  height: number
+  area: number
+}
```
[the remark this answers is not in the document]
zamboney: This is what I wrote in my first comment on this pr, my original thought was the way you suggested but it seems (to me) that the `ImageSettings` is more of a _"user setting"_ and not a _"internal setting"_. If you think this isn't the right place for it, then i will move it as you suggested.

js/packages/core-base/src/utils/transform-target.ts
```diff
@@ -43,3 +49,28 @@ export async function transformTarget({
...
+    height: image.height,
+  }
+
+  const maxImageHeight = limit?.height ?? Infinity
+  const maxImageArea = limit?.area ?? Infinity
+  if (image.height > maxImageHeight) {
+    crop.height = maxImageHeight
+  }
+  if (crop.height * image.width > maxImageArea) {
+    crop.width = maxImageArea / crop.height
```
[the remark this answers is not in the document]
zamboney: I didn't check what the backend was doing. I will check, but i don't think there is any other way to check this...

js/packages/core-base/src/utils/transform-target.ts
```diff
@@ -43,3 +49,28 @@ export async function transformTarget({
...
+  }
+  if (crop.width < image.width || crop.height < image.height) {
```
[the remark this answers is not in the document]
zamboney: Again my thought was to do it so i would not call to `image.corp` if it was not needed. But if you think this is unnecessary, i can remove it.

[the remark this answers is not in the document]
kyryloonufriiev: It is unnecessary because it is already optimized in this function so you don't have to think too much about the optimization when you work with images.

js/packages/core-base/src/utils/transform-target.ts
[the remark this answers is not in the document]
kyryloonufriiev: I was under the impression (after talking to Amit) that we have to check the backend's behavior first. Because there obviously two ways to reduce the area of a rectangle - cut a width, or cut a height.
You are trying to reduce it by cutting a width.
I personally think that the better way to do it is to cut a height. But as we already understood we should preserve the server behavior.

js/packages/core-base/src/utils/transform-target.ts
```diff
@@ -43,3 +49,28 @@ export async function transformTarget({
 
   return target
 }
+
+function cropToLimit(image: Image, limit?: ImageServerLimit): void {
+  if (!limit) {
+    return
+  }
```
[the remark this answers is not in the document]
kyryloonufriiev: The thing is that you should not call it if the limit is `undefined` in the first place, if it would be something that the user can call then, of course, we have to do the validation. But in this case, you call it, and you don't need that reach api.

js/packages/core-base/src/types.ts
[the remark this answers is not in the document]
kyryloonufriiev: I think we should not have any behavior that could not be overridden. This means if a user (I am talking user of `core-base`) wants a different image `maxHeight` or `maxArea` and they are not interfering with whatever the server provides (means we would always choose the smallest limit for height/area) we will use the provided values. The server just helps us to get defaults for some of the values, but never dictates something that could not be otherwise configured.

js/packages/core-base/src/utils/transform-target.ts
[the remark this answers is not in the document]
zamboney: good point i miss this, after a talk with backend i will do what they are doing 10x

js/packages/core-base/src/types.ts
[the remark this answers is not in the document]
zamboney: @kyryloonufriiev just to clarify, what you want is to be part of the `ImageSetting`. taking the minimum of `accountinfo` and `ImageSettings.normalization.cut` provide and place it on `imageSetting.normalization.cut`? is that make sense to you?

kyryloonufriiev:
@zamboney I noticed you committed `NO_CHANGELOG`, I think this feature worth a mention in the changelog

pending-changes.yaml
```diff
@@ -11,22 +11,24 @@ Apollo SDK:
 "@applitools/eyes-universal":
   feature:
     - Add NML support for Android
+    - Crop screenshot image base on account info before upload
```
zamboney, line 14: @benbabayoff8 ☝️

js/packages/core-base/src/check-and-close.ts
```diff
@@ -18,6 +18,20 @@ export function makeCheckAndClose({requests, logger: defaultLogger}: Options) {
...
+        maxImageHeight: Math.min(
+          settings.normalization?.limit?.maxImageHeight ?? Infinity,
+          requests.test.account.maxImageHeight,
+        ),
+        maxImageArea: Math.min(
+          settings.normalization?.limit?.maxImageArea ?? Infinity,
+          requests.test.account.maxImageArea,
+        ),
+      },
+    }
```
kyryloonufriiev, line 34: better use a new nullish coalescing operator
```suggestion
    settings ??= {}
    settings.normalization ??= {}
    settings.normalization.limit ??= {}
    settings.normalization.limit.maxImageHeight = Math.min(
      settings.normalization.limit.maxImageHeight ?? Infinity,
      requests.test.account.maxImageHeight,
    )
    settings.normalization.limit.maxImageArea = Math.min(
      settings.normalization.limit.maxImageArea ?? Infinity,
      requests.test.account.maxImageArea,
    )
```

kyryloonufriiev (changes requested):
This also should be added to `extractText` and `locate`

js/packages/core-base/src/locate-text.ts
```diff
@@ -18,8 +18,21 @@ export function makeLocateText({requests, logger: defaultLogger}: Options) {
...
+    settings.normalization.limit = {
+      maxImageHeight: Math.min(
+        settings.normalization.limit?.maxImageHeight ?? Infinity,
+        requests.test.account.maxImageHeight,
+      ),
+      maxImageArea: Math.min(
+        settings.normalization.limit?.maxImageArea ?? Infinity,
+        requests.test.account.maxImageArea,
+      ),
+    }
```
kyryloonufriiev, line 31: Since `locate`, `extractText` and `locateText` are part of the `Core` now, you don't have a test started there. The way to get account info will be:
```suggestion
    const account = await requests.getAccountInfo({settings, logger})
    settings.normalization ??= {}
    settings.normalization.limit = {
      maxImageHeight: Math.min(
        settings.normalization.limit?.maxImageHeight ?? Infinity,
        account.maxImageHeight,
      ),
      maxImageArea: Math.min(
        settings.normalization.limit?.maxImageArea ?? Infinity,
        account.maxImageArea,
      ),
    }
```

js/packages/core-base/src/utils/transform-target.ts
```diff
@@ -28,13 +28,18 @@ export async function transformTarget({
       if (settings.normalization.scaleRatio) image.scale(settings.normalization.scaleRatio)
       if (settings.normalization.rotation) image.rotate(settings.normalization.rotation)
       if (settings.normalization.cut) image.crop(settings.normalization.cut)
+      if (settings.normalization.limit) {
+        const height = cropToHeightLimit(image, settings.normalization.limit)
+        image.crop({y: 0, x: 0, width: image.width, height})
+      }
```
kyryloonufriiev, line 34: Shouldn't it happen after we will take the region from the screenshot? The initial image could be gigantic, but the region could be a small portion of it. I think the limits validation should happen after all other manipulations.

js/packages/core-base/test/it/check.spec.ts
```diff
@@ -119,4 +122,57 @@ describe('check', () => {
 
     assert.strictEqual(checked, 1)
   })
+
+  it('crop image base on "renderinfo" "maxImageHeight" and "maxImageArea"', async () => {
```
kyryloonufriiev, line 126: Please don't use "renderinfo", it is outdated terminology. In our code base, this is called account info.

js/packages/core-base/test/it/transform-target.spec.ts
```diff
@@ -1,3 +1,4 @@
+import {type Image} from '@applitools/image'
 import {readFileSync} from 'fs'
 import {makeImage} from '@applitools/image'
```
kyryloonufriiev, line 3:
```suggestion
import {readFileSync} from 'fs'
import {makeImage, type Image} from '@applitools/image'
```

kyryloonufriiev (changes requested):

.github/actions/parser/main.js
```diff
@@ -251,8 +251,8 @@ function changedInCurrentBranch() {
   const changedFiles = execSync('git --no-pager diff --name-only origin/master', {encoding: 'utf8'})
   const changedPackageNames = changedFiles.split('\n').reduce((changedPackageNames, changedFile) => {
     const changedPackage = Object.values(packages).find(changedPackage => {
-      const changedFilePath = path.resolve(process.cwd(), changedFile)
-      return changedFilePath.startsWith(changedPackage.path)
+      const changedFilePath = path.resolve(process.cwd(), changedFile, './')
+      return changedFilePath.startsWith(changedPackage.path + '/')
```
kyryloonufriiev, line 255: 👍🏼

js/packages/core-base/src/utils/transform-target.ts
zamboney: isn't it what happen now? am i missing something?

kyryloonufriiev: I understand the confusion. There is `normalization.cut` and `region` they are doing basically the same but serve different purposes. So you apply the limits after `normalization.cut`, but before `region`.
Check out line 37

js/packages/core-base/src/extract-text.ts
```diff
@@ -24,6 +24,12 @@ export function makeExtractText({requests, logger: defaultLogger}: Options) {
     settings = utils.types.isArray(settings) ? settings : [settings]
     const results = await Promise.all(
       settings.map(async settings => {
+        const account = await requests.getAccountInfo({settings, logger})
```
amitzur, line 27: @zamboney @kyrylodolynskyi from a user perspective, they should be able to run `extractText` only after calling `open`. So I think in the normal case, we should already have the account info available somewhere. We can avoid sending this request.

zamboney: @amitzur I understand what you are saying, and indeed it seems from the code like a waste of request, but this call came from the same [getAccountInfoWithCache](https://github.com/applitools/eyes.sdk.javascript1/blob/2e843f9ba6b0abe9ddb614793d47fc6e30f1aa01/js/packages/core-base/src/server/requests.ts#L54-L56)

PR #1285, opened 2023-02-08
Storybook 7 beta

SophieDePaula:
[Breaking changes in version 7](https://github.com/storybookjs/storybook/blob/4a532ec8d8922ddf6b7a7e1cc68d88b35377b0a6/MIGRATION.md#frameworkoptions-renamed)


.github/workflows/test.yml
```diff
@@ -28,6 +28,11 @@ on:
       grep:
         description: grep pattern
         required: false
+      nodevresion:
+        description: node version
+        type: choice
+        options: [14, 16,  18]
+        default: 18
```
kyryloonufriiev, line 35: No need to do this. You can specify any node version with the name. Check `parser` action

kyryloonufriiev (changes requested):

.github/workflows/test.yml
kyryloonufriiev: For example:
- with node version `storybook(node:18)`
- with node version and os `storybook(node:18;os:windows)`
- with node version and framework version `storybook(framework:7;node:18)`

pending-changes.yaml
```diff
@@ -93,7 +93,7 @@ Apollo SDK:
     - 
 "@applitools/eyes-storybook":
   feature:
-    - 
+    - Add Support for storybook 7
```
SophieDePaula, line 96: @benbabayoff8

js/packages/eyes-storybook/test/e2e/happy-config/storybook-csf.versions.config.js
```diff
@@ -18,4 +18,5 @@ module.exports = {
     {width: 640, height: 480, name: 'chrome'},
     {width: 1280, height: 960, name: 'chrome'},
   ],
+  waitBeforeCapture: 3000,
```
amitzur, line 21: why?

js/packages/eyes-storybook/test/fixtures/storybook-versions/next/package.json
```diff
@@ -16,19 +16,35 @@
...
+    "@storybook/preset-create-react-app": "next",
+    "@storybook/react": "next",
+    "@storybook/react-webpack5": "next",
+    "@storybook/testing-library": "0.0.14-next.1",
+    "webpack": "5.75.0"
   },
   "snaptdout": {
     "snapshotsPrefix": "v_next_"
+  },
+  "browserslist": {
```
amitzur, line 38: is this mandatory? if not then I think it's just noise

js/packages/eyes-storybook/test/fixtures/storybookCSF/stories/LoginForm.stories.js
```diff
@@ -11,15 +11,18 @@ const clickSubmitDelay = 1000;
...
   title: 'Examples/Login',
   component: LoginForm,
-  excludeStories,
   args: {
     clickSubmitDelay,
     sbVersion,
   },
-};
+}
+if(parseFloat(sbVersion) < 6.4) {
```
amitzur, line 22: Can you explain why this was needed?

pending-changes.yaml
amitzur: lowercase `s` in `Support`

amitzur (commented):
Reviewed without actually running it yet. First iteration looking good. See comments inline.

js/packages/eyes-storybook/test/fixtures/storybookCSF/stories/LoginForm.stories.js
SophieDePaula: `excludeStories` throws an error with storybook 7, so I exclude it from the object

js/packages/eyes-storybook/test/e2e/happy-config/storybook-csf.versions.config.js
SophieDePaula: With storybook 7, I noticed that we take the screenshot before the `play` is done

SophieDePaula: @amitzur I found the problem and fixed it. The way to access the addons changed.

js/packages/eyes-storybook/src/browser/getClientAPI.js
```diff
@@ -10,12 +10,12 @@ const API_VERSIONS = {
...
   const frameWindow = getFrameWindow();
   const clientAPI = frameWindow.__STORYBOOK_CLIENT_API__;
-  const addons = frameWindow.__STORYBOOK_ADDONS;
+  const addons = frameWindow.__STORYBOOK_ADDONS || frameWindow.__STORYBOOK_ADDONS_PREVIEW;
 
   return getAPI(getStorybookVersion());
 
   function getStorybookVersion() {
-    const addons = frameWindow.__STORYBOOK_ADDONS;
+    const addons = frameWindow.__STORYBOOK_ADDONS || frameWindow.__STORYBOOK_ADDONS_PREVIEW;
```
amitzur, line 18: Here you don't need to perform this change. See where `addons` is used. You may refactor the code a bit so that it's clearer what `addons` serves in this function

js/packages/eyes-storybook/test/fixtures/storybook-versions/copyStoriesToVersionDir.js
```diff
@@ -5,32 +5,46 @@ const pexec = p(exec);
 const path = require('path');
 let testsSupportedVersions = [];
 
+
+exports.copyStoriesToVersionDir = async ({ storybookSourceDir, storybookVersion}) => {
+  const versionDir = path.resolve(__dirname, `./${storybookVersion}`);
+  const V7 = fs.existsSync(path.resolve(versionDir, 'node_modules/.bin/sb')) ? true : false;
```
amitzur, line 11: no need to do `... ? true : false`. `fs.existsSync` returns a boolean value

js/packages/eyes-storybook/test/fixtures/storybook-versions/copyStoriesToVersionDir.js
```diff
@@ -5,32 +5,46 @@ const pexec = p(exec);
...
-const getConfigFilesListByVersion = ver => {
+// sb version 7 and up expects framework in main.js
+  const getConfigFilesListByVersion = ver => {
   switch (ver) {
     case '4.1.18':
     case '5.0.0':
       return ['config.js'];
     default:
-      return testsSupportedVersions.includes(ver) ? ['main.js', 'preview.js'] : [];
+      if(V7){
```
amitzur, line 23: this seems overly complicated, using both switch/case and `if`. Choose one and stick with it. Otherwise there's no point for the switch/case.

js/packages/eyes-storybook/test/fixtures/storybookCSF/.storybook/mainV7.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
```
amitzur, line 1: I think adding `mainV7.js` is complicating things. I would create `.storybook-v7` folder which has what is needed, or even consolidate it with `storybookCSFV7` folder.

js/packages/eyes-storybook/test/fixtures/storybookCSF/stories/LoginForm.stories.js
amitzur: try changing the empty array to `undefined`, it would make the code clearer

SophieDePaula:
@benbabayoff8 , @amitzur we have visual diffs when running the same stories with storybook 7 vs previous versions
https://eyes.applitools.com/app/test-results/00000251724551578266/?accountId=UAujt6tHnEKUivQXIz7G6A~~

amitzur (changes requested):
see inline comments, and let's sit f2f to organize the storybook-versions mechanism

js/packages/eyes-storybook/test/fixtures/storybookCSF/stories/LoginForm.stories.js
SophieDePaula: @amitzur changing the value to `undefined` resulted in the same error message.

amitzur:
> @benbabayoff8 , @amitzur we have [visual diffs](https://eyes.applitools.com/app/test-results/00000251724551578266/?accountId=UAujt6tHnEKUivQXIz7G6A~~) when running the same stories with storybook 7 vs previous versions

I verified that this is because the default UI that storybook ships with specifies the font "Nunito Sans". But in version 6.5 they doesn't actually include this font, so the font-family definition doesn't take effect. Only in version 7 did they start shipping with the "Nunito Sans" font.
The impact of this is that users that rely on Storybook's UI will get diffs when upgrading from 6.5 to 7, and the diff **reflects reality correctly**.
In our tests this is not very interesting, so I recommend that we specify our own css and fonts in the storybook that we test, and make sure that all storybook versions land on the same baseline (stop specifying a different `appName` for different versions).

js/packages/eyes-storybook/src/renderStories.js
```diff
@@ -126,6 +126,7 @@ function makeRenderStories({
       return (
         resultsOrErr.constructor &&
         resultsOrErr.constructor.name !== 'Error' &&
+        resultsOrErr.constructor.name !== 'SyntaxError' &&
```
amitzur, line 129: @SophieDePaula Were you able to reproduce this?

SophieDePaula: @amitzur I was not, I added it because I saw your comment in one of the storybook cards

SophieDePaula: @amitzur I updated this condition to check for any type of Error.

js/packages/eyes-storybook/src/browser/getClientAPI.js
SophieDePaula: done.

SophieDePaula:
@amitzur I addressed all your comments and added listeners to story error events.
Also, I'll update the  `latest`  `package.json` to be compatible with storybook 7 in a different PR when storybook 7 is released and document the process.
Can you please give your final comments / approval so I can merge this?

js/packages/eyes-storybook/src/browser/getClientAPI.js
amitzur: oh, you did it in `dist/getClientApi.js`

amitzur (changes requested):
- See inline comments.
- The `dist` scripts have changes when running `yarn build`.
- I had to switch from Node.js 18 to 16 in order to make tests pass. I'm not sure what's the problem but seems like compatibility issue between Storybook and Node.js 18.
- integration tests fail locally for me because of terminal colors. I'm not sure if there is a need to manually set STDOUT_LEVEL. I'm fine if they pass in CI.
- Worth running the `Test packages` workflow on this branch.

js/packages/eyes-storybook/src/renderStories.js
```diff
@@ -125,7 +125,7 @@ function makeRenderStories({
     function didTestPass({resultsOrErr}) {
       return (
         resultsOrErr.constructor &&
-        resultsOrErr.constructor.name !== 'Error' &&
+        !resultsOrErr.constructor.name.toLowerCase().includes('error') &&
```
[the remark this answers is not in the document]
SophieDePaula: @amitzur I now check if `resultsOrErr` is an array instead

js/packages/eyes-storybook/test/fixtures/storybook-with-errors/stories/LoginForm.stories.js
```diff
@@ -0,0 +1,25 @@
...
+
+export default exports
+// const Template = () => <LoginForm />;
+export const Form = () => <LoginForm />;
+
+Form.play = async ({ canvasElement }) => {
+  // Starts querying the component from its root element
+  const canvas = within(canvasElement);
+
+  await userEvent.type(canvas.getByTestId('wrong-selector'), 'bla', {
```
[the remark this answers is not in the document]
SophieDePaula: @amitzur now I test storybook v 6.5 and v7 in this test. The reason is that `playFunctionThrewException` was introduce with v 7 and with 6.5 `storyThrewException` is the event that is triggered when play throws an exception.

SophieDePaula:
> * See inline comments.
> * The `dist` scripts have changes when running `yarn build`.
> * I had to switch from Node.js 18 to 16 in order to make tests pass. I'm not sure what's the problem but seems like compatibility issue between Storybook and Node.js 18.
> * integration tests fail locally for me because of terminal colors. I'm not sure if there is a need to manually set STDOUT_LEVEL. I'm fine if they pass in CI.
> * Worth running the `Test packages` workflow on this branch.

@amitzur here is response
 - dist folder - I applied my changes to the file in browser dir instead of dist.
 - currenly I see that we are testing against node 14 and 16 so we'll have to check why node 18 is failing on a separate PR.
 - I am aware that locally some integration tests fail due to diffrences in the terminal, I was not able to solve this with `STDOUT_LEVEL`
 - Currenly there is an issue with the test workflow, we get `timeout` when trying to start a local fake eyes server (it does work with publish), it require further investigation.


Can you please lmk if you have any additional comments or if I can merge this?

madiyetov:
Hi everyone!
 
I am having trouble with running screenshot tests using Storybook v7 with `storyStoreV7` turned on. 
No issues section in this repo, so I decided to write it down here:
So the problem is that when I run the visual tests for the first time, it gets stuck with `Reading stories` message endlessly. I checked out the code, it happens in [the navigation step](https://github.com/applitools/eyes.sdk.javascript1/blob/master/js/packages/eyes-storybook/src/eyesStorybook.js#L254). I've run the tests with `puppeteerOptions: {headless: false, devtools: true}` and got this:

image

I guess the `eyes-storybook` starts loading the page before the `Vite` even built the Storybook and gets stuck.
Running the same tests for the second time works as usual. Because it will already have `node_modules/cache/.vite-storybook` cached. 

My tech stack is `Vue 2.7.14` + `Vite 4.2.1` and `@applitools/eyes-storybook:v3.32.0`.

I'd appreciate any ideas and suggestions. Thanks in advance!

SophieDePaula:
@madiyetov it you have a pre build step to build your storybook, then you should built it separately and then start the test with `storybookUrl` arg.
You can try to do something like the below script using `http-server` and `start-server-and-test` packages:

"scripts": {
  "visual-test-storybook": "npm run build-storybook && npx start-server-and-test \"http-server --silent -c-1 storybook-static\" 8080 \"eyes-storybook --storybookUrl http://localhost:8080\""
}

madiyetov:
> if you have a pre build step to build your storybook, then you should built it separately and then start the test with `storybookUrl` arg.
> You can try to do something like the below script using `http-server` and `start-server-and-test` packages:

it helped, thanks @SophieDePaula!

PR #1292, opened 2023-02-13
Core universal

kyryloonufriiev:
In this PR I am merging universal (v4) into the core, making it one package

Some additional breaking changes in the universal protocol were made:
1. `Core.makeCore` changes the shape of options
2. `Server.log` event was renamed to `Logger.log` event
3. The CLI command changed a bit it is now `<binary path> universal [args]`, instead of `<binary path> [args]`

All of the relevant universal documentation can be found in `js/packages/core/docs/universal.md`

TODO:
- [x] Consolidate core and universal
- [x] Move all of the relevant universal tests
- [x] Create core changelog instead of the universal one
- [x] Update publish action for core to do what ever universal was doing


kyryloonufriiev:
This will make https://github.com/applitools/eyes.sdk.javascript1/pull/1139 obsolete.

js/.husky/pre-commit
```diff
@@ -2,4 +2,4 @@
 . "$(dirname "$0")/_/husky.sh"
 
 cd ./js
-yarn lint:staged
\ No newline at end of file
+NODE_OPTIONS="--max-old-space-size=4096" yarn lint:staged
```
tourdedave, line 5: How did you choose this value? It looks like the hosted runners in GH actions can go higher if we need. Or is this a safe number in your mind? I'm not familiar with the typical memory footprint when running everything all in under a test load.

js/packages/core/docs/universal.md
```diff
@@ -1,20 +1,66 @@
...
+```ts
+interface LogPayload {
+  level: 'info' | 'warn' | 'error' | 'fatal',
+  message: string
 }
 ```
 
+## Commands
+
+Some commands payload has `settings` and `config` properties, this is done to simplify work on the clients, while user-facing api has a state an two levels of configuring a command (a.k.a. command settings and configuration), underlining logic of the core doesn't have a huge state with configuration and accepts only `settings` of each command at the moment it called. To avoid making clients to construct `settings` object out of command settings and global configuration we allow it to provide both and the core will do the merge.
```
tourdedave, line 62: A few tweaks to the copy. What do you think?

```
Some command payloads have `settings` and `config` properties. This is done to simplify work on the clients.

The user-facing api has state and two levels of configuring a command (a.k.a. command settings and configuration). But the underlying logic of core doesn't have a large state with configuration (like it used to). It accepts only `settings` of each command at the moment it is called. To avoid making clients do more work (e.g., constructing `settings` objects out of command settings and global configuration) we instead allow them to provide both and the core will handle the merge automatically.
```

js/packages/socket/test/socket.spec.ts
```diff
@@ -0,0 +1,40 @@
+import assert from 'assert'
+import {EventEmitter} from 'events'
+import {makeSocket} from '../src/socket'
+import {makeTransport} from '../src/transports/ee'
+
+describe('socket', () => {
```
tourdedave, line 6: Wouldn't it be helpful to have a test for a non-error command response too?

tourdedave (approved):
Looks good to me. Just a few questions and a suggestion on a copy change in the readme. 

This PR also takes us one step closer to using the spec-driver-webdriver as a central package. Rather than having a one-off spec-driver living inside the universal server code base. So I'm a fan.

js/.husky/pre-commit
kyryloonufriiev: I actually had a memory problem locally. I don't say it is the right solution it is just the one that allowed me to continue working. I think there is some misconfiguration, that makes eslint to consume the tone of memory.

js/packages/socket/test/socket.spec.ts
kyryloonufriiev: Sure. It is just the tests that we had for the universal's socket. And since it is a package on its own I just moved the test to a proper location, with some changes to not depend on websocket, since this package could be used with any transport you want.
I will add some happy path tests as well.

js/packages/core/docs/universal.md
kyryloonufriiev: Makes sense. Thank you!

kyryloonufriiev:
@tourdedave I love that you mentioned the `spec-driver-webdriver` thing. Unfortunately, it will not be possible to completely avoid per-SDK implementation, however, you can check out my best take on utilizing `spec-driver-webdriver` in `eyes-nightwatch`.

SophieDePaula (approved):
Looks good to me

js/.husky/pre-commit
kyryloonufriiev: @tourdedave Here is the actual way to solve the problem I see - https://github.com/applitools/eyes.sdk.javascript1/pull/1292/commits/34b0f091e2db00f96c47c9f6b4be3844612ba97f

tourdedave: Ah, that makes sense.

zamboney:
Looks good to me

PR #1295, opened 2023-02-15
[driver] Optimize driver inititalization 

kyryloonufriiev:
I removed helper initialization from the `Driver#init`, instead I introduced `Driver#getHelper` method to initialize the helper once only when needed.


tourdedave (approved):
LGTM

PR #1302, opened 2023-02-20
[core-base] add density metrics

zamboney:
in order to support `PPI` this pr add to the `api/session/running` the `densityMetrics` property

[trello](https://trello.com/c/6HYzeSfz/1707-dfin-poc-regular-text-classified-as-large-text-for-pdfs)

- [x] tests
- [x] documentation


js/packages/core-base/test/e2e/images.spec.ts
```diff
@@ -89,4 +90,65 @@ describe('images', () => {
...
+
+  describe('works with densityMetrics', () => {
+    let eyes: Eyes
+    beforeEach(async function () {
+      eyes = await core.openEyes({
+        settings: {
+          serverUrl: 'https://eyesapi.applitools.com',
+          apiKey: process.env.APPLITOOLS_API_KEY as string,
+          appName: 'Test App',
+          // @ts-ignore
```
zamboney, line 102: This is so i can use the `core.openEyes` and add a `testName` if you think it's not necessary i will remove it

js/packages/core-base/test/e2e/images.spec.ts
```diff
@@ -89,4 +90,65 @@ describe('images', () => {
     const [result] = await eyes.getResults()
     assert.strictEqual(result.status, 'Passed')
   })
+
+  describe('works with densityMetrics', () => {
```
zamboney, line 94: i have to do 2 `openEyes` because the `accessibilityStatus` is per test and not per `check`

js/packages/core-base/test/unit/requests.spec.ts
```diff
@@ -97,6 +97,11 @@ describe('requests', () => {
         ignoreBaseline: false,
         saveDiffs: false,
         abortIdleTestTimeout: 60_000,
+        densityMetrics: {
```
zamboney, line 100: not sure if this is mandatory but a added it any way

zamboney (commented):
@kyrylodolynskyi, i added some comment on the PR please take a look before read 10x

js/packages/core-base/test/e2e/images.spec.ts
```diff
@@ -89,4 +90,65 @@ describe('images', () => {
...
+          testName: this.currentTest.title,
+          environment: {
+            os: 'Platform',
+            hostingApp: 'TestBrowser',
+            deviceName: 'Machine',
+            viewportSize: {width: 2125, height: 2750},
+          },
+        },
+      })
+    })
```
kyrylodolynskyi, line 112: I doubt there is a reason for the test without property at all, but even if there is. I don't think this is a reasonable way to write a test. The test should be self-contained, so we can easily understand what it does and maintain it (means if you need to open eyes do it in `it`).

js/packages/core-base/test/e2e/images.spec.ts
```diff
@@ -89,4 +90,65 @@ describe('images', () => {
...
+    it('with densityMetrics accessibilityStatus', async () => {
```
kyrylodolynskyi, line 113: The name is very artificial, please write a more readable name for the test that describes what you are doing or what should happen.

kyrylodolynskyi (changes requested):

js/packages/core-base/test/e2e/images.spec.ts
```diff
@@ -89,4 +90,65 @@ describe('images', () => {
...
+            version: 'WCAG_2_0',
+          },
+        },
+      })
+
+      await eyes.close()
+      const [result] = await eyes.getResults()
+      assert.strictEqual(result.accessibilityStatus?.status, 'Passed')
+    })
+  })
```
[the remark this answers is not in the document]
zamboney: This test ensures that once there is no `densityMetrics`, the `result.accessibilityStatus.status` `'Passed"` because. it's making sure that the reason for the change is due to the removal of the `densityMetrics`. i get what you are saying that it does not make sense at first glance, so i will add a very low `densityMetrics` to make the `result.accessibilityStatus.status` `'Passed"`  so it will be clear for the future reader this test

pending-changes.yaml
```diff
@@ -10,7 +10,7 @@ Apollo SDK:
       An example of how to inline `markdown code highlights`
 "@applitools/core":
   feature:
-    - 
+    - Passing density metric for PPI support in the sdk
```
zamboney, line 13: @benbabayoff8 ☝️

js/packages/core-base/test/e2e/images.spec.ts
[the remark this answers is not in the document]
kyrylodolynskyi: I don't think this test is needed. If something should change status due to the fact that we successfully sent `dencityMetrics` it is already enough to prevent future regressions. The test that demonstrates that something works differently without it is pointless. I mean if you want this demonstration it should be a part of one test, not two different tests.

kyrylodolynskyi:
@zamboney Are you going to add user-facing for this feature? I think we definitely should have it.

zamboney:
@kyrylodolynskyi what do you mean by "user-facing"?

kyrylodolynskyi:
@zamboney I mean something in `eyes-api` package. You can see in the trello card that we decided to add `.ppi(...)` method to `CheckSettings`

js/packages/core-base/test/e2e/images.spec.ts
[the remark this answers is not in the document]
zamboney: i changed the test so that it's now it is setting the pixel ratio of the screen to `10x10` so all of the texts are now large, and the accessibility status is `'Passed'`, i still think we should keep it because if this test failed, and the other test is pass, for example, both of the `accessibiltyStatus` is `"Fail"`, it's giving as a clue that the PPI isn't set right (could be our end or the backend).
If, after that explanation, you still think that this test is pointless, i will remove it

zamboney:
@kyrylodolynskyi i added user interface

js/packages/eyes-api/src/input/CheckSettings.ts
```diff
@@ -594,6 +596,11 @@ export class CheckSettingsAutomationFluent<TSpec extends SpecType = SpecType> ex
     return this
   }
 
+  densityMetrics(options: DensityMetrics): this {
+    this._settings.densityMetrics = options
+    return this
+  }
```
kyrylodolynskyi, line 602: It shouldn't go to the `CheckSettingAutomation`, but to the `CheckSettingsImage` a.k.a. `CheckSettingsBase`. Also The api we decided on is `.ppi(...)`, not `.densityMetrics(...)`

js/packages/eyes-api/tsconfig.json
```diff
@@ -1,4 +1,4 @@
 {
   "extends": "../../tsconfig.base.json",
-  "ts-node": {"transpileOnly": true},
+  "ts-node": {"transpileOnly": false}
```
kyrylodolynskyi, line 3: why?

js/packages/eyes-cypress/README.md
```diff
@@ -767,6 +768,22 @@ cy.eyesCheckWindow({lazyLoad: {
...
+- `scaleRatio` - The scale ratio.
+
+```js
+// set density metrics
+cy.eyesCheckWindow(Target.window().densityMetrics({
+  xdpi: 100,
+  ydpi: 100,
+  scaleRatio: 1
+}))
+```
```
kyrylodolynskyi, line 786: Cypress doesn't have this api. In cypress, users do not use `Target.window()`, they have an even worse thing, the custom object of a shape that only cypress supports.

js/packages/eyes-nightwatch/README.md
```diff
@@ -482,6 +483,23 @@ Other details:
...
 
+##### Density metrics (`densityMetrics`)
+
+In order to set the density metrics for the screenshot, use the `densityMetrics` method. This method accepts a object value with the following properties:
+
+- `xdpi` - The exact physical pixels per inch of the screen in the X dimension.
+- `ydpi` - The exact physical pixels per inch of the screen in the Y dimension.
+- `scaleRatio` - The scale ratio.
+
+```js
```
kyrylodolynskyi, line 494: I beliave this don't even worth maintaining since it is going to be removed soon

js/packages/core-base/test/e2e/images.spec.ts
```diff
@@ -89,4 +89,80 @@ describe('images', () => {
...
+        },
+      },
+    })
+
+    await eyes.close()
+    const [result] = await eyes.getResults()
+    result.accessibilityStatus
+    assert.strictEqual(result.accessibilityStatus?.status, 'Failed')
+  })
+  it('should passing the accessibility when the density metrics xdpi and ydpi set to 10 and not the same as the viewportSize width and height', async () => {
```
kyrylodolynskyi, line 131: I still don't understand this test. Wouldn't it work without `densityMetrics`?

js/packages/core-base/test/e2e/images.spec.ts
```diff
@@ -89,4 +89,80 @@ describe('images', () => {
...
+        accessibilitySettings: {
+          level: 'AA',
+          version: 'WCAG_2_0',
+        },
+      },
+    })
+
+    await eyes.close()
+    const [result] = await eyes.getResults()
+    result.accessibilityStatus
```
kyrylodolynskyi, line 128:
```suggestion
```

kyrylodolynskyi (changes requested):

js/packages/core-base/test/e2e/images.spec.ts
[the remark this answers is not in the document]
kyrylodolynskyi: I understand what you mean. The first test could fail for a different reason than the presence of `densityMetrics`, so in this case, the second test would fail as well, so it should bring our attention. I think it is clear that this is not the ideal way, no one would ever understand (without) additional information that we have to look at in those two tests together. This is why I think we should have only one test since this is one case that should be checked and demonstrated together.
The ideal test of course would be just an image that could pass only with `densityMetrics` then you don't need these two steps. If it is possible to get/create such an image this will be the preferable way to proceed, if not then just merging two tests in one will work for me.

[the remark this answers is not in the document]
zamboney: yes of course this is what I want to have an image that could only "Pass" only with `dinsityMetrics` but the issue that this PR is fixing is actually the problem that all of the images "Pass" the `accessibiltyStatus` without `dinsityMetrics`.

good point about merging these two tests, that's what i will do 👍

js/packages/eyes-api/src/input/CheckSettings.ts
zamboney: > It shouldn't go to the `CheckSettingAutomation`, but to the `CheckSettingsImage` a.k.a. `CheckSettingsBase`. 
np i will fix the

> Also The api we decided on is `.ppi(...)`, not `.densityMetrics(...)`

`densityMetrics` is more general than `PPI`, and the unit that the server is currently using is `DPI` ([read more](https://99designs.com/blog/tips/ppi-vs-dpi-whats-the-difference/#:~:text=What%20is%20the%20difference%20between%20PPI%20and%20DPI%3F&text=PPI%20describes%20the%20resolution%20in,the%20quality%20of%20the%20output.)) so I thought that making it more general is a good practice and not misleading. Furthermore i API is not in use so I think the naming can be changed

js/packages/eyes-api/tsconfig.json
zamboney: if not then the test is passing, although the `tsc` `typecheck` is failing [here](https://github.com/applitools/eyes.sdk.javascript1/blob/universal-densityMetrics/js/packages/eyes-api/test/unit/CheckSettings.spec.ts) so i thought it a good idea to do a `typecheck` as well, if you think it's unnecessary i will remove it

kyrylodolynskyi: I know that the test may have some bad typescript, but the reason I added it to all of the ts configs is that with the test we often do some experiments and at this point, it is just annoying to handle the types. Of course, I try to have a valid typescript at the end, but not being able to run the test due to types is not what I would want.

js/packages/eyes-api/src/input/CheckSettings.ts
kyrylodolynskyi: Ok, I see your point. But we don't want to publish an api and then change it in the next iteration. Let's talk about it on today's sync

kyrylodolynskyi (changes requested):
Other than one small comment about the casing, it looks good to me. Let's wait for the sync to finally decide on the user-facing api.

PR #1304, opened 2023-02-23
[ufg-client] Properly resolve urls specified in custom properties

kyrylodolynskyi:
As I recently discovered, relative urls specified in custom properties are related to the document location, not css file location as all other urls.
This means that if you have file `/styles/main.css` with content
```css
:root { --image-url: url('./duck.jpeg'); }
body { background-image: url('./duck.jpeg'); }
```
and your document is simple `/index.html` this means that the body background and `--image-url` prop is actually two different ducks, one background one is `/styles/duck.jpeg`, and the custom prop one is `/duck.jpeg`.

In order to fix the issue I had to replace the css parser from `postcss-value-parser` to `css-tree`. The problem with the old one was that it parses only values, but we need more context in order to understand if the url is contained in the custom property or not.

[Trello](https://trello.com/c/GQw3Y1Ql)

TODO:

- [x] Fix the url extraction on the SDK sdk
- [ ] Fix the url extraction on the dom-snapshot side


kyrylodolynskyi:
This also should be fixed on the grid side, because it has the same problem as we had, so when we resolve url properly, the grid doesn't know which url it has to grab from the resource mapping.

js/packages/ufg-client/test/unit/extract-css-dependency-urls.spec.ts
```diff
@@ -2,95 +2,44 @@ import {extractCssDependencyUrls} from '../../src/utils/extract-css-dependency-u
...
-    const cssText = `@supports (display: grid) {
-      div {
-        display: grid;
-        background: url('hello.jpg');
-      }
-    }`
-    const resourceUrls = extractCssDependencyUrls(cssText)
-    assert.deepStrictEqual(resourceUrls, ['hello.jpg'])
+  it('extracts url from unfinished "url" function', () => {
+    // NOTE: Yep! Browser supports it
```
tourdedave, line 36: 😱

tourdedave (approved):
LGTM. Nice find and well done!

js/packages/ufg-client/test/unit/extract-css-dependency-urls.spec.ts
```diff
@@ -2,95 +2,44 @@ import {extractCssDependencyUrls} from '../../src/utils/extract-css-dependency-u
...
-    assert.deepStrictEqual(resourceUrls, ['/somePath/some.jpg'])
+    const resourceUrls = extractCssDependencyUrls(cssText, {resourceUrl: 'http://this/style.css'})
+    assert.deepStrictEqual(resourceUrls, ['http://this/somePath/some.jpg'])
   })
 
-  it('supports data url', () => {
-    const cssText = `url(data:image/svg+xml,%3Csvg xmlns=%22http://www.w3.org/2000/svg%22 width=%222%3E %3Ccircle cx=%2210%22 cy=%225.5%22 r=%224.5%22/%3E %3C/g%3E %3C/svg%3E);`
-    const resourceUrls = extractCssDependencyUrls(cssText)
+  it('extracts data url from "url" function', () => {
+    const cssText = `div { background-image: url('data:image/svg+xml,%3Csvg xmlns=%22http://www.w3.org/2000/svg%22 width=%222%3E %3Ccircle cx=%2210%22 cy=%225.5%22 r=%224.5%22/%3E %3C/g%3E %3C/svg%3E'); }`
```
amitzur, line 28: I understand that you added `div { background-image: ...}` probably because of `css-tree`. But you also added `'` around the data URL. I think not having the quotes was part of the test.

amitzur (approved):
So if I understand correct, this PR is good but we still need to:
1. Do the same for dom-snapshot
2. Support it on the grid.

Correct? Let's push this forward then.

btw, how come we don't have unit tests for extract-svg-dependency?

js/packages/ufg-client/test/unit/extract-css-dependency-urls.spec.ts
kyrylodolynskyi: I will remove quotes

amitzur:
@kyrylodolynskyi I agree. So I think it's fine to leave dom-snapshot as it is. What needs to be done on the grid side?

js/packages/ufg-client/test/unit/extract-css-dependency-urls.spec.ts
kyrylodolynskyi: It doesn't work without quotes and this makes sense because it is not a valid use of url function. I believe `css-tree` can figure it out.

PR #1305, opened 2023-02-23
use timeout and concurrency when fetching resource

SophieDePaula:
[Trello card](https://trello.com/c/oLyE17zj/1758-bokksu-cypress-poc-resources-missing-with-504)

In this PR we need to:
1) Add a `timeout` to not hang on fetch requests. 
2) We need to add an API for the user to provide us with the max concurrency to fetch resources

TODO:
- [x] write e2e test in core
- [x] Cypress - map fetchConcurrency to the runner
- [x] write a test in cypress to make sure the value is passed correctly
- [ ] update README/s


js/packages/req/src/req.ts
```diff
@@ -288,6 +288,7 @@ export async function req(input: string | URL | Request, options?: Options): Pro
       }
     },
     signal: controller.signal,
+    timeout: options?.timeout,
```
kyrylodolynskyi, line 291: This is not the timeout you need. Take a better look at the code and where `options.timeout` is used. `req` already supports `timeout` functionality.

js/packages/ufg-client/src/resources/fetch-resource.ts
```diff
@@ -13,13 +13,15 @@ import {
...
 import {createUserAgentHeader} from '../utils/create-user-agent-header'
+const FETCH_TIMEOUT = 20 * 1000
 
 export type FetchResourceSettings = {
   referer?: string
   proxy?: Proxy
   autProxy?: Proxy & {mode?: 'Allow' | 'Block'; domains?: string[]}
   cookies?: Cookie[]
   userAgent?: string
+  timeout?: number
```
kyrylodolynskyi, line 24: I don't think there is a point in making it configurable per request. It should be a global config same as we already have for `streamingTimeout`

kyrylodolynskyi (changes requested):

js/packages/ufg-client/test/it/fetch-resource.spec.ts
```diff
@@ -4,22 +4,49 @@ import {makeTestServer, generateCertificate} from '@applitools/test-server'
...
     const resource = await fetchResource({
       resource: makeResource({url: `https://localhost:${server.port}/page/smurfs.jpg`}),
     })
     assert.strictEqual((resource.hash as any).contentType, 'image/jpeg')
   })
+
+  it('does not hang for unresponsive resource', async () => {
+    server = await makeTestServer({
+      port: 9000,
+      ...authority,
```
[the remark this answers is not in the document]
SophieDePaula: @kyrylodolynskyi when I don't send that, I get an `ssl` error when trying to fetch the resource.

js/packages/ufg-client/test/it/fetch-resource.spec.ts
```diff
@@ -4,22 +4,45 @@ import {makeTestServer, generateCertificate} from '@applitools/test-server'
...
+
+    const fetchResource = makeFetchResource({retryLimit: 1, fetchTimeout: 1000})
+    await assert.rejects(
+      fetchResource({
+        resource: makeResource({url: `https://localhost:${server.port}/page/smurfs.jpg`}),
+      }),
+      err => {
+        assert.ok(err.message.includes('network timeout'))
+        return true
+      },
```
kyrylodolynskyi, line 45: This is also not how you do it. You should simply return a `true` if the error is the one you expected.

js/packages/ufg-client/test/it/fetch-resource.spec.ts
```diff
@@ -4,22 +4,45 @@ import {makeTestServer, generateCertificate} from '@applitools/test-server'
...
+
+  it('does not hang for unresponsive resource', async () => {
+    server = await makeTestServer({
+      ...authority,
+      middlewares: [
+        async (_req: any, _res: any, next: any) => {
+          await new Promise(resolve => setTimeout(resolve, 1200))
+          next()
+        },
+      ],
```
kyrylodolynskyi, line 34: There is already a middleware that may fit your need. Will it work for you?
```suggestion
      middlewares: ['slow'],
```

kyrylodolynskyi (changes requested):

js/packages/req/src/req.ts
```diff
@@ -61,6 +61,7 @@ export interface Options {
   hooks?: Hooks<this> | Hooks<this>[]
   signal?: AbortSignal
   fetch?: Fetch
+  fetchTimeout?: number
```
kyrylodolynskyi, line 64: How is it different from the `timeout` property?

js/packages/ufg-client/test/it/fetch-resource.spec.ts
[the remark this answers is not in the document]
SophieDePaula: Got it, I missed it.

js/packages/req/src/req.ts
SophieDePaula: @kyrylodolynskyi I thought the `timeout` property is used for something different and therefore I created a dedicated property for fetch timeout.

pending-changes.yaml
```diff
@@ -12,22 +12,22 @@ Apollo SDK:
   feature:
     - 
   bug-fix:
-    - 
+    - Add a timeout for fetching resources
```
SophieDePaula, line 15: @benbabayoff8

js/packages/ufg-client/test/it/fetch-resource.spec.ts
```diff
@@ -4,22 +4,36 @@ import {makeTestServer, generateCertificate} from '@applitools/test-server'
 import assert from 'assert'
 
 describe('fetch-resource', () => {
-  let server: any
+  let server: any, authority: any
 
   before(async () => {
-    const authority = await generateCertificate({days: 1})
-    server = await makeTestServer({...authority, port: 12345})
+    authority = await generateCertificate({days: 1})
```
kyrylodolynskyi, line 10: You don't need the global `authority`, since you are going to use it only in one test

kyrylodolynskyi (changes requested):

js/packages/req/src/req.ts
SophieDePaula: @kyrylodolynskyi , If I don't send `fetchTimeout` to `fetch`, I get a generic error `user aborted test` instead on `network timeout`, and I think it's confusing and in such a case it's harder to understand what is going on. 
To me it's much clearer, to have a `fetchTimeout` property and get the relevant error message.

kyrylodolynskyi: @SophieDePaula If your problem is with an error message you can work to change that. I 100% with you on that. But this is not the way to do it. `fetch` api doesn't support `timeout`, and the right way to abort requests on timeout is the way that is already implemented.
I also don't like the fact that we cannot specify a custom error when aborting requests. We can change that, let's sync

js/packages/ufg-client/test/it/fetch-resource.spec.ts
```diff
@@ -6,20 +6,31 @@ import assert from 'assert'
...
+    server = await makeTestServer({
+      middlewares: ['slow'],
+    })
+
+    const fetchResource = makeFetchResource({retryLimit: 1, fetchTimeout: 1000})
+    await assert.rejects(
+      fetchResource({
+        resource: makeResource({url: `http://localhost:${server.port}/page/smurfs.jpg`}),
+      }),
+      err => err.message.includes('The user aborted a request'),
```
kyrylodolynskyi, line 33: You would better check that this is `AbortError`, the message could change.

js/packages/ufg-client/test/unit/fetch-resources.spec.ts
```diff
@@ -86,7 +120,7 @@ describe('fetch-resource', () => {
         .delayBody(200)
         .reply(200, mockMediaResource.value, {'content-type': mockMediaResource.contentType})
 
-      const fetchResource = makeFetchResource({streamingTimeout: 80})
+      const fetchResource = makeFetchResource({streamingTimeout: 80, fetchTimeout: 0})
```
kyrylodolynskyi, line 123: Why do you need it here?

js/packages/ufg-client/test/unit/fetch-resources.spec.ts
```diff
@@ -72,6 +73,39 @@ describe('fetch-resource', () => {
...
+      nock(`http://something${i}`).get('/').reply(200, countServerParallelRequests)
+    }
+
+    const fetchResource = makeFetchResource({fetchConcurrency: 5})
+    const resources = []
+    for (const resource of mockResources) {
+      resources.push(fetchResource({resource}))
+    }
+
+    await Promise.all(resources)
```
kyrylodolynskyi, line 97: Use `map` instead of `for` loop
```suggestion
    await Promise.all(mockResources.map(resource => fetchResource({resource})))
```

js/packages/ufg-client/test/unit/fetch-resources.spec.ts
```diff
@@ -72,6 +73,39 @@ describe('fetch-resource', () => {
...
+  it('fetch with concurrency limitation', async () => {
+    const mockResources = []
+    let count = 0,
+      maxCount = 0
+    for (let i = 0; i < 10; i++) {
+      mockResources.push(
+        makeResource({
+          url: `http://something${i}`,
+          contentType: 'some/content-type',
+          value: Buffer.from('bla'),
```
kyrylodolynskyi, line 85: It should have no value or content type. To create url resource you need just `url`.

js/packages/ufg-client/test/unit/fetch-resources.spec.ts
```diff
@@ -72,6 +73,39 @@ describe('fetch-resource', () => {
...
+
+    assert.strictEqual(maxCount, 5)
+
+    async function countServerParallelRequests() {
+      count += 1
+      maxCount = count > maxCount ? count : maxCount
+      await utils.general.sleep(300)
+      count -= 1
+    }
+  })
```
kyrylodolynskyi, line 107: I see what you are doing. It is fine, but I think we should emulate the behavior we see on the customer side.
If there is more than X request in the queue right now respond immediately with an error status code.

kyrylodolynskyi (changes requested):

js/packages/ufg-client/test/unit/fetch-resources.spec.ts
SophieDePaula: @kyrylodolynskyi That's what I wanted to do in the first place, but with `nock` you can't send a `callback` as a status, that's the reason I took this approach.

js/packages/ufg-client/test/unit/fetch-resources.spec.ts
SophieDePaula: @kyrylodolynskyi we don't , I removed it

js/packages/ufg-client/test/unit/fetch-resources.spec.ts
kyrylodolynskyi: Of course, you can. Here is example from our code.
```
    nock('https://fonts.googleapis.com')
      .get('/some-font')
      .once()
      .reply(function () {
        return [200, 'font', {'Content-Type': `application/${this.req.headers['user-agent']}`}]
      })
```

js/packages/ufg-client/test/unit/fetch-resources.spec.ts
```diff
@@ -72,6 +73,36 @@ describe('fetch-resource', () => {
...
+
+    const fetchResource = makeFetchResource({fetchConcurrency: 5})
+    const resResources = await Promise.all(mockResources.map(resource => fetchResource({resource})))
+
+    assert.strictEqual(
+      resResources.filter(resource => {
+        utils.types.has(resource, `errorStatusCode`)
+      }).length,
+      0,
+    )
```
kyrylodolynskyi, line 96: Using filter + length is not the best way to get a number of elements that match your condition, but in this case, this is also not what you need. You actually want to know if there are any failed resources in the array. So I recommend
```suggestion
    assert.strictEqual(resResources.some(resource => utils.types.has(resource, `errorStatusCode`)), false)
```

js/packages/ufg-client/test/unit/fetch-resources.spec.ts
```diff
@@ -72,6 +73,36 @@ describe('fetch-resource', () => {
...
+    const mockResources = []
+    let count = 0
+    for (let i = 0; i < 10; i++) {
+      mockResources.push(
+        makeResource({
+          url: `http://something${i}`,
+        }),
+      )
+      nock(`http://something${i}`).get('/').once().reply(limitServerParallelRequests)
+    }
```
kyrylodolynskyi, line 86: A am still kinda skeptical about this `for` loop. I understand what you tried to do, but I would like us to avoid such usage in general, maybe it is fine for the tests.
Anyway, I want to recommend this, just to not give chance for this practice.

```suggestion
    const mockResources = Array({length: 10}, (_, index) => makeResource({url: `http://something/${index}`}))
    let count = 0
     nock('http://something').get(/\d+/).times(mockResources.length).reply(limitServerParallelRequests)
```

SophieDePaula: Thank you for taking the time to mention that

SophieDePaula: That actually, doesn't work, let's talk about it.

js/packages/core/src/ufg/get-base-eyes.ts
kyrylodolynskyi, line 1: in this file, you should not care about `fetchConcurrency`. This is conceptually wrong.

js/packages/ufg-client/src/client.ts
```diff
@@ -13,16 +13,18 @@ export const defaultResourceCache = new Map<string, any>()
 export function makeUFGClient({
   config,
   cache = defaultResourceCache,
+  fetchConcurrency,
   logger,
 }: {
   config: UFGRequestsConfig
   cache?: Map<string, any>
+  fetchConcurrency?: number
```
kyrylodolynskyi, line 21: I think it should go in the `config` object.

js/packages/ufg-client/src/resources/fetch-resource.ts
```diff
@@ -30,12 +31,16 @@ export type FetchResource = (options: {
...
+  fetchTimeout = 30 * 1000,
+  fetchConcurrency,
   cache = new Map(),
   fetch,
   logger,
 }: {
   retryLimit?: number
   streamingTimeout?: number
+  fetchConcurrency?: number
+  fetchTimeout?: number
```
kyrylodolynskyi, line 43: At this point, you better call it `concurrency` and `timeout`. It will create nice symmetry with other commands (e.g. `render`)

js/packages/core/src/ufg/core.ts
```diff
@@ -21,6 +21,7 @@ type Options<TSpec extends SpecType> = {
   base?: BaseCore
   agentId?: string
   concurrency?: number
+  fetchConcurrency?: number
```
kyrylodolynskyi, line 24: This property should not exist on the core level. Because it could be configured only through the manager.

js/packages/core/src/ufg/get-ufg-client.ts
```diff
@@ -5,16 +5,17 @@ import * as utils from '@applitools/utils'
 type Options = {
   client?: UFGClient
   logger: Logger
+  fetchConcurrency?: number
 }
 
-export function makeGetUFGClient({client, logger: defaultLogger}: Options) {
+export function makeGetUFGClient({client, fetchConcurrency, logger: defaultLogger}: Options) {
```
kyrylodolynskyi, line 11: This property should not be predefined, it should be provided in `getUFGClient`

kyrylodolynskyi (changes requested):
I think an e2e test on the core level would help us to not regress in the future.

js/packages/ufg-client/src/resources/fetch-resource.ts
SophieDePaula: @kyrylodolynskyi I prefer to keep these names for better readability of the code.

js/packages/core/src/ufg/core.ts
SophieDePaula: @kyrylodolynskyi How should I pass it then?

kyrylodolynskyi (changes requested):
You also should add a test that will verify that even two eyes under the same manager will not fetch more than X resources in parallel.

js/packages/core/test/fixtures/browser-fetching-fetchConcurrency/app.ts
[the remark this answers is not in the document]
SophieDePaula: @kyrylodolynskyi notice that I created several endpoints to the same resources so it won't be cached, can I do that with test server?

```
 app.get('/image1', (_req: any, res: any) => {
  res.sendFile(path.join(__dirname, '/smurfs.jpg'));
});
app.get('/image2', (_req: any, res: any) => {
  res.sendFile(path.join(__dirname, '/smurfs.jpg'));
});
app.get('/image3', (_req: any, res: any) => {
  res.sendFile(path.join(__dirname, '/smurfs.jpg'));
});
app.get('/image4', (_req: any, res: any) => {
  res.sendFile(path.join(__dirname, '/smurfs.jpg'));
});
   ```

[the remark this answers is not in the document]
Comment (kyrylodolynskyi) [reply js/packages/core/test/fixtures/browser-fetching-fetchConcurrency/app.ts:1]: You just create a folder for your page where you put as much resources as you need (you can copy the same file N times as well)

File: js/packages/core/test/e2e/ufg/Resource-fetching-concurrency.spec.ts
Comment (kyrylodolynskyi) [inline js/packages/core/test/e2e/ufg/Resource-fetching-concurrency.spec.ts:1]: The file name shouldn't be uppercased

File: js/packages/core/src/ufg/get-ufg-client.ts [hunk 20]
```diff
@@ -5,16 +5,19 @@ import * as utils from '@applitools/utils'
...
+  // we are caching by the server config, therefor if the user creates another Runner / manager with the same server config but different
+  // fetchConcurrency, it will not take any affect.
   const getUFGClientWithCache = utils.general.cachify(getUFGClient, ([options]) =>
     client ? 'default' : [options.config],
   )
```
kyrylodolynskyi, line 16: Hmmm. I think you should exclude fetchConcurrency from the cache key.

js/packages/core/test/e2e/ufg/Resource-fetching-concurrency.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+    })
+    await eyes.check({settings: {renderers: [{name: 'chrome', width: 800, height: 600}], disableBrowserFetching: true}})
+    await eyes.close({settings: {updateBaselineIfNew: false}})
+    const [result] = await eyes.getResults()
+    assert.strictEqual(result.isDifferent, false)
+  })
+
+  it('should limit a number of resources fetched in parallel with two eyes instances ', async () => {
+    const driver = await makeDriver({driver: page, spec})
+    await page.goto(`${baseUrl}/browser-fetching-fetchConcurrency/index.html`)
```
kyrylodolynskyi, line 49: please call the page properly. It has nothing to do with "browser fetching", also you should not use kebab and camel case together.
But even better would be to use the page that we already have, as far as I understand the only requirement for the page is to have multiple resources, we already have a page called `page` that has a bunch of different resources. I think it should work just fine for you.

kyrylodolynskyi (changes requested):

js/packages/core/test/e2e/ufg/Resource-fetching-concurrency.spec.ts
SophieDePaula: @kyrylodolynskyi the page points to a the resources that are hosted under the `app` I wrote so that every time a resource is fetched, we count the request and can respond with the right status code

SophieDePaula: I also changed the name as you requested

js/packages/core/src/ufg/get-ufg-client.ts
kyrylodolynskyi: It should. The fact that it uses the old type is another mistake that should be fixed.

js/packages/core/test/e2e/ufg/Resource-fetching-concurrency.spec.ts
SophieDePaula: I changed it

js/packages/core/src/ufg/get-ufg-client.ts
kyrylodolynskyi: It gets `UFGRequestsConfig` because at the time it was created `UFGClientConfig` type didn't exist. And since you introduced a better type, it is a mistake that it isn't used here as well.

PR #1310, opened 2023-03-02
[eyes-cypress] upgrade to `core@v2`

zamboney:


js/packages/eyes-cypress/.eslintrc.json
```diff
@@ -1,4 +1,5 @@
 {
+  "root": true,
```
kyrylodolynskyi, line 2: 👍🏼

js/packages/eyes-cypress/src/browser/commands.js
```diff
@@ -35,7 +35,8 @@ Cypress.Commands.add('eyesGetAllTestResults', () => {
       return
     }
     await Promise.all(closePromiseArr)
-    const summary = await socket.request('EyesManager.closeManager', {manager, throwErr})
+    await socket.request('Eyes.close', {eyes, throwErr: false})
+    const summary = await socket.request('EyesManager.getResults', {manager, settings: {throwErr}})
```
kyrylodolynskyi, line 39: Why did `Eyes.close` pop up here? Especially in the wrong configuration. It doesn't accept `throwErr`.

js/packages/eyes-cypress/src/browser/commands.js
```diff
@@ -110,17 +108,17 @@ Cypress.Commands.add('eyesOpen', function (args = {}) {
     if (!connectedToUniversal) {
       socket.connect(`wss://localhost:${Cypress.config('eyesPort')}/eyes`)
       connectedToUniversal = true
-      socket.emit('Core.makeSDK', {
-        name: 'eyes.cypress',
-        version: require('../../package.json').version,
-        commands: Object.keys(spec).concat(['isSelector', 'isDriver', 'isElement']), // TODO fix spec.isSelector and spec.isDriver and spec.isElement in driver utils
+      socket.emit('Core.makeCore', {
+        agentId: 'eyes.cypress/' + require('../../package.json').version,
```
kyrylodolynskyi, line 112: Why not template string?

js/packages/eyes-cypress/src/browser/commands.js
```diff
@@ -136,7 +134,7 @@ Cypress.Commands.add('eyesOpen', function (args = {}) {
         name: 'chrome',
       },
     })
-    eyes = await socket.request('EyesManager.openEyes', {manager, driver, config})
+    eyes = await socket.request('EyesManager.openEyes', {manager, target: driver, config: {open: config}})
```
kyrylodolynskyi, line 137: You should just do `settings`, instead of `config.open`

js/packages/eyes-cypress/src/expose.ts
```diff
@@ -19,7 +19,9 @@ type Selector = {selector: string; type?: 'css' | 'xpath'; nodeType?: 'element'
 type Element = HTMLElement | JQuery<HTMLElement>
 type ElementWithOptions = {element: Element; regionId?: string; padding?: any}
 
-export type CypressCheckSettings = api.CheckSettingsAutomationPlain<Element, Selector> & {
```
kyrylodolynskyi, line 22: Use the type from the SpecType from `@applitools/driver` to define your own variation. It is also provided by the core.

kyrylodolynskyi (commented):
Isn't it require any changes in the shape of the object? I believe it should

js/packages/eyes-cypress/src/expose.ts
```diff
@@ -19,9 +20,7 @@ type Selector = {selector: string; type?: 'css' | 'xpath'; nodeType?: 'element'
 type Element = HTMLElement | JQuery<HTMLElement>
 type ElementWithOptions = {element: Element; regionId?: string; padding?: any}
 
-export type SpecType = {driver: unknown; context: unknown; element: Element; selector: Selector}
-
-export type CypressCheckSettings = api.CheckSettingsAutomationPlain<SpecType> & {
+export type CypressCheckSettings = api.CheckSettingsAutomationPlain<core.SpecType> & {
```
kyrylodolynskyi, line 23: This is not what I meant. It is a generic type, so something like `core.SpecType<unknown, unknown, Element, Selector>`

zamboney:
🟢 [build](https://github.com/applitools/eyes.sdk.javascript1/actions/runs/4367131049)

js/packages/eyes-cypress/src/browser/eyesCheckMapping.js
```diff
@@ -1,9 +1,15 @@
 /* global Node */
-function eyesCheckMapValues({args, refer}) {
+function eyesCheckMapValues({args, refer, appliConfFile}) {
   if (typeof args === `string`) {
     args = {tag: args}
   }
+  let renderers = args.renderers || args.browser || appliConfFile.browser
```
amitzur, line 6: why not put this line right before the `if (renderers) {` in line 30, where you continue to populate it? Now it's broken into 2 places so it's confusing (to me)

js/packages/eyes-cypress/test/unit/browser/eyesCheckMapping.test.js
```diff
@@ -80,7 +84,7 @@ describe('eyes check mapping', () => {
 
     const appliConfFile = {}
 
-    const coreConfig = eyesCheckMapValues({args, appliConfFile})
+    const coreConfig = eyesCheckMapValues({args, appliConfFile, defaultBrowser})
```
amitzur, line 87: but `defaultBrowser` is not a parameter of `eyesCheckMapValues`

js/packages/eyes-cypress/test/unit/browser/eyesOpenMapping.test.js
```diff
@@ -16,27 +16,23 @@ describe('eyes open mapping', () => {
       useDom: true,
       ignoreCaret: true,
       ignoreDisplacements: true,
-      accessibilityValidation: {level: 'AAA', guidelinesVersion: 'WCAG_2_0'},
+      accessibilitySettings: {level: 'AAA', version: 'WCAG_2_0'},
```
amitzur, line 19: I don't think this is a good change. You are changing the input of the test! It's now not testing the `accessibilityValidation` input anymore. But the user can still specify it.

js/packages/eyes-cypress/test/unit/browser/eyesOpenMapping.test.js
```diff
@@ -16,27 +16,23 @@ describe('eyes open mapping', () => {
...
       testName,
       defaultMatchSettings: {
-        accessibilitySettings: {level: 'AAA', guidelinesVersion: 'WCAG_2_0'},
         enablePatterns: true,
         ignoreCaret: true,
         ignoreDisplacements: true,
         matchLevel: 'Layout',
         useDom: true,
       },
+      accessibilitySettings: {level: 'AAA', version: 'WCAG_2_0'},
```
amitzur, line 35: core doesn't expect it under `defaultMatchSettings` anymore? If so, what about all the rest - `enablePatterns` etc. Should they stay inside `defaultMatchSettings`?

amitzur (commented):
See inline comments.
Also, I see you work hard to merge arguments from check, open, and config file.
But core's API lets you send in `eyes.check` both the check settings AND the config, so that you don't need to work this hard. So why?

amitzur:
@zamboney I took a closer look at the PR and tried to take a shot at taking it in the direction I was thinking. And instead of improving it, I felt that I was just repeating what you did.
I do think that there is a different - clearer - way to do this, especially around the code in commands.js. But I think there are so many things there that are written backwards, that it requires its proper refactor, so not as part of this PR.

Bottom line, I think we should move forward with this PR. The changes are approved (I pushed a little more code, take a look).
The only thing that remains is to have all the tests passing. Here's are the current failures as I see it:
1. `test:e2e` is failing on Node.js 18 because Cypress 6.5.0 doesn't seem to be supported ([link](https://github.com/applitools/eyes.sdk.javascript1/actions/runs/4457162685/jobs/7828076846))
2. `test:e2e` is also failing on Node.js 16 - only the global hooks tests ([link](https://github.com/applitools/eyes.sdk.javascript1/actions/runs/4457162685/jobs/7828076846))
3. Still waiting for `test:coverage` to finish (seems to be hanging on Node.js 18 - [link](https://github.com/applitools/eyes.sdk.javascript1/actions/runs/4459307157/jobs/7831685258) and now running Node.js 16 - [link](https://github.com/applitools/eyes.sdk.javascript1/actions/runs/4459382403/jobs/7831805568))

---
For reference:
The thing I wanted to improve the payload we pass to core. The API is designed such that `openEyes` and `check` both accept `settings` which are relevant to them, and also the global `config` object, which can contain all the configuration, divided into specific areas.
How the Cypress client (and any other client e.g. Java Selenium) should work with this?
There should be a global config object on the client side, and then the arguments that are sent to core are both the arguments that the user passed to `cy.eyesOpen` and `cy.eyesCheckWindow` (these are the `settings`), AND that configuration (this is `config`).
The mapping functions that we should have on the client side should be mapping from user input (and applitools.config.js file) onto those objects.
I think this should be done as part of a refactor where we will split `commands.js` into several parts (since it's very hard to read and reason about right now), and to add TypeScript where we will use Core types, which will make it easier to understand the mappings we perform.

PR #1315, opened 2023-03-06
[core] fix univesal cli for js user

zamboney:
For a javascript library that use `universal` as a dependency (`eyes-cypress` for example), there is the `makeCoreServerProcess` API This function allow to start a server process that will be used by the library to communicate with the `universal` server by wrap the `universal` cli and using `child_process.fork` to start the `universal` server.

This pr change the process arguments so it will point to the right `cli.js` file and use the new `universal` parameter.


tourdedave:
@zamboney can you please add a description to the PR that talks about what problem it solves?

js/packages/core/src/universal/core-server-process.ts
```diff
@@ -6,10 +6,14 @@ export function makeCoreServerProcess(
   options: Options & {forkOptions?: ForkOptions},
 ): Promise<{port: number; close: () => void}> {
   return new Promise((resolve, reject) => {
-    const server = fork(path.resolve(__dirname, '../../dist/cli.js'), [`--config=${JSON.stringify(options)}`], {
-      stdio: [options.shutdownMode === 'stdin' ? 'inherit' : 'ignore', 'ignore', 'ignore', 'ipc'],
-      ...options.forkOptions,
-    })
+    const server = fork(
+      path.resolve(__dirname, '../../dist/cli/cli.js'),
```
kyrylodolynskyi, line 10: 👍🏼

PR #1321, opened 2023-03-07
bongo: consolidate test reporting and release notifications

tourdedave:
### Objective

- to enable all SDK languages to use the same mechanism for sending both test reports and release notifications
- to enable the JS core (non-generic tests) to report and send release notifications in the same manner as the SDKs

### Before

```
coverage-tests generate ./configuration --name 'sdk-name'
// run tests...
coverage-tests report ./configuration --name 'sdk-name' --reportId $GITHUB_SHA
// send notification (custom for each language)
```

### After

```
coverage-tests generate ./configuration --name 'sdk-name'
// run tests...
bongo report --name 'sdk-name' --reportId $GITHUB_SHA
bongo send-release-notification --name 'sdk name' --version 'sdk version'
```

## TODO

- [x] in core: add test result output in junit xml
- [x] in bongo: add mocked test coverage for `report` and `send-release-notification` commands
- [ ] in bongo: test flows e2e after new core route added by Bongo
- [ ] discuss this changeset with the non-JS teams


node_modules/.yarn-integrity
```diff
@@ -0,0 +1,33 @@
...
+    "@applitools/types",
+    "@applitools/ufg-client",
+    "@applitools/utils",
+    "@applitools/visual-grid-client"
+  ],
+  "topLevelPatterns": [],
+  "lockfileEntries": {},
+  "files": [],
+  "artifacts": {}
+}
```
kyrylodolynskyi, line 33: what is that?

js/packages/bongo/src/cli/index.js
```diff
@@ -236,9 +236,49 @@ yargs
   .command(
     ['send-release-notification', 'hello-world'],
```
kyrylodolynskyi, line 237: Why is there an alias `hello-world`?

kyrylodolynskyi (changes requested):
Looks nice and clean to me. There are a few questions in the comments.
There is also a small request, I would love to get rid of `convertSdkNameToReportName`. We should simply specify the names that @Yarden-Ingber wants to receive, or even better @Yarden-Ingber will accept the real SDK names (maybe with language-specific prefixes), instead of made-up keys.

node_modules/.yarn-integrity
tourdedave: 🤔 I don't know.

tourdedave:
> I would love to get rid of `convertSdkNameToReportName`. We should simply specify the names that @Yarden-Ingber wants to receive, or even better @Yarden-Ingber will accept the real SDK names (maybe with language-specific prefixes), instead of made-up keys.

I'm a fan of the first part, not sure about the second. So I propose we break it into two steps.

1. The first thing is to get rid of this function and update the SDKs in our monorepo to specify the name Bongo's service expects (and let the languages not in the mono know of this breaking change). This would remove us from being a gatekeeper in the report/notification process since we won't need to handle the name translation anymore.

2. Talk with Bongo (separate from this PR) to see if he has an appetite to update his service to accept the names of the SDKs as they are.

kyrylodolynskyi:
@tourdedave Makes perfect sense. 👍🏼 BTW we need to talk to Bongo anyway because there should be a custom endpoint to send the release notif for the core.

kyrylodolynskyi:
@tourdedave Also I would expect this PR to update all of the JS packages to send reports using bongo.

tourdedave:
> BTW we need to talk to Bongo anyway because there should be a custom endpoint to send the release notif for the core.

Oh, really? I thought we would just be able to use the same route, change the group, and it would just work ™️. Why do we need a custom endpoint?

> @tourdedave Also I would expect this PR to update all of the JS packages to send reports using bongo.

Yes, of course.

kyrylodolynskyi:
@tourdedave I believe the release notification doesn't have a `group` property. The `group` property exists in the test report and there we don't need a custom endpoint. Basically, the "custom endpoint" thing is what I heard from Bongo during our sync about it. What we actually need is any way to differentiate core notifications from sdks notifications.

tourdedave:
> @tourdedave I believe the release notification doesn't have a `group` property. The `group` property exists in the test report and there we don't need a custom endpoint. Basically, the "custom endpoint" thing is what I heard from Bongo during our sync about it. What we actually need is any way to differentiate core notifications from sdks notifications.

Got it. Thanks

tourdedave:
I talked to @Yarden-Ingber, he will add a new route for us. It will be `/send_mail/core_sdk`

kyrylodolynskyi (approved):
Looks great!

PR #1329, opened 2023-03-13
[driver] Metadata extraction optimization

kyrylodolynskyi:
Here is an incomplete list of properties a driver operates with

- Running protocol
  - W3C
  - Appium
  - JSONWire
- Host environment
  - Device name
  - Platform name
  - Platform version
  - Browser name
  - Browser version
- Execution environment
  - Web
  - Native
- Viewport
  - Size
  - Orientation
  - Pixel ratio
  - Viewport scale
- Web specific
  - Current frame
- Native specific
  - Bar sizes
  - Safe area (for iOS)

Most of them could change in run time by a user out of our control, but we assume that nothing should change out of our control when a user gives us control over a driver. This is where the current optimization was built, using this assumption we extracted all of the metadata about a driver once we were given control over it. With time the amount of required metadata grew, and the process of extracting *all* of the metadata began taking an alarming amount of time, also not all of the metadata is always needed which creates even more confusion. So it is time to think about a more flexible solution where required metadata will be fetched lazily on demand and cached appropriately.


js/packages/core/src/classic/utils/take-dom-capture.ts
```diff
@@ -25,19 +25,22 @@ export async function takeDomCapture<TSpec extends SpecType>({
   settings?: DomCaptureSettings
   logger: Logger
 }) {
-  const isLegacyBrowser = driver.isIE || driver.isEdgeLegacy
+  const environment = await driver.getEnvironment()
+  const features = await driver.getFeatures()
```
zamboney, line 29: Just to make sure, because before this change there was `?`, is `getFeatures` always returns an `object`?

js/packages/core/src/open-eyes.ts
```diff
@@ -78,7 +79,7 @@ export function makeOpenEyes<TSpec extends SpecType, TDefaultType extends 'class
           testConcurrency: concurrency,
           concurrentRendersPerTest: (config as Config<SpecType, 'ufg'>)?.check?.renderers?.length ?? 1,
           node: {version: process.version, platform: process.platform, arch: process.arch},
-          driverUrl: driver?.remoteHostname,
+          driverUrl: driver?.getDriverUrl(),
```
zamboney, line 82: On purpose?

js/packages/core/src/ufg/get-nml-client.ts
```diff
@@ -10,7 +10,7 @@ type Options = {
 
 export function makeGetNMLClient({client, logger: defaultLogger}: Options) {
   const getNMLClientWithCache = utils.general.cachify(getNMLClient, ([options]) =>
-    client ? 'default' : [options.driver.sessionId, options.config],
+    client ? 'default' : [options.driver.guid, options.config],
```
zamboney, line 13: On purpose?

js/packages/driver/src/driver.ts
```diff
@@ -64,282 +74,389 @@ export class Driver<T extends SpecType> {
...
-    // NOTE: this is here because saucelabs does not provide right capabilities for the first call
-    await this._spec.getCapabilities?.(this.target)
-    const capabilities = await this._spec.getCapabilities?.(this.target)
-    this._logger.log('Driver capabilities', capabilities)
+  async refresh(): Promise<this> {
+    this._driverInfo = undefined
+    this._environment = undefined
+    this._viewport = undefined
+    this._features = undefined
+    this._helper = undefined
```
zamboney, line 96: Why not setting default values?

js/packages/driver/src/driver.ts
```diff
@@ -64,282 +74,389 @@ export class Driver<T extends SpecType> {
...
-          this._driverInfo.platformVersion = browserInfo.platformVersion ?? this._driverInfo.platformVersion
+    async function isEqualElements(
+      context: T['context'],
+      element1: T['element'],
+      element2: T['element'],
+    ): Promise<boolean> {
+      return spec.executeScript(context, snippets.isEqualElements, [element1, element2]).catch(() => false)
+    }
+
+    async function findContextReference(context: T['context'], contextInfo: any): Promise<T['element'] | null> {
```
zamboney, line 151: Is `any` legit here?

zamboney (approved):
Looking good, how much time is it saving?

js/packages/core/src/classic/utils/take-dom-capture.ts
kyrylodolynskyi: It is not needed any more

js/packages/core/src/ufg/get-nml-client.ts
kyrylodolynskyi: Yep. The session id to identify the driver wasn't reliable because not all of the drivers have session ids, also I need it here to be sync, so having a guide of driver instance is good enough.

js/packages/driver/src/driver.ts
kyrylodolynskyi: I don't understand. It is `undefined` by default. And I need to have a function to refresh the driver to re-calculate all of the values.

js/packages/driver/src/driver.ts
kyrylodolynskyi: Not really, but it wasn't the focus of this PR, it was there long before.

js/packages/driver/src/driver.ts
zamboney: So why set `undefined`? Do you check the existing of the property somewhere?

kyrylodolynskyi: I am setting undefined because I need it to be `undefined`. For example
1. I created a drive instance and all of its props are `undefined` by default.
2. I do something with the driver and some/all of the properties are populated. Once properties are calculated, they not going to be recalculated.
3. I reach a point when I need to refresh the state of the driver, so I cannot rely on the calculated properties anymore. This is when I need to refresh the properties.

zamboney: Understood

PR #1339, opened 2023-03-16
[core-base, core, ufg-client] Fix the concurrency issue

kyrylodolynskyi:
I fixed the concurrency issue by limiting the concurrency of the renders by a renderer and not globally as it was before. This removes the need to specify concurrency for `UFGClient`, so the number of concurrent renders per renderer could only be tweaked by the env variable now.
Also, I moved the concurrency limitation of the `openEyesBase` to `EyesBase` instead of overriding this method in the `EyesUFG` implementation.


SophieDePaula:
@kyrylodolynskyi can you explain a bit about `rendererUniqueId`?
Is it a unique id per step or test or per test + env?
If it's unique per test+env then I the changes in `renderer` makes sense to me.

SophieDePaula:
@kyrylodolynskyi I linked these packages to `eyes-storybook` and ran a test, I see that `rendererUniqueId` is `undefined`, where do we get this value from? I don't see that it's returned form `bookRenderer`

kyrylodolynskyi:
@SophieDePaula I've added `rendererUniqueId` value it is unique for test + env

PR #1347, opened 2023-03-20
[storybook] Use eyes manager settings

kyrylodolynskyi:
I am applying the new `EyesManagerSettings` that should work with the latest core.
Also removing some redundancy in the `makeCore` and `makeManager`


kyrylodolynskyi:
@SophieDePaula Please review

SophieDePaula (approved):
@kyrylodolynskyi makes sense to me as we create the `manager` here and use it to call `open` therefore we don't need to send `concurrency` to `makeCore`.
Also this matches the changes I see in master re how we send concurrency to `makeManager`.

amitzur:
@SophieDePaula this LGTM but I would like to see that there is some failing test in Eyes-Storybook without this fix.

amitzur:
@SophieDePaula @kyrylodolynskyi so was it the case that without this PR storybook tests were failing?

kyrylodolynskyi:
> @SophieDePaula @kyrylodolynskyi so was it the case that without this PR storybook tests were failing?

The concurrency test was failing, because concurrency wasn't passed.

PR #1348, opened 2023-03-20
Remove duplicate tests

tourdedave:
[Trello](https://trello.com/c/dUGbA7TW/84-support-cypress-retry-mechanism), [GCT](https://github.com/applitools/sdk.coverage.tests/pull/182)
(a.k.a. support test framework retries)

- [x] add functionality in core
- [x] expose functionality through eyes-api
- [x] stabilize new generic coverage tests for at least one SDK (e.g., wdio)
- [x] add support in cypress
- [ ] finalize property that is exposed to the user (re: [this comment](https://github.com/applitools/eyes.sdk.javascript1/pull/1348#issuecomment-1487412024)) (waiting on a response from @benbabayoff8)
- [ ] incorporate final PR feedback


js/packages/core/src/separate-duplicate-results.ts
kyrylodolynskyi, line 1: This file should go to the utils folder, the root level is just for the actual commands that will be user-facing.

js/packages/core/src/separate-duplicate-results.ts
```diff
@@ -0,0 +1,28 @@
...
+  let duplicates: result[] = []
+  batches.forEach(batchId => {
+    const batchResults = results.filter((r: result) => r?.result?.batchId === batchId)
+    const batchBaselines = new Set(results.map((r: result) => r?.result?.baselineId))
+    batchBaselines.forEach(baselineId => {
+      const batchResultsByBaseline = batchResults
+        .filter((r: result) => r?.result?.baselineId === baselineId)
+        .sort(
+          (a: result, b: result) =>
+            new Date(b?.result?.startedAt as Date).valueOf() - new Date(a?.result?.startedAt as Date).valueOf(),
```
kyrylodolynskyi, line 15: instead on `new Date(startedAt).valueOf()` I recommend `Date.parse(startedAt)`

kyrylodolynskyi (commented):
I understand it is a draft. Just wanna leave here a few comments.

js/packages/core/src/separate-duplicate-results.ts
tourdedave: 💯

js/packages/core/src/get-manager-results.ts
```diff
@@ -48,6 +49,32 @@ export function makeGetManagerResults<TSpec extends SpecType, TType extends 'cla
       }),
     )
 
+    let results
+    const {keepDuplicateTests} = storage[0]?.test
+    if ((settings?.removeDuplicateTestsPerBatch, !keepDuplicateTests)) {
+      const [updatedResults, duplicateResults] = separateDuplicateResults(containers.flat())
+      results = updatedResults
+      if (duplicateResults?.length) {
+        const {serverUrl, apiKey, proxy} = storage[0]?.test?.server
```
kyrylodolynskyi, line 58: This couldn't be right... There might be different server urls and api keys.

js/packages/core/src/get-manager-results.ts
```diff
@@ -48,6 +49,32 @@ export function makeGetManagerResults<TSpec extends SpecType, TType extends 'cla
...
+              settings: {
+                testId: dup?.result?.id as string,
+                batchId: dup?.result?.batchId as string,
+                secretToken: dup?.result?.secretToken as string,
+                serverUrl,
+                apiKey,
+                proxy,
+              },
+            })
+          }
```
kyrylodolynskyi, line 71: Why not delete tests in parallel?

js/packages/core-base/src/types.ts
```diff
@@ -281,6 +282,7 @@ export interface AbortSettings extends ReportSettings {
 export interface GetResultsSettings {
   /** @internal */
   userCommandId?: string
+  removeDuplicateTestsPerBatch?: boolean
```
kyrylodolynskyi, line 285: This is the wrong place for this property. `core-base` doesn't know anything about managers, this type is used in `eyes.getResults()` not `manager.getResults()`

js/packages/core/src/make-manager.ts
```diff
@@ -17,6 +17,7 @@ type Options<TSpec extends SpecType> = {
   agentId?: string
   cwd?: string
   logger: Logger
+  removeDuplicateTestsPerBatch?: boolean
```
kyrylodolynskyi, line 20: As we decided. This is unnecessary here.

js/packages/core/src/get-manager-results.ts
```diff
@@ -48,6 +49,32 @@ export function makeGetManagerResults<TSpec extends SpecType, TType extends 'cla
       }),
     )
 
+    let results
+    const {keepDuplicateTests} = storage[0]?.test
```
kyrylodolynskyi, line 53: This property could be different for every test

js/packages/core/src/get-manager-results.ts
```diff
@@ -48,6 +49,32 @@ export function makeGetManagerResults<TSpec extends SpecType, TType extends 'cla
...
+      results = updatedResults
+      if (duplicateResults?.length) {
+        const {serverUrl, apiKey, proxy} = storage[0]?.test?.server
+        if (serverUrl && apiKey) {
+          for (const dup of duplicateResults) {
+            await core.deleteTest({
+              settings: {
+                testId: dup?.result?.id as string,
+                batchId: dup?.result?.batchId as string,
+                secretToken: dup?.result?.secretToken as string,
```
kyrylodolynskyi, line 65: No need for type casting here.

kyrylodolynskyi (changes requested):

js/packages/core/src/get-manager-results.ts
tourdedave: Okay. So, what do you recommend?

js/packages/core-base/src/types.ts
tourdedave: So where does it need to go?

```
src/get-manager-results.ts:54:20 - error TS2339: Property 'removeDuplicateTestsPerBatch' does not exist on type 'GetResultsSettings<TType>'.

54     if ((settings?.removeDuplicateTestsPerBatch, !keepDuplicateTests)) {
                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error in src/get-manager-results.ts:54
```

kyrylodolynskyi: To the core's (not `core-base`) type - `GetResultsSettings`

js/packages/core/src/get-manager-results.ts
tourdedave:
```
src/get-manager-results.ts:63:17 - error TS2322: Type 'string | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.

63                 testId: dup?.result?.id,
                   ~~~~~~

src/get-manager-results.ts:64:17 - error TS2322: Type 'string | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.

64                 batchId: dup?.result?.batchId,
                   ~~~~~~~

src/get-manager-results.ts:65:17 - error TS2322: Type 'string | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.

65                 secretToken: dup?.result?.secretToken,
                   ~~~~~~~~~~~


Found 3 errors in the same file, starting at: src/get-manager-results.ts:63
```

js/packages/core-base/src/types.ts
tourdedave: But the way core's `GetResultsSettings` are written derive their values from what is in `core-base` through a ternary operator. It's unclear to me what action I should take.

```
export type GetResultsSettings<TType extends 'classic' | 'ufg'> = (TType extends 'ufg'
  ? UFGCore.GetResultsSettings
  : ClassicCore.GetResultsSettings) & {throwErr?: boolean}
```

kyrylodolynskyi: You can see that core's `GetResultsSettings` extends ufg's `GetResultsSettings` or classic's `GetResultsSettings` with `throwErr` property, you should simply add also your property allong the declaration of `throwErr`.

js/packages/core/src/get-manager-results.ts
kyrylodolynskyi: I recommend marking the test results with this flag, so you can understand what to do with the result object (remove or not)

js/packages/core/src/get-manager-results.ts
kyrylodolynskyi: Ah, I admit it. The `TestResult` type has a shitty declaration where every property is optional.

js/packages/core-base/src/types.ts
tourdedave: Okay, thanks for the explanation. But just so you know, this declaration is not easy to read. It wasn't clear until you pointed it out that the `throwErr` property was being applied to either of the `GetResultSettings`. At a glance, it looks like the property is only applied to classic's given the way the ternary operator is laid out. Only after reading it a few times did I see the parens enclosing the ternary.

js/packages/core/src/utils/separate-duplicate-results.ts
```diff
@@ -0,0 +1,23 @@
...
+        .sort((a, b) => Date.parse(b?.result?.startedAt as string) - Date.parse(a?.result?.startedAt as string)) // sort descending
+      if (batchResultsByBaseline.length > 1) duplicates = duplicates.concat(batchResultsByBaseline.slice(1)) // leave the 1st item, store the rest
+    })
+  })
+  return [
+    // tests to keep
+    !duplicates.length ? results : results.filter(r => duplicates.some(dup => dup?.result !== r?.result)),
+    // tests to delete
+    duplicates,
+  ]
```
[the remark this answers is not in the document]
tourdedave: My understanding of the requirement is to 1) find uniques so we can delete them from eyes, and 2) remove the uniques from the results returned to the user.

I don't want to mutate date in place, but rather, return a modified copy. And I want to unit test the behavior not just to exercise it but to make it clear what the contract of the function is to anyone who needs to maintain it later. I also don't want to add additional code to an already large function for someone to reason about when it can easily be abstracted.

The argument of utility versus business case doesn't make sense to me because the utility is the business case. And we currently only need to do it in one place. If we needed to reuse this in additional places and customize it based on different contexts, then I would agree with you that what you're recommending would be a better direction.

js/packages/core/src/get-manager-results.ts
tourdedave: Okay. But before I get to that, seeing a situation when running a GCT where core-base completes `openEyes` and sets `keepDuplicateTests` to `true`, but when core/`getManagerResults` is called, `keepDuplicateTests` is not included in any of the tests. Thoughts on what would cause that?

js/packages/core-base/src/types.ts
kyrylodolynskyi: This is prettier formatting. I agree it is not the best way to write this expression, but the parenthesis should give a hint.

js/packages/core/src/get-manager-results.ts
tourdedave: What do you recommend?

tourdedave:
@benbabayoff8 I've added `removeDuplicateTestsPerBatch` as the RunnerOption users can use to enable this behavior. Please let me know if you'd prefer a different name for this.

js/packages/core-base/src/types.ts
```diff
@@ -66,6 +66,7 @@ export interface TestInfo {
   rendererId?: string
   rendererUniqueId?: string
   rendererInfo?: {type?: 'web' | 'native'; renderer?: Record<string, any>}
+  keepDuplicateTests?: boolean
```
kyrylodolynskyi, line 69: This is always defined, so should not be optional.
I think the name is a little bit misleading, it should not be plural, also you keep the result and not the test, so I would say `keepDuplicate(d?)Result`

js/packages/core/src/types.ts
```diff
@@ -116,6 +116,7 @@ export interface Eyes<TSpec extends SpecType, TDefaultType extends 'classic' | '
 }
 
 export interface EyesManagerSettings {
+  removeDuplicateTestsPerBatch?: boolean
```
kyrylodolynskyi, line 119: We decided that we don't need it here.

js/packages/eyes-api/test/unit/Eyes.spec.ts
```diff
@@ -39,7 +49,33 @@ describe('Eyes', () => {
...
+        {
+          command: 'makeManager',
+          data: {
+            type: 'ufg',
+            settings: {removeDuplicateTestsPerBatch: true, concurrency: undefined, legacyConcurrency: undefined},
+          },
+        },
+      ],
     )
   })
```
kyrylodolynskyi, line 80: Actually, we don't need this. We want to pass it only in `manager.getResults`

js/packages/core/test/it/get-manager-results.spec.ts
```diff
@@ -52,4 +52,16 @@ describe('get manager results', async () => {
...
+  it('should support removing duplicates', async () => {
+    const manager = await core.makeManager({
+      settings: {removeDuplicateTestsPerBatch: true},
+    })
+    const eyes = await manager.openEyes({target: driver, settings: {appName: 'App', testName: 'Test'}})
+    await eyes.check({settings: {name: 'blah'}})
+    await eyes.check({settings: {name: 'blah'}})
+    await eyes.close()
+    const summary = await manager.getResults({settings: {throwErr: false}})
+    assert.deepStrictEqual(summary.results.length, 1)
```
kyrylodolynskyi, line 65: What kind of duplicate is that? It is just a two-step test, we need to open other eyes to create a pseudo-duplicate.
Also, we agreed that `removeDuplicateTestsPerBatch` should go in `getResults`.

js/packages/core/test/utils/fake-base-core.ts
```diff
@@ -143,9 +145,13 @@ export function makeFakeCore({
...
-              results.push({
-                status: steps.every(result => result.asExpected) ? ('Passed' as const) : ('Unresolved' as const),
-                stepsInfo: steps,
+              steps.forEach(step => {
+                results.push({
+                  status: step.asExpected ? ('Passed' as const) : ('Unresolved' as const),
+                  baselineId: 'baseline-id',
+                  batchId: 'batch-id',
+                  stepsInfo: [step],
+                })
```
kyrylodolynskyi, line 154: This is wrong. Steps are not test results, they are steps in the test result, so basically, the way it was before.
This explains the integration test that I also commented on.

js/packages/eyes-api/src/Runners.ts
```diff
@@ -70,13 +72,15 @@ export abstract class EyesRunner {
...
   private _legacyConcurrency?: number
+  private _removeDuplicateTestsPerBatch?: boolean
   /** @internal */
   readonly type = 'ufg' as const
   /** @internal */
   get settings() {
     return {
       concurrency: this._testConcurrency,
       legacyConcurrency: this._legacyConcurrency,
+      removeDuplicateTestsPerBatch: this._removeDuplicateTestsPerBatch,
```
kyrylodolynskyi, line 83: I understand that we don't need it in the settings because we decided to put it in `getResults`.

kyrylodolynskyi (changes requested):
I have one more request, can you update `js/core/docs/universal.md` with the new feature you added?

js/packages/eyes-api/src/Runners.ts
benbabayoff8: @kyrylodolynskyi How would you imagine it within the getAllTestResults (user API)?

kyrylodolynskyi: We are not talking about user-facing api. This is an internal api on which we decided since it is the most flexible one.

js/packages/core/test/it/get-manager-results.spec.ts
tourdedave: Nice catch. I never revisited this test once I got the e2e test working. I'll take a look.

js/packages/core/src/types.ts
tourdedave: Put another way - if I remove this declaration, then `eyes-api` will throw a build error

```
src/Runners.ts:53:74 - error TS2339: Property 'removeDuplicateTestsPerBatch' does not exist on type 'EyesManagerSettings'.

53         settings: {throwErr, removeDuplicateTestsPerBatch: this.settings.removeDuplicateTestsPerBatch},
```

js/packages/eyes-api/test/unit/Eyes.spec.ts
tourdedave: I'm not sure I agree. Waiting on feedback to my comment above.

js/packages/core-base/src/types.ts
tourdedave: Thinking about this more, why should it not be plural? There could be more than one duplicate result. And why is it misleading? The usage of "test" as the object of interest is consistent with what we expose at the user level. Which is better in the long term and less confusing - to be precise (e.g., specific to the implementation detail) or accurate (e.g., consistent with the naming convention of the feature that is exposed to the user)?

js/packages/core/src/types.ts
kyrylodolynskyi: `eyes` package will get a build error because it is wrong as well. We decided that `removeDuplicateTestsPerBatch` will be passed in `.getResults()`, this means it should go in `GetResultSettings` (you already added it there). `EyesManagerSettings` applied to `makeManager`, passing new property there was our first idea that we throw away during our discussion.

js/packages/core-base/src/types.ts
kyrylodolynskyi: I hear your reasoning about `test` vs. `result`. But about plural vs. singular, I have to point out that you specify it per test, so there could not be multiple tests within a test. This flag answers the question - "should I remove this test (singular) if it is considered a duplicate?"

js/packages/core/src/types.ts
tourdedave: We clearly have a misunderstanding and need to sync to discuss. Because yesterday we reviewed my changes in eyes-api on our sync call and you said it aligned with what we discussed. So I am confused, and find this back and forth surprising and frustrating.

js/packages/core-base/src/types.ts
tourdedave: Depends on your definition of `test`, because the property that is exposed for the user to opt-in to this behavior is `removeDuplicateTestsPerBatch`. Again - I was going for consistency that would (in my mind, at least) make this code easier to reason about long term.

tourdedave:
Functionally, everything works as needed for both cypress and all core/universal based SDKs. However, based on the last round of feedback from @kyrylodolynskyi there are some changes that need to be addressed. I'm off tomorrow but back on Friday. If this can wait until Friday, then I will finalize the PR and work on releasing by end-of-the-day then. But - if you (@benbabayoff8, @amitzur, @kyrylodolynskyi) would like to err on the side of caution and ship something sooner, then feel free to make the necessary changes to finalize this PR and ship while I'm away.

js/packages/core/src/get-manager-results.ts
```diff
@@ -48,6 +49,38 @@ export function makeGetManagerResults<TSpec extends SpecType, TType extends 'cla
...
+      logger.log('User opted into removing duplicate tests per batch, checking for duplicates...')
+      const [updatedResults, duplicateResults] = separateDuplicateResults(containers.flat())
+      results = updatedResults
+      if (!duplicateResults.length) logger.log('No duplicate tests found.')
+      else {
+        logger.log(`Duplicates found (${duplicateResults.length} in total), cleaning them up...`)
+        await Promise.all(
+          duplicateResults.map(async dup => {
+            // eslint-disable-next-line @typescript-eslint/no-non-null-asserted-optional-chain
+            const {serverUrl, apiKey, proxy} = dup?.result?.server!
```
amitzur, line 63: @kyrylodolynskyi there's a typescript error on this line because `Property 'server' does not exist on type 'TestResult'`. Where should these values be taken from then?

js/packages/core/src/utils/separate-duplicate-results.ts
```diff
@@ -0,0 +1,23 @@
+import {TestResultContainer} from '../types'
+
+export function separateDuplicateResults(containers: TestResultContainer<'classic' | 'ufg'>[]) {
+  const batches = new Set(containers.filter(r => r.result?.batchId).map(r => r.result?.batchId))
+  let duplicates: TestResultContainer<'classic' | 'ufg'>[] = []
+  batches.forEach(batchId => {
+    const batchResults = containers.filter(r => r.result?.batchId === batchId)
+    const batchBaselines = new Set(containers.map(r => r.result?.baselineId))
```
amitzur, line 8: @kyrylodolynskyi there is a TS error here `Property 'baselineId' does not exist on type 'TestResult'.`
From what I see, `baselineId` is not defined on `TestResult` in `core`, but it is defined on `TestResult` in `core-base`. How do you suggest I mitigate this situation?
In general, would we want to define `core`'s `TestResult` as `extends CoreBase.TestResult` ?

js/packages/core/src/get-manager-results.ts
amitzur: Sorry, it was my bad for not linking the packages.

kyrylodolynskyi:
@amitzur @tourdedave I applied some changes to the types in the `core` and implementation in the `eyes`, also finalized e2e test in cypress.

tourdedave:
Thanks @amitzur & @kyrylodolynskyi for taking this PR the rest of the way!

PR #1349, opened 2023-03-20
fix: cypress@8 with core@2 not working

zamboney:
`nodeVersion` property set the way the `node` process will executed [doc link](https://docs.cypress.io/guides/references/configuration#Node-version)
- if set to `system` it will use the `node` process that the OS have
- if set to `bundled` it will use the `node` process that the `Cypress Helper.app` have

If `nodeVersion` not set to `system` then the process not sending the process CLI arguments as expected

Because `core@2` change the CLI from `eyes-universal` to `eyes universal` the universal server will not start if `nodeVersion` exits and not set to `system`, this is why we pass to the `universal` server the `execPath` of `node`

Because the change in `cypress@9.0.0` `nodeVersion` that change premetly to `system` [link](https://github.com/cypress-io/cypress/pull/18732) this effects only occurred in `cypress@8`


js/packages/eyes-cypress/src/plugin/server.ts
```diff
@@ -38,7 +38,21 @@ export default function makeStartServer({logger}: {logger: Logger}) {
...
+    const isCypressVersionBelow7 = semverLt(cypressVersion, '7.0.0')
+
+    // `nodeVersion` property set the way the `node` process will be executed
+    // if set to `system` it will use the `node` process that the OS have
+    // if set to `bundled` it will use the `node` process that the `Cypress Helper.app` have
+    //
+    // [doc link](https://docs.cypress.io/guides/references/configuration#Node-version)
+    //
+    // if `nodeVersion` not set to `system` the the process not sending the process cli arguments as expected
+    // and because of the change in `eyes-universal` to `eyes universal` the universal server will not start
```
kyrylodolynskyi, line 50: I think this comment is misleading. I appreciate the explanation, but when someone will work with that they will not care about how it was when cli was different.

kyrylodolynskyi (approved):
👍🏼

js/packages/eyes-cypress/src/plugin/server.ts
kyrylodolynskyi: I was talking about line 50

zamboney: fixed

PR #1351, opened 2023-03-21
    fix(closeBatch): send the right argument to universal

zamboney:
[Trello](https://trello.com/c/PvlzTh6B)

### Test

Add `typescript` types to `closeBatch`


js/packages/eyes-cypress/src/plugin/hooks.ts
```diff
@@ -39,7 +44,7 @@ export default function makeGlobalRunHooks({closeManager, closeBatches, closeUni
         }
         if (!config.appliConfFile.dontCloseBatches) {
           await closeBatches({
-            batchIds: [config.appliConfFile.batchId || config.appliConfFile.batch.id],
+            batchId: config.appliConfFile.batchId || config.appliConfFile.batch.id,
```
zamboney, line 47: ☝️ this is the fix ☝️ 

all of the other changes is to add types so that next time this error will prompt in `ttsc`

kyrylodolynskyi (approved):

js/packages/eyes-cypress/src/plugin/hooks.ts
```diff
@@ -9,7 +10,7 @@ declare global {
         dontCloseBatches: boolean
         batch: any
         serverUrl: string
-        proxy: string
+        proxy: any
```
amitzur, line 13: please add `// TODO`

amitzur (commented):
We need an e2e test for this functionality. Other than that the changes are good :100:

zamboney:
> We need an e2e test for this functionality. Other than that, the changes are good 💯

@amitzur, I and @kyrylodolynskyi discuss this, and there is no easy way to test it. I will explain myself:

the `isComplete` property that returns from the `batch` API isn't returned right after the request to `closeBatch` what is this mean is we need to have some kind of pulling mechanism that will check for changes on that property. I think it's doesn't make sense

Another option is to create some kind of `unit`/`integration` test but will just check that the input produce output, which means that you would not know if the API changed, so it's not making more sense, in my opinion.

---

i think that the "right" solution is that the BE will provide us a new property that flags that the request to `closeRequest` has been submitted but didn't process.

@amitzur, if you still think that one of the options above is necessary i will add it to another PR k?

amitzur:
@zamboney ah, yes. Thanks for documenting this. I remember now that this is the reason we don't have e2e tests for this.
We can merge this PR, but let's insist on the e2e test. I'm fine with testing only that we send Eyes yhe correct payload, or we can work with Yotam to get a solution from him.

PR #1370, opened 2023-03-23
js bongo: send report payload for js_images with the correct category

tourdedave:


js/packages/bongo/src/report/create.js
```diff
@@ -1,10 +1,21 @@
...
+function findGroupFromName(name) {
+  switch (name) {
+    case 'core':
+      return 'core'
+    case 'js_images':
+      return 'images'
+    default:
+      return 'selenium'
+  }
+}
```
kyrylodolynskyi, line 12: Didn't we aim to remove this kind of thing? I would prefer to simply specify the group the way it should be in bongo's system. We can keep the `selenium` as default if you want. WDYT?

tourdedave: Oh, right! I'll expose `group` as an argument to `report` and set the default to `selenium` unless otherwise specified.

kyrylodolynskyi (approved):

PR #1375, opened 2023-03-26
fix(formatter): fix print result

zamboney:
the `formatter` `core` pass without his function
[trello](https://trello.com/c/XJbKLTz0)


js/packages/eyes-storybook/test/e2e/eyes-storybook.e2e.test.js
```diff
@@ -26,6 +30,7 @@ describe('eyes-storybook', () => {
   after(async () => {
     await closeTestServer();
     process.env.APPLITOOLS_SHOW_LOGS = '';
+    fs.rmSync(tmpdir, {recursive: true, force: true});
```
kyrylodolynskyi, line 33: @zamboney This is not supported by node 12, or don't we test the storybook with node 12?

kyrylodolynskyi:
This is not the right way to fix the issue.
Let me explain. In past, there was a class `TestFormatter`, that you had to initialize with an array of test results, then you could tell it what kind of report you want to generate from it. Since the new core I removed the class and introduced a set of pure formatted functions, you chose the function (format) and provide an array of results, that will give you a formatted string.
When the storybook was migrated to use a new core this nuance was lost. And what you are doing now is trying to make it work the old way instead of upgrading it to the new one.

zamboney:
@kyrylodolynskyi please take another look, Is this what you mean?

js/packages/eyes-storybook/test/it/processResults.test.js
```diff
@@ -492,12 +492,12 @@ describe('processResults', () => {
         },
       ],
     };
-    const {formatter} = processResults({
+    const {summary: formatter} = processResults({
```
kyrylodolynskyi, line 495: why?

kyrylodolynskyi (approved):
👍🏼

js/packages/eyes-storybook/test/it/processResults.test.js
zamboney: i change the output, bi longer returns the `formatter` but the test `summary`, if it's not making sense to you here, i will change the naming

kyrylodolynskyi: It doesn't make sense for me. Why would you rename `summary` to `formatter` here, it is not a formatter.

PR #1377, opened 2023-03-29
fix: disableBrowserFetching change from open to check

zamboney:
[trello](https://trello.com/c/X2TDz0vG)

add new test in `e2e`


SophieDePaula (commented):
Looks good, I'll add this value in `eyesCheckMapping.test.js` tests

PR #1380, opened 2023-03-30
map checkSettings from open and global config

SophieDePaula:
trellos:
https://trello.com/c/I5HQq3xP/1809-cypress-everfi-defining-matchlevel-in-cyeyesopen-and-applitoolsconfigjs-does-not-work
https://trello.com/c/SZTgzwzL/1807-fanduel-eyes-cypress-3300-no-longer-honors-layoutbreakpoints-in-config-file

TODO:
- [x] merge global and open config in openMapping and send that as `config` to `check`
- [x] add an e2e test for checkSettings in open in `cypress-run-checkSettings-globalConfig.test`
- [x] add an e2e test for `visualGridOptions` from config
- [x] add coverage test for `visualGridOptions` from open
- [x] fix unit tests


SophieDePaula:
@zamboney , in this PR I map all relevant `checkSettings` args from `open` and `applitools.config.js` file. So if you have any additional PR we might not need them

js/packages/eyes-storybook/src/checkSettingsParams.js
```diff
@@ -24,6 +24,7 @@ const checkSettingsParams = [
   'selector',
   'fully',
   'fakeIE',
+  'disableBrowserFetching',
```
zamboney, line 27: Is this on purpose?

zamboney (commented):
looking good!

About the tests. Unit tests like this doesn't really tests because there is no guarantee that something will catch in the test if something changes. All this test do is make sure that the input that you enter will produce the output you expected

js/packages/eyes-storybook/src/checkSettingsParams.js
SophieDePaula: @zamboney yes, I forgot to add this parameter to storybook, and working on this PR reminded me of that

pending-changes.yaml
```diff
@@ -64,6 +64,7 @@ Apollo SDK:
     - 
   bug-fix:
     - disableBrowserFetching change from open to check
+    - map values from open and global config
```
SophieDePaula, line 67: @benbabayoff8

kyrylodolynskyi (changes requested):
I walk through the implementation and understand that it is hard to comment on a specific line of code.
Generally speaking the current state of things in cypress is counterintuitive and the implementation is very spaghetti. Don't get me wrong, it doesn't apply to what you did, I am talking about the way it is in cypress. This sdk suffers from the lack of structure and this causes us nothing but problems. I think we have a good opportunity to make things right. I am not talking about changing everything here, just the way we process user input.
So what I propose:
There should be types for transform and merge functions. Here is the functions I propose:
1. `mergeCypressConfigs(config1: CypressConfig, config2: CypressConfig): CypressConfig`, this function we should use to merge config file with open config (both are of type `CypressConfig`)
2. `transformCypressConfig(config: CypressConfig): core.Config`, this function we should use to create a config object which core can process.
3. `transformCypressCheckSettings(settings: CypressCheckSettings): core.CheckSettings`, this function we should use to create a check settings object which core can process.

I believe these three simple functions will simplify the code a lot. I recommend keeping these functions simple without setting any default values or doing some unnecessary code sharing. The logic is fairly trivial, I recommend something like what we have in `eyes` - https://github.com/applitools/eyes.sdk.javascript1/blob/74cf38583d84778dee2b3602edb681003e08d7e7/js/packages/eyes-api/src/input/Configuration.ts#L1077. Just calling each property by name without any loops and arrays of allowed keys, it will give us a super simple impl that is easy to reason about. Where we will see the important differences with "standard" api, so we can push for deprecating what is different.

js/packages/eyes-cypress/src/browser/utils.js
```diff
@@ -0,0 +1,40 @@
...
+      browser.name = 'chrome'
+    }
+    if (browser.deviceName) {
+      browser = {chromeEmulationInfo: browser}
+    }
+    return browser
+  } else {
+    return browser
+  }
+}
```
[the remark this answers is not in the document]
SophieDePaula: @kyrylodolynskyi that's incorrect. If the user sends viewport without `name` the test hangs for me.

js/packages/eyes-cypress/src/browser/utils.js
```diff
@@ -0,0 +1,40 @@
...
+    'disableBrowserFetching',
+  ]
+  const checkSettingsConf = {}
+  for (const [key, value] of Object.entries(config)) {
+    if (checkSettingsParams.includes(key)) {
+      checkSettingsConf[key] = value
+    }
+  }
+  return checkSettingsConf
+}
```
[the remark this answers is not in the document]
SophieDePaula: @kyrylodolynskyi I wanted to send the relevant props to `config.check` and not the entire config

js/packages/eyes-cypress/src/browser/transformCypressConfig.ts
```diff
@@ -0,0 +1,79 @@
+import * as utils from '@applitools/utils'
+
```
SophieDePaula, line 2: @kyrylodolynskyi please take a look a this and lmk what you think.
I still need to create `transformCheckSettings` and update unit tests.

js/packages/eyes-cypress/src/browser/commands.js
```diff
@@ -135,16 +135,9 @@ Cypress.Commands.add('eyesOpen', function (args = {}) {
 
     const appliConfFile = Cypress.config('appliConfFile')
 
-    openToCheckSettingsArgs = eyesOpenToCheckMapValues(args)
+    openAndGlobalConfig = transformCypressConfig({testName, ...appliConfFile, ...args, shouldUseBrowserHooks})
```
kyrylodolynskyi, line 138: What about merging `properties` and other things a little bit smarter than just overriding? You told me that we were doing it before.

js/packages/eyes-cypress/src/browser/commands.js
```diff
@@ -158,15 +151,15 @@ Cypress.Commands.add('eyesCheckWindow', (args = {}) =>
     Cypress.log({name: 'Eyes: check window'})
 
     const settings = eyesCheckMapValues({
```
kyrylodolynskyi, line 153: What the name of the function means?

js/packages/eyes-cypress/test/coverage/generic/cypress/e2e/custom/AdoptedStyleSheetsOnFirefoxOpenConfig.cy.js
```diff
@@ -0,0 +1,21 @@
+/* global cy, expect*/
```
zamboney, line 1: This test isn't needed anymore because firefox support `adoptedStyleSheets

zamboney (approved):
Looking good 😺
@kyrylodolynskyi  10x for the input

js/packages/eyes-cypress/src/browser/transformCypressConfig.ts
```diff
@@ -0,0 +1,79 @@
...
+import type {CypressEyesConfig, CypressBrowser} from '../expose'
+import type {SpecType, Config} from '@applitools/core'
+import type {DesktopBrowserInfo, ChromeEmulationInfo, IOSDeviceInfo, AndroidDeviceInfo} from '@applitools/eyes-api'
+type coreBrowser = DesktopBrowserInfo | ChromeEmulationInfo | IOSDeviceInfo | AndroidDeviceInfo
```
kyrylodolynskyi, line 6: Types should always start with a capital latter. Also this is not the definition of "core browser"

kyrylodolynskyi (changes requested):
I like the `transformCypressConfig` function. Let's discuss everything else on the sync.

js/packages/eyes-cypress/src/browser/mergeCypressConfigs.ts
```diff
@@ -0,0 +1,37 @@
...
+      return openConfig[prop]
+    } else if (openConfig.batch && openConfig.batch.hasOwnProperty(nestedProp)) {
+      return openConfig.batch[nestedProp]
+    } else if (globalConfig.hasOwnProperty(prop)) {
+      return globalConfig[prop]
+    } else if (globalConfig.batch && globalConfig.batch.hasOwnProperty(nestedProp)) {
+      return globalConfig.batch[nestedProp]
+    }
+    return undefined
+  }
```
kyrylodolynskyi, line 36: This seems like an overcomplication to me. Also, you don't need to transform config here, you should simply merge it, which means you don't need to put `batchId` prop in the `batch` object, you will do it when you transform `CypressConfig` to `Config` (`CoreConfig`).

```suggestion
  return {
    ...globalConfig,
    ...openConfig,
    batch: {...globalConfig.batch, ...openConfig.batch}
  }
```

js/packages/eyes-cypress/src/expose.ts
```diff
@@ -20,7 +20,11 @@ type Selector = {selector: string; type?: 'css' | 'xpath'; nodeType?: 'element'
 type Element = HTMLElement | JQuery<HTMLElement>
 type ElementWithOptions = {element: Element; regionId?: string; padding?: any}
 type SpecType = core.SpecType<unknown, unknown, Element, Selector>
-
+export type CypressBrowser = MaybeArray<
+  | NonNullable<CypressEyesConfig['browsersInfo']>[number]
+  | {deviceName: DeviceName; screenOrientation?: api.ScreenOrientationPlain; name?: string}
```
kyrylodolynskyi, line 25: Is this second type (`{deviceName, screenOrientation, name}`) not in the `CypressEyesConfig` type? It should be there.

SophieDePaula: @kyrylodolynskyi It's there because I added it, I just defined a new type so I can export it and use it elsewhere

js/packages/eyes-cypress/src/browser/mergeCypressConfigs.ts
SophieDePaula: @kyrylodolynskyi:
1) Doing what you suggested and doing `{...globalConfig, ....openConfig}` is the same.
2) The logic I presented here is merging , for instance, if the user set in the global config `batch: {id: '1234', name: 'my batch'}` and in open `batchId: '4321'`, I'll need to overwrite the id from the global with the prop I got from open and still keep the `name`

kyrylodolynskyi: 1. It is not the same, because you will get the `batch` object as a composition of `batch` from `globalConfig` and `openConfig`.
2. You are not exactly merging here, you are transforming. The return value of this function is not `CypressConfig`, it is a different type that doesn't have `batchId`, `batchName`, `batchSequenceName`, `notifyOnCompletion` properties (it still has but we don't need to look at them) because they we already processed and used to construct `batch` object. I don't want to introduce a new type, I want us to use only what is necessary, which means in this case we should minimize the work with cypress specific type. If we going to transform this type into something else anyway, why would we split this work? All of the transformations should be done in the `transfromCypressConfig` function.

js/packages/eyes-cypress/src/expose.ts
kyrylodolynskyi: Ah, I get it. You copied it from the `CypressEyesConfig` declaration, but this line works differently here and there. Basically what you did now has a circular dependency. Also, I think (I maybe wrong) you don't need to declare this type here separately because you don't need to use it elsewhere. For such a type to be declared we need to have a good reason. As I wrote in my previous comment, we need to minimize the work with cypress-specific types.

js/packages/eyes-cypress/src/browser/mergeCypressConfigs.ts
SophieDePaula: @kyrylodolynskyi ok, I see your pooint

js/packages/eyes-cypress/src/expose.ts
SophieDePaula: It used to reference itself even before my change:
```
  browser?: MaybeArray<
    | NonNullable<CypressEyesConfig['browsersInfo']>[number]
    | {deviceName: string; screenOrientation?: api.ScreenOrientationPlain; name?: string}
  >
  ```
  I need this type when I transform the user input to the shape `core` expects in `utils`.
Do you have a better suggestion?

Comment (SophieDePaula) [reply js/packages/eyes-cypress/src/expose.ts:25]: I didn't tag you so , @kyrylodolynskyi 👆

File: js/packages/eyes-cypress/src/browser/transformCypressConfig.ts [hunk 12]
```diff
@@ -0,0 +1,56 @@
...
+      proxy: config.proxy,
+      appName: config.appName,
+      testName: config.testName,
+      displayName: config.displayName,
+      batch: {
+        id: getBatchProperty('batchId', 'id'),
+        name: getBatchProperty('batchName', 'name'),
+        sequenceName: getBatchProperty('batchSequenceName', 'sequenceName'),
+        notifyOnCompletion: getBatchProperty('notifyOnCompletion', 'notifyOnCompletion'),
+      },
```
kyrylodolynskyi, line 20: This way it will be much more clear. Also in your implementation, you are losing all other batch properties, for example, `properties`.
```suggestion
      batch: {
        ...config.batch,
        id: config.batchId ?? config.batch?.id,
        name: config.batchName ?? config.batch?.name,
        sequenceName: config.batchSequenceName ?? config.batch?.sequenceName,
        notifyOnCompletion: conifg.notifyOnCompletion ?? config.batch?.notifyOnCompletion,
      },
```

js/packages/eyes-cypress/src/browser/transformCypressCheckSettings.ts
```diff
@@ -0,0 +1,255 @@
...
+            shadowDomSettings['shadow'] = prevSettings
+          }
+        }
+      }
+      regionSettings = {region: shadowDomSettings}
+    }
+    return regionSettings
+  }
+
+  function convertAccessabilityRegions(
```
kyrylodolynskyi, line 108: I don't like that accessibility, floating, and other regions have their own transformation methods. Most of the logic is the same and should be shared.

js/packages/eyes-cypress/src/expose.ts
```diff
@@ -9,60 +9,80 @@
...
+export type Element = HTMLElement | JQuery<HTMLElement>
+export type ElementWithOptions = {element: Element; regionId?: string; padding?: any}
+export type SelectorWithOptions = {region: Selector; regionId?: string; padding?: number | LegacyRegion}
 type SpecType = core.SpecType<unknown, unknown, Element, Selector>
+export type CypressBrowser = MaybeArray<
+  | NonNullable<CypressEyesConfig['browsersInfo']>[number]
+  | {deviceName: DeviceName; screenOrientation?: api.ScreenOrientationPlain; name?: string}
+>
+export type AccessibilityValidation = NonNullable<CypressEyesConfig['defaultMatchSettings']>['accessibilitySettings']
+export type floatingRegion = MaybeArray<
```
kyrylodolynskyi, line 27: Please call types using PascalCase.

js/packages/eyes-cypress/src/expose.ts
```diff
@@ -9,60 +9,80 @@
...
-    | NonNullable<CypressCheckSettings['accessibilityRegions']>[number]
-    | ((ElementWithOptions | Selector | LegacyRegion) & {accessibilityType?: api.AccessibilityRegionTypePlain})
-  >
+  region?: LegacyRegion
+  ignore?: MaybeArray<NonNullable<Element | ElementWithOptions | LegacyRegion | Selector | EyesSelector>>
+  layout?: MaybeArray<NonNullable<Element | ElementWithOptions | LegacyRegion | Selector | EyesSelector>>
+  content?: MaybeArray<NonNullable<Element | ElementWithOptions | LegacyRegion | Selector | EyesSelector>>
+  strict?: MaybeArray<NonNullable<Element | ElementWithOptions | LegacyRegion | Selector | EyesSelector>>
+  floating?: floatingRegion
+  accessibility?: accessibilityRegion
```
kyrylodolynskyi, line 57: I feel dissonance looking at this. You should use the same approach for typing all of the regions. Also, I see that we don't extend this type from the base Configuration class.

kyrylodolynskyi (changes requested):

js/packages/eyes-cypress/src/expose.ts
SophieDePaula: @kyrylodolynskyi I defined the types according to what I see we allow in our docs

js/packages/eyes-cypress/src/browser/transformCypressCheckSettings.ts
SophieDePaula: @kyrylodolynskyi yes there are similarities in the logic for all regions, but there are also differences, I need to check for a specific props in floating and accessibility and map them differently.

kyrylodolynskyi: But there are more similarities than differences. Floating/accessibility regions are just the normal codded region + additional properties. This should be reflected in types as well as in implementation. Any change in the codded region should get the accessibility/floating regions without any additional work. But with this implementation, we have to remember to do the same thing 3 times.
I think you should start with the type definition, from the discussion below I understood that you decided to redefine some types, I am ok with that, but you should build the types with the idea in mind that codded region is the base type for floating and accessibility region types. Then the organization of the implementation should be more obvious.

js/packages/eyes-cypress/src/expose.ts
kyrylodolynskyi: Let's talk about it in the conversation above.

js/packages/eyes-cypress/src/browser/transformCypressCheckSettings.ts
SophieDePaula: @kyrylodolynskyi I understand this can be optimized, but we need to release this fix. Let's do what you suggested as part of Cypress code improvments

pending-changes.yaml
```diff
@@ -72,7 +72,7 @@ Apollo SDK:
   feature:
     - 
   bug-fix:
-    - 
+    - Map values from open and global config
```
SophieDePaula, line 75: @benbabayoff8

js/packages/eyes-cypress/src/browser/transformCypressCheckSettings.ts
kyrylodolynskyi: As for me, this was a Cypress code improvement. We probably not gonna get time to just sit and do refactoring in the whole SDK.
In my opinion, the current state of cypress (the mess) is the result of a chain of decisions to do the right things later in order to release it faster, I am not a fan. This PR is good, it provides some order in the mess it was, so maybe it could be released in the current state, but I don't like this state of mind in the first place.

SophieDePaula: @kyrylodolynskyi I agree with you, but we need to find the balance between these two things and I think we are in agreement that now we are in the better shape

PR #1386, opened 2023-04-03
Java SDK

IdosApplitools:
Added the Java SDK to the monorepo :tada:


.gitignore
```diff
@@ -3,3 +3,4 @@
 .history
 .DS_Store
 Session.vim
+*.iml
```
kyrylodolynskyi, line 6: Is it java related? In this file, we should specify only stuff related to the environment (os, ide, etc.)

.github/actions/java/tests/action.yml
kyrylodolynskyi, line 1: Correct me if I wrong, but I don't think sub folders are allowed in `.github/actions`

.gitignore
IdosApplitools: The `.iml` file is generated by IntelliJi, which is the Java IDE, when it detected the project. I think I can have the IDE detect only the inner Java folder. I'll look into it.

.github/actions/java/tests/action.yml
IdosApplitools: The tests seem to be running so I guess this is allowed.
e.g. https://github.com/applitools/eyes.sdk.javascript1/actions/runs/4597512655

.gitignore
IdosApplitools: Not having the `.iml` file limits the IDE from searching files that are outside the `/java` folder, however, I can have the IDE ignore these files (so git does not track them) while still having them so I can search the entire repo.

PR #1388, opened 2023-04-07
[core] Return clickable coordinates from `locate`

kyrylodolynskyi:
In this PR I am going to change the way we return regions from `core.locate` method. The change makes sense under the assumption that the `locate` functionality is mostly used by the customers to click on stuff, and the current impl causes problems with that because it returns coordinates related to screenshots, and they not always could be applied directly to do a click.

Also, I find out that core returns located regions in server format `{left, top, width, height}` when we always declare it to be `{x, y, width., height}`. So I changed it the way it was supposed to be with `x` and `y`.


SophieDePaula (approved):
@kyrylodolynskyi looks good to me, just notice that you didn't accept the baseline in `returns correct coordinates for web page` so the test fails.

kyrylodolynskyi:
@SophieDePaula Did you run the tests?

SophieDePaula:
> @SophieDePaula Did you run the tests?


@kyrylodolynskyi yes I did

tourdedave (approved):
LGTM - glad to see we're fixing the x/left, y/top shenanigans.

kyrylodolynskyi:
@SophieDePaula All of the baselines were accepted. Maybe you were running with the wrong api key.

PR #1422, opened 2023-04-10
fix: map the results to result

zamboney:
map it to the right type of Result[]


kyrylodolynskyi (changes requested):
What about other output formats?

kyrylodolynskyi (changes requested):
@zamboney I see the storybook also has the `handleJsonFile` impl. Did you miss it?

zamboney:
@kyrylodolynskyi it's just a map to `JSON.stringify` and the original result didn't change so i think it should keep the same

kyrylodolynskyi:
@zamboney I don't understand, the function accepts one time but you feed it with another. This is not correct. I understand that nothing will blow up due to the implementation is simply a `JSON.stringify` call, but the format of the output is probably different now.

kyrylodolynskyi (approved):

PR #1426, opened 2023-04-11
Non-visual tests

kyrylodolynskyi:
Description coming soon...


js/packages/core/src/universal/core.ts
```diff
@@ -44,6 +44,10 @@ export function makeCore<TSpec extends SpecType, TType extends 'classic' | 'ufg'
       const eyesRef = await socket.request('Core.openEyes', options)
       return makeEyes({socket, core, eyesRef})
     },
+    async openFunctionalSession(_options) {
+      // TODO
```
tourdedave, line 48: What needs to happen here?

js/packages/execution-grid-client/test/e2e/client.spec.ts
```diff
@@ -1,117 +1,56 @@
...
-      await driver.get('https://applitools.com')
-      await driver.quit()
-      driver = await new Builder()
-        .withCapabilities({browserName: 'chrome', 'applitools:tunnel': true})
-        .usingServer(client.url)
-        .build()
-      await driver.get('https://applitools.com')
-      await driver.quit()
-    })
+    await driver.executeScript('applitools:endTest', {status: 'Failed'})
```
tourdedave, line 54: This test is just checking that the command completes, we should also check that the intended result happened in Eyes, no?

tourdedave (changes requested):
Looks great! The distinction between visual and functional tests is clear, the refactoring in the ec-client makes it easier to follow, the executeScript implementation support looks solid, and it's a nice touch that you made it so that we can get the metadata for self-healing out of executeScript.

I saw a TODO comment in core/universal and it wasn't clear to me what needed to be done and if it was necessary for this PR. And as for the execute script e2e test - it seems like we should add an assertion to make sure the desired result was achieved (or else how would we know that it's working as we expect?)

js/packages/core/src/universal/core.ts
kyrylodolynskyi: This is not necessary. It is here just because of the types. Because the universal core should be an exact mapping of the regular core, we don't need it here.

js/packages/execution-grid-client/test/e2e/client.spec.ts
kyrylodolynskyi: I agree! But the problem is that this is basically `closeAsync`, so there are no test results in the response, so we don't have the info to check it on the back-end. Maybe we should introduce another command `applitools:getResults` to use it for debugging at least.

PR #1459, opened 2023-04-18
[java, python] Use new format `Core.getECClient`

kyrylodolynskyi:
I am apologizing in advance for this crazy idea.
The changes are:
- use `Core.getECClient` instead of `Core.makeECClient`
- rename `capabilities` to `options` in the `ECClientSettings` type


IdosApplitools:
Looks good to me. The Java SDK already uses `Core.getECClient` so no changes are required there.
Just wondering, why did you call it `ECCapabilitiesOptions` and not `ECOptions`?

kyrylodolynskyi:
Because this is what the type is called on out side. https://github.com/applitools/eyes.sdk.javascript1/blob/d382a4af3f1ef58320ad3d219966abdad66e015f/js/packages/execution-grid-client/src/types.ts#L28

The logic behind - this is the options of the capabilities, that could be provided in capabilities as 
```jsonc
{
  "browserName": "chrome",
  "applitools:options": {/*this is ECCapabiltiesOptions*/}
}
```

ikseek:
Looks good to me. Is old command name and options attribute supported as well?

kyrylodolynskyi:
You can use `Core.makeECClient` with `options` instead of settings, but I would prefer to remove it once we all migrate to the new `Core.getECClient`

IdosApplitools (approved):
Looks good from Java perspective

ikseek (approved):
Thanks for clarification!
As it is planned to become the first breaking change in core-universal, I would like us to just have it clear how will we manage them. Either we follow semantic versioning and old message and argument names might be cleaned up in next major increment of core-universal, or we give up on semantic versioning and all languages SDKs should pin to specific version of core-universal (likely by bundling executable inside the package) and tie their release cycle to core-universal.

PR #1488, opened 2023-04-21
[req] node-fetch update to v3

kyrylodolynskyi:


tourdedave (approved):
LGTM, but I feel like I just understand the changes at a high level. Probably worth discussing this in-depth in an information sharing session with the team at some point soon to make sure we all understand the fundamentals here.

PR #1489, opened 2023-04-21
fix fetchConcurrency mapping in core-server and test

SophieDePaula:
I missed one place I was suppose to map this value.


pending-changes.yaml
```diff
@@ -61,7 +61,7 @@ Apollo SDK:
   feature:
     - 
   bug-fix:
-    - 
+    - Fix fetchConcurrency mapping
```
SophieDePaula, line 64: @benbabayoff8

kyrylodolynskyi (changes requested):
I don't think we need this PR. You didn't miss anything. You simply should use `settings` in cypress instead of sending everything plain.

PR #1490, opened 2023-04-24
js/eyes-storybook: fix disableBrowserFetching

tourdedave:
[Trello](https://trello.com/c/2PtnFEFc)

There was a typo in the settings creation for takeDomSnapshots that prevented disableBrowserFetching from getting passed correctly. I initially wanted to verify this behavior in an e2e test but found it to be very cumbersome to get right. So instead I opted to abstract the setting creation into a function that's covered by a unit test.

This SDK needs some attention in order to be brought up to the same level of quality that our other SDKs have (e.g., TypeScript, architecture that is easier to reason about, etc.)


js/packages/eyes-storybook/src/utils/prepare-settings.js
```diff
@@ -0,0 +1,10 @@
+function prepareTakeDomSnapshotsSettings({config, options}) {
+  let result = {...options};
+  result.layoutBreakpoints !== undefined ? result.layoutBreakpoints : config.layoutBreakpoints,
+    (result.disableBrowserFetching = config.disableBrowserFetching);
```
kyrylodolynskyi, line 4: This looks strange. You probably meant something like this
```suggestion
  result.layoutBreakpoints = result.layoutBreakpoints !== undefined ? result.layoutBreakpoints : config.layoutBreakpoints;
  result.disableBrowserFetching = config.disableBrowserFetching;
```

tourdedave: Move fast and break things. 😎 

Fixed in https://github.com/applitools/eyes.sdk.javascript1/pull/1490/commits/18c3ca4d2b1a8e08845aad205d270c216c304389

kyrylodolynskyi (approved):

PR #1507, opened 2023-04-30
fix: broken typescript es6 declaration

NoamGaash:
when using `@applitools/eyes-selenium` with typescript, I got the following typescript exception:
image

seems like typescript needs to know (using `es6` in `lib` in the `tsconfig.json`) when an es6 feature is being used:
https://github.com/TypeStrong/atom-typescript/issues/1329

for additional reading:
https://www.typescriptlang.org/tsconfig#lib

I believe this change might fix the issue

### DIFF
--- a/js/tsconfig.base.json
+++ b/js/tsconfig.base.json
@@ -3,7 +3,7 @@
     "module": "commonjs",
     "moduleResolution": "node16",
     "target": "es2019",
-    "lib": ["es2019", "es2020.promise", "es2020.bigint", "es2020.string", "dom"],
+    "lib": ["es2019", "es2020.promise", "es2020.bigint", "es2020.string", "dom", "ES6"],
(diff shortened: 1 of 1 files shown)


kyrylodolynskyi:
Ah, there is a different PR. This makes no sense to me, `es6 === es2015`, but we support `e2019` which is a superset of `es2015`.

kyrylodolynskyi:
@NoamGaash I am closing it. The issue was caused by a bug in our tool that generates d.ts files. If you curious here is the fix - https://github.com/applitools/eyes.sdk.javascript1/commit/4f6d76451bc9d864789d5fb1272b302fd0e1f9d1. It was due to a breaking change ts internal api, so nothing interesting.

NoamGaash:
thank you ! :pray:

PR #1509, opened 2023-04-30
fix: broken typescript es6 decleration

NoamGaash:
this will fix the `npm i` command on new npm versions (I use 9.5.1)
https://stackoverflow.com/questions/74679689/unsupported-url-type-link-while-using-npm

### DIFF
--- a/js/tsconfig.base.json
+++ b/js/tsconfig.base.json
@@ -3,7 +3,7 @@
     "module": "commonjs",
     "moduleResolution": "node16",
     "target": "es2019",
-    "lib": ["es2019", "es2020.promise", "es2020.bigint", "es2020.string", "dom"],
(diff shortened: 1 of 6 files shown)


kyrylodolynskyi:
We use yarn. So it was supposed to work the way it was.

js/tsconfig.base.json
```diff
@@ -3,7 +3,7 @@
     "module": "commonjs",
     "moduleResolution": "node16",
     "target": "es2019",
-    "lib": ["es2019", "es2020.promise", "es2020.bigint", "es2020.string", "dom"],
+    "lib": ["es2019", "es2020.promise", "es2020.bigint", "es2020.string", "dom", "ES6"],
```
kyrylodolynskyi, line 6: Why? We don't support es6. I mean we support newer standards.

PR #1520, opened 2023-05-02
[ec-client] Fixed issue with broken response of find element request

kyrylodolynskyi:
The issue happened because the response of the failed find element request was containing lots of escaped sequences for some chars (e.g. `\u003C`), but they were simplified by the `JSON.parse` call. So when we send the response back after processing, it had a smaller size than it is specified in the response `Content-Length` header, and `selenium-webdriver` impl treat it very seriously, so it was hanging until the promised size is not fulfilled (never).

[Trello](https://trello.com/c/Wae577Nf)


amitzur (approved):
Nice!

tourdedave (approved):
LGTM

PR #1521, opened 2023-05-03
[driver] Refactor user agent parsing mechanism

kyrylodolynskyi:
This PR syncs the user agent parsing mechanism with the one we have on the eyes backend. The new user agent parsing tests are also taken from the server side.


tourdedave (approved):
🥳

PR #1522, opened 2023-05-03
fix coded region with coordinates

SophieDePaula:
[Trello card](https://trello.com/c/SZTgzwzL/1807-fanduel-eyes-cypress-3300-no-longer-honors-layoutbreakpoints-in-config-file)


pending-changes.yaml
```diff
@@ -68,6 +68,7 @@ Apollo SDK:
     - 
   bug-fix:
     - Update internal dependencies
+    - Fix coded regions by coordinates
```
SophieDePaula, line 71: @benbabayoff8

PR #1525, opened 2023-05-06
fix: Cannot destructure property 'selector' of 'calculatedRegions.shift

NoamGaash:
could fix the following bug:
https://trello.com/c/RijdUGer/1856-cannot-destructure-property-selector-of-calculatedregionsshift-as-it-is-undefined


js/packages/core/src/automation/utils/to-base-check-settings.ts
```diff
@@ -38,13 +38,18 @@ export function toBaseCheckSettings<TSpec extends SpecType>({settings}: {setting
       transformedSettings[`${regionType}Regions`] = transformedSettings[`${regionType}Regions`]?.flatMap(reference => {
         const {region, ...options} = utils.types.has(reference, 'region') ? reference : {region: reference}
         if (isRegion(region)) return reference
-        const {selector, regions} = calculatedRegions.shift()!
+        const regionsWithSelectors = calculatedRegions.shift();
```
NoamGaash, line 41: I assumed that the side effect of removing elements from `calculatedRegions` has been done on purpose.
However, you might want to consider this approach - changing objects received by reference could be very confusing for most developers

PR #1526, opened 2023-05-08
Logging refactoring

kyrylodolynskyi:
This PR creates a more trackable log structure.
The new structure:
```
{label} ({"/" separated set of tags} & {other "/" separated set of tags}) | {timestamp} [{log level}] {message}
```
Example:
```
core (eyes/check) | 2023-05-05T13:47:38.461Z [INFO] Command "check" called blalba...
```


kyrylodolynskyi:
@SophieDePaula I don't see how it would be better, I think there is no reason to have the same id for different operations. Also, it would create unnecessary complications in the implementation.
I was going to update `eyes-api`, you didn't need it to try out new logs. Now it is done

kyrylodolynskyi:
Logs example: [logs.log](https://github.com/applitools/eyes.sdk.javascript1/files/11431189/logs.log)

SophieDePaula:
@kyrylodolynskyi I thought we needed a better way to link open, check and close for a specific test to track down problems as when running on multiple browsers on the grid everything is happening asynchronously.
I do like the new logging system you added though :)

SophieDePaula (approved):

kyrylodolynskyi:
@SophieDePaula We have this way. If you called `openEyes` it will generate (`eyes-12345` tad), all of the related log lines will contain this, so you don't care about the id being the same, because tags for each log line will contain the whole path from the moment of initiation to the current command

amitzur (commented):
For now, my review is just on the log file output, not the implementation.
I'd like to get the log structure right, and then the implementation is less concerning.

tourdedave:
Is there a spec that this implementation is written against that aligns with what the field expects?

kyrylodolynskyi:
@amitzur @IdosApplitools @SophieDePaula @tourdedave 
Please review these files:
[logs.log](https://github.com/applitools/eyes.sdk.javascript1/files/11451531/logs.log) - slightly improved log file
[logs-parsed.json](https://github.com/applitools/eyes.sdk.javascript1/files/11451544/logs-parsed.json.log) - parsed log file where each multiline log represented as an object (generated using `cat logs.log | eyes logs`)
[logs-structured.json.log](https://github.com/applitools/eyes.sdk.javascript1/files/11451556/logs-structured.json.log) - parsed and structured log file where each log object is placed under a certain namespace according to tags (generated using `cat logs.log | eyes logs --structure`)

IdosApplitools:
@kyrylodolynskyi I like it, the structured format is clear and I immediately feel I understand what happened in the log by looking at high level keys, like `manager-ufg-n3a/eyes-ufg-d85` shows me two ufg tests with an appropriate close and complete results. 

If I look at a tag `manager-ufg-n3a/eyes-ufg-d85/check-ufg-ojl` for example, am I expected to find things related only to the last tag in the path? seeing `check-ufg-ojl/check-ufg-icm` or `check-ufg-ojl/renderer-ozr/check-ufg-icm` means the second check is linked to the first? 

Also, wondering if its right that under `eyes-ufg-d85` I see `check-ufg-icm` & `check-ufg-ojl` while under `get-eyes-results-8ul` I see only `check-ufg-icm`?

kyrylodolynskyi:
@IdosApplitools Oops, duplicated `check` tag actually appears only inside driver logs, and this is a bug that happens due wrong usage of the logger inside a driver. I will fix it.

PR #1528, opened 2023-05-08
Js/feat/refresh layout breakpoints

tourdedave:
[Trello](https://trello.com/c/3KZDat3M)
[Generic coverage test](https://github.com/applitools/sdk.coverage.tests/pull/191)


js/packages/eyes-api/src/input/CheckSettings.ts
```diff
@@ -67,6 +67,7 @@ export type CheckSettingsAutomation<TSpec extends Core.SpecType> = CheckSettings
   fully?: boolean
   disableBrowserFetching?: boolean
   layoutBreakpoints?: boolean | number[]
+  reloadLayoutBreakpoints?: boolean
```
kyrylodolynskyi, line 70: if the fluent api is `.layoutBreakpoints(breakpoint: number[] | boolean, settings: {reload?: boolean})`, then I think the plain api should be something like `{layoutBreakpoints: number[] | boolean | {breakpoints: number[] | boolean, reload?: boolean}}`. Tell me what you think about it.

kyrylodolynskyi (changes requested):

js/packages/driver/src/driver.ts
```diff
@@ -88,6 +88,12 @@ export class Driver<T extends SpecType> {
     this._currentContext = context
   }
 
+  async reloadPage(): Promise<this> {
+    const context = this.currentContext.target
+    await this._spec.executeScript(context, snippets.reloadPage)
+    return this.refresh()
+  }
```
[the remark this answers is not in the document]
tourdedave: It seemed logical to me. We're reloading the page in a different state. It seemed like there was a potential the driver's knowledge the page/context/etc. to be outdated. Do you think I should remove it?

js/packages/eyes-api/src/input/CheckSettings.ts
tourdedave: I like the consolidation, but do they need to match? It feels a little clunky/redundant using the word `breakpoints` twice. I wasn't sure of a good keyword to use, so I didn't explore this direction.

js/packages/driver/src/driver.ts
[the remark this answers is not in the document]
kyrylodolynskyi: A refresh is needed to calculate to ensure we know the current context, but we know that after reloading current context is the main context, so there is no need for fancy logic. Also, it will remove all of the cached values, I could not say about all of them, but at least the viewport and environment, are not gonna change. I think you should not call `.refresh()` but instead adjust the state of the driver the way it should be after reloading.

js/packages/eyes-api/src/input/CheckSettings.ts
kyrylodolynskyi: They don't necessarily have to match, but I think it is a nicer api if you cannot specify something that doesn't make sense like `reloadLayoutBreakpoints` without `layoutBreakpoints`.

tourdedave: Fair point. I'll update it.

js/packages/driver/src/driver.ts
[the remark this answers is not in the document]
tourdedave: Running with `mainContext` but without `refresh`, gets this error:

`Error: Execution context was destroyed, most likely because of a navigation.`

js/packages/core/src/ufg/utils/take-dom-snapshots.ts
```diff
@@ -96,6 +96,10 @@ export async function takeDomSnapshots<TSpec extends SpecType>({
     logger.log(`taking dom snapshot for width ${requiredWidth}`)
     try {
       await driver.setViewportSize({width: requiredWidth, height: viewportSize.height})
+      if (utils.types.has(settings.layoutBreakpoints, 'reload')) {
```
tourdedave, line 99: @kyrylodolynskyi how can I check that this is true w/o hitting a build error?

pending-changes.yaml
```diff
@@ -10,30 +10,30 @@ Apollo SDK:
       An example of how to inline `markdown code highlights`
 "@applitools/core":
   feature:
-    - 
+    - Add support for reloading the page when using layoutBreakpoints
```
tourdedave, line 13: @benbabayoff8 ☝️

js/packages/core/src/ufg/utils/take-dom-snapshots.ts
kyrylodolynskyi: I see you made core types accepting new types of property. So before checking the property, you should ensure that the variable is of type that may have this property. For example in your case you can check that this is an object first, and then you can be sure that you work with the type `{breakpoints: number[] | boolean; reload?: boolean}`.
But actually the right solution would be to accept only the object type, since this is a superior way to specify layout breakpoint. You can do it just in ds function, or make a breaking change and do it in the core. Both a fine to me, since this was a reason to separate core from user-facing api, so we don't have to bear the burden of legacy api.

js/packages/eyes-api/src/input/CheckSettings.ts
```diff
@@ -528,14 +528,16 @@ export class CheckSettingsAutomationFluent<TSpec extends Core.SpecType = Core.Sp
     return this
   }
 
-  layoutBreakpoints(layoutBreakpoints: boolean | number[] = true): this {
+  layoutBreakpoints(layoutBreakpoints: boolean | number[] = true, settings: {reload?: boolean}): this {
```
kyrylodolynskyi, line 531: This kind of change also should be done in `Configuration`, since this is another way to specify layout breakpoint.

kyrylodolynskyi (changes requested):

js/packages/driver/src/driver.ts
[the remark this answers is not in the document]
kyrylodolynskyi: This is because you didn't adjust the state. 
This means:
1. `this._currentContext = this._mainContext`
2. `this._state = {}` & `this._helper = undefined`

js/packages/eyes-api/src/input/CheckSettings.ts
tourdedave: Right. Nice catch.

js/packages/driver/src/driver.ts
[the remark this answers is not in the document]
tourdedave: So, this?

```
    await this.mainContext.execute(snippets.reloadPage).catch(() => null)
    this._currentContext = this._mainContext
    this._state = {}
    this._helper = undefined
```

It still gets the same error. What am I missing?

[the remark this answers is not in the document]
tourdedave: We discussed this offline. We will keep `refresh()`.

PR #1580, opened 2023-05-16
chore: conventional commits

kyrylodolynskyi:
Adding conventional commits validation. This means we will not be able to commit anything if a commit message is not in a specific format ([Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)). This is the first step to integrating the new release flow.

Once it is merged calling `yarn` on `js` level will be required to install new listing deps.


amitzur (approved):
My notes:
1. `commitlint --edit` means `read last commit message from the specified file or fallbacks to ./.git/COMMIT_EDITMSG`. I had to run `npx commitlint --help` to understand.
2. default types that come with `config-conventional` are nice - https://github.com/conventional-changelog/commitlint/tree/master/@commitlint/config-conventional#type-enum - I encourage us to use them (docs, perf, refactor, test, etc.)
3. Was the husky rule added with `npx husky add .husky/commit-msg` as documented in https://github.com/conventional-changelog/commitlint#add-hook ?

kyrylodolynskyi:
@amitzur I did use the husky cli to add this hook, instead I simply added the file with the script. I kinda forgot about cli, I did use it when I added the code linting hook.
About the rules, we can adjust them, there is a config file that we can use in the similar way as eslint config.

zamboney:
I think that every commit is to rubast, more then 1 commit in pr is the standard I encourage

SophieDePaula:
Fyi, when I merge a PR I use `squash and merge`, therefore that the commit message we will see in `master` is actually the name of the PR.
I also agree with @zamboney about only needing one conventional commit in a PR.

amitzur:
@zamboney @SophieDePaula how will that single commit in a PR become the squashed commit message? I tend to agree on your direction, I want to understand the mechanics.

kyrylodolynskyi:
Why is it a problem to have all of the commits follow the conversation? I think we can enforce the PR title to follow this conversation, but since we are keeping the possibility to push to master where those commit messages will be validated?
In general I think it is just better practice to enforce something, then believe that we will keep convention manually. Plus this is really simple and provides additional context to the commit messages.

kyrylodolynskyi:
@SophieDePaula > Fyi, when I merge a PR I use `squash and merge`, therefore the commit message we will see in `master` is actually the name of the PR.

GitHub generates a commit message out of pr name + all of the commit messages in this PR. It also gives you an opportunity to change it before merging.

amitzur:
1. I don't think it's a big overhead to have all the commits as conventional commits, and on the other hand neither do I see a big value in it, since the PR title is more important.
2. I don't understand the statement from @zamboney that only 1 commit in the PR should be conventional commit. I don't see any reason for it.

I encourage everyone to be as clear as possible in commit messages, but I acknowledge that sometimes there are really small things that we try out and don't want big overhead of commit messages. But I don't think conventional commits incur any overhead in that.

For example, even for the smallest commits, these would still be valid conventional commits inside a PR:
- `fix: lint`
- `fix: test`
- `fix: typo`
- `refactor: getStoryData method`
- `chore: try 1`
- `chore: try 2`
- `chore: try 3`

kyrylodolynskyi:
@amitzur I don't really understand your position. In the same message, you said that you don't see a big value in having it for every commit, but you also articulated that it is not a big overhead to use it for every commit.

I agree that having it for every commit is not a game changer. But I think that it allows us to think less about one more thing. I wouldn't like to think if this commit should be conventional or not.
About the PR title, it is important but again not mandatory, because you decide on the commit message while merging. Also since the push to master is still allowed it is important to keep track of those commits too.
Regarding what Ran meant about "single conventional commit" I think he meant that all of the commits will be part of the squashed commit message and `release-please` will find it there. You can see this example that of multiple features/fixes implemented in a single PR - https://github.com/googleapis/release-please#what-if-my-pr-contains-multiple-fixes-or-features. In my opinion, this link also proves why all of the commits should be conventional because I may do some side fixes while working on a PR, and construction of the commit message is a good moment to give it some thought and note it.

amitzur:
@kyrylodolynskyi 

> it allows us to think less about one more thing. I wouldn't like to think if this commit should be conventional or not.

Yes yes yes, this is the sentence I forgot to add.

In my message I just wanted to represent both sides. Not very productive, I know.
So, given that there's no overhead I say move forward - conventional commits enforcement for all commits.

zamboney:
@kyrylodolynskyi @amitzur I think the main issue with using the commits as a source for the changes is that GitHub not provides us real control over the commit title/message that the "squash and merge" produce.

this is a real issue for any usage of conventional commit and "squash and merge". this is why i think that the solution should be one of these choices:
1. prevent the usage of squash and merge: this way, we can merge/rebase the commits in the pr and then, as @kyrylodolynskyi suggested in this pr, force every commit to be in the conventional commit format.
2. change the default commit message in "squash and merge" to take the PR **title** and **description** : this way, we don't need a conventional commit in the pr, but we do enforce the title of the pr to be in the conventional commit title which will represent the change that PR does.
3. change the default commit message in "squash and merge" to take the pr **title** and **commits details**: this way, we can force a conventional commit message on the commit for the pr, but it will **not** respect it because of the structure of the message, the commit message of the "squash and merge" is not in the same format as the @kyrylodolynskyi  suggested [here](https://github.com/googleapis/release-please#what-if-my-pr-contains-multiple-fixes-or-features) (take a look of a random [pr message](https://github.com/applitools/eyes.sdk.javascript1/commit/76b0a886e38728514c1b721f13dda34cbd05555f) and you can see the difference)

The technic @kyrylodolynskyi suggested (and I mentioned it in my 3rd option) can be used on the 2nd option as part of the pr description and it will represent the change in the changelog (once we move to use `release please`)

Hope its makes sense for you

amitzur:
Thanks @zamboney .
@kyrylodolynskyi - can you share more details on the planned next steps after this?
I don't mind approving this PR and using conventional commits for every commit **even as a standalone step** to standardize our commit messages.
How will this help the release flow, and which of the options that @zamboney suggested should we choose? Do we plan to even use release-please or do we intend to roll our own "inspired by"?

kyrylodolynskyi:
@zamboney @amitzur I don't want to swamp in discussion again. So I am going to merge it now, at least as a new convention for commits.
Regarding what Ran said.
1. We should use "squash and merge" because it is much easier to understand what the change was about later.
2. I see what you saying about the "squash and merge" default message format. We will solve it as you suggested by changing the default formatting of the squash commit message.

PR #1581, opened 2023-05-17
add failOnDiff to let the user fail on errors and not diffs

SophieDePaula:
[Trello card](https://trello.com/c/3sq0h3lW/1872-canvaexitcode-configuration-in-storybook)
As for Canva request that we let the user exit with non-zero code if the test crashes but exit with zero exit code if there are diffs.
It was not clear to me if we need to exit with non zero code in case there was an error during one of the stories and at time of writing this I didn't not get an answer yet from Ranjith.

TODO:
- [ ] write an e2e test
- [ ] update pending changes

### DIFF
--- a/pending-changes.yaml
+++ b/pending-changes.yaml
@@ -68,7 +68,8 @@ Apollo SDK:
     - 
 "@applitools/eyes-storybook":
   feature:
-    - Add the optiosn to not fail on visual differences when exit code is set to true
+    - |
+      Add the option to not fail on visual differences when `exitcode` is set to true
   bug-fix:
     - 
 "@applitools/eyes-browser-extension":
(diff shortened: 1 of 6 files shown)


pending-changes.yaml
```diff
@@ -68,7 +68,7 @@ Apollo SDK:
     - 
 "@applitools/eyes-storybook":
   feature:
-    - 
+    - Add the optiosn to not fail on visual differences when exit code is set to true
```
SophieDePaula, line 71: @benbabayoff8

pending-changes.yaml
```diff
@@ -68,7 +68,8 @@ Apollo SDK:
     - 
 "@applitools/eyes-storybook":
   feature:
-    - 
+    - |
+      Add the option to not fail on visual differences when `exitcode` is set to true
```
SophieDePaula, line 72: @benbabayoff8

PR #1592, opened 2023-05-19
feat: update agentId for component testing

SophieDePaula:


amitzur:
@SophieDePaula we decided not to do it this way, but rather do it via a Splunk event

PR #1670, opened 2023-06-27
chore(java): support release-please

IdosApplitools:
@kyrylodolynskyi I also added the changes needed for publishing, but I think it is not fully ready. Need to add the binaries part as well as the publish strategy. I saw under `rich-workspace.ts` you specified `workspace: 'node' | 'maven'` but I think I need a different strategy to support what we currently do, more specifically the [java-strategy](https://github.com/googleapis/release-please/blob/main/docs/java.md#java-strategy) seems to be exactly what I need (to be able to change a single property inside the `pom.xml` of the parent)


kyrylodolynskyi:
@IdosApplitools Strategies and workspaces are two different things.
I just need to be sure, will you be able to release any package independently using this strategy?

IdosApplitools:
> @IdosApplitools Strategies and workspaces are two different things. I just need to be sure, will you be able to release any package independently using this strategy?

@kyrylodolynskyi I see, then I meant the part responsible for changing the actual version in the manifest file, is that the strategy or workspace?

About releasing a package independently, I did not design it yet as we currently release all packages for every change (client or core). In the release please config, I added a `java/` entry that is representing the parent, which will build all child modules and will also release all child modules so only that entry has `builds` and `releases`. 

Can we make it work that way as a start and improve later on? At least for testing it shouldn't matter

kyrylodolynskyi:
@IdosApplitools You are talking about strategy.
To be honest I am not a fan of this plan, we will not go far by cutting corners like this. I don't see us getting back to work on it once it is "done". I would prefer to do it right and get the real benefits of the integration. The point of the new release flow is not to preserve the way we work, but to change it in order to make it more flexible and comfortable. It also enforces some structure that I believe will make sense for all of us, and make our projects more understandable to one another.
I think we really need all that I say before, because since we merged our repos, we never get to the point of actual integration between our code bases, it will stay that way until some enforcement will be applied, and now we have a great opportunity.

IdosApplitools:
@kyrylodolynskyi I updated our modules to be independent. I used the `build-type` in the config to represent the maven profile to use, but it would be nice to have some way to switch profiles for a different release without changing all the entries in the config.

With this change, the [maven strategy](https://github.com/googleapis/release-please/blob/main/docs/java.md#maven-strategy) should be the right choice.

release-please-config.json
```diff
@@ -488,6 +488,169 @@
...
+    },
+    "java/eyes.universal.core/eyes-universal-core-arm": {
+      "component": "java/eyes-universal-core-arm",
+      "builds": [
+        {"artifacts": ["./target"], "build-type": "maven", "key": "{{componenet}}/{{package-version}}/{{hash}}"}
+      ],
+      "releases": [
+        {"profile": "maven"}
+      ]
+    },
```
kyrylodolynskyi, line 601: Am I understanding it right, you have `eyes.universal.core` package, but you also have a separate package per platform (just with the binary perhaps), my question is, why you don't run binary tests in the binary-specific package, but instead running them in the `eyes.universal.core`

IdosApplitools: Yes, I introduced a feature that can dynamically detect which OS is running during the build cycle of the package (e.g selenium java) and it dynamically uses the matching module with the correct binary. It was just convenient and nothing more, I will change it.

PR #1676, opened 2023-06-28
DotNet SDK

itaibh:
Adding the .net SDK to the monorepo.


itaibh:
There are actions I want to start running under `dotnet/.github/workflows/dotnet-tests.yml` but I don't see them in Actions or Checks. How do I run them?

IdosApplitools:
@itaibh you placed the workflow in the wrong directory. the `.github` directory is found at the root of the repository `./.github/workflows/`.
In addition, I see that you set your workflow to trigger for the following conditions:

```
on:
  push:
  pull_request:
```
This means it will trigger for every push to the repository and for every pull request, which is not good. You should change it so the workflow will be triggered for changes under the `dotnet/` directory or manually using `workflow_dispatch`. The downside of `workflow_dispatch` is that your workflow would have to be present in the `master` branch at least once for it to be used in other branches.

You can see what I did in the [test-java.yml](https://github.com/applitools/eyes.sdk.javascript1/blob/master/.github/workflows/test-java.yml):

```
on:
  pull_request:
    types:
      - opened
      - reopened
      - synchronize
      - ready_for_review
    paths:
      - 'java/**'
```

This will trigger the workflow for each pull request of the above types, with a path `java/**`.
You can check what Python or Ruby did with their workflows.

amitzur:
@itaibh I reviewed the general structure, and merged from master. Once tests are ready you can merge this.

PR #1680, opened 2023-06-28
feat: support dns caching

amitzur:
[Trello](https://trello.com/c/A704uHVc)

For cases that DNS is weak and flaky, this PR provides an opt-in parameter called `useDnsCache`. It causes `dns.lookup` to be run on the Eyes, UFG, and EC server URLs, and then all requests will use the IP address of these servers instead of the server name.

- [x] UFG requests should succeed
- [x] Eyes requests should succeed
- [ ] Broker requests (NML)
- [ ] Tunnel client
- [x] support storybook
- [x] support EC
- [x] support env var


iasisapp:
Hi @amitzur ,
Sending the requests to the IP instead of domain name will break the TLS.
You should use a custom DNS resolver :
https://stackoverflow.com/questions/71074255/use-custom-dns-resolver-for-any-request-in-nodejs

amitzur:
@iasisapp I was thinking that as well (you mean that the certificate could not be verified, right?) but actually @ormeda suggested that I set the `host` header to render-wus.applitools.com, and it works! Perhaps that's exactly what a custom DNS resolver would do?

amitzur:
I can confirm that using a custom DNS resolver removes the need to set the `host` header explicitly.
It also causes the unit test that uses `nock` to fail. I guess `nock` is doing some stuff under the hood.

iasisapp:
If we're already on the matter, 
It would be much better for the SDK to use a single http client for each of our servers (Eyes, UFG ...)
Holding a single http-client during the execution will reuse the http connections and will save time.

amitzur:
@iasisapp I believe you mean "https agent", not "client". It's a good idea, we'll check that. Worth trying it with the `keepAlive` option (https://nodejs.org/api/http.html#new-agentoptions).

amitzur:
My problem now is that I can't test this in any way. The fact that there was a custom lookup function doesn't appear anywhere on the request or the response. This is a problem as I would like to guard that this functionality won't be removed unintentionally.

amitzur:
I was confused by the implementation. So I wrote the following to myself. Maybe it will be useful for anyone:

##### Who cares about `eyes.test.ufgServer`?
- passed to `getUFGClient` (from core/src/ufg: `check`, `check-and-close`, `get-base-eyes`)
- populated in core-base/src/server/requests/getAccountInfo, and now enriched in core/src/ufg/open-eyes

##### Places that `useDnsCache` is used:
1. core/src/ufg/open-eyes: set on eyes.test.ufgServer
2. core/src/open-eyes: passed to core.logEvent
3. core-base/src/server/req-eyes: passed to `makeReq` for `makeReqEyes`
4. req/src/req: passed to makeProxyAgent
5. req/src/proxy-agent: used to create the lookupWithCache function passed to the http/https agent
6. ufg-client/src/server/req-ufg: passed to `makeReq` for `makeReqUFG`
7. execution-grid-client/src/req-proxy: passed to `makeReq` for `makeReqProxy`
8. eyes-storybook/src/eyesStorybook: passed to `renderStory`

##### Types that have `useDnsCache`
1. core-base/src/server/req-eyes: `ReqEyesConfig`
2. core-base/src/types: `ServerSettings`
3. core-base/src/types: `UFGServerSettings`
4. req/types: `Options` (passed to `req` and `makeReq`)
5. ufg-client/src/server/req-ufg: `ReqUFGConfig`
6. execution-grid-client/src/req-proxy: `ReqProxyConfig`
6. execution-grid-client/src/types: `ECClientSettings`

js/packages/req/src/proxy-agent.ts
```diff
@@ -23,7 +31,7 @@ export function makeProxyAgent(proxy: Proxy | undefined | ((url: URL) => Proxy |
         return createHttpProxyAgent(options)
       }
     } else if (url.protocol === 'https:') {
-      return new HttpsAgent({rejectUnauthorized: false})
+      return new HttpsAgent({rejectUnauthorized: false, lookup})
     }
```
kyrylodolynskyi, line 35: What about `HttpAgent` here? Previously we didn't care to create a custom one, because there here no custom props to pass there.

js/packages/req/src/dns-cache.ts
```diff
@@ -0,0 +1,20 @@
...
+import {promisify} from 'util'
+import {type LookupFunction} from 'net'
+const dnsLookup = promisify(dns.lookup)
+
+const dnsCache: Record<string, string> = {}
+
+export const lookup: LookupFunction = async (hostname, _options, callback) => {
+  try {
+    if (!dnsCache[hostname]) {
+      const {address} = await dnsLookup(hostname)
```
kyrylodolynskyi, line 11: Do you avoid passing `options` here for a particular reason?

js/packages/core/src/ufg/open-eyes.ts
```diff
@@ -63,7 +63,7 @@ export function makeOpenEyes<TSpec extends SpecType>({core, spec, logger: mainLo
           batchId: settings.batch?.id,
           keepBatchOpen: settings.keepBatchOpen,
           server: account.server,
-          ufgServer: account.ufgServer,
```
kyrylodolynskyi, line 66: I beliave this should be already added in `core-base`'s `.getAccountInfo()`

kyrylodolynskyi: BTW the same should be done for `account.server` as well.

kyrylodolynskyi (changes requested):

js/packages/req/src/proxy-agent.ts
amitzur: Good catch, but surprisingly the `Agent` from `http` doesn't have `lookup` in the `AgentOptions` type. It is defined in `tls.ConnectionOptions`.
But it does exist in `http.ClientRequestArgs` type.
Is this a bug in Node.js types?
Will it work if I ignore the TS error and still create `HttpAgent({lookup})`?

js/packages/req/src/dns-cache.ts
amitzur: I'm good with the defaults

js/packages/core/src/ufg/open-eyes.ts
amitzur: I was thinking about it. The plus side of it is less handling of it in the code.

On the other hand it creates an implicit decision, which I wanted to avoid - it means that calling `core.getAccountInfo()` with `useDnsCache:true` is interpreted as _"remember this setting for the UFG server requests as well"_. Although this is the behavior I want, I didn't like that it's implicit.

I'm fine changing it, it would simplify the code. I was wondering your opinion about it.

js/packages/req/src/dns-cache.ts
kyrylodolynskyi: But why do we need to interfere with whatever is happening there? I think we must preserve the current behavior as is + add caching, but this can potentially introduce a different behavior.

js/packages/core/src/ufg/open-eyes.ts
kyrylodolynskyi: I think this is precisely what we need because dns caching for me is the same as proxy, another property to tweak the way we send requests. And `getAccountInfo` sent with the proxy will cause `ufgServer` to have proxy info as well, and I made it on purpose to avoid the chain of transformations. Also types-wise this is confusing, to declare a property in one package, but not make the package do something with this property.
So this is why I am voting for putting it to the `.getAccountInfo()`.

js/packages/req/src/dns-cache.ts
```diff
@@ -1,31 +1,29 @@
...
-  }
-
-  if (parseInt(process.versions.node) < 20) {
-    callback(null, dnsCache[hostname], 4)
-  } else {
-    // @ts-ignore
-    callback(null, [{address: dnsCache[hostname], family: 4}])
+    const result = dnsCache[hostname]
+    const addresses = utils.types.isArray(result) ? result : [result]
+    if (options.all) callback(null, addresses)
```
amitzur, line 22: @kyrylodolynskyi weird, I was sure I tried this out. If it works for you then probably I didn't something wrong there

js/packages/req/src/dns-cache.ts
```diff
@@ -1,31 +1,29 @@
...
+    else callback(null, addresses[0].address, addresses[0].family)
+  } catch (error) {
+    callback(error as NodeJS.ErrnoException, '', 0)
   }
 }
+
+export const lookupWithCache = lookup as LookupFunction
```
amitzur, line 29: @kyrylodolynskyi why do you need to export the same thing again with a different name?

kyrylodolynskyi: Because if you take a look at the `LookupFunction` definition, it is a subset of how the actual lookup function is defined. It assumes that lookup will be never called with `all: true`, but it is going to be called this way in node 20, and probably with other nodes in certain conditions. So I have to cast the actual implementation to this type in order to use it for agents' input.

js/packages/req/src/dns-cache.ts
kyrylodolynskyi: I checked it with all possible minors/patches of node 20

PR #1692, opened 2023-06-30
chore(java): improve release-please integration

IdosApplitools:
Improve Java SDKs support for release-please.
- removed `coverage-tests` module
- add a build script that will build dependent modules first for each module


IdosApplitools:
I am unable to build the module artifacts since they rely on the artifacts of the inner modules.
e.g. `eyes.appium.java` uses `eyes.sdk.core` and since they are built with a matrix parallel jobs the build fails.

scripts/download-core-bin.sh
```diff
@@ -4,19 +4,22 @@ while [ $# -gt 0 ]
 do
   case $1 in
   -p|--platform) 
-    platform="$2" ;;
+    platform="$2"
+    shift 2 ;;
```
IdosApplitools, line 8: @kyrylodolynskyi can you please check this change?
I was not able to input both `--platform` and `--dir`, it would only read the first input but this change worked for me - https://github.com/applitools/eyes.sdk.javascript1/pull/1692/files#diff-217459db8436fa0d2d885acf17d7ec77b0dd9d81c5ac463e36ee6ef916ca8c8fR20

kyrylodolynskyi: I am pretty sure it was working for me, but I am fine with this change

kyrylodolynskyi (approved):

PR #1703, opened 2023-07-03
feat: ability to skip deserialize dom snapshot

amitzur:


kyrylodolynskyi (commented):
This is a little confusing to me due to:
1. I would prefer to call the new setting `deserialize` and not `skipDeserialize`, so we don't need to carry this additional historical load.
2. If the property is optional there is no way to determine what the return type is going to be. Also in general it seems to me like lots of code for such a small thing, that is going to be turned off all the time.
3. `DomSnapshot` type declared in `ufg-client` and could potentially have `string` as resource value, but the `ufg-client` itself doesn't work with such type, this means to me that something isn't right.

I want to re-propose the approach I was thinking about during our call. I took a look at the code and see that it is actually much more elegant. In `ufg-client` before working with a resource from a dom snapshot we always create a `ContentfulResource` out of it (`makeResource` impl), I believe this is the place where deserialization should happen. At the same time `DomSnapshot` type has to be easily serializable, which means no `Buffer`s at all, so it should always return base64 `string` as resource value.
What do you think?

amitzur:
@kyrylodolynskyi I'll address your specific points, and then the general note.

1. I did it to avoid a default value of `true`.
2. I agree it's a big change for such a small thing. I didn't see any error coming from the fact that `settings` is optional.
3. I agree, it means that something isn't right :)

In general, I'm not very liking the idea that we keep the values base64. Because it's an implementation detail and actually we know we will need the buffer (except for the case we're now working to solve). I like it when data flow is clear and when data transformations occur at a clear point.
That said, if doing it in `makeResource` will result in clear code, and not prone to future regressions (such as calling `makeResource` unknowingly and thereby performing the data transformation again), it is fine by me. It also has a benefit that the deserialization will move to the non-blocking part of the code (!)

kyrylodolynskyi:
@amitzur I see where you were coming from, but I look at this a bit differently, data flow is still clear we just got the new requirement - the `DomSnapshot` type should be easily serializable. So the way to address it is what I propose, but creating two flows (deserialized vs. not deserialized) is less clear to me. I also don't understand how you didn't see errors while building, there is some parts of code in the core that were not updated, so it couldn't be built, I didn't update them as well but I am really not sure that `takeDomSnapshot` without passing `skipDeserialize` will get that the return value is dom-snapshot with Buffer and not string.

Speaking about your concern, it is clear in the code that the only way to work with the resource is to create a `Resource` object, and those are created only once for each type of resource. So in this case `ContentfulResource` could be created only if it came from dom-snapshot - https://github.com/applitools/eyes.sdk.javascript1/blob/7ba5d32879bfd7b8371e3892182c21a2952c0741/js/packages/ufg-client/src/create-render-target.ts#L64-L69
or when `UrlResource` was successfully fetched - https://github.com/applitools/eyes.sdk.javascript1/blob/7ba5d32879bfd7b8371e3892182c21a2952c0741/js/packages/ufg-client/src/resources/fetch-resource.ts#L86-L90

amitzur:
@kyrylodolynskyi makes sense. I'll do the necessary changes

js/packages/ufg-client/src/create-render-target.ts
```diff
@@ -60,7 +60,7 @@ export function makeCreateRenderTarget({processResources, logger: mainLogger}: O
             (resources, [url, resource]) => {
               return Object.assign(resources, {
                 [url]: utils.types.has(resource, 'errorStatusCode')
-                  ? makeResource({id: url, errorStatusCode: resource.errorStatusCode})
+                  ? makeResource({id: url, errorStatusCode: resource.errorStatusCode as number})
```
kyrylodolynskyi, line 63: This demonstrated a problem with types here. Can I push a fix for it?

amitzur:
@kyrylodolynskyi  please review. It should be helpful to you to read these notes first:
1. take-dom-snapshot.ts - rename `deserializeDomSnapshot` to `transformRawDomSnapshot` and just remove the `Buffer.from(blob.value, 'base64')` part
2. take-dom-snapshot.spec-ts - just stricter TS (this is why I started exporting `RawDomSnapshot`)
3. create-render-target.ts - I couldn't figure out why there's a TS error all of a sudden.
4. resource.ts - (a) add `SerializedResource` type, (b) do the deserialization if needed, (c) switch `.slice` to `.subarray` on big buffers since `.slice` is deprecated since 16.15.0

js/packages/ufg-client/src/create-render-target.ts
```diff
@@ -52,30 +52,28 @@ export function makeCreateRenderTarget({processResources, logger: mainLogger}: O
...
-                      dependencies: resource.dependencies,
-                    }),
-              })
-            },
-            {},
-          ),
-        },
+        resources: Object.fromEntries([
+          ...((snapshot as DomSnapshot).resourceUrls ?? []).map(url => {
+            return [url, {url, renderer: settings?.renderer}]
```
amitzur, line 57: @kyrylodolynskyi what about `makeResource` here?

kyrylodolynskyi: Oops, my bad. Re-added it there

kyrylodolynskyi (approved):

PR #1715, opened 2023-07-06
feat: support custom property per renderer

amitzur:


js/packages/ufg-client/src/types.ts
```diff
@@ -218,12 +218,16 @@ export type AndroidDeviceRenderer = {
...
+type CustomProperty = {name: string; value: string}
+
 export type Renderer = (
   | DesktopBrowserRenderer
   | ChromeEmulationDeviceRenderer
   | IOSDeviceRenderer
   | AndroidDeviceRenderer
 ) & {
+  properties?: CustomProperty[]
+
```
kyrylodolynskyi, line 230: This shouldn't be here, because properties are not going to be used in `ufg-client`. This property is needed only in the core, so it should be added there like that `Renderer $ {properties?: CustomProperty[]}`.

js/packages/core/src/ufg/get-base-eyes.ts
```diff
@@ -42,8 +42,12 @@ export function makeGetBaseEyes<TSpec extends SpecType>({
...
     const environment = await ufgClient.bookRenderer({settings, logger})
+    let properties
+    if (defaultSettings.properties || environment.rendererInfo.renderer.properties) {
+      properties = (defaultSettings.properties ?? []).concat(environment.rendererInfo.renderer.properties ?? [])
+    }
     const baseEyes = await eyes.core.base.openEyes({
-      settings: {...defaultSettings, environment: {...defaultSettings.environment, ...environment}},
+      settings: {...defaultSettings, environment: {...defaultSettings.environment, ...environment}, properties},
       logger,
     })
```
kyrylodolynskyi, line 52: This feels very strange to me because here you rely on `ufg-client` to propagate the value that you already know, it is an unnecessary trip for this property, and I would prefer it be read directly from the `settings`.

kyrylodolynskyi (changes requested):

js/packages/core/src/ufg/get-base-eyes.ts
amitzur: Correct. I will fix this

js/packages/ufg-client/src/types.ts
amitzur: In `core/src/ufg/types.ts` there are multiple references to `Renderer`. They should all be changed, right? Should I create a `RendererWithProps` type in that file?
In addition, how should I handle `RendererSettings`? It's a type that comes from `ufg-client`. But now I need 
something like `RendererWithPropsSettings`. What's the way out of this?

kyrylodolynskyi: There 4 references, two of them are in `CheckResult` and `TestResult` I think those shouldn't be changed. The other one is `getBaseEyes` which actually would use the `RendererSettings` type which should also be patched with custom properties

js/packages/core/src/ufg/types.ts
```diff
@@ -11,13 +11,22 @@ import {
...
 export type SnapshotTarget = MaybeArray<DomSnapshot> | MaybeArray<AndroidSnapshot> | MaybeArray<IOSSnapshot>
 export type Target<TSpec extends SpecType> = SnapshotTarget | AutomationCore.Target<TSpec>
 
+type RendererWithProps = Renderer & {
+  properties?: BaseCore.CustomProperty[]
+}
+
+type RendererSettingsWithProps = RendererSettings & {
+  renderers: RendererWithProps[]
+}
```
kyrylodolynskyi, line 28: I think this is fine, but patching one type and then patching another with it, seems unnecessary. Also the name `RendererSettingsWithProps` is confusing, because we don't really need both, we will need them separately. This is why I suggest doing the `GetBaseEyesSettings` with properties, without patching the `renderers` type.

```suggestion
type GetBaseEyesSettings = RendererSettings & {
  properties?: BaseCore.CustomProperty[]
}
```

js/packages/core/src/ufg/types.ts
```diff
@@ -46,7 +52,7 @@ export interface Eyes<TSpec extends SpecType> extends AutomationCore.Eyes<TSpec>
 }
 
 export type CheckSettings<TSpec extends SpecType> = AutomationCore.CheckSettings<TSpec> & {
-  renderers?: Renderer[]
+  renderers?: RendererWithProps[]
```
kyrylodolynskyi, line 55: Then there is no need in this type anymore

```suggestion
renderers?: (Renderer & {properties?: BaseCore.CustomProperty})[]
```

kyrylodolynskyi (commented):
If you like this direction I would also suggest separating `properties` from the renderer as soon as we start working with it by replacing this in the `check`/`checkAndClose`
```js
const promises = uniqueRenderers.map(async (renderer, index) => {
```
with this
```js
const promises = uniqueRenderers.map(async ({properties, ...renderer}, index) => {
```

amitzur:
@kyrylodolynskyi really nice direction, I like it!
One thing was still an issue, `uniquifyRenders` still returns `Renderer[]`, which is not good for me, and I didn't have a type for it. I was thinking to change `uniquifyRenders` to actually not accept and return `Renderer[]` but rather something like `{id: string, [k: string]: unknown}` (is there a better way to describe an arbitrary object with the `id` property?). But I eventually chose to set it to the type `typeof settings.renderers`. Not sure if there's a cleaner way.

js/packages/core/src/ufg/check.ts
```diff
@@ -58,7 +58,7 @@ export function makeCheck<TSpec extends SpecType>({
       settings,
     })
 
-    const uniqueRenderers = uniquifyRenderers(settings.renderers ?? [])
+    const uniqueRenderers: typeof settings.renderers = uniquifyRenderers(settings.renderers ?? [])
```
kyrylodolynskyi: I've fixed this in my commit

kyrylodolynskyi (approved):
I fixed some unnecessary type declaration and added the new feature to `checkAndClose`

js/packages/core/src/ufg/utils/uniquify-renderers.ts
```diff
@@ -1,13 +1,13 @@
 import {type Renderer} from '@applitools/ufg-client'
 
-export function uniquifyRenderers(renderers: Renderer[]): Renderer[] {
+export function uniquifyRenderers<TRenderer extends Renderer>(renderers: TRenderer[]): TRenderer[] {
```
amitzur, line 3: @kyrylodolynskyi yeah, it was also a thought. I wasn't sure if to take it that far

kyrylodolynskyi: I think this is a good move because it allows us to add any additional properties we may need

PR #1718, opened 2023-07-06
feat: export deserialize dom snapshot

amitzur:


kyrylodolynskyi:
Why do we need it?

amitzur:
@kyrylodolynskyi for external consumers of takeDomSnapshots

kyrylodolynskyi:
I don't understand. Why would they use it if we already made core to work without it.

amitzur:
@kyrylodolynskyi oy, you're right. I was confused :facepalm:

PR #1719, opened 2023-07-06
Python release please integration

ikseek:


release-please-config.json
```diff
@@ -714,6 +714,56 @@
       "releases": [
         {"profile": "maven", "notification": {"name": "java_playwright"}}
       ]
+    },
+    "python/core_universal": {
+      "release-type": "python",
+      "skip-snapshot": true,
+      "component": "python/core-universal",
+      "builds": [
+        {"runner": "linux", "artifacts": ["./dist"], "key": "{{component}}/{{package-version}}/{{hash}}"}
```
kyrylodolynskyi, line 723: no need to specify `runner` if this is a default one.

release-please-config.json
```diff
@@ -714,6 +714,56 @@
       "releases": [
         {"profile": "maven", "notification": {"name": "java_playwright"}}
       ]
+    },
+    "python/core_universal": {
+      "release-type": "python",
+      "skip-snapshot": true,
```
kyrylodolynskyi, line 720: this is a java-only property

.github/actions/setup/main.ts
```diff
@@ -102,6 +103,33 @@ async function main() {
...
+          const manifest = ini.parse(await fs.readFile(setupConfigFilePath, 'utf-8'))
+          let versionModule = null
+          for (let module of ['__version__.py', '__init__.py']) try {
+              let modulePath = path.resolve(packagePath, `applitools/${packageDir}/${module}`)
+              versionModule = await fs.readFile(modulePath, 'utf8');
+              break
+          } catch (e) {
+          }
+          versionModule = versionModule ?? await fs.readFile(path.resolve(packagePath, "src/EyesLibrary/__version__.py"), 'utf8')
+          let [_, version] = versionModule.match(/__version__\s+=\s+"([0-9.]+)"/) as RegExpMatchArray
```
kyrylodolynskyi, line 118: This seems like a very complicated way to get the version. So many variations, isn't it the same in all python packages?

.github/actions/setup/main.ts
```diff
@@ -102,6 +103,33 @@ async function main() {
...
+          return packages.then(packages => {
+              packages["core-universal"] = {
```
kyrylodolynskyi, line 120: The package name has to be a key here.

.github/actions/python/tests/action.yml
```diff
@@ -0,0 +1,24 @@
...
+  working-directory:
+    description: path to the package directory relative to repository root
+    required: true
+  test-type:
+    description: test type
+    default: all
+    required: false
+  test-filter:
+    description: test filter substring
+    required: false
```
kyrylodolynskyi, line 14: What does it mean? If this is just a pattern string to decide what test to run, then we already have a "standard" name for this - `grep`. It is not required for us to have the same naming here, but I think this will really nice.

kyrylodolynskyi (changes requested):

.github/actions/setup/main.ts
ikseek: This variation is for historical/compatibility reasons. There is a change in code structure not merged yet that makes all packages have version in applitools/pkg_name/__init__.py. But robot will be different, it has their own naming convention.

kyrylodolynskyi: What about the version in `setup.cfg`? Isn't in what we need?

.github/workflows/ci.yml
```diff
@@ -307,8 +307,116 @@ jobs:
...
+      - name: Check-out repository
+        uses: actions/checkout@v3
+        with:
+          ref: ${{github.event.pull_request.head.ref}}
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{matrix.python-version}}
+      - name: Restore build artifacts
+        if: ${{matrix.builds}}
```
kyrylodolynskyi, line 333: it is always an array so it never be falsy
```suggestion
        if: ${{join(matrix.builds)}}
```

kyrylodolynskyi (changes requested):
Looks good to me. But you will have to add `"builds": ["js/core/x64/{{environment}}", "js/core/arm64/{{environment}}"]` to your build jobs that need to have binaries to create a proper build.

.github/actions/python/tests/action.yml
kyrylodolynskyi, line 1: Can you call this action `test` instead of `tests`? It is a small thing, I just want to keep the terminology the same across languages at least for central actions.

kyrylodolynskyi (commented):
Few more questions:
1. Do you use any tools for reporting / sending notifications to the qa server other than the action called `bongo` (this one https://github.com/applitools/eyes.sdk.javascript1/blob/master/.github/actions/bongo/action.yml)? 
2. Did you integrate linting into `python/build` action as we discussed so we can avoid running `pre-commit` workflow separately? 
3. Are you planning to remove `py-tests` workflow once fully migrated to the new release flow?

ikseek:
> Few more questions:
> 
> 1. Do you use any tools for reporting / sending notifications to the qa server other than the action called `bongo` (this one https://github.com/applitools/eyes.sdk.javascript1/blob/master/.github/actions/bongo/action.yml)?
Not anymore, i removed all other reporting calls. But this bongo action will need a change addressing the issue @Artem-applitools was solving by a smart report merge
> 2. Did you integrate linting into `python/build` action as we discussed so we can avoid running `pre-commit` workflow separately?
Not yet, I don't remember that was needed. I will do it in separate PR, this one is big enough already.
> 3. Are you planning to remove `py-tests` workflow once fully migrated to the new release flow?
Sure, together with py-publish.

python/Makefile
```diff
@@ -65,15 +65,15 @@ install_unit_tests_requirements: install_tests_requirements
...
 	python -m pip install 'pillow' "webdriver-manager==3.7.0; python_version>='3'" "futures<4; python_version<'3.0'"
 
 
-install_installation_tests_requirements: install_tests_requirements node_modules
+install_installation_tests_requirements: install_tests_requirements
 	python -m pip install 'pytest-venv==0.2.1' 'virtualenv==20.11'
 
 
-install_selenium_generated_tests_requirements: install_functional_tests_requirements node_modules
+install_selenium_generated_tests_requirements: install_functional_tests_requirements
```
Artem-applitools, line 76: How it generate tests now without node_modules install?

PR #1760, opened 2023-07-12
Py fix report

Artem-applitools:
meta file generated by the coverage tests tool was renamed in the latest 3.1 release and caused failures...


Artem-applitools:
https://github.com/applitools/eyes.sdk.javascript1/actions/runs/5531644649 python tests run, report was working fine)

kyrylodolynskyi:
@Artem-applitools I think we better start using this action instead - https://github.com/applitools/eyes.sdk.javascript1/tree/master/.github/actions/bongo. Here is an example - https://github.com/applitools/eyes.sdk.javascript1/blob/95e52686b94c56564a4e083189f235ff4e8c0e23/.github/workflows/ci.yml#L143-L155

Artem-applitools:
@kyrylodolynskyi it won't work for the python, it needs one report from 2 jobs. Action you mention action which working only in 1 job.

kyrylodolynskyi:
Ah. Ok, then. I don't understand why though.

ikseek:
@kyrylodolynskyi The thing is, I generate separate images, appium, and selenium test suites now using the coverage-tests tool. When all of them are executed in parallel and reported one by one, we get incorrect reports in the report spreadsheet.

As far as I understand, the problem is triggered like this: when I generate "appium" tests, "selenium" and "images" tests get marked ask skipped. When I generate "appium" tests, "selenium" and "images" tests are marked as skipped. When those suites are executed, normally selenium tests finish first and report marking appium tests as disabled is sent. Later when appium suite finishes and reports results, it overwrites passed selenium tests as skipped and marks previously skipped appium tests as passed.

What @Artem-applitools did is added an additional job that collects results from test jobs, merges them correctly and reports just once.

kyrylodolynskyi:
Ah. So this is a work around, this is definitely not what we are going to do with new release flow

Artem-applitools:
@kyrylodolynskyi how it will be handled in the new flow? Does we planning to run python appium and selenium runs in one job or bongo will be updated to have few report xml and meta json files?

PR #1763, opened 2023-07-12
perf(js/eyes-storybook): improve runtime performance

tourdedave:
[Trello](https://trello.com/c/bpQ6KkFp)

When running in a resource constrained machine we see poor performance with the storybook SDK. This PR is a first step in trying to remedy that.

A few hotspots:
 - page pool initialization (e.g., we were doing this in series when running with > 3 concurrent tabs)
 - page keep-alive (e.g., when a page is in an unhealthy state the page pool tries to kill and restart it -- this was happening even after the SDK closed the browser, causing needless retries to attempt to restore pages)
 - the browser (e.g., running with the `'new'` Chrome headless adds a boost, as does running with the latest major version of Puppeteer which pulls in a newer version of Chrome)
 
The setup:
- running on an `m5d.xlarge` AWS EC2 node
- using a large and heavy customer provided storybook
- running _without_ calls to Eyes (e.g., focusing primarily on speed of setup, JS execution, and teardown -- which accounts for 90% of the SDK's runtime)
- capturing how long it takes for a run to complete with the `time` unix command
- running with 10 concurrent tabs
- running with `disableBrowserFetching: true`
- running with the new Chrome headless

Using a combination of these code and configuration changes showed a ~38% improvement in total runtime (e.g., dropping from ~13 minutes down to ~8 minutes).


SophieDePaula (commented):
@tourdedave I agree with the approach to resolve the browser throwing an exception when we try to create a new page after closing. However, I did something similar on a different PR and Amit seemed to not like it, however, what you did is more minimal.

As for the performance improvement, what affected the performance here? The upgrade to the new Puppeteer version?
Also I am not sure why in `eyesStorybook` we separated how we created pages according to the number or tabs and why now we don't

tourdedave:
> @tourdedave I agree with the approach to resolve the browser throwing an exception when we try to create a new page after closing. However, I did something similar on a different PR and Amit seemed to not like it, however, what you did is more minimal.
> 
> As for the performance improvement, what affected the performance here? The upgrade to the new Puppeteer version? Also I am not sure why in `eyesStorybook` we separated how we created pages according to the number or tabs and why now we don't

I'm open to suggestions on how to better handle closing the page pool. It feels like the need for it is actually a symptom of a bug in `processStoryLoop`.

I've updated the PR description with more details - hopefully that help answers your question regarding what affected performance.

As for the rationale around behavior based on the number of tabs - there is a bit of an explainer in Amit's PR description which is when this behavior was introduced although I think it's focused primarily on what was done in `initPage`, not how it's called ([link](https://github.com/applitools/eyes.sdk.javascript1/pull/1560)).

SophieDePaula:
> > @tourdedave I agree with the approach to resolve the browser throwing an exception when we try to create a new page after closing. However, I did something similar on a different PR and Amit seemed to not like it, however, what you did is more minimal.
> > As for the performance improvement, what affected the performance here? The upgrade to the new Puppeteer version? Also I am not sure why in `eyesStorybook` we separated how we created pages according to the number or tabs and why now we don't
> 
> I'm open to suggestions on how to better handle closing the page pool. It feels like the need for it is actually a symptom of a bug in `processStoryLoop`.
> 
> I've updated the PR description with more details - hopefully that help answers your question regarding what affected performance.
> 
> As for the rationale around behavior based on the number of tabs - there is a bit of an explainer in Amit's PR description which is when this behavior was introduced although I think it's focused primarily on what was done in `initPage`, not how it's called ([link](https://github.com/applitools/eyes.sdk.javascript1/pull/1560)).

@tourdedave so essentially, the new Puppeteer version is the major reason for the performance improvement?
Also, why did you decide to change the behavior Amit introduced re the tabs number?

js/packages/eyes-storybook/src/eyesStorybook.js
```diff
@@ -161,19 +161,12 @@ async function eyesStorybook({
...
-      await Promise.all(
-        new Array(CONCURRENT_TABS).fill().map(async () => {
-          const {pageId} = await pagePool.createPage();
-          pagePool.addToPool(pageId);
-        }),
-      );
-    } else {
-      for (const _x of new Array(CONCURRENT_TABS).fill()) {
+    await Promise.all(
+      new Array(CONCURRENT_TABS).fill().map(async () => {
```
kyrylodolynskyi, line 165:
```suggestion
     Array.from({length: CONCURRENT_TABS}, async () => {
```

tourdedave:
> @tourdedave so essentially, the new Puppeteer version is the major reason for the performance improvement? Also, why did you decide to change the behavior Amit introduced re the tabs number?

Not exactly.
1. Running with the new version Puppeteer shaved off about 2 minutes in execution time.
2. Changing the tab behavior Amit introduced switched from spawning pages from in serial to in parallel - shaved off another 90 seconds or so (it's not clear to me why he introduced this since his rationale was not included in his PR and there are no code comments or tests to explain or illustrate the benefit, but objectively running it in parallel consistently performs faster).
3. Fixing the issue with the page pool trying to re-spawn pages after browser was closed saved some additional time (around 15 to 20 seconds).
4. Any additional speed gains are attributable to the new headless execution.

SophieDePaula (approved):
Thanks for the explanation Dave, looks good to me.

PR #1773, opened 2023-07-13
fix(js/req,js/ufg-client): support retries on stuck requests

tourdedave:
[Trello](https://trello.com/c/A704uHVc)

In some customer networks, a request to check the render status of a job in the UFG can exceed the allotted timeout of 60 seconds. When this happens the SDK aborts the job rather than retrying the request (since no response has been received for the request -- so, no HTTP status or error codes to potentially trigger a reply). This PR attempts to remedy that by enabling retries on ~~`AbortError` to be done~~ stuck requests, opt-in by specifying ~~it in `codes` (see [my test](https://github.com/applitools/eyes.sdk.javascript1/blob/fa6dc81342c8afd168ed137ebb0fa72f83bde3cc/js/packages/req/test/it/req.spec.ts#L188) for an example)~~ a new `stuckTimeout` option (see [my test](https://github.com/applitools/eyes.sdk.javascript1/blob/070fb2f672232820b24bdbb3faf14e34aba46b2e/js/packages/req/test/it/req.spec.ts#L172) for an example).

In order to support this behavior, a new `AbortController` needed to be created for each `req` (e.g., inside of the inner `req` function instead of the outer `req` function). But, not at the expense of preserving the original timeout `AbortController` behavior.

The interesting parts are a) being able to distinguish between the reason for why the request was aborted (since the version of `AbortController` we're using doesn't support passing a reason when calling `abort()`), and b) being able to only pass one `AbortController` signal to the request.


kyrylodolynskyi (commented):
This makes no sense to me, first of all, I intentionally put timeout higher than retries, because this is what we want to limit the amount of time per whole operation, not just for one request, this way there is less sense to set timeout because you will still don't know how many time it will take with all of the retries.
But an even more critical point is that AbortError !== timeout, there are other reasons for the request to about. Obviously, when you intentionally abort a request it should not be retried.

js/packages/req/test/it/req.spec.ts
```diff
@@ -169,17 +169,22 @@ describe('req', () => {
...
-      () =>
-        req('https://eyesapi.applitools.com/api/hello', {
-          retry: {validate: ({response}) => response?.status === 500},
-          timeout: 1000,
-        }),
+      () => req('https://eyesapi.applitools.com/api/hello', {stuckTimeout: 500, timeout: 1000}),
       (error: Error) => error.constructor.name === 'AbortError',
     )
+    assert.deepStrictEqual(index, 2)
   })
```
kyrylodolynskyi, line 188: Why do you remove this test? Didn't we decide to preserve the functionality we already have?

js/packages/ufg-client/src/server/requests.ts
```diff
@@ -181,6 +181,7 @@ export function makeUFGRequests({config, logger}: {config: UFGRequestsConfig; lo
       body: renders.map(render => render.renderId),
       expected: 200,
       timeout: 60000,
+      stuckTimeout: 15000,
```
kyrylodolynskyi, line 184: I understand that this number is coming from Or perception of how much time this request should take. But I think we should be slightly less optimistic here because we don't know the customers' network condition and other client-side crap. I think the original timeout was written without an understanding of the separation between stuck timeout and overall request timeout, which means the 1-minute timeout should be used as a stuck timeout and the overall timeout should be set to a higher number. But maybe you would argue that this is  "too safe", so I suggest timeout = 90_000 and stuckTiomeout = 30_000

js/packages/req/src/types.ts
```diff
@@ -48,6 +48,10 @@ export interface Options {
    * @example 7000
    */
   timeout?: number
+  /**
+   * Timeout to wait on a stuck request before aborting and retrying it
+   */
+  stuckTimeout?: number
```
kyrylodolynskyi, line 54: I like the name, but I think in terms of api it should be done a little bit differently. I will try to explain, by looking at this api it seems like I can interchange stuck and overall timeouts and the difference is in semantics. But, as your implementation proves, the only reason to use `stuckTimeout` is to retry when it happens. And I think that the api should reflect it, this is why I suggest putting this property in the `Retry` type, this will also allow configuring amount of attempts and an interval.
I would like to discuss it today over zoom to hear your arguments.

kyrylodolynskyi (changes requested):
One more thing I would like to request is since we already tracking the reasons for abortions we can throw better errors from req, by replacing the `AbortError` thrown from fetch.

js/packages/req/test/it/req.spec.ts
tourdedave: My mistake - I thought I was modifying a test I had added. I'll fix it.

js/packages/req/src/req.ts
```diff
@@ -23,14 +24,20 @@ export function makeReq<TOptions extends Options = Options, TBaseOptions extends
...
 
   if (options.hooks) options.hooks = utils.types.isArray(options.hooks) ? options.hooks : [options.hooks]
   if (options.retry) options.retry = utils.types.isArray(options.retry) ? options.retry : [options.retry]
   if (options.headers)
     options.headers = Object.fromEntries(Object.entries(options.headers).filter(([_, value]) => value))
 
-  const controller = new AbortController()
-  const timeout = options.timeout ? setTimeout(() => controller.abort(), options.timeout) : null
+  const connectionTimeoutController = new AbortController()
+  const timeout = options.connectionTimeout
```
kyrylodolynskyi, line 35: Can you use the same prefix for both, so it will be clear that they are connected? I mean rename `timeout` to `connectionTimeoutTimer` or to `connectionTimer`, but then `connectionTimeoutController` to `connectionController`.

js/packages/req/src/req.ts
```diff
@@ -103,6 +120,12 @@ export async function req(input: string | URL | Request, ...requestOptions: Opti
       response = await afterResponse({request, response, options})
       return response
     } catch (error: any) {
+      switch (abortReason) {
+        case AbortReasons.requestTimeout:
+          error = new RequestTimeoutError()
+        case AbortReasons.connectionTimeout:
+          error = new ConnectionTimeoutError()
+      }
```
kyrylodolynskyi, line 128: This is a remote scenario, but I think you should check if the error cached is AbortError, because what if we will have some other error happening between calling `controller.abort()` and the actual error thrown?

js/packages/req/src/req-errors.ts
```diff
@@ -0,0 +1,20 @@
+export const AbortReasons = {
+  requestTimeout: 'STUCK_REQUEST',
+  connectionTimeout: 'MAX_TIMEOUT_REACHED',
+}
```
kyrylodolynskyi, line 4: You should use an enum for that.

kyrylodolynskyi (changes requested):

js/packages/req/src/req.ts
```diff
@@ -39,6 +46,16 @@ export async function req(input: string | URL | Request, ...requestOptions: Opti
   }
 
   async function req(input: string | URL | Request, options: Options): Promise<Response> {
+    const controller = options.requestTimeout ? new AbortController() : connectionTimeoutController
+    const requestTimeout = options.requestTimeout
+      ? setTimeout(() => {
+          // only set it if not populated
+          // if populated, it will be because the top-level timeout set it
+          if (!abortReason) AbortReasons.requestTimeout
```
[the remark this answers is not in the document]
tourdedave: Right! I didn't update the test correctly, so this went unnoticed. Nice catch!

js/packages/req/src/req.ts
```diff
@@ -39,6 +46,16 @@ export async function req(input: string | URL | Request, ...requestOptions: Opti
...
+          controller.abort()
+        }, options.requestTimeout)
+      : null
+
     const fetch = options.fetch ?? globalFetch
 
     if (options.signal) options.signal.onabort = () => controller.abort()
```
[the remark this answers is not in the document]
tourdedave: If both are set *and* the `code` `STUCK_REQUEST` is specified in the retry options, then it will respect the `connectionTimeout`. But if the `code` is not provided, then yes, it seems like something else will happen.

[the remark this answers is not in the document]
tourdedave: I've fixed it so it will only retry on stuck requests with both the `requestTimeout` and retry.codes includes `STUCK_TIMEOUT` is specified. In either case, the `connectionTimeout` will be respected.

js/packages/req/src/req.ts
tourdedave: Wouldn't we only get an `AbortError` if we call `controller.abort()`? If there is some other error, then the error would be untouched and pass through as-is. If you can think of a concrete example then we can add a test for it to try and expose/fix this scenario. I'll leave it as-is for now.

PR #1780, opened 2023-07-15
refactor: ufg client

kyrylodolynskyi:
As part of this PR:
1. I changed the naming inside of ufg client to make them simpler and more reasonable, for example, `bookRenderer` is renamed to `getRenderEnvironement`, `RenderEnvironemnt` type was written specifically to be a subset of properties of `Environment` type form `core`, so it could be directly passed to the core.
2. I get rid of `rendererInfo` thing (it was an object that combined `renderer` and `type` of the renderer `'native' | 'web'`), instead there is a new type `MobileRenderer` combines Android and iOS renderer types with additional `type` property (this is needed when you request rendering on ios device which could be either native for nmg, or web)
3. I introduced better names for various `serverUrl` properties, since today we have many more different servers these days (eyes, ufg, tunnel, ec)
4. I replaced `fetchResourceFromTunnel` function with a req-hook that will modify the request if certain properties were passed  during the creation of `fetchResource`
5. I noticed that we processing results in various requests of `core-base`, I made it the way results processing now happens only when we call `getResults`, and this caused me to refactor a bit `checkAndClose` function since it always was more like `checkAndCloseAndGetResults`. So this function is no longer going to return results, instead `getResults` should be called. (I am wondering what it will cause in the storybook...)


SophieDePaula (approved):
Looks good to me, I see that there are braking changes to `eyes-storybook` that we will need to address

PR #1823, opened 2023-07-25
fix: consider response headers and status code which are returned from the EC resource handler

amitzur:
https://trello.com/c/CdZBNKav/813-resource-handler-service#comment-64bf92bdd888de91a76c1010


js/packages/ufg-client/src/resources/fetch-resource.ts
```diff
@@ -141,6 +141,19 @@ function handleTunneledResource({
...
+      for (const [name, value] of response.headers) {
+        if (name.startsWith('x-fetch-response-header')) {
+          headers[name.replace('x-fetch-response-header-', '')] = value
+        }
+      }
+      return new Response(response.body, {
+        status: Number(response.headers.get('x-fetch-status-code')),
+        statusText: response.headers.get('x-fetch-status-text')!,
+        headers,
+      })
```
kyrylodolynskyi, line 155: I prefer to use that syntax instead because how we work with the `Response` and `Request` classes is tricky.
```suggestion
      return {
        body: response.body,
        status: Number(response.headers.get('x-fetch-status-code')),
        statusText: response.headers.get('x-fetch-status-text')!,
        headers,
      }
```

js/packages/ufg-client/src/resources/fetch-resource.ts
```diff
@@ -1,6 +1,6 @@
 import type {Cookie} from '../types'
 import {type Logger} from '@applitools/logger'
-import {makeReq, type Fetch, type Proxy, type Hooks} from '@applitools/req'
+import {makeReq, type Fetch, type Proxy, type Hooks, Response} from '@applitools/req'
```
kyrylodolynskyi, line 3: If you going to accept what I proposed you would have to remove it as well.
```suggestion
import {makeReq, type Fetch, type Proxy, type Hooks} from '@applitools/req'
```

js/packages/ufg-client/src/resources/fetch-resource.ts
amitzur: I don't mind accepting your suggestion. But what is tricky about it? I saw that if I return a response, then that is used: https://github.com/applitools/eyes.sdk.javascript1/blob/cf34ad1a5b3cba0b29b3509616b20a2b1313c62f/js/packages/req/src/req.ts#L216 what is the difference?

tourdedave (approved):
LGTM

js/packages/ufg-client/src/resources/fetch-resource.ts
kyrylodolynskyi: It works only because req is bundled with node-fetch, maybe in the future we will be able to avoid it, the api I suggested was created to overcome this limitation, also this is the way that we use in all other places.

amitzur: Got it. So I will accept your suggestion. Thanks!

PR #1836, opened 2023-07-31
feat: drop support of programmatic debug screenshot config

kyrylodolynskyi:


amitzur:
@kyrylodolynskyi please see my note here: https://trello.com/c/XjaVesSa/105-debug-screenhosts-support-env-var-and-deprecate-api

PR #1841, opened 2023-08-01
feat: mark session as component

SophieDePaula:
https://trello.com/c/HnPWRB4u/94-cypress-component-testing

We now send `isComponentAgent` to `open` from cypress, and in `core-base` we map this value to `startInfo`


js/packages/core/src/open-eyes.ts
```diff
@@ -65,6 +65,7 @@ export function makeOpenEyes<TSpec extends SpecType, TDefaultType extends 'class
     settings.baselineBranchName ??= utils.general.getEnvValue('BASELINE_BRANCH')
     settings.ignoreBaseline ??= false
     settings.compareWithParentBranch ??= false
+    settings.isComponentAgent = config?.open.isComponentAgent
```
kyrylodolynskyi, line 68: You don't need to set it here, because it it already there, take a look at line 48

js/packages/core-base/src/types.ts
```diff
@@ -169,6 +169,8 @@ export interface OpenSettings extends EyesServerSettings {
   connectionTimeout?: number
   removeSession?: boolean
   isFunctionalTest?: boolean
+  /** @internal */
+  isComponentAgent?: boolean
```
kyrylodolynskyi, line 173: I think the property should call `isComponentTest` because agent terminology is not used on our side. Also, it would create a nice symmetry with the `isFunctionalTest` flag.

js/packages/eyes-cypress/test/components/cypress.config.js
```diff
@@ -1,13 +1,20 @@
-const webpackConfig = require('./webpack.config.js');
+const webpackConfig = require("./webpack.config.js");
```
kyrylodolynskyi, line 1: Why?

js/packages/eyes-cypress/src/plugin/pluginExport.ts
```diff
@@ -100,7 +100,9 @@ export default function makePluginExport({
         }
       }
 
-      return Object.assign({}, eyesConfig, {eyesPort: port}, moduleExportsResult)
+      return Object.assign({}, eyesConfig, {eyesPort: port}, moduleExportsResult, {
+        isComponent: cypressConfig?.testingType === 'component',
+      })
```
kyrylodolynskyi, line 105: What is the type of this object?

kyrylodolynskyi (changes requested):

js/packages/eyes-cypress/test/components/cypress.config.js
SophieDePaula: @kyrylodolynskyi I'm actually not sure how it happened, I'll revert it

js/packages/eyes-cypress/src/plugin/pluginExport.ts
SophieDePaula: @kyrylodolynskyi the returned Object? It doesn't have a type

js/packages/core-base/src/types.ts
SophieDePaula: Ok, I'll change it

SophieDePaula:
@kyrylodolynskyi please let me know if you have any other comments

kyrylodolynskyi (approved):
I think this is fine

PR #1842, opened 2023-08-02
feat: rework log event on opent eyes

kyrylodolynskyi:


kyrylodolynskyi:
@ikseek @IdosApplitools Take a look at this, this will allow you to get rid of calling `Core.logEvent` command. I would also love all of us to add the `framework: {name: string, version: string}` property so we can keep track of what version of the framework is used. For your languages versions, I suggest `python: string` and `java: string` respectively. We will add `node: string, platform: string, arch: string` on top of whatever you send.

IdosApplitools:
@kyrylodolynskyi looks good to me. Just to make sure I follow, please confirm this payload example:

```
{
    "agentId": "eyes.sdk.java/5.63.4",
    "cwd": "/Users/idos/IdeaProjects/eyes.sdk.javascript1/java/eyes.selenium.java",
    "spec": "webdriver",
    "environment": {
        "framework": {"name": "Selenium", "version": "4.11.0"},
        "java": "11.0.12"    
    }
}
```

kyrylodolynskyi:
@IdosApplitools That's correct!

kyrylodolynskyi:
@IdosApplitools @ikseek  FYI, feel free to work on this branch. We should move in the direction where we implement features all together in the same PR.

ikseek:
I suggest making the environment a flat uniform map of components and their versions so we can add everything relevant on the side of the binding, like :
```
environment: {
  "python": "3.11",
  "selenium": "4.8",
  "appium-python-client": "2.2",
  "robotframework": "6.11"
  "robotframework-seleniumlibrary": "6.1",
}
```
And js side can extend it by adding general entries like `windows: "10.0.22"` and relevant js packages versions.

Or we could keep os and mappings of packages for different languages under their own keys if there is a risk of collision. The main proposal is to be able to track more language-specific pieces of environment information.

If you agree, I'd be happy to implement what's needed for that, if someone guides me through building and local testing process of JS packages.

kyrylodolynskyi:
@ikseek Does nested structure create any kind of problem for you? I think a nested structure is better because it is easier to search in. Does your motivation caused by the need to address the case when two frameworks are used in the same SDK?

ikseek:
@kyrylodolynskyi no, there is no problem with nesting, it's just that I wanted to make sure I can add python-specific components and their versions. Mapping is the simplest structure I can imagine, but it can of course be a list of dicts like `[{"name": "python", "version": "3.11"}, {"name":"selenium", "version":"4.8"}]`.
The thing is, I'd like to see how wide is the range of versions of certain packages that we should care to support.

kyrylodolynskyi:
Actually, on second thought, I agree with your first suggestion. So the structure will be key-value as you suggested. The only thing I want to avoid is having a single list that contains everything (dependency packages, language info, os name, ci information). How would you like this
```json
{
  "environment": {
    "dependencies": {
      "selenium": "4.8",
      "appium-python-client": "2.2",
      "robotframework": "6.11",
      "robotframework-seleniumlibrary": "6.1"
    },
    "python": "3.11"
  }
}
```

And then after core postprocessing we will have
```json
{
  "environment": {
    "dependencies": {
      "selenium": "4.8",
      "appium-python-client": "2.2",
      "robotframework": "6.11",
      "robotframework-seleniumlibrary": "6.1",
      "core": "4.0..0"
    },
    "python": "3.11",
    "node": "18.0.0",
    "platform": "win32",
    "arch": "x64",
    "ci": "GitHub Actions",
    "driverUrl": "http://localhost:4444/wd/hub"
  }
}
```

amitzur:
@kyrylodolynskyi @ikseek I'd rename `depedencies` to `versions`, like in Node.js `process.versions`. Then I suggest clients to send version numbers of everything (including Java/Python version) inside `versions`.
Then any other information is flat next to `versions`.
WDYT?

kyrylodolynskyi:
@amitzur I like that. Done ✅ 
I also added some additional information about the environment (like ram and better platform info)
@ikseek @IdosApplitools I think we agreed on the shape, so please do your part

IdosApplitools:
@kyrylodolynskyi Java is ready for testing

js/packages/core/src/open-eyes.ts
```diff
@@ -71,19 +72,13 @@ export function makeOpenEyes<TSpec extends SpecType, TDefaultType extends 'class
...
-        agentId: settings.agentId,
-        useDnsCache: settings.useDnsCache,
+        ...(settings as OpenSettings<'ufg' | 'classic'>),
         level: 'Notice',
         event: {
-          type: 'runnerStarted',
-          testConcurrency: concurrency,
-          concurrentRendersPerTest: (config as Config<SpecType, 'ufg'>)?.check?.renderers?.length ?? 1,
-          node: {version: process.version, platform: process.platform, arch: process.arch},
+          type: 'openEyes',
```
amitzur, line 78: @kyrylodolynskyi this must be communicated and coordinated with Product. The name of the event changed, and the payload changed. They currently have queries that rely on this data. Please communicate the necessary info.

PR #1844, opened 2023-08-03
feat: add the option to fail on diffs after all spec files

SophieDePaula:
[Trello card](https://trello.com/c/dTlEpkgc/1978-fujitsu-cypress-tests-taking-too-long)

To reduce execution time, we now expose a new flag `failCypressOnDiffAfterAll`, when this flag is enabled, we will not call `getResults` in the spec `after` hook, but only in the global hooks.
This will cause the spec file to pass even if there are diffs, but we'll still get an exception at the end of the execution.


SophieDePaula:
@amitzur @benbabayoff8 ,

here is the PR re the Cypress execution time we were talking about.
I named the flag `failCypressOnDiffAfterAll` as it aligns with the current flag we have `failCypressOnDiff` and it is clear from the name what's the functionality.
@amitzur I think the name you suggested `failCypressAfterAllSpecs` is also very informative but a bit too long, I am ok with using it if you think what I suggested is not clear enough.

js/packages/eyes-cypress/src/plugin/hooks.ts
```diff
@@ -15,6 +15,8 @@ declare global {
         batchId: string
         tapDirPath: string
         tapFileName: string
+        failCypressOnDiffAfterAll?: boolean
+        showLogs?: boolean
```
amitzur, line 19: where did this pop up?

js/packages/eyes-cypress/src/plugin/hooks.ts
```diff
@@ -38,10 +40,12 @@ export default function makeGlobalRunHooks({
         if (!(config as Cypress.Config).isTextTerminal) return
         const summaries = await closeManager()
 
-        let testResults
+        let testResults = []
```
amitzur, line 43: wow, the code before this change was really doing nonesense.
I think you should use more functional code style here:
```
const testResults = summaries.map(
  ({results}) => results.map(({result}) => result)
).flat()
```

js/packages/eyes-cypress/src/plugin/hooks.ts
```diff
@@ -57,6 +61,15 @@ export default function makeGlobalRunHooks({
             tapFileName: config.appliConfFile.tapFileName,
           })
         }
+        const resultConfig = {
+          showLogs: config.appliConfFile.showLogs,
+          eyesFailCypressOnDiff: config.appliConfFile.failCypressOnDiffAfterAll,
```
amitzur, line 66: This way to write code leads to confusing code IMO. Renaming a parameter like that, I mean.
We should be aware to make changes to parameter names to make them be more sensible. In this case, for example, the parameter to `printTestResults` makes sense to call `shouldThrowError`.
I'm sorry to say, this is just a drop in an ocean of spaghetti unmaintained code.

js/packages/eyes-cypress/src/browser/commands.js
```diff
@@ -66,7 +66,10 @@ Cypress.Commands.add('eyesGetResults', (args = {}) => {
   })
 })
 
-if (shouldUseBrowserHooks || Cypress.config('eyesFailCypressOnDiff')) {
+if (
+  (shouldUseBrowserHooks || Cypress.config('eyesFailCypressOnDiff')) &&
+  !Cypress.config('appliConfFile').failCypressOnDiffAfterAll
```
amitzur, line 71: How do you feel about this condition? Does it look like it would be easy to maintain?

amitzur (changes requested):
@SophieDePaula I left some inline comments. 2 general points:
1. We keep making patches in the Cypress codebase. We can't continue to do this. There are so many confusing and things I cannot explain in the code, that I don't know where to start providing feedback.
2. Not sure about the naming. `failCypressOnDiffAfterAll` sounds too verbose and too similar to `failCypressOnDiff`. There is a difference between them. `failCypressOnDiff` is responsible for the switch fail-->non-fail. `failCypressOnDiffAfterAll` is responsible to make the failure happen at a specific point. This begs the question - what happens if you set `failCypressOnDiff` to false, and `failCypressOnDiffAfterAll` to true? IMO there should be no failure, since it's set to false. This is why I think it should be called `failCypressAfterAllSpecs` or `waitForAllSpecs`, to emphasize that it's talking about **when** to do the thing.

js/packages/eyes-cypress/src/plugin/hooks.ts
SophieDePaula: @amitzur I suppose that this type was created to address the current needs of this file and does not reflect all the properties that can be added to the configuration file (`applitools.config.js`) I modified it to be closer to reality.

js/packages/eyes-cypress/src/plugin/hooks.ts
SophieDePaula: @amitzur yes, we had a bug there , I used your code instead of mine and added a test to make sure we include all test results in the tap file.

js/packages/eyes-cypress/src/plugin/hooks.ts
SophieDePaula: @amitzur I took your advice and now we check for `shouldThrowError` rather than `eyesFailCypressOnDiff`.
Let's talk about this SDK in more details in our next meeting.

js/packages/eyes-cypress/src/browser/commands.js
SophieDePaula: @amitzur I now use a variable for this condition instead.

SophieDePaula:
@amitzur I addressed your inline comments, as for your general comments:

1) I changed the name of the new flag to `failCypressAfterAllSpecs`.  The logic is if `failCypressOnDiff: false` and `failCypressAfterAllSpecs: true` then we check for diffs only after all specs are done
2) As I mentioned in one of my previous comments, let's discuss the state of this SDK in our next meeting.

js/packages/eyes-cypress/test/e2e/cypress-run-CypressFailOnDiffAfterAllSpecs.test.js
```diff
@@ -0,0 +1,54 @@
...
+
+  after(async () => {
+    fs.rmdirSync(targetTestAppPath, {recursive: true})
+  })
+
+  it('failCypressOnDiffAfterAllSpecs', async () => {
+    const config = {...applitoolsConfig, failCypressAfterAllSpecs: true}
+    fs.writeFileSync(`${targetTestAppPath}/applitools.config.js`, 'module.exports =' + JSON.stringify(config, 2, null))
+    const [err, _v] = await presult(runCypress('index-run.js', 'helloworldDiffs.js'))
+    const normalizedStdout = err.stdout.replace(/\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])/g, '')
```
amitzur, line 50: `strip-ansi`

amitzur (commented):
Much much better :)

PR #1852, opened 2023-08-06
feat: support getting userAgent externally

amitzur:
When taking the DOM snapshot separately from the `check` command, there is a need to still fetch resources with a specidied user agent. Normally, this is taken from the driver. But when there is no driver, it needs to be provided from the user.


kyrylodolynskyi (commented):
I am not sure I like this approach in terms of api. I would rather prefer it to be part of the `DomSbapshot` object, than add so niche property to `CheckSettings`

amitzur:
@kyrylodolynskyi I thought about it but then I thought it would provide more flexibility to use a different `userAgent` to fetch resources than the one used for taking the dom snapshot. WDYT?

PR #1865, opened 2023-08-09
chore: improve java release-please integration

IdosApplitools:


kyrylodolynskyi:
@IdosApplitools Makes sense to me. I like the fact that we keep the possibility to release each module independently. Can you explain why is it important to release the parent module to the maven central? Does it even have any source code at all?

IdosApplitools:
> @IdosApplitools Makes sense to me. I like the fact that we keep the possibility to release each module independently. Can you explain why is it important to release the parent module to the maven central? Does it even have any source code at all?

@kyrylodolynskyi to my understanding it could have source code but we do not have code there. The parent, as the name suggests, acts as a parent module to all the child modules, but it is a module like the rest. In each child, we specify its parent and this allows us to specify common plugins dependencies, etc.. If this parent pom is not found in central then this should cause compile errors when building a child module. 

This behavior in Maven essentially allows any module to become a parent of a different module. A parent can have another parent and that parent could have another parent of its own.

kyrylodolynskyi:
@amitzur I read about it a little bit, @IdosApplitools correct me if I am wrong. Parent pom is listed in the children's poms so maven will have to be able to resolve it. I believe this is done to be able to share dependency and plugin configs.
This is something like we have a root `package.json` and we install some shared dev deps there, the only difference is that node resolves those modules regardless and does not require to provide any reference from child to parent. The release of the parent module could be avoided only if the parent pom will not be used as a parent but only as an aggregator, which means no shared deps, but still possible to use nicer scripts to build modules in the proper order.

kyrylodolynskyi:
@IdosApplitools @amitzur This is the source I used - https://stackoverflow.com/questions/58521838/how-to-deploy-child-module-without-also-deploying-parent-module/72830959#72830959

IdosApplitools:
> @IdosApplitools what eventually happened with this initiative?

I plan to deliver this together with the generics changes

PR #1877, opened 2023-08-15
feat: show aborted tests in Eyes dashboard

SophieDePaula:
TODO:

- [x] Add test
- [x] Fix `unit` and `it` tests
- [x] Add unit tests for `renderStory` and `processResults`
- [x] Handle `processResults` when all stories fail ([Trello card](https://trello.com/c/KXs3dW5O/2013-canvaeditor-build-failures))


js/packages/eyes-storybook/src/processResults.js
```diff
@@ -15,21 +15,36 @@ function processResults({
...
     {err: result.error, title: result.result.name},
   ]);
-  errors = flatten(errors).filter(
-    ({err}) =>
-      err &&
-      !err.message.includes('detected differences') &&
-      !err.message.includes('Please approve the new baseline'),
-  );
+
+  const resultsOrErr = results.results
```
amitzur, line 31: What use case is this solving? Did we miss some errors until now? Why is a variable that holds only errors called `resultsOrErr`? Is `.stack` really the way to identify errors in this array? Why do all other places in this function process `results.summary.results` and only this place processes `results.results`?

amitzur (changes requested):
@SophieDePaula See inline comments. There is missing/wrong implementation.
In general, I would like to see unit tests for the new code in `renderStory` and `processResults`.

js/packages/eyes-storybook/src/renderStory.js
```diff
@@ -130,11 +130,20 @@ function makeRenderStory({
         throttle(async () => {
           try {
             await eyes.checkAndClose({
-              target: snapshots,
+              target: snapshots ?? {},
```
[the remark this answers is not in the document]
SophieDePaula: @amitzur the reason I was calling `check` with an empty `snapshot` was because the `storage` was empty otherwise.
Where would you call `baseEyes = await eyes.getBaseEyes()` ? I don't have an `eyes` instance.

js/packages/eyes-storybook/src/processResults.js
```diff
@@ -15,21 +15,36 @@ function processResults({
   let outputStr = '\n';
   const pluralize = utils.general.pluralize;
   let testResults = flatten(results.summary.results);
-  const unresolved = testResults.filter(r => r.result.isDifferent);
-  const passedOrNew = testResults.filter(r => r.result.status === 'Passed' || r.result.isNew);
-  const newTests = testResults.filter(r => r.result.isNew);
+  const unresolved = testResults.filter(r => r.result.isDifferent && !r.result.isAborted);
+  const passedOrNew = testResults.filter(
+    r => (r.result.status === 'Passed' || r.result.isNew) && !r.result.isAborted,
```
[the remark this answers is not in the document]
SophieDePaula: @amitzur no, but we check here if it's `Passed` or `New`, I can make this condition clearer.

js/packages/eyes-storybook/src/renderStory.js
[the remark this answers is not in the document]
SophieDePaula: @amitzur re the `API` change, I was thinking about the use experience here, if they see that some of their tests are aborted than it's a better experience to see the error immediately in the console and not start looking for the error in the probably huge logs, or even worse , not having logs enable in the first place, rerunning the test and then maybe or maybe not reproducing the same issue.

js/packages/eyes-storybook/src/processResults.js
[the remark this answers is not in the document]
amitzur: That's where I was aiming :)

js/packages/eyes-storybook/src/renderStory.js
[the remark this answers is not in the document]
SophieDePaula: @amitzur I spoke with Kyrylo and now we only call `open` and `abort` with additional logic in `abort`

js/packages/core/src/ufg/abort.ts
```diff
@@ -41,14 +44,35 @@ export function makeAbort<TSpec extends SpecType>({
...
         logger.warn('Command "abort" received an error during waiting for eyes instances in background', error)
         await error?.info?.eyes?.abort({settings, logger})
       }
-    })
+
+      // adding to storage for test results
+      uniqueRenderers.map((renderer, index) => {
+        const key = JSON.stringify(renderer)
+        storage.set(key, [...(storage.get(key) ?? []), promises[index]])
+      })
```
SophieDePaula, line 75: @kyrylodolynskyi this is what we discussed plus some additions and changes. When I called `eyes.abort` from line 61, `fetch` crashes.
Also, I had to add the test to the storage as I need to the `testResults` and I need to know which test was aborted.

js/packages/eyes-storybook/src/renderStory.js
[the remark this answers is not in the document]
SophieDePaula: Re the second point, I removed it even though I think it can only be beneficial for the user, I suppose it's not super important.

js/packages/eyes-storybook/src/processResults.js
SophieDePaula: @amitzur I changed this condition to check for the logic around our `TestResults`object structure rather than the error object structure.
[Type](https://github.com/applitools/eyes.sdk.javascript1/blob/master/js/packages/core/src/ufg/types.ts#L64) and [returned value](https://github.com/applitools/eyes.sdk.javascript1/blob/master/js/packages/core/src/ufg/get-results.ts#L19)

js/packages/core/src/ufg/abort.ts
```diff
@@ -41,14 +44,35 @@ export function makeAbort<TSpec extends SpecType>({
       }
     }
 
-    storage.forEach(async promises => {
+    if (storage.size > 0) {
+      storage.forEach(async promises => {
+        try {
+          const {eyes} = await Promise.race(promises)
```
amitzur, line 50: @SophieDePaula don't duplicate this code. Try to organize the code to write it once.

js/packages/eyes-storybook/src/processResults.js
```diff
@@ -15,21 +15,38 @@ function processResults({
...
+    })
+    .filter(Boolean);
+
+  errors = flatten(errors)
+    .filter(
+      ({err}) =>
+        err &&
+        !err.message.includes('detected differences') &&
+        !err.message.includes('Please approve the new baseline') &&
+        !err.message.includes('is failed! See details at'),
```
amitzur, line 47: we should really try to find another way to do this.

js/packages/eyes-storybook/test/e2e/snapshots/storybook-aborted.e2e.test.json
```diff
@@ -0,0 +1,22 @@
+{
+  "stdout": [
+    "Using @applitools/eyes-storybook version <version>.",
+    "",
+    "",
+    "[EYES: TEST RESULTS]:",
+    "",
+    "Single category: Single story [Chrome] [640x480] - Aborted",
+    "",
+    "Single category: Single story - Failed Failed to get story data for Single category: Single story, test was aborted",
```
amitzur, line 10: This `Failed Failed` output doesn't look good. Let's try to get a better result.

js/packages/eyes-storybook/test/unit/renderStory.test.js
```diff
@@ -546,6 +551,54 @@ describe('renderStory', () => {
...
     expect(results).to.eql(expected);
   });
+
+  it('should abort if snapshot is undefined', async () => {
+    const openEyes = async openParams => {
+      const result = {openParams};
+      return {
+        checkAndClose: async checkParams => (result.checkParams = checkParams),
+        getResults: async () => result,
+        abort: async () => {
```
amitzur, line 561: @SophieDePaula I wrote a comment about this code style already in this PR. Either do `return Promise.reject(...)` or better yet `throw new Error(...)`.

amitzur (changes requested):
See inline comments

js/packages/eyes-storybook/test/e2e/snapshots/storybook-aborted.e2e.test.json
SophieDePaula: @amitzur I added a `.` after the first `Failed`, so now it's `Single category: Single story - Failed. Failed to get ...`

js/packages/eyes-storybook/src/processResults.js
SophieDePaula: @amitzur I rearranged that to :

```
  const errMessagesToExclude = [
    'detected differences',
    'Please approve the new baseline',
    'is failed! See details at',
  ];
  errors = flatten(errors)
    .filter(({err}) => err && !errMessagesToExclude.some(msg => err.message.includes(msg)))
    .concat(flatten(resultsErr));
```

js/packages/core/src/ufg/abort.ts
SophieDePaula: fixed

js/packages/eyes-storybook/test/unit/renderStory.test.js
SophieDePaula: Fixed

js/packages/core/src/ufg/abort.ts
```diff
@@ -41,14 +44,39 @@ export function makeAbort<TSpec extends SpecType>({
...
-    storage.forEach(async promises => {
+    if (storage.size > 0) {
+      storage.forEach(async promises => {
+        await waitForPromisesAndAbort({promises, logger})
+      })
+    } else if (settings.renderers) {
+      const uniqueRenderers = uniquifyRenderers(settings.renderers ?? [])
+
+      const promises = uniqueRenderers.map(async ({properties, ...renderer}) => {
+        const [baseEyes] = await eyes.getBaseEyes({settings: {renderer, properties}, logger: logger})
```
amitzur, line 55: @kyrylodolynskyi why does `getBaseEyes` return an array?
@SophieDePaula please change `logger: logger` to just `logger`.

js/packages/core/src/ufg/abort.ts
```diff
@@ -41,14 +44,39 @@ export function makeAbort<TSpec extends SpecType>({
...
+
+        // adding to storage for test results
+        const key = JSON.stringify(renderer)
+        storage.set(key, [new Promise(resolve => resolve({renderer, eyes: baseEyes}))])
```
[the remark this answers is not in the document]
SophieDePaula: Ah yes, we discussed this

js/packages/eyes-storybook/src/processResults.js
```diff
@@ -15,31 +15,51 @@ function processResults({
...
-      err &&
-      !err.message.includes('detected differences') &&
-      !err.message.includes('Please approve the new baseline'),
-  );
 
-  const hasResults = unresolved.length || passedOrNew.length;
+  const resultsErr = results.results
+    .map(result => {
+      if (
+        !(Array.isArray(result.resultsOrErr) && utils.types.has(result.resultsOrErr[0], 'renderer'))
```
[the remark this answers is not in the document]
SophieDePaula: @amitzur it can either be an `error` or [an Array](https://github.com/applitools/eyes.sdk.javascript1/blob/master/js/packages/core/src/ufg/get-results.ts#L19) of [TestResult](https://github.com/applitools/eyes.sdk.javascript1/blob/master/js/packages/core/src/ufg/types.ts#L64)

js/packages/eyes-storybook/src/processResults.js
```diff
@@ -130,7 +150,9 @@ function testResultsOutput(results, warnForUnsavedNewTests) {
   sortedTestResults.forEach(result => {
     const storyTitle = `${result.result.name} [${result.result.hostApp}] [${result.result.hostDisplaySize.width}x${result.result.hostDisplaySize.height}] - `;
 
-    if (result.result.isNew) {
+    if (result.result.isAborted) {
```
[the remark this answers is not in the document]
SophieDePaula: Sure, I guess it make sense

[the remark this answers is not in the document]
SophieDePaula: @amitzur actually when I look at the code I see that if `isAborted` is not the first one, then I'll have to add ` && !result.result.isAborted` to `isNew` and `!result.result.isDifferent`.
So leaving it as is make more sense to me.

js/packages/core/src/ufg/abort.ts
kyrylodolynskyi: Because there maybe many eyes if it was started like this, when you open automation eyes you can pass it an array of base eyes, then when you call `getBaseEyes` you will get them back.

js/packages/eyes-storybook/src/processResults.js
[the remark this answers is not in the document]
amitzur: But the output to the user is more important than the code, won't you agree?

js/packages/core/src/ufg/abort.ts
```diff
@@ -41,14 +44,39 @@ export function makeAbort<TSpec extends SpecType>({
...
-    storage.forEach(async promises => {
+    if (storage.size > 0) {
+      storage.forEach(async promises => {
+        await waitForPromisesAndAbort({promises, logger})
+      })
+    } else if (settings.renderers) {
+      const uniqueRenderers = uniquifyRenderers(settings.renderers ?? [])
+
+      const promises = uniqueRenderers.map(async ({properties, ...renderer}) => {
+        const [baseEyes] = await eyes.getBaseEyes({settings: {renderer, properties}, logger})
```
amitzur, line 55: @kyrylodolynskyi why does `getBaseEyes` return an array?

SophieDePaula:
@amitzur I addressed the scenario where all test fails and we don't have test results (AKA ` Cannot read properties of undefined (reading 'isDifferent')`).
Please let me know if you have any additional comments.

amitzur (changes requested):
See inline comments. I'm not 100% done yet, I still need to go over `processResults.js`

js/packages/eyes-storybook/test/e2e/snapshots/storybook-aborted.e2e.test.json
```diff
@@ -0,0 +1,27 @@
...
+    "Using @applitools/eyes-storybook version <version>.",
+    "",
+    "",
+    "See details at <some_url>",
+    "",
+    "[EYES: TEST RESULTS]:",
+    "",
+    "Single category: Single story [Chrome] [640x480] - Aborted",
+    "",
+    "Single category: Single story - Failed. Failed to get story data for Single category: Single story, test was aborted",
```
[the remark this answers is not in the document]
SophieDePaula: @amitzur I have a [test in `it`](https://github.com/applitools/eyes.sdk.javascript1/blob/storybook-abort-failed-stories/js/packages/eyes-storybook/test/it/snapshots/processResults.test.json#L147) that shows the output of two tests, one `Passed` and on `Aborted`.

As for the second part of your question, it is showed twice, once with the status (`Passed` / `Unresolved` / `Aborted`) and we also specify the `browserInfo` as we can have some stories that have different statuses on different environments. The second time with show the `title` and the reason for the failure.

js/packages/eyes-storybook/test/unit/renderStory.test.js
```diff
@@ -546,6 +551,91 @@ describe('renderStory', () => {
...
+        story,
+        snapshots: undefined,
+        url: 'url',
+      }),
+    );
+
+    expect(err.message).to.equal('test was aborted');
+  });
+
+  it('should throw an error when openEyes fails', async () => {
```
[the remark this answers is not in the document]
SophieDePaula: @amitzur it is not , but I didn't see a test for it and I wanted to make sure this woks properly as it is related to the [second trello card](https://trello.com/c/KXs3dW5O/2013-canvaeditor-build-failures) fix that is included in this PR.

js/packages/eyes-storybook/test/unit/renderStory.test.js
```diff
@@ -546,6 +551,91 @@ describe('renderStory', () => {
...
+
+    const renderStory = makeRenderStory({
+      ...defaultSettings,
+      logger,
+      performance,
+      timeItAsync,
+      openEyes,
+    });
+
+    const baseStory = {
```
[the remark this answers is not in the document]
SophieDePaula: @amitzur I only need the `baselineName` and an empty `config`, I'll remove everything else.

js/packages/eyes-storybook/test/it/processResults.test.js
```diff
@@ -314,17 +319,7 @@ describe('processResults', () => {
...
-            status: 'Passed',
-            isDifferent: false,
-            name: 'My Component | Button2',
-            hostApp: 'Chrome',
-            hostDisplaySize: {width: 10, height: 20},
-            appUrls: {batch: 'https://eyes.com/results'},
-          },
-          new Error('another error messgae !'),
-        ],
+        resultsOrErr: [new Error('another error messgae !')],
```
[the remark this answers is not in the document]
SophieDePaula: @amitzur I done what we spoke about , I threw an exception for one of the renderers in `render.ts` in `checkRenderResults` and the result was that we fail all environments. Therefore, the scenario that we have an array with one entry of `TestResult` and one with error is not possible.

js/packages/eyes-storybook/src/processResults.js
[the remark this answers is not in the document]
amitzur: @SophieDePaula if `result.resultsOrErr` is an array, then it cannot be an error, right? So why `utils.types.has(result.resultsOrErr[0], 'renderer')`?

js/packages/eyes-storybook/src/processResults.js
```diff
@@ -14,26 +14,41 @@ function processResults({
...
+      ) {
+        return {err: result.resultsOrErr, title: result.title};
+      }
+    })
+    .filter(Boolean);
 
-  const hasResults = unresolved.length || passedOrNew.length;
+  errors = flatten(errors)
+    .filter(({err}) => err && !errMessagesToExclude.some(msg => err.message.includes(msg)))
+    .concat(flatten(resultsErr));
```
amitzur, line 46: @SophieDePaula in this `concat` aren't you probably creating duplicates?

js/packages/eyes-storybook/src/processResults.js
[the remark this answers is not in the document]
SophieDePaula: @amitzur that's correct, I removed it.

js/packages/eyes-storybook/src/processResults.js
SophieDePaula: @amitzur while looking into this I found, that if we get an error in the `TestResult` (for example when a render fails), the structure is different than what we expected. First, we might not have `result` anymore, also we can only get the `testName` from [`userTestId`](https://github.com/applitools/eyes.sdk.javascript1/blob/master/js/packages/core/src/open-eyes.ts#L55) in `eyes` so I fixed that . As for dups, we use `uniq` on `title` which is `baselineName` in both cases so we don't print dup errors per story.

amitzur: Ok, `userTestId` is 100% not a good way to get the test name. Please try to find another way.

amitzur: @SophieDePaula I don't understand. What does `eyes.getResults` return in the case of a render failure?

js/packages/eyes-storybook/test/it/snapshots/processResults.test.json
```diff
@@ -54,8 +54,8 @@
     "My Component | Button1 [Chrome] [10x20] - Aborted",
     "My Component | Button2 [Chrome] [10x20] - Aborted",
     "",
-    "My Component | Button1 - Failed. Error: some error messgae !",
-    "My Component | Button2 - Failed. Error: another error messgae !",
+    "My Component | Button1 - Failed. some error messgae !",
```
amitzur, line 57: @SophieDePaula what caused this change in the output?

js/packages/eyes-storybook/src/processResults.js
SophieDePaula: @amitzur when we fail to render it seems we add `info` to the `Error` object, [see this](https://github.com/applitools/eyes.sdk.javascript1/blob/master/js/packages/core/src/ufg/check-and-close.ts#L310).

Also, why is `userTestId` is 100% not a good way to get the test name? I don't have the test name unfortunatly if the render fails and I see that in `open-eyes` that `userTestId` is  `settings.userTestId ??= ${settings.testName}--${utils.general.guid()}`. If this is not good enough I am afraid that I'll have to make changes to `core` to include the test name in the returned object.

amitzur: @SophieDePaula the reason it's not good is because it's not an API, it's an implementation detail. Changing that line in core will break your code and there is no way to detect it.

js/packages/eyes-storybook/src/processResults.js
```diff
@@ -15,12 +15,14 @@ function processResults({
   let outputStr = '\n';
   const pluralize = utils.general.pluralize;
   const testResults = flatten(results.summary.results).filter(Boolean);
-  const unresolved = testResults.filter(r => r.result.isDifferent && !r.result.isAborted);
+  const unresolved = testResults.filter(
```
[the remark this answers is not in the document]
SophieDePaula: @amitzur I now filter results that doesn't have `result` property. However, I needed to create another var for testResultsWithErrors as they are no longer included in `testResults`.

js/packages/eyes-storybook/src/processResults.js
amitzur: I think we should make `eyes.test` available in `error.info`, and `VisualTest` should have all the OpenSettings.
@kyrylodolynskyi does that sound right?

js/packages/eyes-storybook/test/it/snapshots/processResults.test.json
SophieDePaula: @amitzur because now that the `Error` is not in an Array (as it should be) , then we have `.message` and we don't just  stringify the error object, [see this](https://github.com/applitools/eyes.sdk.javascript1/blob/master/js/packages/eyes-storybook/src/processResults.js#L53)

js/packages/eyes-storybook/src/processResults.js
SophieDePaula: @amitzur I added `testName` to `VisualTest` type and now I use it in `processResults`.
I didn't add all `OpenSettings` as it includes many properties and I didn't want to start mapping everything before we get to a final decision  
https://github.com/applitools/eyes.sdk.javascript1/blob/storybook-abort-failed-stories/js/packages/core-base/src/server/requests.ts#L171

PR #1879, opened 2023-08-16
chore: update release-please-config and config location

SophieDePaula:


amitzur:
@SophieDePaula is this because Cypress behaves like universal SDK's rather than JS SDK's?

SophieDePaula:
> @SophieDePaula is this because Cypress behaves like universal SDK's rather than JS SDK's?

@amitzur , no, we also have `generic` in `tests` in our JS SDK's. For example see [puppteer](https://github.com/applitools/eyes.sdk.javascript1/blob/08c98f9ea2771d42d0976b4928bab50f45cee4e5/release-please-config.json#L1069).

I think it was just missed when we updated the way we are now running our coverage tests.

amitzur:
@SophieDePaula thanks. I suggest consistency - all other SDK's have `test/generic/config.mjs`. I don't want Cypress to have `test/coverage/generic/config.mjs`.

PR #1886, opened 2023-08-22
feat(js/eyes-storybook,js/core): add support for passing a DOM mapping file to deobfuscate RCA

tourdedave:
[Trello](https://trello.com/c/4S87RPc0)


js/packages/core-base/src/types.ts
```diff
@@ -278,6 +278,7 @@ export interface CheckSettings<TRegion = Region> extends ImageSettings<TRegion>
   forceMismatch?: boolean
   /** @internal */
   forceMatch?: boolean
+  domMapping?: Buffer | URL | string
```
amitzur, line 281: This type can be kept if you move the download functionality to core (see comment in validateAndPopulateConfig)

js/packages/eyes-storybook/src/validateAndPopulateConfig.js
```diff
@@ -67,6 +68,26 @@ async function validateAndPopulateConfig({config, packagePath = '', logger = mak
     }
   }
 
+  if (config.domMapping) {
```
amitzur, line 71: It makes more sense to me that core will handle the download. You already [set the type](https://github.com/applitools/eyes.sdk.javascript1/pull/1886/files#diff-209bedd00374a3221f48b91d395004991fec556c460a9810b2f5e0d4868ae2daR281) of `domMapping` to `Buffer | URL | string`, so you might as well provide this functionality from core.

As a side note, if you did keep this implementation then I would advise to change `domMapping` type in core to `Buffer`. The only reason to add `| URL | string` was because you piggyback on it to save the uploaded URL. But that is wrong since you didn't really support receiving a string or URL from the outside.
In general, try to keep types tight so the interface is clear.

The only challenge is to show a warning in Storybook's console if there was an issue. We don't write to console from core, but for universal SDK's we have a mechanism for Core to instruct clients to log something. Maybe we can use something like that (@kyrylodolynskyi wdyt?)

js/packages/eyes-storybook/src/validateAndPopulateConfig.js
```diff
@@ -14,6 +14,7 @@ const {isIE} = require('./shouldRenderIE');
 const {makeLogger} = require('@applitools/logger');
 const findNpmModuleCommandPath = require('./findNpmModuleCommandPath');
 const {resolve} = require('path');
+const fetch = require('node-fetch');
```
amitzur, line 17: The download implementation should be similar to downloading UFG resources. It should at least take into consideration a proxy configuration. This is why I don't suggest to use `node-fetch` directly, but use some preconfigured `req`.

js/packages/eyes-storybook/test/e2e/dom-mapping.e2e.test.js
```diff
@@ -0,0 +1,45 @@
...
+    //const stderr = err ? err.stderr : result.stderr;
+    const output = stdout
+      .replace(/Total time\: \d+ seconds/, 'Total time: <some_time> seconds')
+      .replace(
+        /See details at https\:\/\/.+.applitools.com\/app\/test-results\/.+/g,
+        'See details at <some_url>',
+      )
+      .replace(version, '<version>')
+      .replace(/\d+(?:\.\d+)+/g, '<browser_version>');
+    await snap(output, 'wait before capture config');
```
amitzur, line 43: what is this test verifying?
I think it's hard to test this e2e in storybook. The most you could test is in `eyesStorybook.it.test.js` to get the test results, then download the session data from Eyes and see that `domMappingUrl` is populated.

amitzur (changes requested):
See inline comments

js/packages/core-base/src/types.ts
kyrylodolynskyi, line 281: This shouldn't be part of the `CheckSettings`, but rather `Target` same as `image` and `dom`.

kyrylodolynskyi (changes requested):

js/packages/core-base/src/types.ts
amitzur: I was thinking this as well, initially. But the fact is that that it's not an output of the page. It's more of a settings sort of input.

kyrylodolynskyi: It should not be an output of the page, but this is clearly an output of something. `Target` describes something on which the backend will run a verification. I agree that this is also not exactly a definition of "dom mapping", but I still think that this is closer to `Target` than `CheckSettings`. In general I think this is a `Target` and not a setting, because there is nothing to set, it is a piece of information about the page that could exist or not, also, keep in mind that this information is needed only in combination with dom capture that also provided in `Target`.

js/packages/eyes-storybook/src/validateAndPopulateConfig.js
kyrylodolynskyi: @amitzur Completely agree with you about types. I think the implementation here should be the same as for the image provided as url, which means `core-base` should download it and then upload it to the proper location.
Regarding the logs, the core writes logs to the console, there is `logger.console` functionality, this is how we write warnings when we fail to resize the viewport and in other cases.

js/packages/eyes-storybook/test/e2e/dom-mapping.e2e.test.js
tourdedave: Yeah - right now the test isn't verifying anything helpful. I was just using it to spot check in the logs that the value in the payload was getting correctly populated. The next step would be to do as you said.

js/packages/core-base/src/types.ts
tourdedave: Just to clarify - if putting it on Target is the direction, then wouldn't that have implications for the API used in storybook and what gets handled in core (not just core-base)?

E.g.,
```
await eyes.checkAndClose({
  target: {snapshots, domMapping},
  settings: {...checkParams, ...closeSettings},
});
```

Instead of what it is currently...
```
await eyes.checkAndClose({
  target: snapshots
  settings: {...checkParams, ...closeSettings}, // which has checkParams.domMapping
});
```

If so, can I get some guidance on where the type change(s) should happen? I tried a few things that seemed intuitive, but running into friction which makes me challenge my assumption about what you meant.

Following the change above to its conclusion, this is the changeset I assumed we'd need:

```
diff --git a/js/packages/core/src/ufg/check-and-close.ts b/js/packages/core/src/ufg/check-and-close.ts
index d214ba692..94bd8559b 100644
--- a/js/packages/core/src/ufg/check-and-close.ts
+++ b/js/packages/core/src/ufg/check-and-close.ts
@@ -163,7 +163,11 @@ export function makeCheckAndClose<TSpec extends SpecType>({

       await cleanupGeneratedSelectors?.()
     } else {
-      snapshots = !utils.types.isArray(target) ? Array(uniqueRenderers.length).fill(target) : target
+      if (utils.types.isObject(target) && utils.types.has(target, 'snapshots')) {
+        snapshots = !utils.types.isArray(target?.snapshots) ? Array(uniqueRenderers.length).fill(target?.snapshots) : target?.snapshots
+      } else {
+        snapshots = !utils.types.isArray(target) ? Array(uniqueRenderers.length).fill(target) : target
+      }
       snapshotUrl = utils.types.has(snapshots[0], 'url') ? snapshots[0].url : undefined
     }
     regionToTarget ??= (elementReferenceToTarget as Selector) ?? (settings.region as Region)
@@ -279,7 +283,7 @@ export function makeCheckAndClose<TSpec extends SpecType>({
           }

           await baseEyes.checkAndClose({
-            target: {...baseTarget, isTransformed: true},
+            target: {...baseTarget, isTransformed: true, domMapping: target?.domMapping},
             settings: baseSettings,
             logger: rendererLogger,
           })
diff --git a/js/packages/core/src/ufg/types.ts b/js/packages/core/src/ufg/types.ts
index 0921f2f32..e6bdaaca3 100644
--- a/js/packages/core/src/ufg/types.ts
+++ b/js/packages/core/src/ufg/types.ts
@@ -16,7 +16,7 @@ import {

 export * from '../automation/types'

-export type SnapshotTarget = MaybeArray<DomSnapshot> | MaybeArray<AndroidSnapshot> | MaybeArray<IOSSnapshot>
+export type SnapshotTarget = MaybeArray<DomSnapshot> | MaybeArray<AndroidSnapshot> | MaybeArray<IOSSnapshot> | {snapshots: MaybeArray<DomSnapshot>, domMapping?: string}
 export type Target<TSpec extends SpecType> = SnapshotTarget | AutomationCore.Target<TSpec>

 export interface Core<TSpec extends SpecType> extends AutomationCore.Core<TSpec> {
```

But when trying to build, I get this error:

```
src/ufg/check-and-close.ts:286:78 - error TS2339: Property 'domMapping' does not exist on type 'ImageTarget | SnapshotTarget | { domMapping?: string | undefined; } | Driver<TSpec> | NonNullable<TSpec["driver"]>'.
  Property 'domMapping' does not exist on type 'DomSnapshot'.

286             target: {...baseTarget, isTransformed: true, domMapping: target?.domMapping},
                                                                                 ~~~~~~~~~~


Found 1 error in src/ufg/check-and-close.ts:286
```

amitzur: Good question, waiting for @kyrylodolynskyi 's answer

js/packages/core-base/src/utils/transform-target.ts
```diff
@@ -13,6 +13,15 @@ export async function transformTarget({
 }): Promise<ImageTarget> {
   if (target.isTransformed) return target
   if (target.image instanceof URL) target.image = target.image.href
+  if (utils.types.isString(target.domMapping)) {
+    const str = target.domMapping
+    if (utils.types.isHttpUrl(str)) {
+      const response = await req(target.domMapping, {proxy: settings?.autProxy})
+      target.image = Buffer.from(await response.arrayBuffer())
```
amitzur, line 20: @kyrylodolynskyi you mean `target.domMapping`

amitzur:
@kyrylodolynskyi I previously didn't have a strong opinion about where `domMapping` should be. But now that I see the implementation when it's part of `Target`, then I like it less. You only handle the case that the snapshots are passed to `check`. What about a future case where you want to pass both a `domMapping` **and** take the DOM snapshots? It doesn't sit well to me that `driver` and `snapshots` are both possible values for the `target` argument. I always find it hard to interpret `driver` as `target`, since it's not the actual target - it's just a tool to get the target.

Apart of the above, I really don't have a strong opinion about where `domMapping` should be - I agree with both reasoning. But as long as it makes sense in the code.

js/packages/core-base/src/utils/transform-target.ts
kyrylodolynskyi: Why me? This is not something I wrote

kyrylodolynskyi:
@amitzur OK, I see what you mean. It makes sense, putting `domMapping` in the target would require a bigger change than I thought in the `Target` type.
It still doesn't feel like a setting to me, there previously was a clear separation between driver and other types, because a driver was the ultimate type that could extract any data we may need. This feature breaks the assumption.
Re your problem with interpreting a driver as a target, I think the misunderstanding lies in your understanding of what target means here. For me target is a representation of the testing environment, this consolidate all of the different variations of this type:
1. image + dom + other metadata is obviously a representation of what we are going to test, it also what the backend need to run a test
2. dom snapshot is a representation of a page that allow us to get the first type through the grid api
3. driver is an ultimate representation that allows you to get the first type directly or take a dom snapshot and get the first type through the second type path

Under this paradigm, I think that `domMapping` falls under the target definition, the only challenging part here is that it couldn't be taken on our own using driver.

js/packages/core-base/src/types.ts
kyrylodolynskyi: I am not sure I know how to answer all of it in text. I was suggesting the `core-base` api, so it was clear you just added another property to the object already full of properties. here is a little bit more complicated, because the `target` in `ufg-core` is `DomSnapshot` of course you can add the new property to the dom-snapshot itself, but I like your direction more. 
I pushed a commit that solved the problem. I also renamed `snapshots` to `snapshot`, because the plural name usually means array input, but we have "maybe-array" input.

js/packages/core-base/src/utils/transform-target.ts
amitzur: Oh, right. @tourdedave :)

tourdedave: 👌 copy/paste error

tourdedave:
@amitzur is this the property we need to make sure that Yotam includes in the back-end? Just want to confirm before following up with him.

https://github.com/applitools/eyes.sdk.javascript1/blob/js/feat/dom-mapping/js/packages/eyes-storybook/test/e2e/dom-mapping.e2e.test.js#L57

amitzur:
@tourdedave yes, `domMappingUrl` on `actualAppOutput` is exactly it.

amitzur:
@kyrylodolynskyi I can see the reasoning of your explanation of `target`. But I think it gives really wide interpretation of what `target` is, and IMO too wide.
It's true - in most cases you would not need a driver if you already provide the "target" (e.g. image+dom). But jumping from that to actually deciding that they are mutually exclusive - I think that's too big a jump.
I agree that conceptually, `domMapping` is more a part of the target than a setting. But for being able to move forward now, I suggest we add it to `BaseCore.CheckSettings`.

kyrylodolynskyi:
@tourdedave I am okay with the direction @amitzur suggested.

js/packages/core-base/src/utils/transform-target.ts
```diff
@@ -13,6 +13,15 @@ export async function transformTarget({
 }): Promise<ImageTarget> {
   if (target.isTransformed) return target
   if (target.image instanceof URL) target.image = target.image.href
+  if (utils.types.isString(settings?.domMapping)) {
```
amitzur, line 16: @tourdedave the type for `domMapping` is `Buffer | URL | string`. What about the case that it's a `Buffer` or `URL`?

js/packages/eyes-storybook/test/e2e/dom-mapping.e2e.test.js
```diff
@@ -0,0 +1,60 @@
...
+      },
+      logger,
+      performance,
+      timeItAsync,
+      outputStream: stream,
+    });
+
+    results = results.results.flatMap(r => r.resultsOrErr);
+    for (const testResults of results) {
+      const sessionUrl = `${testResults.apiUrls.session}?format=json&AccessToken=${testResults.secretToken}&apiKey=${process.env.APPLITOOLS_API_KEY}`;
```
amitzur, line 53: There is a utility function `getTestInfo` in `test-utils` for this

js/packages/eyes-storybook/test/e2e/dom-mapping.e2e.test.js
```diff
@@ -0,0 +1,60 @@
...
+
+      const session = await fetch(sessionUrl).then(r => r.json());
+      const [actualAppOutput] = session.actualAppOutput;
+      expect(actualAppOutput).to.have.property(domMappingUrl);
```
amitzur, line 57: I would go farther to actually download the dom mapping file and see that it's the content of `dom-mapping.json`. You might need assistance from Yotam to figure how to make the file accessible to you.

js/packages/core-base/src/utils/transform-target.ts
amitzur: I'm missing the unit tests for `transformTarget` for this case.

js/packages/eyes-storybook/test/fixtures/dom-mapping/preview-body.html
amitzur, line 1: you can avoid this file and render the story like so:
```
return <>
  <div class="abcd qwerty">hi there</div>
  <style>
    .qwerty {
      --zxcv: blue;
    }
    .abcd {
      padding: 60px;
      background: var(--zxcv);
    }
  </style>
</>
```

js/packages/eyes-storybook/test/fixtures/dom-mapping/index.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+
+storiesOf('dom-mapping', module).add('dom-mapping page', () => {
```
amitzur, line 4: The content here doesn't really need to be correlated with the dom mapping file. The SDK has no interest in it.
However I see value to do it like this, to demonstrate the use case that this is solving.
I would add a comment about it here - that this content is just for demonstration purposes.

amitzur (changes requested):
See inline comments.
One general comment about `transformTarget` -
I agree that what is added in this PR needs to be done as part of `transformTarget`, but the name is not suitable anymore since you are actually modifying `settings`.
I think `transformTarget` is a confusing name in general, without this PR I would have called it `transformImageTarget` or `prepareImageTarget`, but now with this change I would rename it to `prepareTargetAndSettings`. @kyrylodolynskyi WDYT?

js/packages/eyes-storybook/test/e2e/dom-mapping.e2e.test.js
tourdedave: Right. I copied the test from another e2e test. I'll see about cleaning them all up.

js/packages/core-base/src/utils/transform-target.ts
tourdedave: I don't see unit tests for this. I think you meant integration tests. I'll add it.

tourdedave:
> See inline comments.
> One general comment about `transformTarget` -
> I agree that what is added in this PR needs to be done as part of `transformTarget`, but the name is not suitable anymore since you are actually modifying `settings`.
> I think `transformTarget` is a confusing name in general, without this PR I would have called it `transformImageTarget` or `prepareImageTarget`, but now with this change I would rename it to `prepareTargetAndSettings`. @kyrylodolynskyi WDYT?

I think `transform` is a better word than `prepare`. What about `transformTargetAndSettings`?

js/packages/eyes-storybook/test/e2e/dom-mapping.e2e.test.js
tourdedave: Okay. I added this to my request to Yotam in the Trello card. Waiting for it to get prioritized on his board.

amitzur:
@tourdedave `transformTargetAndSettings` is just fine for me.

js/packages/core-base/src/utils/transform-target.ts
```diff
@@ -13,6 +13,15 @@ export async function transformTarget({
...
+    const str = settings?.domMapping
+    if (utils.types.isHttpUrl(str)) {
+      const response = await req(settings!.domMapping, {proxy: settings?.autProxy})
+      settings!.domMapping = Buffer.from(await response.arrayBuffer())
```
amitzur, line 20: @tourdedave how do you know that `settings` exists? Using `!` is a red flag. If it's really not optional, then change the input type for it not to be optional

js/packages/core-base/src/utils/transform-target.ts
```diff
@@ -13,6 +13,15 @@ export async function transformTarget({
...
+    } else {
+      settings!.domMapping = await fs.readFile(settings!.domMapping)
```
amitzur, line 22: @tourdedave also here

js/packages/core-base/src/utils/transform-target.ts
```diff
@@ -13,6 +13,15 @@ export async function transformTarget({
 }): Promise<ImageTarget> {
   if (target.isTransformed) return target
   if (target.image instanceof URL) target.image = target.image.href
+  if (utils.types.isString(settings?.domMapping)) {
+    const str = settings?.domMapping
+    if (utils.types.isHttpUrl(str)) {
+      const response = await req(settings!.domMapping, {proxy: settings?.autProxy})
```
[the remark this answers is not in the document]
amitzur: @tourdedave using `autProxy` directly like this is not correct.
See the usage of `autProxy` in the SDK: https://github.com/applitools/eyes.sdk.javascript1/blob/372cb96b905a0661c36e2fa10a7855208fb55bb0/js/packages/ufg-client/src/resources/fetch-resource.ts#L82

Ideally that logic would be extracted to a utility function. Depends on how much you want to invest in it. I think it's an edge case.

js/packages/core-base/src/utils/transform-target.ts
tourdedave: I know that `settings` exists because of the entry into the conditional ([here](https://github.com/applitools/eyes.sdk.javascript1/blob/js/feat/dom-mapping/js/packages/core-base/src/utils/transform-target.ts#L16)). It checks that `settings` exists and that `settings.domMapping` is a string. The compiler thinks that `settings` could potentially be `undefined` but we know that to not be the case. Using `!` is an approach I've seen used in other parts of our code base when we know that the thing we want exists but the compiler isn't satisfied.

`settings` is optional for the function, but inside of this conditional block, it must exist -- I'm just helping the compiler along. Is there a different approach I should be taking?

@kyrylodolynskyi anything to add?

js/packages/core-base/src/utils/transform-target.ts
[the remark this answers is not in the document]
tourdedave: I'll defer to @kyrylodolynskyi on this one, since I'm just doing the same thing as was done for downloading the image.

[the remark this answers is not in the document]
amitzur: @tourdedave where is `autProxy` used for downloading the image?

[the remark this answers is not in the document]
tourdedave: https://github.com/applitools/eyes.sdk.javascript1/blob/js/feat/dom-mapping/js/packages/core-base/src/utils/transform-target.ts#L28

js/packages/core-base/src/utils/transform-target.ts
tourdedave: 👌

js/packages/core-base/src/utils/transform-target.ts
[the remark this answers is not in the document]
amitzur: well, that's not good eithre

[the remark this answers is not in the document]
tourdedave: 😂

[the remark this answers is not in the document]
tourdedave: `proxy` doesn't exist on `CoreBase.ImageMatchSettings`, only `autProxy`. Also - it's type is just `Proxy` ([link](https://github.com/applitools/eyes.sdk.javascript1/blob/js/feat/dom-mapping/js/packages/core-base/src/types.ts#L239)). Not the amended type like in ufg-client ([link](https://github.com/applitools/eyes.sdk.javascript1/blob/372cb96b905a0661c36e2fa10a7855208fb55bb0/js/packages/ufg-client/src/resources/fetch-resource.ts#L27)).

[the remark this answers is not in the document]
kyrylodolynskyi: We use `autProxy` to download resources for the test, the image is a resource in this case, plus `autProxy` populated from the proxy if it is missing.

kyrylodolynskyi:
I just want to share that I don't like where it went. With an attempt to skip some steps, it deviated from an original idea that I believe was thoughtful enough. I not always can restore all of the deep meanings from the first glance, but now I see it. `CheckSettings` is not meant to be transformed, it is supposed to be a simple structure that describes what to do with the target (it could be done by us or by the server), since it is a simple structure it is not meant to be transformed. Also, settings shouldn't cause any additional processing of them like (downloading from the internet or reading from a file). I really think my initial thought that this thing has to be part of the target is very valid and was the right thing to do in the long term. 
Because with this PR merged we will have a function that transforms everything, it may not be a big deal right now, but for the next feature and feature after that, I see how other stuff in the settings going to be transformed randomly here and there. This will happen because there is already a precedent, and no one is going to remember that it was done as an attempt to cut corners.

amitzur:
I'm not sure I relate to this perspective. The important thing to enforce IMO is that transformations of user input should occur as soon as possible. I don't really mind that it sits on `settings` or `target`.
In this instance, for example, I would even say that `domMapping` is a user-facing feature, and actually `domMappingUrl` is the right property for the internal `CheckSettings` that is passed around inside core/core-base.

A relevant point to bring up is that sometimes there is even implicit transformation that takes place, which is hard to track. The example is `autProxy` mentioned above.
Dave pointed out correctly that it's used as `Proxy` type in `core-base`, as opposed to `Proxy & {...}` (let's call it `AUTProxy`). in UFG core. How can it happen that user input `autProxy` which is `AUTProxy` is "transformed" into a `Proxy` type? It happens because `toBaseCheckSettings` treats `settings` as `AutomationCore.CheckSettings`, so it doesn't know about `AUTProxy`. But the actual value could be of type `AUTProxy`. It works seamlessly because `AUTProxy` can be narrowed to `Proxy`. But this is also an implicit sort of transformation of a setting that the user passed.

tourdedave:
So where does this leave us @amitzur @kyrylodolynskyi?

amitzur:
@tourdedave now that the backend is ready, let's ship this without making any additional big changes.
Please note the instructions from Yotam in the card, on how to verify the output.
To address @kyrylodolynskyi 's concern, we can add a comment in the transform function that references this discussion, and calls for not extending this function into a general input transformation point.

tourdedave:
> @tourdedave now that the backend is ready, let's ship this without making any additional big changes.
> Please note the instructions from Yotam in the card, on how to verify the output.
> To address @kyrylodolynskyi 's concern, we can add a comment in the transform function that references this discussion, and calls for not extending this function into a general input transformation point.

@amitzur I updated `getTestDom` in `test-utils` to support using the API Yotam mentioned in Trello, and updated the e2e test in `eyes-storybook` to compare the download with what was uploaded. However, I no longer have access to `testeyes` because of my email address change and the change in auth to use Google. So I'm unable to verify everything works since the backend change was only published to `testeyes` (right now the request to get the domMapping will 404 - which shows up as a `Unexpected end of JSON input` in the e2e test). Can you set me up with a testeyes account and/or help me verify? I hate be a blocker in merging/shipping this since others are waiting on it.

tourdedave:
Okay, I was working on updating the e2e test for storybook to verify the payload by downloading it and comparing it to the file that was uploaded. The problem is that the API in the backend needs us to send a body containing the `imageId`. I've verified this works with curl, but when trying to do it in `node-fetch` it's not supported (see [here](https://github.com/whatwg/fetch/issues/551#issuecomment-1229275789)). The recommended alternative in that thread is to use `axios`. I tried that, but it doesn't seem to be sending my headers correctly.

Also, it looks like the file path is being uploaded to the back-end (instead of the JSON file contents).

I'll pick this up again tomorrow.

tourdedave:
> Okay, I was working on updating the e2e test for storybook to verify the payload by downloading it and comparing it to the file that was uploaded. The problem is that the API in the backend needs us to send a body containing the `imageId`. I've verified this works with curl, but when trying to do it in `node-fetch` it's not supported (see [here](https://github.com/whatwg/fetch/issues/551#issuecomment-1229275789)). The recommended alternative in that thread is to use `axios`. I tried that, but it doesn't seem to be sending my headers correctly.
> 
> Also, it looks like the file path is being uploaded to the back-end (instead of the JSON file contents).
> 
> I'll pick this up again tomorrow.

Okay, everything is working now.

I had to use `axios` for the reason mentioned in my previous comment. The trick to get it working (re: the header issue I was seeing) was to use `decompress: false` in the request options. As for the incorrect thing getting uploaded to the backend, a couple of things were happening. There was a premature return in the transform function (since `target.isTransformed` was `true` so the conditional for modifying the settings was never reached), and the transform function modifying the settings but not returning them. So I separated the transform into two functions (e.g., `transformTarget` and `transformSettings`). I also renamed the file to `transform-inputs.ts`. Seemed like an intuitive choice. Let me know if it's too misleading/cringy for you guys.

amitzur:
@tourdedave before reviewing everything, a couple of points:
1. Where does it specify that we need to do a `GET` request with body? It doesn't make sense to me.
2.  I would keep `transformTarget` as is and create another file `transform-settings`. And I would support it also in `check` and not just `checkAndClose`.

tourdedave:
> @tourdedave before reviewing everything, a couple of points:
> 
> 1. Where does it specify that we need to do a `GET` request with body? It doesn't make sense to me.

In the swagger API docs for Eyes ([link](https://eyes.applitools.com/swagger/index.html)).

Screenshot 2023-09-27 at 14 42 51

> 2. I would keep `transformTarget` as is and create another file `transform-settings`. And I would support it also in `check` and not just `checkAndClose`.

I was initially going to add it in check, but for the purposes of this PR/storybook we only need it for checkAndClose, and adding it to check looked to lead to some additional typing issues to sort out. I'll take a closer look though.

amitzur:
@tourdedave I think that's an issue with swagger. Eyes' swagger is not a good source of truth.
I looked for the error we get - `incorrect header check`. Found this - https://github.com/node-fetch/node-fetch/issues/93#issuecomment-200791658
So you can get the content without needing to pass `imageId` to `getTestDom`. The only change needed is to pass `{compress: false`:
```
async function getTestDom(result, domId) {
  // ...
  const response = await fetch(url, {compress: false})
```
I don't know **why** this happens. Maybe Eyes is returning something wrong in this case.
I committed a change to rollback the use of axios. See if it makes sense to you.
There's still a small quirk I had to overcome with that `getNodesByAttribute` which gets returned.

tourdedave:
> @tourdedave I think that's an issue with swagger. Eyes' swagger is not a good source of truth. I looked for the error we get - `incorrect header check`. Found this - [node-fetch/node-fetch#93 (comment)](https://github.com/node-fetch/node-fetch/issues/93#issuecomment-200791658) So you can get the content without needing to pass `imageId` to `getTestDom`. The only change needed is to pass `{compress: false`:
> 
> ```
> async function getTestDom(result, domId) {
>   // ...
>   const response = await fetch(url, {compress: false})
> ```
> 
> I don't know **why** this happens. Maybe Eyes is returning something wrong in this case. I committed a change to rollback the use of axios. See if it makes sense to you. There's still a small quirk I had to overcome with that `getNodesByAttribute` which gets returned.

Ah, okay. Thanks! LGTM. Ready to ship it shabtay or do you have more to review?

js/packages/core-base/src/utils/transform-settings.ts
```diff
@@ -3,6 +3,20 @@ import {promises as fs} from 'fs'
...
 
+// NOTE:
+//
+// There was much debate about whether to start transforming CheckSettings
+// and whether or not domMapping should live on `target` or `settings`.
+//
+// Where we landed was this - we didn't want to create a general input
+// transformation point in the code base. The compromise was to separate this
+// transform out into a separate function, making it explicit, and to reference
+// the rationale in a code comment (*wave*).
```
amitzur, line 14: @tourdedave should be `:wave:`

js/packages/core-base/test/it/transform-target.spec.ts
```diff
@@ -228,7 +230,7 @@ describe('transform-target', () => {
 
     it('image file url object in bmp format', async () => {
       const result = await transformTarget({
-        target: {image: pathToFileURL(resolvePath('./test/fixtures/screenshot.bmp'))},
+        target: {image: pathToFileURL(resolvePath('./test/fixtures/screenshot.bmp')) as URL},
```
amitzur, line 233: @kyrylodolynskyi @tourdedave I wonder if this will be fixed in new Node.js types. Because the issue is that there are 2 `URL` references - one is a class and one is an interface in `global`.

amitzur:
@tourdedave I pushed some stuff and now it seems ready to me. Have a look and ready to ship from my perspective.

PR #1901, opened 2023-09-01
fix: suppport coded regions with layoutBreakpoints reload

SophieDePaula:
[Trello card](https://trello.com/c/nHRL3ux8/1992-goldman-sachs-java-layout-breakpoint-reload-stale-element)

 <h3>Background:</h3>

When we send coded regions to the grid, we need to do three things in the SDK:
1) Mark the elements in the `DOM` using `data-applitools-selector`
2) Send a list of these elements to `render`
3) Remove `data-applitools-selector` from the page after we are done taking all the snapshots.

 <h3>The problems reloadPage creates: </h3>

1) All markers are removed before we capture the snapshots
2) We crash when we try to remove the markers as the elements are stale

 <h3>The solution</h3>

I extracted the relevant code that marks the elements (there is a new file that is called `calculate-regions`) so that I can call this method right after we reload the page and then when we take the snapshot we'll have the markers.
`take-dom-snapshots` now returns an array of objects of type `snapshotResult` that includes the snapshot and the regions refrences.
In `check` and `checkAndClose` we extract all the relevant info about the snapshot and the regions from the returned object from `take-dom-snapshots` and if we got the snapshot as the `target` instead of a `driver` then we simply use that snapshot.

**TODO**:

- [x] Update `check-and-close` 
- [x] Add tests


amitzur (commented):
I was able to understand it to some degree. I will finalize my thoughts tomorrow.
In the meantime I left a few inline comments.
I'm not fond of the `const shouldAddIds = !ids.length` condition. I think it's implicit.
I think you are on the right track, but there is something about the new API of `generateSafeSelectors` that doesn't sit well for me.

js/packages/core/src/ufg/check-and-close.ts
```diff
@@ -83,24 +88,17 @@ export function makeCheckAndClose<TSpec extends SpecType>({
     let regionToTarget: Selector | Region | undefined
     let scrollRootSelector: Selector | undefined
     let selectorsToCalculate: {safeSelector: Selector | null; originalSelector: Selector | null}[]
+    let cleanupGeneratedSelectors
```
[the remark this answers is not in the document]
SophieDePaula: @amitzur I think it moved due to a change I made and then reverted. I'll move it back to it's original place.

js/packages/core/src/ufg/check-and-close.ts
```diff
@@ -171,7 +174,9 @@ export function makeCheckAndClose<TSpec extends SpecType>({
       snapshotUrl = await driver.getUrl()
       snapshotTitle = await driver.getTitle()
 
-      await cleanupGeneratedSelectors?.()
+      if (!settings.layoutBreakpoints?.reload) {
```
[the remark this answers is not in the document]
SophieDePaula: @amitzur because if we reload then we also reload after we take all the snapshots [here](https://github.com/applitools/eyes.sdk.javascript1/blob/master/js/packages/core/src/ufg/utils/take-dom-snapshots.ts#L126). So we don't have the markers any more

js/packages/core/src/ufg/check-and-close.ts
```diff
@@ -144,6 +142,11 @@ export function makeCheckAndClose<TSpec extends SpecType>({
                 })
               }
             },
+            async beforeEachSnapshot() {
+              if (settings.layoutBreakpoints?.reload) {
+                await generateSafeSelectors(driver.currentContext)
```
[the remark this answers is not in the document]
SophieDePaula: @amitzur we call `generateSafeSelectors` in `check` and if `reload: true` then again after we reload in `takeDomSnpashots`. So it's twice, but we have to call it again after `reloadPage`. Were you talking about that or something else?

SophieDePaula:
> I was able to understand it to some degree. I will finalize my thoughts tomorrow. In the meantime I left a few inline comments. I'm not fond of the `const shouldAddIds = !ids.length` condition. I think it's implicit. I think you are on the right track, but there is something about the new API of `generateSafeSelectors` that doesn't sit well for me.

@amitzur I responded and addressed your comments.
As for the API, I would like to hear why it does not sit well with you, I can try to explain the reason I made my changes:
1) `makeGenerateSafeSelectors`, should get the `elementReferences` as when we call the returned function we always want to make sure we iterate over the same elements.
2) `generateSafeSelectors`, receives the `context` as we need to current context after each reload to be able to find the elements in the DOM.

Re `const shouldAddIds = !ids.length` , I can rename the var to `hasIds` so that it's clearer that we already have `ids` and therefore we don't want to regenerate them, but I am not sure if it's the name that bothers you. Can you explain in more details what you mean?

js/packages/core/src/ufg/check-and-close.ts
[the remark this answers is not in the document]
amitzur: Got it. Makes sense.

js/packages/core/src/ufg/check-and-close.ts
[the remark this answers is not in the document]
amitzur: Now I see that there is one use case that you do it unnecessarily - if one of the layout breakpoints is the existing width, then you call it once in `check`, and then it will be called before the first snapshot, without any reload happening between them, from here - https://github.com/applitools/eyes.sdk.javascript1/blob/a46f5dcaba6fbe8094f77dc4a96aac063d702a09/js/packages/core/src/ufg/utils/take-dom-snapshots.ts#L84

amitzur:
@SophieDePaula I realized what doesn't sit well for me. There is an assumption that all element reference (a.k.a. original selectors) will return exactly the same elements after every reload. But this is not a true assumption. It's probable, but not guaranteed.
It would be nice if every snapshot would have the same ID's for safe selectors, but it's not robust.

I think the ultimate solution is a bigger change than what was done here. It's to incorporate `generateSafeSelectors` into `takeDomSnaphots`, and have different generated selectors for each snapshot. In order to not complicate `takeDomSnapshots` itself, I think I would refactor this code out of `check`, and provide a function that both calls `takeDomSnapshots`, and returns the generated selectors per snapshot (and potentially other info as well like `regionToTarget`, I'm not sure).
Once that's done, then [this line](https://github.com/applitools/eyes.sdk.javascript1/blob/77683e8002f624c030145f07b3056ccb0b06aab9/js/packages/core/src/ufg/check.ts#L261) will need to change in order to take the selectors to calculate from each snapshot.

js/packages/core/src/ufg/check.ts
```diff
@@ -153,36 +125,47 @@ export function makeCheck<TSpec extends SpecType>({
...
 
-      await currentContext.focus()
       snapshotUrl = await driver.getUrl()
       snapshotTitle = await driver.getTitle()
-
-      await cleanupGeneratedSelectors?.()
     } else {
-      snapshots = !utils.types.isArray(target) ? Array(uniqueRenderers.length).fill(target) : target
-      snapshotUrl = utils.types.has(snapshots[0], 'url') ? snapshots[0].url : undefined
+      snapshotResults = !utils.types.isArray(target) ? Array(uniqueRenderers.length).fill({snapshot: target}) : target
```
amitzur, line 163: @SophieDePaula it's wrong here. You treat the array case as `DomSnapshot` and the non-array case as `SnapshotResult`. I will fix it.

amitzur: The outcome of this mistake is that a test for layout breakpoints in storybook should fail.

amitzur: Ah, that's not entirely correct. There **is** a mistake here but there is no code path currently which passes a single target. `takeDomSnapshots` always returns an array, this is why no test fails.

amitzur: Now, because this **has** to be correlated with the renderers array, I am changing this to require an array. There is no use case for passing a single snapshot, just like there is no use case for passing a single renderer (non-array).

amitzur: On second thought, I will keep it. It should be nice to have an API that "checks a snapshot on several renderers".

js/packages/core/src/ufg/utils/take-dom-snapshots.ts
```diff
@@ -39,13 +47,31 @@ export async function takeDomSnapshots<TSpec extends SpecType>({
     }
   }
   await hooks?.beforeSnapshots?.()
-
+  if (!settings.layoutBreakpoints?.reload && settings.calculateRegionsOptions) {
+    const regionsResults = await calculateRegions({
+      ...settings.calculateRegionsOptions,
+      driver,
+    })
+    referencesResults = {
```
amitzur, line 55: @SophieDePaula the whole point about keeping the naming consistent between the return value of `calculateRegions` and `takeDomSnapshot` is to not need this white list assignment

js/packages/core/src/ufg/utils/take-dom-snapshots.ts
```diff
@@ -39,13 +47,31 @@ export async function takeDomSnapshots<TSpec extends SpecType>({
...
+    cleanupGeneratedSelectors = regionsResults.cleanupGeneratedSelectors
+  }
   if (!settings.layoutBreakpoints) {
     logger.log(`taking single dom snapshot`)
     await hooks?.beforeEachSnapshot?.()
     await waitBeforeCapture()
     const snapshot = await takeDomSnapshot({context: currentContext, settings, logger})
-    return Array(settings.renderers.length).fill(snapshot)
+    if (referencesResults) {
+      await cleanupGeneratedSelectors?.()
```
amitzur, line 68: @SophieDePaula either you add the `?` without the `if`, or you don't add the `?` and put it in the `if`. But not both.

js/packages/core/src/ufg/check.ts
SophieDePaula: @amitzur yes you are right, I had to change the way we fill the array as well. I kept that option as I saw we "support" passing single dom-snapshot in master and didn't want to break backwards comparability in case there is some client for it.

js/packages/core/src/ufg/utils/take-dom-snapshots.ts
SophieDePaula: @amitzur you are right

js/packages/core/src/ufg/utils/take-dom-snapshots.ts
SophieDePaula: @amitzur you are right

PR #1913, opened 2023-09-06
feat: change cache to support distributed cache

zamboney:
To add support for a cache outside of the node instance, a fallback function is implemented.
This function generates the required data to store in a distributed cache system. When the same request is made again, the function will return the cached data instead of executing the fallback logic again.

- [x] verify tests and usage around passing `cache` to `makeUFGClient` inside settings


js/packages/ufg-client/.mocharc.js
```diff
@@ -1,7 +1,7 @@
 const group = process.env.MOCHA_GROUP
 
 module.exports = {
-  timeout: 0,
+  timeout: 3000,
```
zamboney, line 4: no reason for more the 3sec waiting for test in this pacakge

js/packages/ufg-client/src/resources/process-resources.ts
```diff
@@ -1,11 +1,11 @@
-import type {Renderer} from '../types'
+import type {Renderer, CacheFunctionWithCallback} from '../types'
 import {
   makeResource,
   type UrlResource,
   type ContentfulResource,
   type HashedResource,
-  type KnownResource,
   type FailedResource,
+  KnownResource,
```
amitzur, line 8: @zamboney it's still a type

js/packages/ufg-client/src/types.ts
```diff
@@ -302,3 +303,14 @@ export interface RenderResult {
   fullViewSize?: Size // full size of the view/page
   selectorRegions?: Region[][]
 }
+
+type KnownResourceWithoutValue = KnownResource & {value?: never}
```
amitzur, line 307: This doesn't make sense. `value` is not in `KnownResource`.

js/packages/ufg-client/src/types.ts
```diff
@@ -302,3 +303,14 @@ export interface RenderResult {
...
+
+export type CacheFunctionWithCallback = (
+  cacheKey: string,
+  createResourceCallback: () => Promise<KnownResource | FailedResource | null>,
+) => Promise<KnownResourceWithoutValue | FailedResource | null>
```
amitzur, line 312: Why not `Promise<KnownResource | FailedResource | null>`? `KnownResource` already doesn't have the `value` property.

js/packages/ufg-client/src/types.ts
```diff
@@ -37,6 +37,7 @@ export interface UFGClientSettings extends UFGServerSettings {
   eyesServerUrl?: string
   apiKey?: string
   fetchConcurrency?: number
+  getValueFromCache?: CacheFunctionWithCallback
```
amitzur, line 40: I think this is leftovers that you forgot to remove

js/packages/ufg-client/src/client.ts
```diff
@@ -14,9 +14,16 @@ export function makeUFGClient({
   settings,
   cache = defaultResourceCache,
```
amitzur, line 15: Why are you still receiving the cache? You replaced it with `getCachedResource`.

js/packages/ufg-client/src/resources/process-resources.ts
```diff
@@ -65,37 +68,52 @@ export function makeProcessResources({
...
     for (const [url, processedResource] of Object.entries(processedResources)) {
       mapping[url] = processedResource.hash
-      ready.push(processedResource.ready)
     }
-
-    return {mapping, promise: Promise.all(ready).then(() => mapping)}
+    return {
+      mapping,
+      promise: Promise.all(waitUploadResources)
+        .then(() => (waitUploadResources = []))
```
amitzur, line 81: I assume you reset this array for the same reason we suspect `ready` was turned from promise to boolean? In order to not keep a reference for this promise forever?

js/packages/ufg-client/src/resources/process-resources.ts
```diff
@@ -65,37 +68,52 @@ export function makeProcessResources({
...
-          logger.log(`Freezing gif image resource with id ${resource.id}`)
-          resource = makeResource({...resource, value: await freezeGif(resource.value)})
-        } catch (error) {
-          logger.warn(`Failed to freeze gif image resource with id ${resource.id} due to an error`, error)
+  async function processContentfulResource(
+    {
+      resource,
+      logger = mainLogger,
+    }: {
+      resource: ContentfulResource | FailedResource | null
```
amitzur, line 91: why can it be null?

js/packages/ufg-client/src/resources/process-resources.ts
```diff
@@ -65,37 +68,52 @@ export function makeProcessResources({
...
+      logger?: Logger
+    },
+    skipCache = false, // if the callback cache is called form another callback, we don't want to cache it
```
amitzur, line 94: Not good coding convention to have this as a second parameter. The function is already receiving a single destructured parameter.

js/packages/ufg-client/src/resources/process-resources.ts
```diff
@@ -65,37 +68,52 @@ export function makeProcessResources({
...
+  ): ReturnType<CacheFunctionWithCallback> {
+    async function callback(): ReturnType<CacheFunctionWithCallback> {
```
amitzur, line 96: Put inner functions at the end so that it's clear what the body of the function is.

js/packages/ufg-client/src/resources/process-resources.ts
```diff
@@ -106,34 +124,30 @@ export function makeProcessResources({
...
-        if (utils.types.has(fetchedResource, 'value')) {
-          const dependencies = await extractDependencyUrls({
-            resource: fetchedResource,
-            settings: {sourceUrl: settings?.headers?.Referer},
-            logger,
-          })
-          logger.log(`dependencyUrls for ${resource.url} --> ${dependencies}`)
-          fetchedResource.dependencies = dependencies
+  }): ReturnType<CacheFunctionWithCallback> {
+    async function callback(): ReturnType<CacheFunctionWithCallback> {
```
amitzur, line 128: inner functions at the end

amitzur (changes requested):
Inline comments mainly about code style and some questions.
But the main point is that I don't get this solution.
Why are you calling `getCachedResource` inside `processContentfulResource`?
Now I'm ready to get an explanation about this solution

zamboney:
@amitzur

About the inline comments i will address them inline 😄.

> Why are you calling `getCachedResource` inside `processContentfulResource`?

The way things work (as far as my narrow understanding) is that we want to **"handles prefilled resources with dependencies"**

I think it's easier to understand it by the tests that are failing if you comment `getCachedResource` [here](https://github.com/applitools/eyes.sdk.javascript1/blob/bfecc76ec798db66ef31cd8e5e6f1de0a7184062/js/packages/ufg-client/src/resources/process-resources.ts#L115-L116) and just return the `callback`

```diff
       }
       return resource
     }
-    if (skipCache) return callback()
-    return getCachedResource(resource?.id || (resource?.hash as {hash: string}).hash, callback)
+    return callback()
   }
 
   async function processUrlResource({
```

Once done 3, tests failed

```
yarn test:unit --reporter tap | grep 'not ok'

not ok 54 processResources handles the case when the same resource appears as prefilled resource and as a dependency of another url resource
not ok 62 processResources handles resources with errorStatusCode (non-200 resources) from cache
not ok 63 processResources handles prefilled resources with dependencies
```


Before, this issue did not occur, but we changed the cache mechanism and no longer can update the dependencies of a resource in the cache. You can see the changes [here](https://github.com/applitools/eyes.sdk.javascript1/blob/f8a36db1f5359491b1f71e8221997824bd2ed7e6/js/packages/ufg-client/src/resources/process-resources.ts#L177-L180) and [here](https://github.com/applitools/eyes.sdk.javascript1/blob/f8a36db1f5359491b1f71e8221997824bd2ed7e6/js/packages/ufg-client/src/resources/process-resources.ts#L195).


This is what I currently understand about the topic. I can delve deeper to gain a better understanding of "prefilled" resources and their dependencies, but it goes beyond the scope of this task (i think).

js/packages/ufg-client/src/resources/process-resources.ts
```diff
@@ -222,4 +230,19 @@ export function makeProcessResources({
       return []
     }
   }
+  // this function is for a special case of cdt
+  function getResourceId({
```
amitzur, line 234: I don't like this function. I don't understand why it's needed. We saw yesterday that a resource should always has an `id`. The only case we saw it doesn't is for the dom resource, which could and should be fixed.

js/packages/ufg-client/src/resources/process-resources.ts
```diff
@@ -222,4 +230,19 @@ export function makeProcessResources({
...
+    resource,
+    logger = mainLogger,
+  }: {
+    resource: {value?: Buffer; hash?: HashedResource | {errorStatusCode: number}; id?: string}
```
amitzur, line 238: I don't like this usage of some ad-hoc type for a resource.

js/packages/ufg-client/src/resources/process-resources.ts
```diff
@@ -36,10 +37,14 @@ export type ResourceMapping = Record<string, HashedResource | {errorStatusCode:
 
 export function makeProcessResources({
   fetchResource,
-  getCachedResource,
+  getCachedResource: getCachedResourcePlain,
   uploadResource,
   logger: mainLogger,
 }: Options): ProcessResources {
+  const getCachedResource = (
```
amitzur, line 44: Why make it complicated? You can just use `getCachedResource` and pass it `resource.id`.

js/packages/ufg-client/src/resources/process-resources.ts
```diff
@@ -56,7 +61,10 @@ export function makeProcessResources({
       async (processedResourcesPromise, [url, resource]) => {
         if (utils.types.has(resource, 'value') || utils.types.has(resource, 'errorStatusCode')) {
           // process contentful resource or failed resource
-          const processedResource = await processContentfulResource({resource, logger})
+          // and set the responses in cache
+          const processedResource = await getCachedResource(resource, () =>
```
amitzur, line 65: No. `getCachedResource` is only meant to be used for fetching resources that you don't have the value for.
There is no need to set anything in the cache if it wasn't fetched in this operation.
Why did you add it?

js/packages/ufg-client/src/resources/resource.ts
```diff
@@ -45,6 +45,10 @@ export type HashedResource = {
   contentType: string
 }
 
+export function isHashedResource(resource: unknown): resource is HashedResource {
```
amitzur, line 48: There is no need for this function.

amitzur (changes requested):

js/packages/ufg-client/src/client.ts
zamboney: i want to keep the existing flow, also it's not really in use. if you think this is unnecessary i will remove it

zamboney (commented):
@amitzur in general you need to make a decision:
1. we are keeping the tests and logic intact and push this change faster and agree to deal with the issue about the change in logic later
2. we are understanding what the author did and fix the logic issue that raise when we change the cache to support in a distributed way

PR #1925, opened 2023-09-14
fix: add support to shadow dom iframe

zamboney:
When the frame is inside `shadowRoot` it throws `"Unable to find out the chain of frames"`.
This pr will fix this issue by checking the `shadowRoot` on **each element** on the page and make sure that there are no shadow root iframe that are missed when `refresh` the driver

- [ ] check for performance issue
- [x] tests

[terllo](https://trello.com/c/g2Xr5iNE)


js/packages/snippets/src/browser/getChildFramesInfo.js
```diff
@@ -1,10 +1,14 @@
...
+  const iframeElements = document.querySelectorAll('frame, iframe')
+  iframeElements.forEach(iframe => frames.push(iframe))
+  const iframeElementsInShadowRoot = Array.prototype.filter
+    .call(document.querySelectorAll('*'), element => element.shadowRoot)
+    .map(element => element.shadowRoot.querySelectorAll('frame, iframe'))
+  iframeElementsInShadowRoot.forEach(element => {
+    element.forEach(iframe => frames.push(iframe))
+  })
+  return frames.map(frameElement => [frameElement, !frameElement.contentDocument, frameElement.src])
 }
```
kyrylodolynskyi, line 12: Shouldn't it be the recursive implementation? I mean calling the same function repeatedly inside every shadow root? This will also cover the case when the iframe is located in a nested shadow root document (the one that is already inside a shadow root)

kyrylodolynskyi (changes requested):

js/packages/snippets/src/browser/getChildFramesInfo.js
zamboney: sure i will do that

zamboney:
@kyrylodolynskyi changed as requested please take anther look, 10x!

js/packages/snippets/src/browser/getChildFramesInfo.js
```diff
@@ -1,10 +1,11 @@
-function getChildFramesInfo() {
-  const frames = document.querySelectorAll('frame, iframe')
-  return Array.prototype.map.call(frames, frameElement => [
-    frameElement,
-    !frameElement.contentDocument,
-    frameElement.src,
-  ])
+function getChildFramesInfo(root) {
+  root = root || document
```
kyrylodolynskyi, line 2: @zamboney Pay attention to the convention among all snippets, it always accepts only one (or none) argument and this argument is an array. So I recommend this

```suggestion
function getChildFramesInfo([root = document] = []) {
```

zamboney:
I try your suggestion but when I do I get a `null` as an argument and it's over writing the default value, I will give it a try tomorrow...

js/packages/snippets/src/browser/getChildFramesInfo.js
zamboney: I try your suggestion but when I do I get a `null` as an argument and it's over writing the default value, I will give it a try tomorrow...

kyrylodolynskyi: This is how it works everywhere else (you can check yourself), so I will be surprised if this isn't going to work

zamboney (commented):
@kyrylodolynskyi made some changes please take a look again 10x

js/packages/snippets/src/browser/getChildFramesInfo.js
zamboney: @kyrylodolynskyi, your suggestion didn't work because the parameter that is supplied by the driver is `null`, but I made some changes by moving the recursive function inside. I hope this better meets your needs

kyrylodolynskyi (approved):

PR #1929, opened 2023-09-19
fix: async cache key - avoid undefined keys and ready promise

amitzur:
Changes in this PR:
1. Fixing the undefined key was about adding an `id` to the dom resource. I wanted the ID to contain the hash, so I had to set it **after** calling `makeResource`. I didn't see value in doing it inside `makeResource`. Not sure about that.
2. Avoiding the async cache entry to contain `ready` promise.
3. Add an important test for `createRenderTarget` with async cache.
4. Added an e2e test in core that was missing. I had to copy/paste the fake async cache code, since it isn't exported from `ufg-client`, as it is related to tests.


js/packages/ufg-client/src/resources/process-resources.ts
```diff
@@ -252,4 +252,26 @@ export function makeProcessResources({
...
+    key: string,
+    callback: () => Promise<CacheableKnownResource & {ready?: boolean | Promise<boolean>}>,
+  ): Promise<CacheableKnownResource> {
+    return new Promise(async (resolve, reject) => {
+      try {
+        const resourceFromCache = await asyncCache!.getCachedResource(key, async () => {
+          const ret = await callback()
+          resolve(ret)
+          const retWithoutReady = {...ret}
+          delete retWithoutReady.ready
```
zamboney, line 268:
```suggestion
          const {ready, ...retWithoutReady} = ret
```
i think that spread and delete is a bit over kill...

js/packages/core/test/e2e/ufg/async-cache.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+    getUploadCache: () => uploadCache,
+  }
+
+  function cacheFunc(cache: Map<string, any>) {
+    return async (key: string, callback: () => Promise<unknown>) => {
+      assert.ok(key, 'cache key cannot be undefined')
+      if (!cache.get(key)) {
+        cache.set(
+          key,
+          wait(100).then(async () => {
```
zamboney, line 29: i think we are missing here that the `cache.set` is **not** a sync call. I understand that it's done really fast, but for sure it's done outside the event loop. maybe add `wait(0)`?

zamboney (commented):
Looking good! 👍
The only thing that is missing is a test for a true `async` set key in cache, like there would be in a real distributed cache mechanism .

js/packages/ufg-client/src/resources/process-resources.ts
amitzur: yes, thanks!

js/packages/ufg-client/src/resources/process-resources.ts
```diff
@@ -252,4 +252,26 @@ export function makeProcessResources({
...
+
+  // Note: this function is needed since the contract with async cache is to return a serializeable object.
+  // So we need to peel off `ready` which is a promise, but still use it as the return value from the callback.
+  function getCachedResource(
+    key: string,
+    callback: () => Promise<CacheableKnownResource & {ready?: boolean | Promise<boolean>}>,
+  ): Promise<CacheableKnownResource> {
+    return new Promise(async (resolve, reject) => {
+      try {
+        const resourceFromCache = await asyncCache!.getCachedResource(key, async () => {
```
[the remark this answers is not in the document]
amitzur: I'm not sure if `logger.warn` actually propagates to the user. But definitely worth writing to the log in this case. I will add more logs in general for the async cache scenario.

js/packages/core/test/e2e/ufg/async-cache.spec.ts
amitzur: It's a good point. But I think it would test the cache and not the SDK.
I'll explain: If there's a race condition, then the cache might call the callback when it shouldn't. But that's a bug of the cache. The SDK in this case would just behave as if there's no cache, and I believe that scenario is well tested.

zamboney: I don't think the race condition is the issue here, the issue is that you have a wrong assumption here that the js event loop is always locked until the cache is set. This is why I ask for `wait(0)`. Just let the event loop go forward to the next call without locking it until the cache is set. It mimics the cache mechanism closer to production.
I can't think of a good example that this could affect us now, but it's not testing the way the cache system behaves, and it could meet us in the future for an exotic bug that could be found if we test it as closely as passable to production.

zamboney (approved):
only think that we are not agree is about the `cache.set` but i don't think is that important for me to not approve this change
🎉

js/packages/core/test/e2e/ufg/async-cache.spec.ts
amitzur: @zamboney the event loop is freed because `cacheFunc` returns an async function. So when the SDK calls is, it gets a promise and awaits it.
I did try it with `wait(0)` before the line `if (!cache.get(key)) {`. All tests passed and I didn't notice anything weird.
Do you still think I should have it? I don't mind having it and adding a comment with a link to this thread.

js/packages/ufg-client/src/resources/process-resources.ts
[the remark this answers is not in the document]
amitzur: added more logs and verification that key is not falsy

js/packages/ufg-client/test/unit/process-resources.spec.ts
```diff
@@ -12,6 +12,40 @@ import nock from 'nock'
...
+            cache.set(key, result)
+            return result
+          }),
+        )
+      }
+      await wait(100)
+      return cache.get(key)
+    }
+  }
+}
```
kyrylodolynskyi, line 47: Exporting something from the test file seems odd to me. Don't you want to put this piece of code in the utils folder as we doing in other packages where we have some mock/helper code for testing?

js/packages/ufg-client/src/resources/process-resources.ts
```diff
@@ -252,4 +253,39 @@ export function makeProcessResources({
...
+          logger.log(`async cache callback called for ${key}`)
+          const ret = await callback()
+          resolve(ret)
+
+          // We also want to make sure the value of the resource is not included in the cache. So better yet, just whitelist the properties of KnownResource
+          // See note in KnownResource type
+          return {
+            hash: ret.hash,
+            dependencies: ret.dependencies,
+          }
```
kyrylodolynskyi, line 281: Did you decide to not use the technic with `Required<KnownResource>`?

kyrylodolynskyi (approved):
Look good to me. It is not critical, but I think you better address my comment about exporting stuff from the test file. I think it will be much better to have a separate place to reason about the mock async cache impl.

js/packages/ufg-client/test/unit/process-resources.spec.ts
amitzur: Fair enough. I was lazy :) Fixed

js/packages/ufg-client/src/resources/process-resources.ts
amitzur: I thought about it but I can't track back in my mind how to use it

PR #1939, opened 2023-09-27
fix: force native on get environment

zamboney:
force `isNative` based on the name of the `switchWorld` variable.

[trello](https://trello.com/c/akyUyzIG)


js/packages/driver/src/driver.ts
```diff
@@ -102,13 +102,32 @@ export class Driver<T extends SpecType> {
...
 
-  async refresh({reset}: {reset?: boolean} = {}): Promise<this> {
+  async refresh({
+    reset,
+  }: {
+    reset?:
+      | {
+          [K in keyof Pick<
+            typeof Driver.prototype,
+            'getDriverInfo' | 'getEnvironment' | 'getViewport' | 'getFeatures' | 'getHelper' // I'm not sure if we need all of these options
```
zamboney, line 112: I'm not sure if we need all of these options

js/packages/driver/src/driver.ts
```diff
@@ -914,7 +933,7 @@ export async function makeDriver<T extends SpecType>(options: {
   driver: Driver<T> | T['driver'] | T['secondary']['driver']
   spec?: SpecDriver<T>
   customConfig?: {useCeilForViewportSize?: boolean}
-  reset?: boolean
+  reset?: Extract<Parameters<typeof Driver.prototype.refresh>[0], 'reset'> | boolean // TODO: WDYT about this? if this is unnecessary, i an remove it IMO we need to remove the boolean option
```
zamboney, line 936: WDYT about this? If this is unnecessary, I can remove it IMO we need to remove the boolean option

kyrylodolynskyi (commented):
I am not sure what you are trying to achieve. We are detecting exactly this when extracting the environment. Are you trying to do an optimization?
In other words, is there any other reason, except optimization, to know that this is a native context before the call to `getEnvironment`?

zamboney:
@kyrylodolynskyi take a look at the test. If the user changes the context to `NATIVE_APP` and then `check`, the test throws an error of `Method not implemented` because the `switchWorld` change to native while the caps are set to `web`. This is an issue because it's trying to get the `userAgent` using `executeScript` which is not supported in `NATIVE_APP`.  🤷‍♂️ 
anyway, I'm trying to fix it from the `spec-driver`... I think it's make more sense

kyrylodolynskyi:
@zamboney It wasn't clear from the test. So basically you are saying that the problem is that we do not reset capabilities. Right?
I don't see how you can fix it from spec-driver side 🤔

js/packages/spec-driver-webdriver/src/spec-driver.ts
```diff
@@ -327,7 +327,8 @@ export async function childContext(driver: Driver, element: Element): Promise<Dr
   return driver
 }
 export async function getDriverInfo(driver: Driver): Promise<DriverInfo> {
-  return {sessionId: driver.sessionId}
+  const isNative = (await driver.getContext()).toString() === 'NATIVE_APP'
+  return {sessionId: driver.sessionId, environment: {isNative}}
```
zamboney, line 331: @kyrylodolynskyi ☝️

kyrylodolynskyi: So you would have to do it in every spec-driver. The spec-driver is meant to be minimal, and everything that could be implemented elsewhere should be implemented elsewhere. This has already been implemented in `getEnvironment` method of the `Driver`

zamboney: Sure makes sense I will change it...

kyrylodolynskyi: Doing it in spec-driver isn't the right way, and I told you the reason why in my previous message. This is not required to be there. Spec-driver already provides instruments to extract this information, this is basically the definition of "not required" for the spec-driver. What is really important for spec driver is to be minimal

js/packages/driver/src/driver.ts
```diff
@@ -264,6 +264,7 @@ export class Driver<T extends SpecType> {
       const capabilitiesEnvironment = capabilities ? extractCapabilitiesEnvironment(capabilities) : null
       this._logger.log('Extracted capabilities environment', capabilitiesEnvironment)
       this._environment = {...this._environment, ...capabilitiesEnvironment}
+      this._environment.isNative = (await this.getCurrentWorld()) === 'NATIVE_APP'
```
kyrylodolynskyi, line 267: Just above this line, we are basically doing the same with some reasonable optimizations. As I understand from our conversation the `!this._environment.browserName` statement is not `true` in this case. Right?
So my question is why it is not `true`? Isn't it happening because even after reset we still have capabilities cached?

zamboney: not in the cache in caps. the user supply `app` as `chrome`/`safari` but switch to `NATIVE_APP`

kyrylodolynskyi: Ok. Now I see it in the repro in the ticket. Anyway, you have to consolidate, the old way to extract this info and the new way.
With the requirement to support this, it means that we just always have to extract the current context, no matter whether we are in the browser or not.

zamboney: Not sure if I understand what you mean...

kyrylodolynskyi: My point is you should consolidate the line you added with the `if` statement on line 269. Otherwise, you extract the current world twice, also do it only when `this._environment.isMobile` is `true` seems reasonable.

kyrylodolynskyi (changes requested):

kyrylodolynskyi (changes requested):
LGTM now 👍🏼 , but please use the same name for the test and the fixture.

kyrylodolynskyi (approved):

PR #1940, opened 2023-09-27
fix: avoid saving data urls in cache (both sync and async)

amitzur:
This resource has dependency URL's that weigh 12MB, which blew up the distributed cache: https://xometry.eu/wp-content/uploads/2021/07/IM-Finishes-Handout-EN4-01.svg

Now we avoid saving those to the cache. I did it in `processContentfulResource` in order to catch every possible data flow.


kyrylodolynskyi:
Doesn't it break the whole idea? You need to know the dependencies in order to be able to reuse cached resources later without parsing them. As I understand after this change if css resource `style.css` contains dependency with data URL (let's say an image), we will not get the image when we read the resource from the cache.

amitzur:
@kyrylodolynskyi data URL's should not be in the resource mapping. This is why it's fine to ignore them when processing resources. If `style.css` references `data:image/png;base64,...`, then it's fine to get `dependencies: []` from the cache. I don't need to process that dependency, and even if I did, `processUrlResource` would return `null` and `processedResourcesWithDependencies` will not contain the data URL resource.

---
Actually, there could be a potential bug here: if there is a data URL which has references to real URL's. For example:
```
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 800 600" width="800" height="600">
    <link
        xmlns="http://www.w3.org/1999/xhtml" rel="stylesheet" type="text/css"
        href="data:text/css,circle{fill:url('http://localhost:8080/smurfs.jpg');}" />
    <circle cx="100" cy="100" r="100"/>
</svg>
```
This means that we wouldn't look inside the data URL to find `http://localhost:8080/smurfs.jpg`. I think we have this bug even today.
I tried to reproduce this in a dozen variations, but couldn't find any example that actually applies the style correctly in the browser.
So I think it's a super edge case.

PR #1947, opened 2023-10-01
fix: move lazyload option to take-dom-snapshots

zamboney:
For autonomous and in general, there no reason for it to be there.


js/packages/core/src/ufg/check-and-close.ts
```diff
@@ -98,23 +97,13 @@ export function makeCheckAndClose<TSpec extends SpecType>({
             layoutBreakpoints: settings.layoutBreakpoints,
             renderers: uniqueRenderers as UFGRenderer[],
             skipResources: ufgClient.getCachedResourceUrls(),
+            lazyLoad: environment.isWeb && settings.lazyLoad,
```
amitzur, line 100: this whole block is inside `if (isWeb)`

js/packages/core/src/ufg/check.ts
```diff
@@ -100,23 +99,13 @@ export function makeCheck<TSpec extends SpecType>({
             layoutBreakpoints: settings.layoutBreakpoints,
             renderers: uniqueRenderers as UFGRenderer[],
             skipResources: ufgClient.getCachedResourceUrls(),
+            lazyLoad: environment.isWeb && settings.lazyLoad,
```
amitzur, line 102: see comment in `check-and-close`

amitzur (approved):
1. Happy to see there are no hiding complications.
2. I would actually remove `beforeSnapshots` hook altoghether. I don't know why such a thing is needed. We have another hook, `beforeEachSnapshot`. That hook makes more sense, but nobody is using it, so I would give up on that as well.
3. I don't like the fact that the type for `lazyLoad` is copied twice. Maybe `CheckSettings['lazyLoad']`? Although we don't write this way normally in the SDK. Or make a type for it like there is `LazyLoadOptions` in the user-facing `@applitools/eyes` package.
4. I'm missing a test for `takeDomSnapshots` itself that checks lazyLoad works.

js/packages/core/src/ufg/utils/take-dom-snapshots.ts
```diff
@@ -7,15 +7,15 @@ import * as utils from '@applitools/utils'
...
+import {type LazyLoadSettings, waitForLazyLoad} from '../../automation/utils/wait-for-lazy-load'
 
 export * from './take-dom-snapshot'
 export type DomSnapshotsSettings = DomSnapshotSettings & {
   renderers: Renderer[]
   waitBeforeCapture?: number | (() => void)
   layoutBreakpoints?: {breakpoints: number[] | boolean; reload?: boolean}
   calculateRegionsOptions?: CalculateRegionsOptions
-  lazyLoad?: boolean | {scrollLength?: number; waitingTime?: number; maxAmountToScroll?: number}
+  lazyLoad?: boolean | Omit<LazyLoadSettings, 'executionTimeout' | 'pollTimeout'>
```
amitzur, line 18: I honestly don't know what `executionTimeout` and `pollTimeout` are doing in `LazyLoadSettings`.
They cannot be configured by any user-facing API.
Perhaps @tourdedave remembers?

zamboney: @amitzur i change the signature

zamboney:
@amitzur fix all of your requests except (4), i don't see a reason to add a `unit`/`integration` to make sure that the API stays as is. if you think this is mandatory I will add a new test in another pr

amitzur:
@zamboney I'm fine with doing it in another PR. I don't mean to test that the API stays as is. I mean - it's a new API in `takeDomSnapshots`, and it should be tested in the tests for `takeDomSnapshots`.
It's a different question whether we have good tests for `takeDomSnapshots` as a unit (I don't think we do), or how easy it is to test it as a unit.

zamboney:
@amitzur i don't see a reason to add a new test just for `takeDomSnapshots`. The current tests check this flow and make sure it works. The test you mean (i think) is to check the flow directly through `takeDomSnapshots`. For this, i don't see the point...

PR #1949, opened 2023-10-01
Fix typo in the Cypress defineConfig instructions

knorrium:
When running `npx eyes-setup` the instructions say `export default eyesPlugin(definedConfig({` where it should be `defineConfig` per the official docs https://docs.cypress.io/guides/tooling/plugins-guide#Real-World-Example


amitzur:
@knorrium thank you for this contribution!
I have filed the same fix in PR #1953 in order for our CI to run successfully. So I'm closing this one.

PR #1955, opened 2023-10-03
chore: add test to lazyload in take-dom-snapshots

zamboney:


zamboney:
This test is not end-to-end, but I couldn't find a better way to do it without breaking the testing folder structure. All the other `e2e` tests are directly from the `makeCore` function, this is the first one that isn't.
It feels like this is the best way to keep the test structure intact without mixing `e2e` and `it` tests together.

@amitzur @kyrylodolynskyi WDYT

amitzur:
@zamboney please see #1956 . This is my proposition of how to add e2e test for `takeDomSnapshots`.
IMO it's important to have e2e for `takeDomSnapshots` because it's an external API of core which has users and needs to be well tested. This is the first instance of this test, I think we should have had something like this already even for the basic case without lazyload, but we can evolve it gradually.

amitzur (approved):

zamboney:
@amitzur i change it to `e2e` as you requested. also move the html to fixture

kyrylodolynskyi (commented):
Why is it essential to use an inline page?  I think it is much harder to read, I would prefer us to use pages from the demo repo or create local pages in the `fixtures` directory.
___
UPD: Ah, I see you changed it

zamboney:
@kyrylodolynskyi The test page moved to `fixture` and the reason that it is not on `demo` because this is a smaller lazy load page so the test will be faster (+1m to ~9s).

kyrylodolynskyi (changes requested):
Hmm. It actually got me thinking. I see what you are doing for validation, but this does not actually test what we care about. I would better check that all of the image resources are returned as a result of dom snapshot.

zamboney:
@kyrylodolynskyi that was @amitzur firs idea, i will fix it

js/packages/core/test/e2e/ufg/take-dom-snapshots.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+      },
+      logger,
+    })
+    assert.deepStrictEqual(
+      snapshots?.[0]?.snapshot.cdt
+        .filter(({nodeName}) => nodeName === 'IMG')?.[0]
+        .attributes.filter(({name}: {name: string}) => name === 'src')?.[0].value,
+      'https://applitools.github.io/demo/TestPages/LazyLoad/George_W_Bush.jpeg',
+      'should change the image src once lazy loading in the cdt',
+    )
```
kyrylodolynskyi, line 53: Hmmm... This is fine, however, I had a less verbose idea to simply check the resource in `resourceUrls` or `resourceContents` props. but if you still see value in your approach then feel free to keep it, but at least change `.filter(...)[0]` to `.find(...)`.

js/packages/core/test/e2e/ufg/take-dom-snapshots.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+      'should be 1 lazy load images',
+    )
+    const snapshots = await takeDomSnapshots({
+      driver,
+      settings: {
+        lazyLoad: true,
+        renderers: [{width: 100, height: 100, name: 'chrome'}],
+        layoutBreakpoints: {
+          breakpoints: true,
+        },
```
kyrylodolynskyi, line 43: Is there any reason to have it here?

kyrylodolynskyi (changes requested):

js/packages/core/test/e2e/ufg/take-dom-snapshots.spec.ts
zamboney: not really i will remove it

kyrylodolynskyi (approved):

PR #1964, opened 2023-10-04
fix: add .cmd for storybook binary for windows for local binaries

SophieDePaula:
[Trello card](https://trello.com/c/RMf0QkcX/1959-storybook-fails-on-windows)

The issue here is that the SDK failed to start storybook because the path to the binary was missing `.cmd` for Windows.


amitzur (commented):
Are you sure that `findNpmModuleCommandPath` will still find the file it needs on windows? Because now it's not adding `.cmd`.
As part of this fix, I require that we will run a basic test on windows as part of storybook CI.

amitzur (changes requested):
need to run simple e2e test on windows

SophieDePaula:
@amitzur 
1) looking at the `.bin` folder in a Windows machine, I see we have both binaries with and without `.cmd`, so we should find the file and then we add `.cmd` in case it's Windows at the end of the `determineStorybookVersion`.

2) I thought that since we have a test for Windows under the support matrix, then we should not have one here as well.

kyrylodolynskyi:
@SophieDePaula ahah, I think you just said out loud (item number 2) what @amitzur was scared about. I think that we are so far from running the "support matrix" before release, so we should not rely on it at all in our work. I agree with Amit that such a thing should be tested on our side by us.

